PR #12, opened 2017-01-20
Add support for Options and refactor

dpvc:
This PR adds support for default and user-supplied options for Document, InputJax, and OutputJax objects.  In particular, `MathJax.HandlerFor()` now takes options that select the input and output jax, so the main program must load the desired jax and pass them in (otherwise, default, do-nothing jax are used).

This also refactors generic code from `HTMLDocument.js` to `Document.js` and `HTMLMathItem.js` to `MathItem.js`.

Finally, Document and MathItem types and names are no longer passed to the constructor, they are taken from the object itself.

### DIFF
--- a/v3/main.js
+++ b/v3/main.js
@@ -2,13 +2,20 @@ import {MathJax} from "mathjax/mathjax.js";
 export {MathJax} from "mathjax/mathjax.js";
 
 import "mathjax/handlers/html.js";
+import {TeX} from "mathjax/input/tex.js";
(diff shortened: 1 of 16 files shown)


zorkow:
Here's a basic question: 
How much will this code become part of the final code base?
If it will, we should start working with some more common coding style.

dpvc:
Closing this since translation of ES6 core to Typescript makes it obsolete.

PR #13, opened 2017-02-06
Use 'as' for typecasting

jasongrout:
I personally think `as` is more readable, especially since `<>` already has another meaning in the very same expression. See the bottom of https://www.typescriptlang.org/docs/handbook/basic-types.html.

### DIFF
--- a/v3/TexParser/ts/map_handler.ts
+++ b/v3/TexParser/ts/map_handler.ts
@@ -81,7 +81,7 @@ export default class MapHandler {
    * @return {T} A boolean, Character, or Macro.
    */
   public lookup<T>(symbol: string): T {
-    let map = <AbstractSymbolMap<T>>(this.applicable(symbol));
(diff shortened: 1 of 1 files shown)


v3/TexParser/ts/map_handler.ts
```diff
@@ -81,7 +81,7 @@ export default class MapHandler {
    * @return {T} A boolean, Character, or Macro.
    */
   public lookup<T>(symbol: string): T {
-    let map = <AbstractSymbolMap<T>>(this.applicable(symbol));
+    let map = this.applicable(symbol) as AbstractSymbolMap<T>;
     return map ? map.lookup(symbol) : null;
```
zorkow: Thanks. I wasn't aware of that syntax. I am still very much used to closure/Java/Ocaml styles...

jasongrout: No problem - I was just scanning the code to check out the Typescript use.

PR #16, opened 2017-02-12
Fix checking for msub/msup and munder/mover

dpvc:
This should resolve the JSON problem that Volker reported with msub in place of msup.

Test case:

    node load.js samples/json.js "x'"

should not produce msup rather than msub.


zorkow:
lgtm.

PR #17, opened 2017-02-17
Add mml jax operator table external files

dpvc:
This adds the rest of the files used by the mml jax for the operator table (when characters that are not in the portion that is preloaded are used).


dpvc:
Closing this since translation of ES6 core to Typescript makes it obsolete.

PR #18, opened 2017-03-22
Internal MathML tree

dpvc:
This is my version of the internal MathML tree structure, in Typescript.  It implements all the functionality that was in the old ElementJax (though not always in the same way).  It includes node factory code to make it possible to create these nodes (in a customizable way), and visitor functions to be able to traverse the tree (again, in a customizable way).

There is a test file in `samples/tex-nodes.js` that can be used to test the format using TeX input.  Use

    node load samples/tex-nodes.js 'tex'

where `tex` is TeX code to be parsed into the internal format.  The output will deserialized MathML with inherited attributes enclose in `[...]`, extra properties applied by the TeX input jax in `[[...]]`, and some of the important internal values enclosed in `{{...}}`.


v3/MmlTree/ts/MmlNodes/mrow.ts
```diff
@@ -0,0 +1,213 @@
...
+        if ((this.getProperty('open') != null || this.getProperty('close') != null) &&
+            (!prev || prev.getProperty('fnOp') != null)) {
+            //
+            // <mrow> came from \left...\right
+            // so treat as subexpression (TeX class INNER).
+            //
+            this.getPrevClass(prev);
+            prev = null;
+            for (const child of this.childNodes) {
+                prev = child.setTeXclass(prev);
```
zorkow: Why do we set TeXclass of the first element to null in this case?
This needs some extra commenting.

v3/MmlTree/ts/MmlNodes/mrow.ts
```diff
@@ -0,0 +1,213 @@
...
+    /*
+     * @return {boolean}  This is inferred
+     */
+    public get isInferred() {
+        return true;
+    }
+
+    /*
+     * @return {boolean}  This node is not considered a parent node
+     */
```
zorkow: override
l

v3/MmlTree/ts/Attributes.ts
```diff
@@ -0,0 +1,191 @@
...
+        return value;
+    }
+
+    /*
+     * @param {string} name  The value of the attribute whose value is to be returned
+     * @return {Property}  The attribute whse name was given if it is explicit on the
+     *                      node (not inherited or defaulted), null otherwise
+     */
+    public getExplicit(name: string) {
+        if (!this.attributes.hasOwnProperty(name)) {
```
dpvc: Well, it's really an enhanced PropertyList -- one with a prototype chain that inherits from the `inherited` and `default` PropertyLists.  It acts like a PropertyList, but the properties may come from any of three places:  the explicit properties in `attributes`, or the ones in `inherited` or the ones in `defaults`.  This is why we don't have to implement the old `Get` function ourselves; object inheritance does that for us.

So to determine if an attributes has been given explicitly (rather than being inherited), we need to use `hasOwnProperty()`.

v3/MmlTree/ts/Node.ts
```diff
@@ -0,0 +1,355 @@
...
+ *  attributes and other properties
+ */
+export type Property = string | number | boolean;
+export type PropertyList = {[key: string]: Property};
+
+/*
+ *  The basic Node and NodeClass types
+ */
+export type Node = INode;
+export type NodeClass = INodeClass;
```
dpvc: OK.

v3/MmlTree/ts/Node.ts
```diff
@@ -0,0 +1,355 @@
...
+/*
+ *  The abstract Node class
+ */
+
+export abstract class ANode implements INode {
+
+    /*
+     * The parent node for this one
+     */
+    public parent: Node = null;
```
dpvc: I don't understand the comment.  Can you explain further what you mean?  (Note that `Node` is the same as `INode` by the definition of `Node`).

v3/MmlTree/ts/Node.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    /*
+     * The properties for this node
+     */
+    protected properties: PropertyList = {};
+
+    /*
+     * The NodeFactory to use to create additional nodes, as needed
+     */
+    public factory: NodeFactory = null;
```
dpvc: It didn't seem necessary to add that overhead in order to prevent something we weren't going to do anyway.

v3/MmlTree/ts/Node.ts
```diff
@@ -0,0 +1,355 @@
...
+     * @param {NodeFactory} factory  The NodeFactory to use to create new nodes when needed
+     * @param {PropertyList} properties  Any properties to be added to the node, if any
+     * @param {Node[]} children  The initial child nodes, if any
+     * @return {Node}  The newly created node
+     *
+     * @constructor
+     * @implements {INode}
+     */
+    constructor(factory: NodeFactory, properties: PropertyList = {}, children: Node[] = []) {
+        this.factory = factory;
```
dpvc: I'm not sure I understand what `public` means for the argument of `constructor()`.  I don't know what you mean by `declaration of factory field class body`.  If this is just part of the "use a getter to prevent changing the value", then I get it, but if it is something else, I don't.

v3/MmlTree/ts/Node.ts
```diff
@@ -0,0 +1,355 @@
...
+        return newChild;
+    }
+
+
+    /*
+     * @override
+     */
+    public childIndex(node: Node) {
+        let i = 0;
+        for (const child of this.childNodes) {
```
dpvc: I'm too used to avoiding features not available in older versions of IE.

v3/MmlTree/ts/Node.ts
```diff
@@ -0,0 +1,355 @@
...
+     */
+    public replaceChild(newChild: Node, oldChild: Node) {
+        return oldChild;
+    }
+
+    /*
+     * @override
+     */
+    public childIndex(node: Node) {
+        return null as number;
```
dpvc: Yes.  Otherwise it gets a return type of `any`, and typescript (with our configuration) complains.

v3/mathjax/legacy/jax/element/MmlNode.js
```diff
@@ -0,0 +1,110 @@
...
+    'isError',
+    'multiline',
+    'variantForm',
+    'autoOP'
+  ];
+
+  MML.mbase.Augment({
+    toMmlNode: function (factory) {
+      var m = this.data.length;
+      var kind = this.type; if (kind === 'texatom') kind = 'TeXAtom';
```
dpvc: This is a legacy file that is only used as glue to get from the old internal format to the new internal format, so this is not going to end up as part of v3.  So I'm not too worried about format, here.

v3/mathjax/legacy/jax/element/MmlNode.js
```diff
@@ -0,0 +1,110 @@
...
+      if (this.inferred) kind = 'inferredMrow';
+      var node = factory.MML[kind]();
+      for (var i = 0; i < m; i++) {
```
dpvc: Again, this file is only temporary.

v3/mathjax/legacy/jax/element/MmlNode.js
```diff
@@ -0,0 +1,110 @@
...
+    nodeAddAttributes: function (node) {
+      var defaults = (this.type === "mstyle" ? MML.math.prototype.defaults : this.defaults);
+      var names = (this.attrNames||MML.copyAttributeNames),
+          skip = MML.skipAttributes,
+          copy = MML.copyAttributes;
+      if (!this.attrNames) {
+        for (var id in defaults) {
+          if (!skip[id] && !copy[id] && defaults.hasOwnProperty(id)) {
+            if (this[id] != null && this[id] !== defaults[id]) {
+              if (this.Get(id,null,1) !== this[id]) node.attributes.set(id,this[id]);
```
dpvc: ditto.

v3/mathjax/legacy/jax/element/MmlNode.js
```diff
@@ -0,0 +1,110 @@
...
+        if (child) node.appendChild(child.toMmlNode(factory));
+      }
+      this.nodeAddAttributes(node);
+      return node;
+    }
+  });
+  
+  MML.munderover.Augment({
+    toMmlNode: function (factory) {
+      var kind = (this.data[this.under] == null ? 'mover' :
```
dpvc: Yes, but in the next line, not this one.

v3/MmlTree/ts/Visitor.ts
```diff
@@ -0,0 +1,173 @@
...
+        for (const kind of factory.getKinds()) {
+            let method = (this as IVisitor)[Visitor.methodName(kind)] as VisitorFunction;
+            if (method) {
+                this.nodeHandlers.set(kind, method);
+            }
+        }
+    }
+
+    /*
+     * Visit the tree by visiting the root node
```
dpvc: I don't really know what `@override` does, or what is needed for it.  I would think that in some cases, a new comment would be needed to explain what the overridden function does (i.e., what its additional functionality is that isn't in the original).

v3/MmlTree/ts/MmlNodes/mrow.ts
dpvc: > Why do we set TeXclass of the first element to null in this case?

We don't.  We set the *previous* element to null, because within `\left...\right`, the first child has no previous element to affect its spacing.

In TeX, the spacing is based on two adjacent items, and `setTeXclass(prev)` isn't setting the class to be `prev`, it is setting the TeX class given that the previous item is `prev`.  There is none in this case.

If that's not what you are meaning, let me know.

dpvc:
@zorkow, OK, I've addressed all the comments, either in line above, or by changes in the code.  Some of the changes don't seem to have closed the comments above, however, but most did.

So you can do a second pass at this point.

dpvc:
@zorkow, thanks!

pkra:
:fireworks: fireworks: fireworks:

PR #20, opened 2017-04-28
Convert ES6 to Typescript

dpvc:
This PR translates all the ES6 files to Typescript, and reorganizes the directory structure to accommodate that.  Unfortunately the renaming causes extra complication in the file differences, since they include the removal of the old ones as well as the creation of the new ones.  Not sure how to make that work better at this point.

There is now a directory `mathjax2` for the legacy code (non-typescript), and some glue files for accessing it (these weren't translated to Typescript, as they will be going away when everything is converted to v3 anyway).  The `mathjax3` directory contains the compiled javascript output of the typescript, and `mathjax-ts` is where all the typescript files live.


v3/mathjax3-ts/core/Handler.ts
```diff
@@ -0,0 +1,130 @@
...
+    Create(document: any, options: OptionList): MathDocument;
+}
+
+/*****************************************************************/
+/*
+ *  The Handler class interface
+ */
+
+export interface HandlerClass {
+    /*
```
dpvc: As discussed face to face, I'm removing the *Class interfaces.

v3/mathjax3-ts/core/Handler.ts
```diff
@@ -0,0 +1,130 @@
...
+/*
+ *  The Handler interface
+ */
+
+export abstract class AbstractHandler implements Handler {
+
+    /*
+     * The name of this class
+     */
+    public static NAME: string = 'generic';
```
dpvc: This is a class (static) variable, but the getter is an instance method.  One might still want to do `myClass.NAME` for `myClass` a subclass of `AbstractHandler`.

v3/mathjax3-ts/core/Handler.ts
```diff
@@ -0,0 +1,130 @@
...
+    constructor(priority: number = 5) {
+        this.priority = priority;
+    }
+
+    /*
+     * @return{string}  The name of this handler class
+     */
+    public get name() {
+        return (this.constructor as HandlerClass).NAME;
+    }
```
dpvc: Just to get Typescript to allow access to the `NAME` property of `this.constructor`.  I have changed to `typeof AbstractHandler` instead, as we discussed.

v3/mathjax3-ts/util/LinkedList.ts
```diff
@@ -0,0 +1,356 @@
...
+     * @constructor
+     */
+    constructor (data: any = null) {
+        this.data = data;
+    }
+}
+
+/*****************************************************************/
+/*
+ *  The ListList class interface
```
dpvc: OOPS!

v3/mathjax3-ts/util/LinkedList.ts
```diff
@@ -0,0 +1,356 @@
...
+/*
+ *  Implements the generic ListList class
+ */
+
+export class LinkedList<DataClass> {
+
+    /*
+     * The linked list
+     */
+    protected list: ListItem<DataClass>;
```
dpvc: What is wrong with `list`?  this points to the actual linked list.  I could see using `root`, though `dummy` doesn't make sense to me, and I don't see that `anchor` describes what it actually is.

v3/mathjax3-ts/util/LinkedList.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    /*
+     *  Used for sorting and merging lists (Overridden by subclasses)
+     *
+     * @param{DataClass} a   The first item to compare
+     * @param{DataClass} b   The second item to compare
+     * @return{boolean}      True if a is before b, false otherwise
+     */
+    public isBefore(a: DataClass, b: DataClass) {
+        return (a < b);
```
dpvc: `DataClass` could be `string` or `number` or some non-object class without methods of their own.  For a linked list of objects, one would override this method (like the `MathList` does).

v3/mathjax3-ts/util/LinkedList.ts
```diff
@@ -0,0 +1,356 @@
...
+        this.list.next = this.list.prev = this.list;
+        return this;
+    }
+
+    /*
+     * Make the list iterable and return the data from the items in the list
+     *
+     * @return{{next: Function}}  The object containing the iterator's next() function
+     */
+    public [Symbol.iterator]() {
```
dpvc: This is what makes an object iterable (i.e., can be used in for-of statements).

See [this description](http://2ality.com/2015/02/es6-iteration.html#implementing-iterables) for more details.

v3/mathjax3-ts/util/LinkedList.ts
```diff
@@ -0,0 +1,356 @@
...
+     */
+    public sort(isBefore: SortFn<DataClass> = null) {
+        if (isBefore === null) {
+            isBefore = this.isBefore.bind(this);
+        }
+        //
+        //  Make an array of singleton lists
+        //
+        let lists: LinkedList<DataClass>[] = [];
+        for (const item of this) {
```
dpvc: Yes, that is the purpose of the `[Symbol.iterator]()` method above.  This is what required Typescript 2.3, since earlier versions didn't implement it for the ES5 target (but did for ES6 target).

v3/mathjax3-ts/core/FindMath.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/*****************************************************************/
+/*
+ *  The FindMath class interface
+ */
+export interface FindMathClass {
+    /*
+     * The default options
+     */
+    OPTIONS: OptionList;
```
dpvc: It is.  The `FindMathClass` interface describes the FindMath class object, so `OPTIONS` here means that the abstract class as `OPTIONS` as a static variable, as is indicated in [line 63](https://github.com/mathjax/mathjax-v3/pull/20/files/2e76d08cf358b652cc143f3edd9bde9f8027ca4b#diff-49dd24eef8a13d63cfad88e0117f3d2dR63) below.

v3/mathjax3-ts/core/FindMath.ts
```diff
@@ -0,0 +1,88 @@
...
+     * The actual options for this instance
+     */
+    protected options: OptionList;
+
+    /*
+     * @param {OptionList} options  The user options for this instance
+     */
+    constructor(options: OptionList) {
+        let CLASS = this.constructor as FindMathClass;
+        this.options = userOptions(defaultOptions({}, CLASS.OPTIONS), options);
```
dpvc: In order to be able to access the `OPTIONS` property in line 75 (as `CLASS.OPTIONS`).  Without that, Typescript will not allow access to the property.

v3/mathjax3-ts/core/InputJax.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.options = userOptions(defaultOptions({}, CLASS.OPTIONS), options);
+        this.preFilters = new FunctionList();
+        this.postFilters = new FunctionList();
+    }
+
+    /*
+     * @return{string}  The name of this input jax class
+     */
+    public get name() {
+        return (this.constructor as InputJaxClass).NAME;
```
dpvc: It *is* the class member `NAME`, but it is the one for the class of whatever subclass of `AbstractInputJax` is the instance that is calling this method.  That is, this gets the proper name from the class of whatever the instance object is an instance of.

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+     */
+    processed: {[name: string]: boolean};
+
+    /*
+     * An array of input jax to run on the document
+     */
+    inputJax: InputJax[];
+
+    /*
+     * The output jax to use for the document
```
dpvc: Are you asking about the InputJax or the OutputJax?  For InputJax, yes, documents can use more than one input format.  For OutputJax, I don't see a use case for more than one.  (I have provisions for that in v2, but it never came up in 8 years of usage.

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+    outputJax: OutputJax;
+
+    /*
+     * Locates the math in the document and constructs the MathList
+     *  for the document.
+     *
+     * @param{OptionList} options  The options for locating the math
+     * @return{MathDocument}       The math document instance
+     */
+    findMath(options?: OptionList): MathDocument;
```
dpvc: It is more complicated than that.  For one thing, the handler has to create the array of strings from the document for the FindMath object to search, so it is the handler that would call FindMath.  See our email for more.

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+     * @return{MathDocument}  The math document instance
+     */
+    typeset(): MathDocument;
+
+    /*
+     * Add any event handlers to the typeset math
+     *
+     * @return{MathDocument}  The math document instance
+     */
+    addEventHandlers(): MathDocument;
```
dpvc: The result of the OutputJax is HTML nodes, or SVG nodes, or MathML nodes, regardless of the document type, and it may be the the typeset math is not actually inserted back into the original document.  For Markdown, you would first need to process the markdown to get HTML and *then* insert the resulting HTML.  But you will need to add the event handlers in any case.  So I think this *does* belong at the interface level.

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+
+    /*
+     * Updates the document to include the typeset math
```
dpvc: the `updateDocument()` method should do whatever is needed to get the math into the document.  For an HTML document, that would include not only replacing the original math text with the typeset version, but also adding whatever stylesheet is needed.  For a Markdown document, it could also include replacing the math with markers (that don't include any Markdown commands), processing the markdown to get HTML, and then inserting the typeset math at the location of the markers.  (That is how the current editor/preview works in StackExchange, for example.)

So I think "update" is the right word rather than "replace".

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+     */
+    reset(): MathDocument;
+
+    /*
+     * Reset the processed values and clear the MathList (so that new math
+     * can be processed in the document).
+     *
+     * @return{MathDocument}  The math document instance
+     */
+    clear(): MathDocument;
```
dpvc: `removeFromDocument()` takes the math back out of the document, putting back the original math.  `clear()` clears the processed status indicators so that you can process new math added to the document (without losing track of the math already in the document, and leaving the typeset math in the document).  `reset()` clears the math list as well as the processed status, i.e., forgets everything it knows about the page, but leaves the typeset math in place.

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+
+    /*
+     * The default options for this type of document
+     */
+    OPTIONS: OptionList;
+
+    /*
+     *  The document states
+     */
+    STATE: {[name: string]: number};
```
dpvc: The possible set of states is not a fixed list.  It may be augmented by the subclasses of MathDocument, as needed.

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+}
+
+/*****************************************************************/
+/*
+ *  The booleans used to keep track of what processing has been
+ *  performed.
+ */
+
+export interface MathProcessed {
+    findMath: boolean;
```
dpvc: I think it is probably just a `type`.  That seems to work.

v3/mathjax3-ts/core/MathDocument.ts
```diff
@@ -0,0 +1,401 @@
...
+     * @override
+     */
+    public concat(list: MathList) {
+        this.math.merge(list);
+        return this;
+    }
+
+}
+
+let STATE = AbstractMathDocument.STATE;
```
dpvc: Left over from pre-Typescript implementation.  ES6 doesn't handle static variables in the same way Typescript does.

v3/mathjax3-ts/input/asciimath/FindAsciiMath.ts
```diff
@@ -0,0 +1,173 @@
...
+     * Search a string for math delimited by one of the delimiter pairs.
+     *
+     * @param{ProtoItem[]} math  The array of proto math items located so far
+     * @param{number} n          The index of the string being searched
+     * @param{string} text       The string being searched
+     */
+    protected findMathInString(math: ProtoItem[], n: number, text: string) {
+        let start, match;
+        this.start.lastIndex = 0;
+        while ((start = this.start.exec(text))) {
```
dpvc: Without these, some browsers (Firefox or Chrome, I forget which), give a warning about this, since it might look like a typo for `while (start == this.start.exec(text)) {` with `==` instead of `=`.  The extra parenthesis are what it says to use to prevent the warning.

v3/mathjax3-ts/core/MathItem.ts
```diff
@@ -0,0 +1,331 @@
...
+ *  The MathItem interface
+ *
+ *  The MathItem is the object that holds the information about a
+ *  particular expression on the page, including pointers to
+ *  where it is in the document, its compiled version (in the
+ *  internal format), its typeset version, its bounding box,
+ *  and so on.
+ */
+
+export interface MathItem {
```
dpvc: No, `ProtoItem` is a separate type whose data is used to create the actual `MathItem` alter.

v3/mathjax3-ts/core/MathItem.ts
```diff
@@ -0,0 +1,331 @@
...
+
+    /*
+     * @override
+     */
+    public addEventHandlers() {}
+
+    /*
+     * @override
+     */
+    public updateDocument(document: MathDocument) {}
```
dpvc: No, just like `addEventHandlers()`, this method does nothing if you don't implement it.  You don't *have* to implement it.

v3/mathjax3-ts/core/MathItem.ts
```diff
@@ -0,0 +1,331 @@
...
+            containerWidth: cwidth,
+            lineWidth: lwidth,
+            scale: scale
+        };
+    }
+
+    /*
+     * @override
+     */
+    public state(state: number = null, restore: boolean = false) {
```
dpvc: Calling `state()` returns the state without setting it.

v3/mathjax3-ts/core/MathItem.ts
```diff
@@ -0,0 +1,331 @@
...
+                this.inputData = {};
+            }
+            this._state = state;
+        }
+        return this._state;
+    }
+
+}
+
+let STATE = AbstractMathItem.STATE;
```
dpvc: Right, different document classes can have their own states.  For example, a Markdown document might have states for whether the item has been encoded in the document (i.e., replaced by a token that identifies the math but has no Markdown commands in it, like source TeX can) or not.  I don't know what all those states may be, so can't have just a fixed set.  I suspect the ones here are sufficient, but don't know for sure.

v3/mathjax3-ts/handlers/html/HTMLDocument.ts
```diff
@@ -0,0 +1,239 @@
...
+     * Get a list of containers (DOM nodes) to be searched for math.  These can be
+     *  specified by CSS selector, or as actual DOM elements or arrays of such.
+     *
+     * @param{(string | Element | Element[])[]} nodes  The array of items to make into a container list
+     * @param{Document} document                       The document in which to search
+     * @return{Element[]}                              The array of containers to search
+     */
+    protected getElements(nodes: (string | Element | Element[])[], document: Document) {
+        let containers: Element[] = [];
+        let NODELIST = document.childNodes.constructor as typeof NodeList;
```
dpvc: In node.js, there is no `window` or `document` object, so there is no global `NodeList` class to refer to (in a browser, `NodeList` would be on the `window` object).  Normally, you could look up the `window` from `document.defaultView`, but for a document created by `DOMParser.parseFromString()`, `document.defaultView` is null.  So I am getting the `NodeList` class from an instance of it (the `document.childNodes` array).

The reason we want this is we want to be able to process the results of calls like `document.selectorQueryAll('...')` without having to have the user convert to an array manually.  I really need to handle `HTMLCollection` as well, so will need to do more here anyway.  Let me see what I can do it improve the situation...

v3/mathjax3-ts/handlers/html/HTMLDomStrings.ts
```diff
@@ -0,0 +1,285 @@
...
+        let CLASS = this.constructor as HTMLDomStringsClass;
+        this.options = userOptions(defaultOptions({}, CLASS.OPTIONS), options);
+        this.init();
+        this.getPatterns();
+    }
+
+    /*
+     * Set the initial values of the main properties
+     */
+    protected init() {
```
dpvc: Doesn't it figure that out itself?

v3/mathjax3-ts/handlers/html/HTMLHandler.ts
```diff
@@ -0,0 +1,114 @@
...
+import {AbstractHandler} from '../../core/Handler.js';
+import {HTMLDocument} from './HTMLDocument.js';
+import {OptionList} from '../../util/Options.js';
+import {DOM} from '../../util/DOM.js';
+
+/*****************************************************************/
+/*
+ *  Typescript's Window doesn't seem to include these constructors,
+ *  so use this fake interface to access them.
+ */
```
dpvc: Taken care of differently via `util/DOM.ts`

v3/mathjax3-ts/handlers/html/HTMLHandler.ts
```diff
@@ -0,0 +1,114 @@
...
+     * by a zerialized HTML document rather than an already parsed one.
+     */
+    protected parser: DOMParser;
+
+    /*
+     * Some classes that can be used as the document for HTMLHandler.
+     */
+    protected DOCUMENT: typeof Document;
+    protected HTMLELEMENT: typeof HTMLElement;
+    protected FRAGMENT: typeof DocumentFragment;
```
dpvc: Removed, and taken care of differently now.

v3/mathjax3-ts/handlers/html/HTMLHandler.ts
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public handlesDocument(document: any) {
+        if (typeof(document) === 'string') {
+            try {
+                document = this.parser.parseFromString(document, 'text/html');
+            } catch (err) {}
+        }
+        if (document instanceof this.DOCUMENT ||
+            document instanceof this.HTMLELEMENT ||
+            document instanceof this.FRAGMENT) {
```
dpvc: Ditto.

v3/mathjax3-ts/handlers/html/HTMLMathItem.ts
```diff
@@ -0,0 +1,131 @@
...
+    public updateDocument(html: HTMLDocument) {
+        if (this.state() < STATE.INSERTED) {
+            if (this.inputJax.processStrings) {
+                let node = this.start.node as Text;
+                if (node === this.end.node) {
+                    if (this.end.n < this.end.node.nodeValue.length) {
+                        this.end.node.splitText(this.end.n);
+                    }
+                    if (this.start.n) {
+                        node = (this.start.node as Text).splitText(this.start.n);
```
dpvc: No, `this.start.node` is the node in which the start of the math expression occurs, not the start of the document.  The math may span several nodes (e.g., several Text nodes, or two text nodes separated by ` <br>` node), so we may need to traverse several nodes to get the text of the math.  We split the node where the start occurs (at the start location, which is `this.start.n`, and use the second half as the beginning of the math.  Then collect text nodes until we get to the `this.end.node`, and split that to get the final piece of the text.

Note that we aren't searching through the document at this point, as we have pointers to the nodes already (the searching was done earlier with `findMath()`).

v3/mathjax3-ts/input/mathml.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /*
+     * Throw an error
+     *
+     * @param{string} message  The error message to produce
+     */
+    protected error(message: string) {
+        // FIXME:  should this be creating merror nodes instead?
+        throw new Error(message);
```
dpvc: Not always.  Some calls are not because it didn't parse, but because of other errors (like the root node isn't a `<math>` node).

v3/mathjax3-ts/input/tex/FindTeX.ts
```diff
@@ -0,0 +1,261 @@
...
+     *   or by \being{env}...\end{env}, or \eqref{...}, \ref{...}, \\, or \$.
+     *
+     * @param{ProtoItem[]} math  The array of proto math items located so far
+     * @param{number} n          The index of the string being searched
+     * @param{string} text       The string being searched
+     */
+    protected findMathInString(math: ProtoItem[], n: number, text: string) {
+        let start, match;
+        this.start.lastIndex = 0;
+        while ((start = this.start.exec(text))) {
```
dpvc: Same reason as above.

v3/mathjax3-ts/util/Options.ts
```diff
@@ -0,0 +1,218 @@
...
+ */
+export const APPEND = Symbol('Append to option array');
+
+
+/*****************************************************************/
+/*
+ * Get all keys and symbols from an object
+ *
+ * @param{Optionlist} def        The object whose keys are to be returned
+ * @return{(string | symbol)[]}  The list of keys for the object
```
dpvc: `symbol` is the primitive type (the actual symbol), whereas `Symbol` is the constructor for a `symbol`.

v3/mathjax3-ts/util/PrioritizedList.ts
```diff
@@ -0,0 +1,121 @@
...
+    /*
+     * Add an item to the list
+     *
+     * @param{DataClass} item   The data for the item to be added
+     * @param{number} priority  The priority for the item
+     * @return{DataClass}       The data itself
+     */
+    public add(item: DataClass, priority: number = PrioritizedList.DEFAULTPRIORITY) {
+        let i = this.items.length;
+        do { i--; } while (i >= 0 && priority <= this.items[i].priority);
```
dpvc: A waste of space and harder to read, but OK.

v3/mathjax3-ts/util/PrioritizedList.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /*
+     * Remove an item from the list
+     *
+     * @param{DataClass} item   The data for the item to be removed
+     */
+    public remove(item: DataClass) {
+        let i = this.items.length;
+        do { i--; } while (i >= 0 && this.items[i].item !== item);
```
dpvc: Ditto.

v3/mathjax3-ts/util/PrioritizedList.ts
```diff
@@ -0,0 +1,121 @@
...
+        if (i >= 0) {
+            this.items.splice(i, 1);
+        }
+    }
+
+    /*
+     * Typescript < 2.3 targeted at ES5 doesn't handle
+     *
+     *     for (const x of this) {...}
```
dpvc: I've used it now that we switched to Typescript 2.3, but left the method in place in case anyone wants to get the array (so they don't have to do `Array.from()` by hand, or in case they are using Typescript < 2.3).  I can remove the method(s) if you prefer.

v3/mathjax3-ts/core/FindMath.ts
dpvc: In any case, the `FindMathClass` interface has been removed, along with the other `*Class` ones.

v3/mathjax3-ts/core/FindMath.ts
dpvc: In any case, I've cast to `typeof FindMath` instead.

v3/mathjax3-ts/handlers/html/HTMLDocument.ts
dpvc: OK, I've worked around that in a different way.

dpvc:
@zorkow, I believe that I have addressed your comments in the branch `ts-core-review1`, except for the issue of separating FindMath from InputJax, which we still need to discuss.  

You can see the [differences](https://github.com/mathjax/mathjax-v3/compare/ts-core...ts-core-review1?expand=1), but I haven't merged the changes in hopes that this comparison will be easier to view than if I merged the changes to `ts-core` first.  If you would rather that I make the merge first, let me know.

dpvc:
OK, I've merged the changes that I've made, so you can do round 2 of review when you get the time.

v3/mathjax3-ts/util/LinkedList.ts
zorkow: That still feels incomplete as a binary predicate does not resolve disambiguity for a sorted list. What if a=b; will I get two copies of the same element?

v3/mathjax3-ts/core/MathDocument.ts
zorkow: Make sense

v3/mathjax3-ts/core/MathDocument.ts
zorkow: But it supposes that we are in some sort of live DOM structure, where events actually exist.
Actually 
> The result of the OutputJax is HTML nodes, or SVG nodes, or MathML nodes, regardless of the 
> document type, and it may be the the typeset math is not actually inserted back into the original 
> document. 
sounds to me that event handling should be separated from the document.
Is it that you are worried that expressions would need to be processed twice?

v3/mathjax3-ts/core/MathDocument.ts
zorkow: Fine.

zorkow:
Still two minor requests/questions. O/w it's good to go.

v3/mathjax3-ts/util/LinkedList.ts
dpvc: Yes, the sorted list will have the same number of elements as the original, even if some are "equal".  There is nothing about the list that requires the elements to be unique.  If the two are equal, then their order doesn't really matter and their order (among those of the same value) is essentially random.  If there should be a distinction among them, then you would want to make your `isBefore()` function take that into account (e.g., `return  a.i < b.i || (a.i === b.i && a.j < b.j)`).

v3/mathjax3-ts/core/MathDocument.ts
dpvc: If you really are concerned about this, I can take it out of the abstract class and put it into the HTML handler and any others for which it is important, but I wanted you to be able to make the major calls no matter what output jax was being used (and the ones that didn't make sense would not do anything).  But I don't have strong feelings for that.

pkra:
Hooray!

Issue #21, opened 2017-05-09
directory structure


pkra:
Move the webworker code into another repository and move the v3 folder contents to the top level.

pkra:
Ping.

dpvc:
Doing this while the ts-core PR is open would cause difficulties with the merge, and complicate the development processes while waiting for that PR to be reviewed.  We need to wait until there are no outstanding major PRs to do this.

pkra:
There are no PRs open. Can we please fix this?

PR #23, opened 2017-07-27
Move v3 directories to the top level. #21

dpvc:
This removes the webworker folder and transfer the contents of the v3 folder to the top level, as requested by @pkra in issue #21.


dpvc:
I don't think this really requires review, so I'm merging it (so I can start basing the new development off of it).  It looks like margin branch from before this change should rename (existing) files properly, but new files might need to be renamed by hand after merging with the renamed folders.

PR #24, opened 2017-07-30
Misc fixes

dpvc:
These are a couple of small things I found while writing tests for the utility functions.


dpvc:
Since this is just typos in comments, and one incorrect `<=` that should be `<`, I don' think this require review, so I'm merging it.

PR #25, opened 2017-08-01
Make Factory object use generic types

dpvc:
This implements factories for generic types, as the `Factory` object in `mathjax3-ts/core/Tree/Factory.ts`.

Most of the functionality of `NodeFactory` has been moved to `Factory` and made generic.  `Factory<N, C>` takes a node type `N` and a node class `C` which give the interfaces for the node instance and the node constructors.  We need both for two reasons:  first, you can't use `typeof N` to get `C`, since `N` is a type not an object, and if `N` has static members, we may want to access them from the results of `getNodeClass(kind)` (this is done in MmlNodes, for example).

Most of `MmlFactory` can now be removed, since it is in the generic `Factory<MmlNode, MmlNodeClass>` object and was only in `MmlFactory` to set the types from `Node` to `MmlNode` and `NodeClass` to `MmlNodeClass`, and to set the default for `nodes` in the constructor (this is now handled via a static property).

Finally, `mathjax3-ts/core/Tree/Node.js` and `mathjax3-ts/core/Tree/Visitor.js` needed to have their references to `NodeFactory` to be updated.  The `Node` and `Visitor` objects themselves need to be made generic, which will be handled in a later PR (since Volker asked for small changes).

### DIFF
--- a/mathjax3-ts/core/Tree/Factory.ts
+++ b/mathjax3-ts/core/Tree/Factory.ts
@@ -30,15 +30,34 @@ export interface FactoryNode {
     readonly kind: string;
 }
 
+/*
+ * @template N  The Node type being created by teh factory
+ */
 export interface FactoryNodeClass<N extends FactoryNode> {
+    /*
+     * @param{Factory<N, FactoryNodeClass<N>>} factory  The factory for creating more nodes
+     * @param{any[]} args  Any additional arguments needed by the node
+     * @return{N}  The newly created node
+     */
     new(factory: Factory<N, FactoryNodeClass<N>>, ...args: any[]): N;
 }
 
 /*****************************************************************/
 /*
  * The Factory interface
+ *
+ * Factory<N, C> takes a node type N and a node class C, which give
+ * the interfaces for the node instance and the node constructors. We
+ * need both for two reasons: first, you can't use typeof N to get C,
+ * since N is a type not an object, and if N has static members, we
+ * may want to access them from the results of getNodeClass(kind)
+ * (this is done in MmlNodes, for example).
  */
 
+/*
+ * @template N  The node type created by the factory
+ * @template C  The class of the node being constructed (for access to static properties)
+ */
 export interface Factory<N extends FactoryNode, C extends FactoryNodeClass<N>> {
     /*
      * @param {string} kind  The kind of node to create
@@ -53,11 +72,13 @@ export interface Factory<N extends FactoryNode, C extends FactoryNodeClass<N>> {
      * @param {C} nodeClass  The class for the given kind
      */
     setNodeClass(kind: string, nodeClass: C): void;
+
     /*
      * @param {string} kind  The kind of node whose class is to be returned
      * @return {C}  The class object for the given kind
      */
     getNodeClass(kind: string): C;
+
     /*
      * @param {string} kind  The kind whose definition is to be deleted
      */
@@ -83,6 +104,10 @@ export interface Factory<N extends FactoryNode, C extends FactoryNodeClass<N>> {
  *   (needed for access to defaultNodes via the constructor)
  */
 
+/*
+ * @template N  The node type created by the factory
+ * @template C  The class of the node being constructed (for access to static properties)
+ */
 interface AbstractFactoryClass<N extends FactoryNode, C extends FactoryNodeClass<N>> extends Function {
     defaultNodes: {[kind: string]: C};
 }
@@ -93,6 +118,10 @@ interface AbstractFactoryClass<N extends FactoryNode, C extends FactoryNodeClass
  * The generic AbstractFactory class
  */
 
+/*
+ * @template N  The node type created by the factory
+ * @template C  The class of the node being constructed (for access to static properties)
+ */
 export abstract class AbstractFactory<N extends FactoryNode, C extends FactoryNodeClass<N>> implements Factory<N, C> {
 
     /*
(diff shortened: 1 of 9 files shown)


mathjax3-ts/core/Tree/Factory.ts
```diff
@@ -0,0 +1,176 @@
...
+/*****************************************************************/
+/*
+ * The Factory node interfaces (one for the node instance, one for the node class)
+ */
+
+export interface FactoryNode {
+    readonly kind: string;
+}
+
+export interface FactoryNodeClass<N extends FactoryNode> {
```
zorkow: Missing JSDOC.

mathjax3-ts/core/Tree/Factory.ts
```diff
@@ -0,0 +1,176 @@
...
+     * An object containing functions for creating the various node kinds
+     */
+    protected node: {[kind: string]: (...args: any[]) => N} = {};
+
+    /*
+     * @override
+     */
+    constructor(nodes: {[kind: string]: C} = null) {
+        if (nodes === null) {
+            nodes = (this.constructor as AbstractFactoryClass<N, C>).defaultNodes;
```
zorkow: We cast a constructor to be an extended function.

mathjax3-ts/core/Tree/Factory.ts
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * @fileoverview  The generic Factory class for creating arbitrary objects
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+/*****************************************************************/
+/*
+ * The Factory node interfaces (one for the node instance, one for the node class)
+ */
```
zorkow: These exist for generics extension.

mathjax3-ts/core/Tree/Factory.ts
```diff
@@ -0,0 +1,176 @@
...
+    create(kind: string): N;
+
+    /*
+     * Defines a class for a given node kind
+     *
+     * @param {string} kind  The kind whose class is being defined
+     * @param {C} nodeClass  The class for the given kind
+     */
+    setNodeClass(kind: string, nodeClass: C): void;
+    /*
```
zorkow: Whitespace.

mathjax3-ts/core/Tree/Factory.ts
```diff
@@ -0,0 +1,176 @@
...
+     * @param {string} kind  The kind of node whose class is to be returned
+     * @return {C}  The class object for the given kind
+     */
+    getNodeClass(kind: string): C;
```
zorkow: Whitespace.

mathjax3-ts/core/Tree/NodeFactory.ts
```diff
@@ -22,119 +22,34 @@
...
 import {Node, NodeClass, PropertyList} from './Node.js';
+import {Factory, FactoryNodeClass, AbstractFactory} from './Factory.js';
 
 /*****************************************************************/
 /*
  * The NodeFactory interface
  */
 
-export interface NodeFactory {
+export interface NodeFactory<N extends Node, C extends FactoryNodeClass<N>> extends Factory<N, C> {
```
zorkow: Docs for N and C?

mathjax3-ts/core/Tree/Factory.ts
dpvc: I'm not sure I understand the comment.  What are you asking to be done, here?

mathjax3-ts/core/Tree/Factory.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+
+/*****************************************************************/
+/*
+ * The generic AbstractFactoryClass interface
+ *   (needed for access to defaultNodes via the constructor)
+ */
+
+interface AbstractFactoryClass<N extends FactoryNode, C extends FactoryNodeClass<N>> extends Function {
```
dpvc: It turns out that the casting in line 123 below needs it.  

```
(this.constructor as AbstractFactoryClass<N, C>)
```

produces the error:

```
Type 'Function' cannot be converted to type 'AbstractFactoryClass<N, C>'.
  Property 'defaultNodes' is missing in type 'Function'.
```

If you know a better way to accomplish this, let me know.

mathjax3-ts/core/Tree/Factory.ts
dpvc: This is to gain access to the `defaultNodes` static property of whatever subclass is being instantiated).  I'm not sure what your comment is asking me to do.

mathjax3-ts/core/Tree/Factory.ts
dpvc: Added

dpvc:
I've made the requested updates (at least I think so).

mathjax3-ts/core/Tree/Factory.ts
zorkow: Not a comment. That was more a remark for me, as I mentioned in our meeting. I did not remove them as I wanted to get the review out before the meeting.

mathjax3-ts/core/Tree/Factory.ts
zorkow: ditto.

mathjax3-ts/core/Tree/Factory.ts
dpvc: OK, now I understand.  Sorry!

PR #26, opened 2017-08-01
Fix some minor linting issues.

dpvc:
I don't think these will be controversial...


dpvc:
> It's odd that we require semi-colons after type blocks.

I don't see anything in the tslint configuration for this.  Is it a problem?

PR #27, opened 2017-08-08
Minor fixes

dpvc:
This fixes up a few minor issues:

* In legacy TeX input jax, the class wasn't being passed along

* In MmlNode attributes, the global attributes need to be part of the lookup chain (and some comments were wrong)

* For mi elements, the mathvariant should be set only if it isn't being inherited

* For mstyle elements, they should set the displaystyle if it is explicitly included.

### DIFF
--- a/mathjax2/legacy/jax/element/MmlNode.js
+++ b/mathjax2/legacy/jax/element/MmlNode.js
@@ -48,6 +48,7 @@
             }
           }
         }
+        if (this['class']) node.attributes.set('class',this['class']);
       }
       for (var i = 0, m = names.length; i < m; i++) {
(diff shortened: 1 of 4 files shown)


mathjax2/legacy/jax/element/MmlNode.js
```diff
@@ -48,6 +48,7 @@
             }
           }
         }
+        if (this['class']) node.attributes.set('class',this['class']);
```
zorkow: Missing brackets.

dpvc: Note:  This is in the mathjax2 legacy code, which (a) will not be in the final version, and (b) does not use the same formatting restrictions that are in the v3 code.  So I think we can leave this one.

PR #28, opened 2017-08-08
Add utility for constructing HTML nodes

dpvc:
This is a utility for creating HTML nodes in a particular document.  The document is cached, since HTML elements are created via `document.createElement()` and `document.createTextNode()`, so we need to know the document in order to create HTML elements.  The output jax will instantiate an HTMLNodes object for the document it is working on in order to generate the needed HTML output.

### DIFF
--- a/mathjax3-ts/util/HTMLNodes.ts
+++ b/mathjax3-ts/util/HTMLNodes.ts
@@ -32,27 +32,14 @@ export class HTMLNodes {
     /*
      * The document in which the HTML nodes will be created
      */
-    protected _document: Document;
+    public document: Document;
 
     /*
      * @param{Document} document  The document in which the nodes will be created
      * @constructor
      */
     constructor(document: Document = null) {
-        this._document = document;
-    }
-
-    /*
-     * Set or get the document in which the elements are created
-     *
-     * @param{Document} document  The document in which to create elements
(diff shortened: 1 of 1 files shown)


mathjax3-ts/util/HTMLNodes.ts
```diff
@@ -0,0 +1,104 @@
...
+        this._document = document;
+    }
+
+    /*
+     * Set or get the document in which the elements are created
+     *
+     * @param{Document} document  The document in which to create elements
+     * @return{Document}  The document currently used for creating elements
+     */
+    public document(document: Document = null) {
```
zorkow: To be consistent with the rest of the code base this should be done with set and get.

mathjax3-ts/util/HTMLNodes.ts
```diff
@@ -0,0 +1,104 @@
...
+/*****************************************************************/
+/*
+ *  Implements the HTMLNodes class for creting HTML elements
+ */
+
+export class HTMLNodes {
+    /*
+     * The document in which the HTML nodes will be created
+     */
+    protected _document: Document;
```
dpvc: It was because there was a `document()` method to set/get it.  I will change it to just be a public value.

PR #29, opened 2017-08-10
Beginnings of CHTML output jax

dpvc:
This is the main framework for the CHTML output jax.  It implements a "Wrapper" object that is the solution to not being able to add methods to the MmlNode objects.  It is essential a shadow hierarchy that mirrors the MmlNode hierarchy, so that you can inherit methods, but not modify the MmlNode classes themselves.  I added generic `Wrapper` and `WrapperFunction` classes in `core/Tree`, and the CHTML-specific ones in `output/chtml`. 

The `CHTMLWrapper` is the base class for CHTML output, and includes all the generic functionality needed to produce CHTML output.  The node-specific methods are in the `output/chtml/Wrappers` directory.  There are only three at the moment, as examples (and the mfrac is just a preliminary version, as more needs to be done there).

This doesn't include the creation of the CSS needed to handle the output yet (I use a fixed CSS file for testing, but it is not included here), but this does produce the proper HTML.  If you want to test actual output, I can send the CSS file(s) and a load.html replacement that includes them.

So this is still preliminary and a work in progress, but since you asked for smaller chunks to review, I thought it would be easier to get through things at this point before all the details are in place.  This does include the HTMLNodes file that is part of a separate PR, so you can ignore that here and deal with it only in the other PR.

### DIFF
--- a/mathjax3-ts/output/chtml.ts
+++ b/mathjax3-ts/output/chtml.ts
@@ -28,7 +28,7 @@ import {MathItem} from '../core/MathItem.js';
 import {MmlNode} from '../core/MmlTree/MmlNode.js';
 import {HTMLNodes} from '../util/HTMLNodes.js';
 import {CHTMLWrapperFactory} from './chtml/WrapperFactory.js';
-import {BIGDIMEN, percent} from '../util/lengths.js';
+import {percent} from '../util/lengths.js';
 
 /*****************************************************************/
 /*
--- a/mathjax3-ts/output/chtml.ts
+++ b/mathjax3-ts/output/chtml.ts
@@ -52,7 +52,7 @@ export class CHTML extends AbstractOutputJax {
     public factory: CHTMLWrapperFactory;
 
     /*
-     * The MatahDocument for the math we find
+     * The MathDocument for the math we find
      * and the MathItem currently being processed
      */
     public document: MathDocument;
--- a/mathjax3-ts/output/chtml.ts
+++ b/mathjax3-ts/output/chtml.ts
@@ -22,10 +22,13 @@
  */
 
 import {AbstractOutputJax} from '../core/OutputJax.js';
-import {LegacyCHTML} from '../../mathjax2/output/CommonHTML.js';
 import {OptionList} from '../util/Options.js';
 import {MathDocument} from '../core/MathDocument.js';
 import {MathItem} from '../core/MathItem.js';
+import {MmlNode} from '../core/MmlTree/MmlNode.js';
+import {HTMLNodes} from '../util/HTMLNodes.js';
+import {CHTMLWrapperFactory} from './chtml/WrapperFactory.js';
+import {BIGDIMEN, percent} from '../util/lengths.js';
 
 /*****************************************************************/
 /*
@@ -36,41 +39,110 @@ export class CHTML extends AbstractOutputJax {
 
     public static NAME: string = 'CHTML';
     public static OPTIONS: OptionList = {
-        ...AbstractOutputJax.OPTIONS
+        ...AbstractOutputJax.OPTIONS,
+        scale: 1,                      // Global scaling factor for all expressions
+        skipAttributes: {},            // RFDa and other attributes NOT to copy to CHTML output
+        CHTMLWrapperFactory: null      // The CHTMLWrapper factory to use
     };
 
     /*
-     * @override
+     *  Used to store the HTMLNodes factory and the CHTMLWraper factory.
      */
-    public typeset(math: MathItem, html: MathDocument) {
-        return LegacyCHTML.Typeset(math, html);
+    public nodes: HTMLNodes;
+    public factory: CHTMLWrapperFactory;
+
+    /*
+     * The MatahDocument for the math we find
+     * and the MathItem currently being processed
+     */
+    public document: MathDocument;
+    public math: MathItem;
+
+    /*
+     * Get the WrapperFactory and connect it to this output jax
+     * Get the HTMLNodes instance
+     *
+     * @param{OptionList} options  The configuration options
+     * @constructor
+     */
+    constructor(options: OptionList = null) {
+        super(options);
+        this.factory = this.options.CHTMLWrapperFactory || new CHTMLWrapperFactory();
+        this.factory.chtml = this;
+        this.nodes = new HTMLNodes();
     }
 
     /*
-     * Handle an escaped character
-     *  (put it in a span so that it won't be a delimiter if
-     *  the page is typeset again).
+     * Save the math document and the math item
+     * Set the document where HTMLNodes will be created
+     * Recusrively set the TeX classes for the nodes
+     * Create the container mjx-chtml node
+     * Create the CHTML output for the root MathML node in the container
      *
      * @override
      */
+    public typeset(math: MathItem, html: MathDocument) {
+        this.document = html;
+        this.math = math;
+        this.nodes.document(html.document);
+        math.root.setTeXclass(null);
+        let node = this.html('mjx-chtml', {'class': 'MathJax_CHTML MJX-TEX'});
+        const scale = math.metrics.scale * this.options.scale;
+        if (scale !== 1) {
+            node.style.fontSize = percent(scale);
+        }
+        this.toCHTML(math.root, node);
+        return node;
+    }
+
+    /*
+     * @override
+     */
     public escaped(math: MathItem, html: MathDocument) {
-        let span = html.document.createElement('span');
-        span.appendChild(html.document.createTextNode(math.math));
-        return span;
+        this.nodes.document(html.document);
+        return this.html('span', {}, [this.text(math.math)]);
     }
 
     /*
      * @override
      */
     public getMetrics(html: MathDocument) {
-        return LegacyCHTML.GetMetrics(html);
+
     }
 
     /*
      * @override
      */
     public styleSheet(html: MathDocument) {
-        return LegacyCHTML.StyleSheet(html);
+        return null as Element;
+    }
+
+    /*
+     * @param{MmlNode} node  The MML node whose HTML is to be produced
+     * @param{HTMLElement} parent  The HTML node to contain the HTML
+     */
+    public toCHTML(node: MmlNode, parent: HTMLElement) {
+        return this.factory.wrap(node).toCHTML(parent);
+    }
+
+    /*
+     * @param{string} type  The type of HTML node to create
+     * @param{OptionList} def  The properties to set on the HTML node
+     * @param{Node[]} content  Array of child nodes to set for the HTML node
+     *
+     * @return{HTMLElement} The newly created HTML tree
+     */
+    public html(type: string, def: OptionList = {}, content: Node[] = []) {
+        return this.nodes.node(type, def, content);
+    }
+
+    /*
+     * @param{string} text  The text string for which to make a text node
+     *
+     * @return{HTMLElement}  A text node with the given text
+     */
+    public text(text: string) {
+        return this.nodes.text(text);
     }
 
 }
--- a/mathjax3-ts/output/chtml.ts
+++ b/mathjax3-ts/output/chtml.ts
@@ -84,7 +84,7 @@ export class CHTML extends AbstractOutputJax {
     public typeset(math: MathItem, html: MathDocument) {
         this.document = html;
         this.math = math;
-        this.nodes.document(html.document);
+        this.nodes.document = html.document;
         math.root.setTeXclass(null);
         let node = this.html('mjx-chtml', {'class': 'MathJax_CHTML MJX-TEX'});
         const scale = math.metrics.scale * this.options.scale;
@@ -99,7 +99,7 @@ export class CHTML extends AbstractOutputJax {
      * @override
      */
     public escaped(math: MathItem, html: MathDocument) {
-        this.nodes.document(html.document);
+        this.nodes.document = html.document;
         return this.html('span', {}, [this.text(math.math)]);
     }
 
--- a/mathjax3-ts/output/chtml/BBox.ts
+++ b/mathjax3-ts/output/chtml/BBox.ts
@@ -1,0 +1,156 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2017 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  Implements a bounding-box object and operations on it
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+import {BIGDIMEN, length2em} from '../../util/lengths.js';
+
+/*
+ *  CSS styles that affect BBoxes
+ */
+export const BBoxstyleAdjust = [
+    ['borderTopWidth', 'h'],
+    ['borderRightWidth', 'w'],
+    ['borderBottomWidth', 'd'],
+    ['borderLeftWidth', 'w', 0],
+    ['paddingTop', 'h'],
+    ['paddingRight', 'w'],
+    ['paddingBottom', 'd'],
+    ['paddingLeft', 'w', 0]
+];
+
+/*
+ *  The data used to initialie a BBox
+ */
+export type BBoxData = {
+    w?: number,
+    h?: number,
+    d?: number
+};
+
+/*****************************************************************/
+/*
+ *  The BBox class
+ */
+
+export class BBox {
+    /*
+     *  These are the data stored for a bounding box
+     */
+    public w: number;
+    public h: number;
+    public d: number;
+    public x: number;
+    public y: number;
+    public scale: number;
+    public rscale: number; // scale relative to the parent's scale
+    public L: number;      // extra space on the left
+    public R: number;      // extra space on the right
+    public pwidth: string; // percentage width (for tables)
+
+    /*
+     * @return{BBox}  A BBox initialized to zeros
+     */
+    public static zero() {
+        return new BBox({h: 0, d: 0, w: 0});
+    }
+
+    /*
+     * @return{BBox}  A BBox with height and depth not set
+     */
+    public static empty() {
+        return new BBox();
+    }
+
+    /*
+     * @param{BBoxData} def  The data with which to initialize the BBox
+     *
+     * @return{BBox}  The newly created BBox
+     *
+     * @constructor
+     */
+    constructor(def: BBoxData = {w: 0, h: -BIGDIMEN, d: -BIGDIMEN}) {
+        this.w = ('w' in def ? def.w : 0);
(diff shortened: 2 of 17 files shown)


pkra:
>  If you want to test actual output, I can send the CSS file(s) and a load.html replacement that includes them.

Yes, please!

mathjax3-ts/output/chtml.ts
```diff
@@ -36,41 +39,110 @@ export class CHTML extends AbstractOutputJax {
...
     };
 
     /*
+     *  Used to store the HTMLNodes factory and the CHTMLWraper factory.
+     */
+    public nodes: HTMLNodes;
+    public factory: CHTMLWrapperFactory;
+
+    /*
+     * The MatahDocument for the math we find
```
zorkow: typo

mathjax3-ts/output/chtml/BBox.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    /*
+     * @param{BBoxData} def  The data with which to initialize the BBox
+     *
+     * @return{BBox}  The newly created BBox
+     *
+     * @constructor
+     */
+    constructor(def: BBoxData = {w: 0, h: -BIGDIMEN, d: -BIGDIMEN}) {
+        this.w = ('w' in def ? def.w : 0);
```
zorkow: def.w || 0; 
should be enough.

mathjax3-ts/output/chtml/BBox.ts
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /*
+     * @param{BBox} cbox  The bounding box to use to overwrite this one
+     */
+    public updateFrom(cbox: BBox) {
+        this.h = cbox.h;
+        this.d = cbox.d;
+        this.w = cbox.w;
+        if (cbox.pwidth !== '') {
```
zorkow: (cbox.pwidth) 
should be enough.

mathjax3-ts/output/chtml/TeX.ts
```diff
@@ -0,0 +1,57 @@
...
+    sup_drop:         .386108,
+    sub_drop:         .05,
+    delim1:          2.39,
+    delim2:          1.0,
+    axis_height:      .25,
+    rule_thickness:   .06,
+    big_op_spacing1:  .111111,
+    big_op_spacing2:  .166666,
+    big_op_spacing3:  .2,
+    big_op_spacing4:  .45, // .6,  // better spacing for under arrows and braces
```
zorkow: Is that comment still relevant? Should this entry be parameterised then?

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+        this.getVariant();
+        this.getScale();
+        this.childNodes = node.childNodes.map((child: Node) => {
+            return this.wrap(child as MmlNode);
+        });
+    }
+
+    /*
+     * @param{MmlNode} node  The node to the wrapped
+     * @param{CHTMLWrapper} parent  The parent wrapped node
```
zorkow: I am not sure what this is. Is this the wrapped parent node of node? Why do we have to supply it explicitly then?

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+        }
+        for (const child of this.childNodes) {
+            child.toCHTML(chtml, WHD);
+        }
+    }
+
+    /*******************************************************************/
+    /*
+     * Return the wrapped node's bounding box, either the cached one, if it exists,
+     *   or computed direcctly if not.
```
zorkow: typo

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+     *
+     * @param{boolean} save  Whether to cache the bbox or not (used for stretchy elements)
+     * @return{BBox}  The computed bounding box
+     */
+    public getBBox(save: boolean = true) {
+        if (!this.bboxComputed) {
```
zorkow: Rewrite to 
if (this.bboxComputed) {
return this.bbox;
}
....

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+
+    /*******************************************************************/
+
+    /*
+     * Add the style attribute, but remove any font-related styles
+     *   (since these are handled separately by the variant)
+     */
+    protected getStyles() {
+        const styleString = this.node.attributes.getExplicit('style') as string;
+        if (styleString) {
```
zorkow: rewrite to 
if (!styleString) {
return;
}
to save one layer of indentation.

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+                }
+            }
+        }
+    }
+
+    /*
+     * Get the mathvariant (or construct one, if needed).
+     */
+    protected getVariant() {
+        if (this.node.isToken) {
```
zorkow: ditto.

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+                    const style = this.removedStyles;
+                    if (style.fontFamily) values.family = style.fontFamily;
+                    if (style.fontWeight) values.weight = style.fontWeight;
+                    if (style.fontStyle)  values.style  = style.fontStyle;
+                }
+                if (values.fontfamily) values.family = values.fontfamily;
+                if (values.fontweight) values.weight = values.fontweight;
+                if (values.fontstyle)  values.style  = values.fontstyle;
+                if (values.weight && values.weight.match(/^\d+$/)) {
+                    values.weight = (parseInt(values.weight) > 600 ? 'bold' : 'normal');
```
zorkow: Superfluous parentheses.

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+     * as a property of the node (e.g., no "onlick", etc.).  If a name in the
+     * skipAttributes object is set to false, then the attribute WILL be copied.
+     */
+    protected handleAttributes() {
+        const attributes = this.node.attributes;
+        const globals = attributes.getAllGlobals();
+        const skip = CHTMLWrapper.skipAttributes;
+        for (const name of attributes.getExplicitNames()) {
+            if (skip[name] === false || (!globals.hasOwnProperty(name) && !skip[name] &&
+                                         typeof((this.chtml as {[name: string]: any})[name]) === 'undefined')) {
```
zorkow: I am not sure I fully understand all these cases. Can you spread it out a bit and have some inline comments?

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+    /*
+     * @param{string} text  The text to turn into unicode locations
+     * @return{number[]}  Array of numbers represeting the string's unicode charater positions
+     */
+    protected unicodeChars(text: string) {
+        return unicodeChars(text);
+    }
+
+    /*
+     * @param{number} n  A unicode code point to be converred to a character reference for use with the
```
zorkow: typo

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -0,0 +1,604 @@
...
+    protected length2em(length: Property, size: number = 1, scale: number = null) {
+        if (scale === null) {
+            scale = this.bbox.scale;
+        }
+        return LENGTHS.length2em(length as string, size, scale, this.metrics.em);
+    }
+
+    /*
+     * @param{string} text  The text to turn into unicode locations
+     * @return{number[]}  Array of numbers represeting the string's unicode charater positions
```
zorkow: typo

mathjax3-ts/output/chtml/Wrappers.ts
```diff
@@ -0,0 +1,34 @@
+/*************************************************************
```
zorkow: Just for clarification: Wrappers combines all wrapper implementations for the different node types.

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+    }
+
+    /*
+     * @override
+     */
+    public computeBBox() {
+        //
+        // Fake a bbox for now (eventually, this will get looked up in the font table)
```
zorkow: Any reason for those values?

mathjax3-ts/output/chtml/Wrappers/mfrac.ts
```diff
@@ -0,0 +1,75 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  Implements the CHTMLmfracr wrapper for the MmlMfrac object
```
zorkow: Typo? CHTMLmfrac?

mathjax3-ts/output/chtml/Wrappers/mfrac.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /*
+     * @override
+     */
+    public computeBBox() {
+        const bbox = BBox.empty();
+        const nbox = this.childNodes[0].getBBox();
+        const dbox = this.childNodes[1].getBBox();
+        const pad = (this.node.getProperty('withDelims') as boolean ? this.TeX.nulldelimiterspace : 0);
```
zorkow: Superfluous parens

mathjax3-ts/util/HTMLNodes.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @fileoverview  Support functions for creating DOM elements
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+import {OptionList} from './Options.js';
+
+/*****************************************************************/
+/*
+ *  Implements the HTMLNodes class for creting HTML elements
```
zorkow: typo

mathjax3-ts/util/HTMLNodes.ts
```diff
@@ -0,0 +1,91 @@
...
+    /*
+     * @param{Document} document  The document in which the nodes will be created
+     * @constructor
+     */
+    constructor(document: Document = null) {
+        this.document = document;
+    }
+
+    /*
+     * @param{string} type     The tag name of rhe HTML node to be created
```
zorkow: typo

mathjax3-ts/util/HTMLNodes.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    /*
+     * @param{HTMLElement} node  The HTML element whose properties are to be set
+     * @param{OptionList} def    The properties to set on that node
+     */
+    public setProperties(node: HTMLElement, def: OptionList) {
+        if (def.style) {
+            let style = node.style as OptionList;
+            for (let key of Object.keys(def.style)) {
+                style[key.replace(/-([a-z])/g, (m, c) => c.tpUpperCase())] = def.style[key];
```
zorkow: toUpperCase !

mathjax3-ts/util/lengths.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  Utility functions for handing dimensions (lengths)
```
zorkow: typo

mathjax3-ts/util/lengths.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+/*
+ *  A very large number
+ */
+
```
zorkow: remove line

mathjax3-ts/util/lengths.ts
```diff
@@ -0,0 +1,141 @@
...
+
+/*
+ * @param{Property} length  A dimension (giving number and units) to be converted to ems
+ * @param{number} size      The default size of the dimension (for percentage values)
+ * @param{number} scale     The current scaling factor (to handle absolute units)
+ * @return{number}          The dimension converted to ems
+ */
+export function length2em(length: string, size: number = 0, scale: number = 1, em: number = 16) {
+    if (length === '' || length == null) return size;
```
zorkow: Bracing around if statements

mathjax3-ts/util/string.ts
```diff
@@ -23,16 +23,39 @@
...
  */
 export function quotePattern(text: string) {
     return text.replace(/([\^$(){}+*?\-|\[\]\:\\])/g, '\\$1');
 }
 
+/*
+ * Convert a UTF-8 string to an array of unicode code points
+ *
+ * @param{string} text  The string to be turned into unicode positions
+ * @return{number[]}  Array of numbers represeting the string's unicode charater positions
```
zorkow: typo

mathjax3-ts/output/chtml/TeX.ts
dpvc: The original .6 is from the TeX fonts themselves, and the comment is to mention that we are using a different value from TeX.  Since I haven't done the munderover element yet, I'm not sure if it is still needed, but it is from the v2 CHTML output, so I suspect it will be.

I'm not sure what you mean by parameterizing it, here.

mathjax3-ts/output/chtml/Wrappers.ts
dpvc: Wrappers is an object that indexes all the wrappers for the different node types, yes.  It is used by an instance of a  WrapperFactory in order to know what types of wrappers it can create.

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
dpvc: They are just temporary (and in fact have been replaced by the actual values in a later pull request already).

They are roughly the height of a capital letter, the depth of a letter with a descender, and the average width of a character (in ems).

mathjax3-ts/output/chtml/Wrappers/mfrac.ts
dpvc: Yes, thanks.

mathjax3-ts/output/chtml/Wrappers/mfrac.ts
dpvc: I have always preferred to wrap `.. ? .. : ..` in parens, especially for an assignment, because the precedences of the operators are not obvious.  This makes is clear that it is `x = (...)` rather than `(x = ...) ? ... : ...`.

mathjax3-ts/util/HTMLNodes.ts
dpvc: Argh!

mathjax3-ts/output/chtml/Wrapper.ts
dpvc: Yes, it is the wrapped parent node.  Because there is no connection from the node to the wrapper (just from the wrapper to the node), we need to store a pointer to the wrapped parent node in order to be able to access the wrapper for the parent of node.

Originally, wrapping was going to be done at the time a wrapped node was needed, but now that the entire tree is wrapped in the `constructor()`, I suppose the parents could be set at that point rather than here.  Would you prefer that?

mathjax3-ts/output/chtml/Wrapper.ts
dpvc: But this introduces an extra operation (the negation), which impacts performance (slightly).  Since `getStyles()` is used for every element being displayed, I would rather have the indentation than the extra operation.

dpvc: I also prefer there to be only one exit to a function, when possible, rather than multiple ones.

mathjax3-ts/output/chtml/Wrapper.ts
dpvc: ditto.

mathjax3-ts/output/chtml/Wrapper.ts
dpvc: Ditto.

mathjax3-ts/output/chtml/Wrapper.ts
dpvc: This is changed in a later PR, so let's put off these edits until then, in order to avoid merge conflicts in the future.  I'll add some comments then.

mathjax3-ts/util/string.ts
dpvc: Two in fact!

dpvc:
> I am convinced what we have is more of a decorator pattern than a sort of adaptor. Anyway, that's nitpicking.

Does that mean the the name should be `Decorator` rather than `Wrapper`?

mathjax3-ts/output/chtml/Wrapper.ts
zorkow: Testing for bit existence or non-existence should not really impact performance.
I would rather keep nesting down to improve code readability. It's a "philosophical" thing.

In general, my philosophy is to concentrate on the main logic in a code block and dispense with the special cases beforehand (return immediately or continue in a loop). That generally keeps nesting down and it often makes it easier to see that one has covered all special cases.

mathjax3-ts/output/chtml/Wrapper.ts
zorkow: OK.

mathjax3-ts/output/chtml/Wrappers/mfrac.ts
zorkow: Fine, we can adopt that as a style.

mathjax3-ts/output/chtml/TeX.ts
zorkow: I.e., have a parameter UNDER_ARROW_SPACING that can be changed whenever necessary.
Not that it is necessary.

mathjax3-ts/output/chtml/Wrapper.ts
dpvc: OK, I'll do it if I can use

    if (!styleString) return;

The extra bracing and line breaks make it harder for me to read, so if this is about readability, I would like the one-line version.

mathjax3-ts/output/chtml/TeX.ts
dpvc: It shouldn't need to change.  The comment is just indicating that it is not the same as the original value in the TeX fonts.  My feeling is that this *is* already parameterized, as `big_op_spacing4` (the TeX name for this font parameter), and this is where you would change it if you wanted a different value.  I'm not sure what you gain by having another name for it (and why you wouldn't do the same for all the others if you did).

mathjax3-ts/output/chtml/Wrapper.ts
zorkow: Sounds good to me.

zorkow:
After the "nesting" changes.

dpvc:
> After the "nesting" changes.

Thanks!  I'll take care of it.

dpvc:
OK, I've made the nesting changes, and will merge it (after resolving the conflict involving a misspelling in a comment).

PR #30, opened 2017-08-15
Minor fixes

dpvc:
This is a small set of changes to existing files:

* Fixes a problem where the default attributes were not being set in `core/MmlTree/Attributes.js` (due to changes in a previous commit).

* Adds a `isSet()` method to `core/MmlTree/Attributes.js` to be able to test if an attribute is present explicitly either on the node itself or on a parent `mstyle` or `math` node (as opposed to being set via default values).

* Makes the `mstyle` default for `scriptminsize` match that for the `math` node.

* Puts the em and ex values in the right order, and makes them slightly larger (so `scriptminsize` will not come into play).  These are fake values for testing in any case.

### DIFF
--- a/mathjax3-ts/handlers/html/HTMLDocument.ts
+++ b/mathjax3-ts/handlers/html/HTMLDocument.ts
@@ -228,7 +228,7 @@ export class HTMLDocument extends AbstractMathDocument {
     public TestMath(text: string, display: boolean = true) {
         if (!this.processed['TestMath']) {
             let math = new HTMLMathItem(text, this.inputJax[0], display);
-            math.setMetrics(6, 14, 1000000, 1000000, 1);
+            math.setMetrics(16, 8, 1000000, 1000000, 1);
             this.math.push(math);
             this.processed['TestMath'] = true;
         }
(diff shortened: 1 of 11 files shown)


dpvc:
Note:  This PR is being targeted at the `chtml` branch rather than `master`, since `chtml` hasn't yet been reviewed and merged.

mathjax3-ts/handlers/html/HTMLDocument.ts
```diff
@@ -228,7 +228,7 @@ export class HTMLDocument extends AbstractMathDocument {
     public TestMath(text: string, display: boolean = true) {
         if (!this.processed['TestMath']) {
             let math = new HTMLMathItem(text, this.inputJax[0], display);
-            math.setMetrics(6, 14, 1000000, 1000000, 1);
+            math.setMetrics(16, 8, 1000000, 1000000, 1);
```
zorkow: Do we want to name these magic numbers? E.g., defaultEm, default,Ex

dpvc: I'm not sure I understand how.  Do you mean by passing an object instead?  They are, of course, named in the definition of `setMetrics()` in [`core/MathItem.ts`](https://github.com/mathjax/mathjax-v3/blob/master/mathjax3-ts/core/MathItem.ts#L172).  Do you have something else in mind?

(Also, this particular function is temporary, and is only here for testing purposes, so these particular values are not going to be in the final version, since this function will be removed.)

PR #32, opened 2017-08-21
Finish stretchy character support and add bbox support

dpvc:
This PR finishes the stretchy character support (by adding the multi-character support).  It also adds a font data object and uses that to hold the information about delimiters, character bounding boxes, and TeX font parameters for the font, so the TeX font parameters have been moved from where they were.

This also implements bbox computations for TextNodes and mo nodes (which should make the bonding boxes correct for mrow as well, but mfrac still needs work).

This PR is targeted to the stretchy-chars branch, so the you don't see all those original files as well.  If you want, I can merge this into stretchy-chars and you can review them as a whole (but I was trying to do incremental updates to keep the reviews small).  Since this does make changes to stretchy-chars, it might be easier not to have to look at the originals but only final versions at this point.  Let me know how you want to handle it.


dpvc:
Just realized I forgot to tslint and add some of the jsdoc comments, so hold off on this until I do that.  Sorry!

dpvc:
OK, I've linted and added the missing jsdoc comments.  So this should be ready to review.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    /*
+     * Create a collection of variants
+     *
+     * @param{string[][]} variants  Array of [name, inherit?, link?] values for
+     *                              the variants to define
+     */
+    public createVariants(variants: string[][]) {
+        for (const variant of variants) {
+            this.createVariant.apply(this, variant);
```
zorkow: Is there a reason why you are using .apply here?

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+    public params: FontParameters;
+
+    /*
+     * Copies the data from the defaults to the instance
+     *
+     * @constructor
+     */
+    constructor() {
+        let CLASS = (this.constructor as typeof FontData);
+        this.params = Object.assign({}, CLASS.defaultParams);
```
zorkow: Use spread: 
this.params = {...Class.defaultParams};
(At least I believe that was the syntax.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    /*
+     * Defines new character data in a given variant
+     *
+     * @param{srtring} name   The variant for these characters
+     * @param{CharMap} chars  The characters to define
+     */
+    public defineChars(name: string, chars: CharMap) {
+        let variant = this.variant[name];
+        Object.assign(variant.chars, chars);
```
zorkow: dito

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+        for (const link of variant.linked) {
+            Object.assign(link, chars);
```
zorkow: dito

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+        }
+    }
+
+    /*
+     * Defines strety delimiters
+     *
+     * @param{DelimiterMap} delims  The delimiters to define
+     */
+    public defineDelimiters(delims: DelimiterMap) {
+        Object.assign(this.delimiters, delims);
```
zorkow: dito

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /*
+     * @param{number} n  The delimiter character number whose variant is needed
+     * @param{number} i  The index in the size array of the size whose variant is needed
+     * @return{string}   The variant of the i-th size for delimiter n
+     */
+    public getSizeVariant(n: number, i: number) {
+        if (this.delimiters[n].variants) {
+            i = this.delimiters[n].variants[i];
```
zorkow: This can be undefined. Is that a problem?

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+        }
+        return this.sizeVariants[i];
```
zorkow: dito

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+     *
+     * @param{DelimiterMap} delims  The delimiters to define
+     */
+    public defineDelimiters(delims: DelimiterMap) {
+        Object.assign(this.delimiters, delims);
+    }
+
+    /*
+     * @param{number} n  The delimiter character number whose data is desired
+     * @return{DelimiterData}  The data for that delimiter (or null)
```
zorkow: or undefined not null

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+        if (this.delimiters[n].variants) {
+            i = this.delimiters[n].variants[i];
+        }
+        return this.sizeVariants[i];
+    }
+
+    /*
+     * @param{string} name  The variant whose character data is being querried
+     * @param{number} n     The unicode number for the character to be found
+     * @return{CharData}    The data for the given character (or null)
```
zorkow: dito

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+     */
+    public getChar(name: string, n: number) {
+        return this.variant[name].chars[n];
+    }
+
+    /*
+     * @param{string} name   The name of the variant whose data is to be obtained
+     * @return{VariantData}  The data for the requested variant (or null)
```
zorkow: dito

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+        return this.sizeVariants[i];
+    }
+
+    /*
+     * @param{string} name  The variant whose character data is being querried
+     * @param{number} n     The unicode number for the character to be found
+     * @return{CharData}    The data for the given character (or null)
+     */
+    public getChar(name: string, n: number) {
+        return this.variant[name].chars[n];
```
zorkow: Case that variant[name] does not exist missing.

zorkow:
I found some documentation on my proposed changes:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html
under "spread and rest", as well as here:
https://github.com/tc39/proposal-object-rest-spread/blob/master/Spread.md

mathjax3-ts/output/chtml/FontData.ts
dpvc: Yes, that's right.  I'll change it.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+     *
+     * @param{string} name     The new variant to create
+     * @param{string} inherit  The variant to use if a character is not in this one
+     * @param{string} link     A variant to search before the inherit one (but only
+     *                           its top-level object).
+     */
+    public createVariant(name: string, inherit: string = null, link: string = null) {
+        let variant = {
+            linked: [] as CharMap[],
+            chars: (inherit ? Object.create(this.variant[inherit].chars) : {}) as CharMap
```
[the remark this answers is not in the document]
dpvc: This is not a clone, it creates an object having a prototype of the given object.  This is a lookup chain.  This allows a variant to have its own data (for for example for bold) but fall back on another variant (e.g., normal).  That way, the bold variant doesn't have to repeat all the data in the normal variant.

It is important to have inheritance of variants in this way, since, for example, one can do

```
<mstyle mathvariant="fraktur">
  <mi>x</mi>
  <mo>&gt;</mo>
  <mi>y</mi>
</mstyle>
```
so all three inner elements get `mathvariant="fraktur'`.  But there is no glyph for a plus sign in the MathJax Fraktur font, so it should default back to the normal version.  The prototype chain is what makes that work.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+        };
+        if (link && this.variant[link]) {
+            Object.assign(variant.chars, this.variant[link].chars);
```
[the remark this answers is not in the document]
dpvc: It is a merge, and the properties are the character data arrays (the mapping from unicode number to CharData array).

The reason for this extra link is that for a `mathvariant` like `bold-italic`, you want to inherit from both the `bold` and `italic` variants, but the prototype chain can only inherit from one.  So for `bold-italic`, we make an object that has a prototype consisting of a *copy* of the `bold` data, and add the `italic` data as the prototype chain.  (Since this *is* a copy, we keep a record of this link so that if `bold` is changed later, we update this copy.  That is not needed for the prototype chain, since the prototypes are the actual objects, not copies.)  We then use this bold-plus-italic object as the prototype chain for the `bold-italic` object

That means that `bold-italic` will first look in its own object for specifically bold-italic glyphs that are defined there, then in the copy of the bold glyphs (only its top level is copied, not is prototype chain), and then the specifically italic glyphs, and then the prototype chain for italics, which is the normal glyphs.  Effectively, this means `bold-italic` looks for bold-italic, then bold, then italic, then normal glyphs in order to find the given character.

Something like `bold-script` will link the `script` font and use `bold` as its prototype.  So `bold-script` will look for specifically bold-script glyphs, then `script` glyphs, then `bold` glyphs, then `normal` ones.

This is set up to reduce the amount of duplication in the lists, as well as the amount of copying that must be done when new glyphs are added to variants that are in the default variants (e.g., normal, bold, italic).

Because `variant.chars` is an object that may have a prototype already, we can't use spread to merge the linked chars.  E.g., 

```
variant.chars = {...this.variant[link].chars};
```
would lose the prototype to the inherited characters.

mathjax3-ts/output/chtml/FontData.ts
dpvc: Because variant is an array that contains the `name`, `inherit`, and `link` values for the `createVariant()` call, so they need to be broken out.  I suppose you could do

```
this.createVariant(variant[0], variant[1], variant[2]);
``` 
or
```
this.createVariant(...variant);
```
if you prefer.  (I admit that I don't always think of the spread operator; It is hard to unlearn past paradigms.)

mathjax3-ts/output/chtml/FontData.ts
dpvc: Note that `getSizeVariant()` is only called when the character is known to be stretchy, i.e., `this.delimiters[n]` is known to exist, so isn't undefined.  For a properly constructed `DelimiterData`, the `variants` array should be the same length as the `sizes` array, but you are correct that there is no guarantee of that.

One could, I suppose, use
```
i = this.delimiters[n].variants[i] || 0;
```
but I think I'd rather get an error when the `undefined` is used as a number in order to find out that the data are not correct rather than hiding it and getting a wrong glyph.

mathjax3-ts/output/chtml/FontData.ts
dpvc: Same here as above.  The `sizeVariants` array should be as large as the largest `sizes` array in any of the delimiters, but again, there is no guarantee that this is true.  One could do
```
return this.sizeVariants[i] || 'normal';
```
I suppose, but again, that is hiding an inconsistency in the font data that should be corrected, not hidden.

mathjax3-ts/output/chtml/FontData.ts
dpvc: Shouldn't happen, and if it does, it should produce an error (as it will here) so that the inconsistency in the data can be corrected.  This is a function that will be called for every character displayed, so any unnecessary overhead should be avoided.

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
```diff
@@ -51,13 +51,24 @@ export class CHTMLTextNode extends CHTMLWrapper {
...
-        // Fake a bbox for now (eventually, this will get looked up in the font table)
-        //
-        this.bbox.w = .5;
-        this.bbox.h = .8;
-        this.bbox.d = .2;
-        return this.bbox;
+        const variant = this.parent.variant;
+        let bbox = this.bbox;
+        if (variant === '-explicitFont') {
+            // FIXME:  measure this using DOM, if possible
```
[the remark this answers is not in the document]
dpvc: This is the location where the code to measure the characters using the DOM will go (or be linked in). The comment is there to indicate that there is more coming in the future.  If it would make you happier, I can add a call to something like `this.getDomBBox()` at this point, and make an empty `getDomBBox()` method.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -0,0 +1,328 @@
...
+            this.defineChars(name, CLASS.defaultChars[name]);
+        }
+    }
+
+    /*
+     * Creates the data structure for a variant (an object with prototype chain
+     *   that includes a copy of the linked variant, and then the inherit variant chain.
+     *   The linked copy is updated automatically when the link variant is modified.
+     *   (The idea is to be able to have something like bold-italic inherit from both
+     *   bold and intalic by having the prototype chain include a copy of bold plus
```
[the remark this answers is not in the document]
dpvc: I don't see it.  Can you be more specific?

mathjax3-ts/output/chtml/FontData.ts
dpvc: It turns out Typescript doesn't like the `this.createVariant(...variant)` form, so I used the other more verbose one.

mathjax3-ts/output/chtml/FontData.ts
dpvc: Again, using spread would destroy the prototype chain that is part of `variant.chars`.  Also, since `variant.chars` may not be empty, you would have to use something like
```
variant.chars = {...variant.chars, ...chars};
```
which would mean you are copying and recreating the object every time `defineChars()` is called.  Adding the new characters into the existing object is more efficient, and is required in order to keep the inheritance intact.

mathjax3-ts/output/chtml/FontData.ts
dpvc: ditto

mathjax3-ts/output/chtml/FontData.ts
dpvc: While there is no inheritance in this case, the `delimiters` may not be empty, and using spread would require copying and recreating the object.  There is no need for that.  This is what `Object.assign()` is for.

mathjax3-ts/output/chtml/Wrappers/mo.ts
```diff
@@ -65,44 +36,112 @@ export class CHTMLmo extends CHTMLWrapper {
...
+
+    /*
+     * Create the HTML for a multi-character stretchy delimiter
+     *
+     * @param{HTMLElement} chtml  The parent element in which to put the delimiter
+     * @param{boolean} symmetric  Whether delimiter should be symmetric about the math axis
+     */
+    protected stretchHTML(chtml: HTMLElement, symmetric: boolean) {
+        const c = this.getText().charCodeAt(0);
+        const C = this.font.getDelimiter(c).stretch;
```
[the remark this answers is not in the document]
dpvc: OK, I changed `C` to `delim`.

dpvc:
I have addressed all the issues either by making the change, or explaining the reason for the original code (except for one typo that I couldn't find).  Can you take another look?

mathjax3-ts/output/chtml/FontData.ts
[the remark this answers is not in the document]
zorkow: Thanks! This is a great explanation, and I believe should be kept for posterity.
Could you please make a comment out of \S 2+3?

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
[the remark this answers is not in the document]
zorkow: Not necessary.

zorkow:
I approve. But somehow my approve button is gone.

zorkow:
It's probably the merge conflict. Anyway, some of your comments to my review should probably be kept for posterity. They are very good!

dpvc:
I've added the comments that you asked for into `FontData.js`.  Let me know if you want any of the others as well.

I've fixed the merge conflict, but I don't see an approval button anymore either, so I'm just going to merge it.

PR #33, opened 2017-08-24
Support more mml nodes

dpvc:
This PR implements several more MathML node types:  ms, mpadded, msqrt, mroot, semantics, annotation, annotation-xml, and (the MathJax-specific) TeXAtom nodes.  Note that some of the support for annotation-xml as the first element of semantics is not yet in place, and will be added later.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/mo.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/mo.ts
@@ -38,7 +38,7 @@ export class CHTMLmo extends CHTMLWrapper {
      * The font size that a stretched operator uses.
      * If -1, then stretch arbitrarily, and bbox gives the actual height, depth, width
      */
-    protected size: number = null;
+    public size: number = null;
 
     /*
      * @override
--- a/mathjax3-ts/output/chtml/Wrappers/mpadded.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/mpadded.ts
@@ -88,7 +88,7 @@ export class CHTMLmpadded extends CHTMLWrapper {
      */
     getDimens() {
         const values = this.node.attributes.getList('width', 'height', 'depth', 'lspace', 'voffset');
-        const bbox = (this.childNodes.length ? this.childNodes[0].getBBox() : BBox.zero());
+        const bbox = super.computeBBox();  // get unmodified bbox of children
         let {w, h, d} = bbox;
         let W = w, H = h, D = d, x = 0, y = 0;
         if (values.width !== '')   w = this.dimen(values.width, bbox, 'w', 0);
--- a/mathjax3-ts/output/chtml/Wrappers/mpadded.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/mpadded.ts
@@ -27,13 +27,6 @@ import {MmlMpadded} from '../../../core/MmlTree/MmlNodes/mpadded.js';
 import {MmlNode} from '../../../core/MmlTree/MmlNode.js';
 import {Property} from '../../../core/Tree/Node.js';
 
-/*
- * Needed in order to access BBox properties by variable
- */
-interface IBBox extends BBox {
-    [name: string]: number | string | Function;
-}
-
 /*****************************************************************/
 /*
  *  The CHTMLmpadded wrapper for the MmlMpadded object
@@ -63,8 +56,8 @@ export class CHTMLmpadded extends CHTMLWrapper {
             style.margin = this.em(dh) + ' 0 ' + this.em(dd);
         }
         //
-        // If there is a horizotnal or vertical shift,
-        //   use relative poisitioning to move the contents
+        // If there is a horizontal or vertical shift,
+        //   use relative positioning to move the contents
         //
         if (x || y) {
             style.position = 'relative';
@@ -112,7 +105,7 @@ export class CHTMLmpadded extends CHTMLWrapper {
     protected dimen(length: Property, bbox: BBox, d: string = '', m: number = null) {
         length = String(length);
         const match = length.match(/width|height|depth/);
-        const size = (match ? (bbox as IBBox)[match[0].charAt(0)] : (d ? (bbox as IBBox)[d] : 0)) as number;
+        const size = (match ? bbox[match[0].charAt(0) as (keyof BBox)] : (d ? bbox[d as (keyof BBox)] : 0)) as number;
         let dimen = (this.length2em(length, size) || 0);
         if (length.match(/^[-+]/) && d) {
             dimen += size;
--- a/mathjax3-ts/output/chtml/Wrappers/mpadded.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/mpadded.ts
@@ -1,0 +1,138 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2017 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  Implements the CHTMLmpadded wrapper for the MmlMpadded object
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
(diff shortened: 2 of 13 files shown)


pkra:
Should this PR be redirected to master now?

dpvc:
It could be, but I didn't want to change anything, since Volker could have a partial review done.  I think changing the target would interfere with that.  

If not, I'm happy to change it over.

zorkow:
I did start the review, but it only has one trivial comment so far. So we can do an experiment:
Change it over and we see what happens to my partial review.
Might be useful to know in the future.

zorkow:
Thanks. My comment is still there.

dpvc:
OK, I switched it over.

BTW, I noticed yesterday that in my copying-and-pasting to create new files, I forgot to change some of the comments indicating which MathML nodes were being handled, so I know about those, and have fixed them in a later PR.  So no need to mark them here.  Sorry about that.

dpvc:
> Thanks. My comment is still there.

Good to know.

mathjax3-ts/output/chtml/Wrappers/mpadded.ts
```diff
@@ -0,0 +1,138 @@
...
+        }
+        //
+        // If the height or depth changed, use margin to make the change
+        //
+        if (dh || dd) {
+            style.margin = this.em(dh) + ' 0 ' + this.em(dd);
+        }
+        //
+        // If there is a horizotnal or vertical shift,
+        //   use relative poisitioning to move the contents
```
zorkow: typo

mathjax3-ts/output/chtml/Wrappers/mo.ts
```diff
@@ -38,7 +38,7 @@ export class CHTMLmo extends CHTMLWrapper {
      * The font size that a stretched operator uses.
      * If -1, then stretch arbitrarily, and bbox gives the actual height, depth, width
      */
-    protected size: number = null;
+    public size: number = null;
```
zorkow: Should probably be private with a getter/setter.

mathjax3-ts/output/chtml/Wrappers/mpadded.ts
```diff
@@ -0,0 +1,138 @@
...
+import {CHTMLWrapper, StringMap} from '../Wrapper.js';
+import {BBox} from '../BBox.js';
+import {MmlMpadded} from '../../../core/MmlTree/MmlNodes/mpadded.js';
+import {MmlNode} from '../../../core/MmlTree/MmlNode.js';
+import {Property} from '../../../core/Tree/Node.js';
+
+/*
+ * Needed in order to access BBox properties by variable
+ */
+interface IBBox extends BBox {
```
zorkow: BBox is a class, so IBBox should not extend it.
Interfaces should extend other interfaces and classes should implement interfaces.
You should probably have an interface declaration in BBox.ts that BBox implements. 
Maybe for the time being it's also sufficient to declare IBBox as a type, e.g.:
export type IBBox [name: string]: number | string | Function;

mathjax3-ts/output/chtml/Wrappers/mpadded.ts
```diff
@@ -0,0 +1,138 @@
...
+     * @param{Property} length   The value to be converted to a length in ems
+     * @param{BBox} bbox         The bbox of the mpadded content
+     * @param{string} d          The default dimension to use for relative sizes ('w', 'h', or 'd')
+     * @param{number} m          The minimum value allowed for the dimension
+     * @return{number}           The final dimension in ems
+     */
+    protected dimen(length: Property, bbox: BBox, d: string = '', m: number = null) {
+        length = String(length);
+        const match = length.match(/width|height|depth/);
+        const size = (match ? (bbox as IBBox)[match[0].charAt(0)] : (d ? (bbox as IBBox)[d] : 0)) as number;
```
zorkow: If this is the only reason for IBBox, the type definition should be sufficient.
However, it's still very confusing. I believe we should have a enum for the properties h,d,w of a bbox and use this in the definition of  the BBoxData type.

mathjax3-ts/output/chtml/Wrappers/ms.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    /*
+     * Create a text wrapper with the given text;
+     *
+     * @param{string} text  The text for the wrapped element
+     * @return{CHTMLWrapper}   The wrapped text node
+     */
+    protected createText(text: string) {
+        const mmlFactory = (this.node as AbstractMmlNode).factory;
+        const node = this.wrap((mmlFactory.create('text') as TextNode).setText(text));
```
zorkow: I believe I've seen that pattern a few time before.  Maybe a convenience method createTextNode returning the correct type might be useful.

mathjax3-ts/output/chtml/Wrappers/msqrt.ts
```diff
@@ -0,0 +1,193 @@
...
+        this.addRoot(ROOT, root, sbox);
+        surd.toCHTML(SURD);
+        base.toCHTML(BASE);
+    }
+
+    /*
+     * Add root HTML (overridden in mroot)
+     *
+     * @param{HTMLElement} ROOT   The container for the root
+     * @param{CHTMLWrapper} root  The wreapped MML root content
```
zorkow: typo

mathjax3-ts/output/chtml/Wrappers/msqrt.ts
```diff
@@ -0,0 +1,193 @@
...
+     */
+    public computeBBox() {
+        const BBOX = BBox.empty();
+        const sbox = this.childNodes[this.surd].getBBox();
+        const bbox = new BBox(this.childNodes[this.base].getBBox());
+        const [p, q] = this.getPQ(sbox);
+        const [x] = this.getRootDimens(sbox);
+        const t = this.font.params.rule_thickness;
+        const H = bbox.h + q + t;
+        bbox.h += q + 2 * t;  // FIXME:  should take into accound minimums for this.px() used above
```
zorkow: typo

mathjax3-ts/output/chtml/Wrappers/msqrt.ts
```diff
@@ -0,0 +1,193 @@
...
+        this.combineRootBBox(BBOX, sbox);
+        BBOX.combine(sbox, x, H - sbox.h);
+        BBOX.combine(bbox, x + sbox.w, 0);
+        BBOX.clean();
+        return BBOX;
```
zorkow: I find bbox vs BBOX very confusing.

mathjax3-ts/output/chtml/Wrappers/mpadded.ts
dpvc: I worked out how to get around needing `IBBox`.  It was just to be able to access a `BBox`'s properties via a variable, `bbox[x]`, with would throw an error because `BBox` has no index structure.  But I can use `bbox[x as (keyof BBox)]` and that works.

mathjax3-ts/output/chtml/Wrappers/mpadded.ts
```diff
@@ -0,0 +1,138 @@
...
+            style.width = this.em(W + dw);
+        }
+        //
+        // If the height or depth changed, use margin to make the change
+        //
+        if (dh || dd) {
+            style.margin = this.em(dh) + ' 0 ' + this.em(dd);
+        }
+        //
+        // If there is a horizotnal or vertical shift,
```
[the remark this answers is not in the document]
dpvc: fixed.

mathjax3-ts/output/chtml/Wrappers/mpadded.ts
dpvc: Taken care of via `bbox[match[0].charAt(0) as (keyof BBox)]`.

mathjax3-ts/output/chtml/Wrappers/mo.ts
dpvc: Why the overhead, if it's just going to do the same thing as reading (or setting) the value anyway?  Should there come a need to do something algorithmic, we can add 

```
protected _size: number;

get size() {return this._size}
set size(size) {this._size = size; return size}
```

or whatever is needed at that time, and nothing external that uses `size` will have to change.  But why add that overhead now for no reason?

mathjax3-ts/output/chtml/Wrappers/ms.ts
dpvc: OK, good idea.  (It did appear twice, but will appear again in several other mml nodes.)  I've added it to the CHTMLWrapper class.

mathjax3-ts/output/chtml/Wrappers/msqrt.ts
dpvc: If one can't use different capitalizations, why does the language have case-sensitive names.  :-)

OK, I've changed `BBOX` to just `box`.  This will probably cause merge conflicts in a later PR, since I have modified this function once I had the actual display working and found out that the layout needed work.

dpvc:
OK, I've update the PR.  See if that passes muster.

PR #36, opened 2017-09-04
Create css styles for CommonHTML

dpvc:
This pull request adds support for generating the CSS styles needed for CHTML output.

The various Wrapper classes each can add their own CSS, and the main Wrapper class includes some generic CSS used by CHTML.  The TeX font object now creates the CSS for the characters and delimiters automatically (in `tex.ts`), and some font data was moved from `Wrapper.ts` and `FontData.ts` to `tex.ts`.

I took Volker's suggestion of moving the skew and italic-correction data from the `CharData` array to the options object, and have made a `CharOptions` for that, now that the data needed for the CSS creation has been worked out.  The `DelimiterData` has been updated to include some additional information that was needed for the CSS as well.  (There is still a bit more to come, as there need to be some additional offset values for the multi-character delimiters, but that is for the future).

The character and delimiter data itself has been moved from `tex.js` into separate files for the various math variants, and the data files have been generated (programmatically from the v2 CommonHTML data files).  Eventually, the v2 fonts will be replaced by new ones for v3 (based on LatinModern, I expect), and the data files will change, but for now, they use the old fonts.

The data files are in a separate PR, in order to make this PR easier to work with.  The data files are just data files, so should not need a review.

### DIFF
--- a/mathjax3-ts/output/chtml.ts
+++ b/mathjax3-ts/output/chtml.ts
@@ -53,7 +53,7 @@ export class CHTML extends AbstractOutputJax {
     };
 
     /*
-     *  Used to store the HTMLNodes factory, the CHTMLWraper factory,
+     *  Used to store the HTMLNodes factory, the CHTMLWrapper factory,
      *  the FontData object, and the CssStyles object.
      */
     public nodes: HTMLNodes;
--- a/mathjax3-ts/output/chtml.ts
+++ b/mathjax3-ts/output/chtml.ts
@@ -31,6 +31,7 @@ import {CHTMLWrapper} from './chtml/Wrapper.js';
 import {CHTMLWrapperFactory} from './chtml/WrapperFactory.js';
 import {FontData} from './chtml/FontData.js';
 import {TeXFont} from './chtml/fonts/tex.js';
+import {CssStyles} from './chtml/CssStyles.js';
 import {BIGDIMEN, percent} from '../util/lengths.js';
 import {BBox} from './chtml/BBox.js';
 
@@ -47,15 +48,18 @@ export class CHTML extends AbstractOutputJax {
         scale: 1,                      // Global scaling factor for all expressions
         skipAttributes: {},            // RFDa and other attributes NOT to copy to CHTML output
         CHTMLWrapperFactory: null,     // The CHTMLWrapper factory to use
-        font: null                     // The FontData object to use
+        font: null,                    // The FontData object to use
+        cssStyles: null                // The CssStyles object to use
     };
 
     /*
-     *  Used to store the HTMLNodes factory, the CHTMLWraper factory, and FontData object.
+     *  Used to store the HTMLNodes factory, the CHTMLWraper factory,
+     *  the FontData object, and the CssStyles object.
      */
     public nodes: HTMLNodes;
     public factory: CHTMLWrapperFactory;
     public font: FontData;
+    public cssStyles: CssStyles;
 
     /*
      * The MatahDocument for the math we find
@@ -83,6 +87,7 @@ export class CHTML extends AbstractOutputJax {
         this.factory = this.options.CHTMLWrapperFactory || new CHTMLWrapperFactory();
         this.factory.chtml = this;
         this.nodes = new HTMLNodes();
+        this.cssStyles = this.options.cssStyles || new CssStyles();
         this.font = this.options.font || new TeXFont();
     }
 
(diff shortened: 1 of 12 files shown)


mathjax3-ts/output/chtml.ts
```diff
@@ -47,15 +48,18 @@ export class CHTML extends AbstractOutputJax {
...
         skipAttributes: {},            // RFDa and other attributes NOT to copy to CHTML output
         CHTMLWrapperFactory: null,     // The CHTMLWrapper factory to use
-        font: null                     // The FontData object to use
+        font: null,                    // The FontData object to use
+        cssStyles: null                // The CssStyles object to use
     };
 
     /*
-     *  Used to store the HTMLNodes factory, the CHTMLWraper factory, and FontData object.
+     *  Used to store the HTMLNodes factory, the CHTMLWraper factory,
```
zorkow: typo

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -325,4 +359,15 @@ export class FontData {
         return this.variant[name];
     }
 
+    /*
+     * @param{number} n  A unicode code point to be converred to a character reference for use with the
```
zorkow: typo

mathjax3-ts/output/chtml/fonts/tex.ts
```diff
@@ -80,103 +118,332 @@ export class TeXFont extends FontData {
...
+     * @param{number} n          The unicode character being defined
+     * @param{CharData} data     The bounding box data and options for the character
+     */
+    protected addCharStyles(styles: StyleList, vclass: string, n: number, data: CharData) {
+        const [h, d, w, options] = data as [number, number, number, CharOptions];
+        const css: StyleData = {};
+        if (options.css) {
+            if (options.css & CSS.width)   css.width = this.em(w);
+            if (options.css & CSS.padding) css.padding = this.em(h) + ' 0 ' + this.em(d);
+            if (options.css & CSS.content) css.content = '"' + (options.c || this.char(n, true)) + '"';
```
zorkow: If styling.

mathjax3-ts/output/chtml/CssStyles.ts
```diff
@@ -0,0 +1,113 @@
...
+/*
+ * A list of selectors and their data (basically a stylesheet)
+ */
+export type StyleList = {
+    [selector: string]: StyleData;
+};
+
+/******************************************************************************/
+/*
+ * The CssStyles class (for managing a collection oof CSS style definitions
```
zorkow: typo + missing )

mathjax3-ts/output/chtml/CssStyles.ts
```diff
@@ -0,0 +1,113 @@
...
+ * The data for a selector
+ */
+export type StyleData = {
+    [property: string]: string | number;
+};
+
+/*
+ * A list of selectors and their data (basically a stylesheet)
+ */
+export type StyleList = {
```
zorkow: That's not really a list. Maybe StyleSet?

mathjax3-ts/output/chtml/CssStyles.ts
```diff
@@ -0,0 +1,113 @@
...
+    constructor(styles: StyleList = null) {
+        this.addStyles(styles);
+    }
+
+    /*
+     * @param{StyleList} styles  The styles to combine with the existing ones
+     */
+    public addStyles(styles: StyleList) {
+        if (!styles) return;
+        for (const style of Object.keys(styles)) {
```
zorkow: use for ... in.

mathjax3-ts/output/chtml/fonts/tex.ts
dpvc: Ok, but it is harder to read what is happening there.  You can't see the symmetry of the three lines as clearly.

Forcing one-line if-then statements onto separate lines seems like a stupid rule to me.  Personally, I think it makes things more difficult to read.  (I removed it from the `tslint.config` some time ago -- you didn't complain.  :-)

mathjax3-ts/output/chtml/CssStyles.ts
dpvc: It's as much a list as `ProperList`, `OptionList`, and the other `*List` types that we have elsewhere.  Do you really think it needs to change?

mathjax3-ts/output/chtml/CssStyles.ts
dpvc: That may enumerate things that we don't want.  For example, if someone adds a method to the `Object.prototype`, `for ... in` would list that as one of the keys.  So we would have to add and `if-then` for `styles.hasOwnProperty()` to make sure it was a top-level key (that would increase the nesting level, and I know you don't like that).  `Object.keys()` does that already.  I'm not sure what is gained by going to a more complicated (and deeply nested) construct.

dpvc:
OK, I've made the changes, pending your responses to my comments above.

christianp:
One huge thing you can do is to make every selector pick descendants of `.MathJax_CHTML`, to avoid the name clash problems we've encountered in MathJax v2. I don't feel completely safe just prefixing each class with `mjx-`.

dpvc:
@christianp, note that the selectors are all *element* based, not *class* based.  That is, the selectors are for `<mjx-mi>` not `<span class="mjx-mi">`.  These would only leak to other content if the page included custom elements starting with `mjx-`, and I consider that highly unlikely.

christianp:
Ahh, that's a good point! Maybe I wrote that comment a bit too early in the day!

mathjax3-ts/output/chtml/CssStyles.ts
zorkow: If it's consistent with the rest, that's fine.

zorkow:
That's ready to merge.

PR #37, opened 2017-09-04
Add font data files and actual font files.

dpvc:
Add font data files and actual font files.  These are in a separate branch for now to make reviewing previous pull requests easier.

These are just data, so no review should be needed.


pkra:
Some thoughts

* is `MathJax_Zero.otf` still needed?
* this would perhaps be better a separate repository
* we might want to exclude woff files as they can be built in whatever build process people (including ourselves) use

dpvc:
> is MathJax_Zero.otf still needed?

Yes.

> this would perhaps be better a separate repository

I'm not sure what "this" refers to.  MathJax_Zero?  All the fonts?  The font data files?

> we might want to exclude woff files as they can be built in whatever build process people (including ourselves) use

All the `mathjax2` content is temporary and eventually will be removed.  Just like the legacy v2 code that is shimmed into v3, these are the v2 fonts that are being used for now until I can do a complete replacement of them (as mentioned in the message for the previous PR that mentioned this branch was coming).  And just as the v2 code is temporarily in this repository, so, too, are these fonts for now.

Once we have a new font-development codebase, we can separate things out.  But for the time being, I think it is easiest to work with them here.

pkra:
> All the mathjax2 content is temporary and eventually will be removed. Just like the legacy v2 code that is shimmed into v3, these are the v2 fonts that are being used for now until I can do a complete replacement of them (as mentioned in the message for the previous PR that mentioned this branch was coming). And just as the v2 code is temporarily in this repository, so, too, are these fonts for now.

Thanks. I had missed the part about `mathjax2`. Sorry for the noise.

dpvc:
No problem.  We certainly want to come back to this in the future when we do make the new set of fonts.

zorkow:
Same here.

PR #39, opened 2017-09-07
Fixes the superfluous mrow problem for \left \right fences.

zorkow:
This reverts to the original way of appending elements to the mrow that contains the required case split.    
The case split has been removed between the following two commits:

      git diff 6b217c70629b1adc1f5fbb23e9b23dd4a56dcf37 3394af4c6ed4c7adb45f315acdcadf886521cd69 mathjax2/legacy/jax/input/TeX/jax.js

### DIFF
--- a/mathjax2/legacy/jax/input/TeX/jax.js
+++ b/mathjax2/legacy/jax/input/TeX/jax.js
@@ -2285,8 +2285,9 @@
     fenced: function (open,mml,close) {
       var mrow = MML.mrow().With({open:open, close:close, texClass:MML.TEXCLASS.INNER});
       mrow.Append(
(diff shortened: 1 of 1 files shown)


dpvc:
I'd rather make the changes in the actual MathJax v2 repo, and then move the files here as needed.  That way, the MathJax repo is always the master copy, and this is just a copy.  If we start making fixes here, then we run the risk of not back porting them to the official MathJax.

Also, you need to check not only if it is an mrow, but if it is an *inferred* mrow before unwrapping it.  Otherwise, you re-introduce the bug that the change originally fixed.

zorkow:
I agree. I've seen your issue. Please also see my email sent a bit earlier. Sorry for talking at cross-purposes.

dpvc:
No problem.

PR #40, opened 2017-09-08
Update v2 files

dpvc:
This moves the change that fixes mathjax/MathJax#1829 into the mathajx2 legacy copy here.


dpvc:
Sorry, included too much.  Will resubmit correctly.

PR #49, opened 2017-09-20
Add support for mub/msup/msubsup

dpvc:
This PR adds support for `msubsup`, `msub`, and `msup` elements.  It currently doesn't handle italic correction (so positions relative to something like an integral sign will be bad).

There is a base class `CHTMLscriptbase` that is the basis for all three elements.  It will also be the basis for the `munderover`, `munder` and `mover` elements, since the `movablelimits` attribute can cause the under-over elements to render as sub-sup elements in non-display modes.  So the latter under-over elements will be subclasses of the sub-sup classes, and the common under-over methods will be added to the `CHTMLscriptbase` class when under-over support is added.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/msubsup.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/msubsup.ts
@@ -1,0 +1,183 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2017 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


mathjax3-ts/output/chtml/Wrappers/scriptbase.ts
```diff
@@ -0,0 +1,145 @@
...
+ *  wrapper implementations
+ */
+
+export class CHTMLscriptbase extends CHTMLWrapper {
+    public static kind = 'scriptbase';
+
+    /*
+     * @return{CHTMLWrapper}  The base element's wrapper
+     */
+    public get base() {
```
zorkow: Can we make a name distinction as in mmlnodes base is a position.

mathjax3-ts/output/chtml/Wrappers/msubsup.ts
```diff
@@ -0,0 +1,183 @@
...
+        this.sub.toCHTML(stack);
+    }
+
+    /*
+     * @override
+     */
+    public computeBBox(bbox: BBox) {
+        const basebox = this.base.getBBox();
+        const subbox  = this.sub.getBBox();
+        const supbox  = this.sup.getBBox();
```
zorkow: ditto

dpvc: OK, changed to `baseChild`, `supChild`, `subChild`.

dpvc:
Changes made.

zorkow:
Lgtn

PR #56, opened 2017-10-11
Math tag support

dpvc:
This PR adds a wrapper for the math tag that handles display math as a block-style element.  It also adds support for percentage-width tables (which have to bubble the 100% percent width up through the parent elements to the containing math element).  The TeX input uses percentage-width tables for the multline environment.  You can also specify a fixed width for tables.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/mrow.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/mrow.ts
@@ -49,15 +49,18 @@ export class CHTMLmrow extends CHTMLWrapper {
      * @override
      */
     public toCHTML(parent: HTMLElement) {
-        let chtml = parent;
+        let chtml = this.chtml = parent;
         if (!this.node.isInferred) {
             chtml = this.standardCHTMLnode(parent);
         }
         let hasNegative = false;
         for (const child of this.childNodes) {
             child.toCHTML(chtml);
-            if (child.bbox && child.bbox.w < 0) {
+            if (child.bbox.w < 0) {
                 hasNegative = true;
+            }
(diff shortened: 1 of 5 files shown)


mathjax3-ts/output/chtml/Wrappers/mrow.ts
```diff
@@ -49,16 +49,19 @@ export class CHTMLmrow extends CHTMLWrapper {
...
-        let chtml = parent;
+        let chtml = this.chtml = parent;
         if (!this.node.isInferred) {
             chtml = this.standardCHTMLnode(parent);
         }
         let hasNegative = false;
         for (const child of this.childNodes) {
             child.toCHTML(chtml);
-            if (child.bbox && child.bbox.w < 0) {
+            if (child.bbox.w < 0) {
```
zorkow: Is child.bbox guaranteed to exist?

mathjax3-ts/output/chtml/Wrappers/mrow.ts
```diff
@@ -68,6 +71,15 @@ export class CHTMLmrow extends CHTMLWrapper {
         }
     }
 
+    /*
+     * Handle the case where a child as a percentage width by
```
zorkow: type: has

mathjax3-ts/output/chtml/Wrappers/mrow.ts
dpvc: Yes, it is created in the Wrapper constructor at [line 266](https://github.com/mathjax/mathjax-v3/blob/master/mathjax3-ts/output/chtml/Wrapper.ts#L266)

PR #57, opened 2017-10-11
Italic correction

dpvc:
This PR adds support for italic correction, based on TeX's rules.  We use the difference between a characters right-bearing and its width as a surrogate for italic-correction (since we don't have the actual TeX metrics for this).

The italic correction is added to the last character in `mi` and `mo` nodes (and should probably be added in `mn` as well, but numbers are usually in roman and so don't have italic correction).  So an `mi` wrapper is added and the `mo` wrapper modified to do that.  The italic correction is added via the CSS.

Italic correction is *not* to be used in some superscript and under-over situations.  So there is CSS set up to remove the italic correction when needed in those cases.

Also, the italic correction is used to properly place super- and subscripts and under-over constructs, as per TeX placement rules.  We adjust the italic correction value in theses cases to more closely match the actual TeX placements (since we don't have the actual TeX values used).

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/TextNode.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/TextNode.ts
@@ -82,7 +82,7 @@ export class CHTMLTextNode extends CHTMLWrapper {
     /*
      * @param{string} variant   The variant in which to look for the character
      * @param{number} n         The number of the character to look up
-     * @param{CharData}         The full CharData object, with CharOptions guaranteed to be defined
+     * @return{CharData}        The full CharData object, with CharOptions guaranteed to be defined
      */
     protected getChar(variant: string, n: number) {
         const char = this.font.getChar(variant, n) || [0, 0, 0, null];
--- a/mathjax3-ts/output/chtml/Wrappers/TextNode.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/TextNode.ts
@@ -24,6 +24,7 @@
 import {CHTMLWrapper} from '../Wrapper.js';
 import {BBox} from '../BBox.js';
 import {TextNode} from '../../../core/MmlTree/MmlNode.js';
+import {CharOptions} from '../FontData.js';
 
 /*****************************************************************/
 /*
@@ -63,17 +64,29 @@ export class CHTMLTextNode extends CHTMLWrapper {
             // FIXME:  measure this using DOM, if possible
         } else {
             const chars = this.unicodeChars((this.node as TextNode).getText());
-            let [h, d, w] = this.font.getChar(variant, chars[0]) || [0, 0, 0];
+            let [h, d, w, data] = this.getChar(variant, chars[0]);
             bbox.h = h;
             bbox.d = d;
             bbox.w = w;
+            bbox.ic = data.ic || 0;
             for (let i = 1, m = chars.length; i < m; i++) {
-                [h, d, w] = this.font.getChar(variant, chars[i]) || [0, 0, 0];
+                [h, d, w, data] = this.getChar(variant, chars[i]);
                 bbox.w += w;
                 if (h > bbox.h) bbox.h = h;
                 if (d > bbox.d) bbox.d = d;
+                bbox.ic = data.ic || 0;
             }
         }
+    }
+
+    /*
+     * @param{string} variant   The variant in which to look for the character
+     * @param{number} n         The number of the character to look up
+     * @param{CharData}         The full CharData object, with CharOptions guaranteed to be defined
+     */
+    protected getChar(variant: string, n: number) {
+        const char = this.font.getChar(variant, n) || [0, 0, 0, null];
+        return [char[0], char[1], char[2], char[3] || {}] as [number, number, number, CharOptions];
     }
 
     /******************************************************/
--- a/mathjax3-ts/output/chtml/Wrappers/msubsup.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/msubsup.ts
@@ -50,7 +50,7 @@ export class CHTMLmsub extends CHTMLscriptbase {
      * @override
      */
     protected getOffset(bbox: BBox, sbox: BBox) {
-        return -this.getV(bbox, sbox);
+        return [0, -this.getV(bbox, sbox)];
     }
 
 }
@@ -62,6 +62,8 @@ export class CHTMLmsub extends CHTMLscriptbase {
 
 export class CHTMLmsup extends CHTMLscriptbase {
     public static kind = MmlMsup.prototype.kind;
+
+    public static useIC: boolean = true;
 
     /*
      * @override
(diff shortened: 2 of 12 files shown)


mathjax3-ts/output/chtml/fonts/tex.ts
```diff
@@ -441,7 +441,14 @@ export class TeXFont extends FontData {
             }
         }
         if (options.f !== undefined) css['font-family'] = 'MJXZERO, MJXTEX' + (options.f ? '-' + options.f : '');
-        styles['.MJX-TEX' + vclass + ' mjx-c[c="' + this.char(n) + '"]::before'] = css;
+        const char = vclass + ' mjx-c[c="' + this.char(n) + '"]';
+        styles['.MJX-TEX' + char + '::before'] = css;
+        if (options.ic) {
+            styles['.MJX-TEX mjx-mi:not([noIC="true"])' + char.substr(1) + ':last-child::before'] =
+            styles['.MJX-TEX mjx-mo:not([noIC="true"])' + char.substr(1) + ':last-child::before'] = {
+                width: this.em(w + options.ic)
```
zorkow: This is confusing. Can we have some auxiliary variables handling substrings? E.g., everything after mo/mi is the same.
Also, I think multiple assignments for initialisation is  fine, but I would avoid them for more meaningful assignments.

mathjax3-ts/output/chtml/Wrappers/scriptbase.ts
```diff
@@ -192,12 +232,24 @@ export class CHTMLscriptbase extends CHTMLWrapper {
...
-    protected getDeltaW(boxes: BBox[]) {
+    protected getDeltaW(boxes: BBox[], delta: number[] = [0, 0, 0]) {
         const widths = boxes.map(box => box.w * box.rscale);
         const w = Math.max(...widths);
-        return widths.map(width => (w - width) / 2);
+        const dw = [];
+        let m = 0;
+        for (const i of widths.keys()) {
+            dw[i] = (w - widths[i]) / 2 + delta[i];
+            if (dw[i] < m) m = -dw[i];
```
zorkow: braces

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
```diff
@@ -63,19 +64,31 @@ export class CHTMLTextNode extends CHTMLWrapper {
...
                 if (d > bbox.d) bbox.d = d;
+                bbox.ic = data.ic || 0;
             }
         }
     }
 
+    /*
+     * @param{string} variant   The variant in which to look for the character
+     * @param{number} n         The number of the character to look up
+     * @param{CharData}         The full CharData object, with CharOptions guaranteed to be defined
```
zorkow: `@return{CharData}` ?

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
```diff
@@ -63,19 +64,31 @@ export class CHTMLTextNode extends CHTMLWrapper {
...
+     */
+    protected getChar(variant: string, n: number) {
+        const char = this.font.getChar(variant, n) || [0, 0, 0, null];
```
zorkow: Why null and not {}?

mathjax3-ts/output/chtml/Wrappers/msubsup.ts
```diff
@@ -162,8 +171,9 @@ export class CHTMLmsubsup extends CHTMLscriptbase {
         if (this.UVQ) return this.UVQ;
         const tex = this.font.params;
         const t = 3 * tex.rule_thickness;
-        let [u, v] = (this.isCharBase() ? [0, 0] : [this.getU(basebox, supbox),
-                       Math.max(basebox.d + tex.sub_drop * subbox.rscale, tex.sub2)]);
+        const subscriptshift = this.length2em(this.node.attributes.get('subscriptshift'), tex.sub2);
+        const drop = (this.isCharBase() ? 0 : basebox.d + tex.sub_drop * subbox.rscale);
```
zorkow: These (also below) seem important computations. Can you add a comment what exactly goes on for the benefit of others (e.g, this reviewer).

mathjax3-ts/output/chtml/Wrappers/munderover.ts
```diff
@@ -73,8 +80,9 @@ export class CHTMLmunder extends CHTMLmsub {
         const basebox = this.base.getBBox();
         const underbox = this.script.getBBox();
         const [k, v] = this.getUnderKV(basebox, underbox);
+        const delta = DELTA * this.baseCore.bbox.ic / 2;
```
zorkow: delta vs DELTA

mathjax3-ts/output/chtml/Wrappers/scriptbase.ts
```diff
@@ -105,10 +145,10 @@ export class CHTMLscriptbase extends CHTMLWrapper {
      *
      * @param{BBox} bbox   The bounding box of the base element
      * @param{BBox} sbox   The bounding box of the script element
-     * @return{number}     The vertical offset for the script
+     * @return{number[]}   The horizontal and vertical offsets for the script
      */
     protected getOffset(bbox: BBox, sbox: BBox) {
-        return 0;
+        return [0, 0];
```
zorkow: I am wondering if we should have a type for pairs and triples of numbers. They appear a lot and might help finding bugs. number[] could be too general. Just a thought, maybe for the future.

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
dpvc: Right.

mathjax3-ts/output/chtml/Wrappers/TextNode.ts
dpvc: The default in the data arrays is `null`.  It gets set to `{}` in the next line anyway, so does it really matter?

mathjax3-ts/output/chtml/Wrappers/msubsup.ts
dpvc: These are the computations given in the TeXbook (Appendix G).  I've cited that in the comment for the function itself.  It's not always clear to me what they are up to, either, as the comments in the TeXBook are pretty sparse as well.  I'll see what I can do.

mathjax3-ts/output/chtml/Wrappers/munderover.ts
dpvc: If you can't use the distinction between upper and lower case, why have them be distinct?  The all-caps version is a constant, and the lower case one is a variable.  I'm not sure why that is such a problem.  In any case, I've changed the `delta` to something less meaningful.

mathjax3-ts/output/chtml/Wrappers/scriptbase.ts
dpvc: OK, we can consider that for the future.

dpvc:
I've made the requested changes.

PR #59, opened 2017-10-15
Better tall square roots

dpvc:
Make sqrt line overlap with tall surd top corner glyph for better attachment.  Only affects very tall roots that requires multi-character surd.  The top corner has always been difficult to align properly with the overline, and this can improve the look (originally suggested in MathJax issue 1838).

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/msqrt.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/msqrt.ts
@@ -158,6 +158,11 @@ export class CHTMLmsqrt extends CHTMLWrapper {
         surd.toCHTML(SURD);
         base.toCHTML(BASE);
         if (surd.size < 0) {
+            //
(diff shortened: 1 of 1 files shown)


mathjax3-ts/output/chtml/Wrappers/msqrt.ts
```diff
@@ -153,6 +157,9 @@ export class CHTMLmsqrt extends CHTMLWrapper {
         this.addRoot(ROOT, root, sbox);
         surd.toCHTML(SURD);
         base.toCHTML(BASE);
+        if (surd.size < 0) {
```
zorkow: Can you put a comment here that contains the gist of the explanation for the PR, for future reference?

dpvc: Done.

PR #60, opened 2017-10-15
Add support for horizontal stretchy characters in munderover

dpvc:
This PR adds support for horizontal stretchy characters in `munderover` elements (I still need to do `mtd` elements).  Most of the code for generating these was already there in the `mo` element, but the CSS needed to be generated, and the `munderover` needed to check for stretchy characters and ask them to stretch.

As part of this, the `stretch` property of the `CHTMLWrapper` was changed from a `DIRECTION`, to holding the actual delimiter data for the stretchy character instead (which includes the stretch direction).  This is because some delimiters change their character to another one (e.g., `^` uses the stretchy characters stored at U+02C6 in the TeX fonts).  So this keeping this data allows the child text node of the `mo` to use the correct glyph and its proper bounding box.


mathjax3-ts/output/chtml/fonts/tex.ts
```diff
@@ -382,23 +382,37 @@ export class TeXFont extends FontData {
      * @param{DelimiterData} data  The data for the delimiter whose CSS is to be added
      */
     protected addDelimiterStyles(styles: StyleList, n: number, data: DelimiterData) {
+        const c = this.char(n);
+        if (data.c && data.c !== n) {
+            styles['.MJX-TEX .mjx-stretched mjx-c[c="' + c + '"]::before'] = {content: '"' + this.char(data.c,true) + '"'};
```
zorkow: Linelength

mathjax3-ts/output/chtml/fonts/tex.ts
```diff
@@ -414,12 +428,45 @@ export class TeXFont extends FontData {
...
+    }
+
+    /*
+     * @param{StyleList} styles  The style object to add styles to
+     * @param{string} c          The vertical character whose part is being added
+     * @param{string} part       The name of the part (beg, ext, end, mid) that is being added
+     * @param{number} n          The unicode character to use for the part
+     */
+    protected addDelimiterHPart(styles: StyleList, c: string, part: string, n: number, force: boolean = false) {
+        if (!n) return 0;
```
zorkow: Missing braces.

mathjax3-ts/output/chtml/Wrappers/mo.ts
```diff
@@ -191,27 +198,28 @@ export class CHTMLmo extends CHTMLWrapper {
...
-        if (this.stretch && this.size < 0) return;
+        if (stretchy && this.size < 0) return;
         super.computeBBox(bbox);
         const child = this.childNodes[this.childNodes.length-1];
         if (child && child.bbox.ic) {
             bbox.ic = child.bbox.ic;
             if (!this.noIC) bbox.w += bbox.ic;
         }
-        if (symmetric) {
+        if (this.node.attributes.get('symmetric') && this.stretch.dir !== DIRECTION.Horizontal) {
```
zorkow: Linelength.

mathjax3-ts/output/chtml/fonts/tex.ts
dpvc: I thought we agreed that returns could be on the same line.

PR #65, opened 2017-11-16
Fix bbox for msqrt.

dpvc:
This fixes a problem with the bounding box for msqrt.  It didn't take the base bounding box height into account correctly.  (This happened when renaming bounding boxes for Volker's request — I missed one in the renaming.)


dpvc:
PS, this fixes the parenthesis height problem in the Ramanujan example.

PR #66, opened 2017-11-16
Add missing data for vertically stretchy braces, and fix a display is…

dpvc:
This fixes an issue with the size of vertically stretchy braces, as well as the missing glyph for the central piece.  It also works around an issue in Safari with the stretchy pieces not displaying in some situations.  (I also deconstructed the `data.stretch` array to make it easier and more mnemonic to refer to the pieces.)


pkra:
=> marked for next alpha release

Issue #64, opened 2017-11-10
Large parens missing extender on Edge


NSoiffer:
Using https://mathjax.github.io/mj3-demos/mj3-tex2html-global.html, with input

`x = -\left( \begin{array}{cc} a & b \\ c & d\\ x-1 & x_2 +x^{3^4} \\ z 4\\ a & b\end{array}  \right)`

the extender is missing in **Edge** (Microsoft Edge 38.14393.0.0; Microsoft EdgeHTML 14.14393). Picture below:

![image](https://user-images.githubusercontent.com/1545836/32676103-368c6826-c60e-11e7-975a-051e82e75fd5.png)

Some other stretchy delims (left { and completely missing right vertical line):
`x = -\left\{ \begin{array}{cc} a & b \\ c & d\\ x-1 & x_2 +x^{3^4} \\ z 4\\ a & b\end{array}  \right|`

![image](https://user-images.githubusercontent.com/1545836/32676193-90fbf52e-c60e-11e7-94c3-5794540f62a1.png)

zorkow:
The bug with braces also occurs on Firefox and Chrome. Other extenders are fine, though.

pkra:
This reminds me of problems I saw early on so I'm not sure if the build is perhaps out of date. Could you try build a fresh one?

Another random idea would be check if webfonts vs local fonts matter (I remember seeing similar issues in other projects).

FYI Davide has taken a look but can't reproduce.

dpvc:
I suspect that the build is not up to date in some way as well.  I will try to make a new one locally myself and see what happens.

As for the fonts, I'm pretty sure local fonts are not used, even if installed, for this version of the CSS.

pkra:
@dpvc I'm reasonably sure I merged `stretchy-cells` and `tex_input_typescript` branches. Are those not the latest?

zorkow:
We've played with the demo linked from the demo repo.

dpvc:
I was wondering if you ran `tsc` after the merge, or after checking out your branch.  I've fallen for that repeatedly myself, and gotten weird results due to partially compiled (or differently compiled) pieces.  There are also some small corrections in some other branches based on Volker's reviews.  And one other branch that is from master to master not in stretchy-cells.

pkra:
@dpvc wrote 

> I was wondering if you ran tsc after the merge, or after checking out your branch. 

[...]

PR #73, opened 2017-12-06
Improve CSS for stretchy braces in IE and Edge

dpvc:
This adjusts the CSS to avoid the gaps pointed out by Neil in #64, and improves the situation for horizontal placement of extenders in Firefox (though not for Chrome).

### DIFF
--- a/mathjax3-ts/output/chtml/fonts/tex.ts
+++ b/mathjax3-ts/output/chtml/fonts/tex.ts
@@ -404,8 +404,9 @@ export class TeXFont extends FontData {
         const Hb = this.addDelimiterVPart(styles, c, 'beg', beg);
         this.addDelimiterVPart(styles, c, 'ext', ext);
         const He = this.addDelimiterVPart(styles, c, 'end', end);
+        let Hm;
         if (mid) {
-            this.addDelimiterVPart(styles, c, 'mid', mid);
+            Hm = this.addDelimiterVPart(styles, c, 'mid', mid);
             styles['.MJX-TEX mjx-stretchy-v[c="' + c + '"] > mjx-ext'] = {height: '50%'}
         }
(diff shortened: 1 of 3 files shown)


mathjax3-ts/output/chtml/fonts/tex.ts
```diff
@@ -414,10 +415,11 @@ export class TeXFont extends FontData {
...
         if (He) {
             css['border-bottom-width'] = this.em0(He - .03);
-            css['margin-bottom'] = this.em(-He);
+            styles['.MJX-TEX mjx-stretchy-v[c="' + c + '"] > mjx-end'] = {'margin-top': this.em(-He)};
             if (mid) {
-                styles['.MJX-TEX mjx-stretchy-v[c="' + c + '"] > mjx-ext:last-of-type'] = {
-                    'margin-top': this.em(-He)
+                styles['.MJX-TEX mjx-stretchy-v[c="' + c + '"] > mjx-mid'] = {
+                    'margin-top': this.em(-Hm/2),
+                    'margin-bottom': this.em(-Hm/2)
```
zorkow: Why can't the ``mid`` case from lines 407-411 be handled here?
Can we ever get a mid part without end part?
Or is it important that the styles is set earlier?

dpvc: Not sure why it is in the location.  Probably due to doing one part at one time and the other at later time.  I've moved them to the same location and tidied up a bit.

dpvc: Oh, I see why it was the way.  The original CSS was for a different element, but was changed to the mid element, and I didn't catch that it could be moved earlier.

dpvc:
This is ready for re-review.

Issue #74, opened 2017-12-12
TypeError: reduce of empty array with no initial value


ridgeworks:
TypeError: reduce of empty array with no initial value 

generated using Chrome/OSX with mj3-demos-master/mj3-mml2html-global.html

with input: 

```
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
<mtable></mtable>
</math>
```

This is the simplest case but others also fail in similar way, e.g.,:

```
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
	<mo>(</mo><mtable><mtr><mtd><mn>1</mn></mtd></mtr></mtable>
</math>
```

Full traceback:

```
TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>)
    at s.computeBBox (mj3-mml2html-global.dist.js:1:192096)
    at s.e.getBBox (mj3-mml2html-global.dist.js:1:15975)
    at s.stretchChildren (mj3-mml2html-global.dist.js:1:171881)
    at new s (mj3-mml2html-global.dist.js:1:170829)
    at node.(anonymous function) (mj3-mml2html-global.dist.js:1:47178)
    at s.t.create (mj3-mml2html-global.dist.js:1:46975)
    at s.wrap (mj3-mml2html-global.dist.js:1:155364)
    at s.e.wrap (mj3-mml2html-global.dist.js:1:15509)
    at mj3-mml2html-global.dist.js:1:15163
```

pkra:
Thanks for the report!

ridgeworks:
No problem.

While MathJax suits my current purpose, ideally I'd like to see a MathML to HTML translator (String -> String) that could be executed in a Web Worker, i.e., no DOM. Is this a usage scenario that mj3 (or some piece of it) would work for?

Rick

ridgeworks:
Further testing on OSX shows this seems to only be a problem on Chrome (V62). Both Safari and Firefox (current versions) are fine.

pkra:
FWIW, I saw it on both Chrome and Firefox on Linux.

ridgeworks:
Yes; my bad. It happens everywhere. I'll leave the diagnosis to the experts.

dpvc:
The `issue74` branch resolves this issue.

PR #75, opened 2017-12-13
Add initial value to reduce() calls to avoid problems with empty arrays. #74

dpvc:
Resolves issue #74 by adding initial values to the `reduce()` calls.


dpvc:
> Can you please stretch that out a bit

OK, see if that is better.

PR #84, opened 2018-03-05
Abstract the DOM actions to support virtual DOMs

dpvc:
This PR touches many files, but mostly because we need to use type templates in order to tell the various components what the types to use for `HTMLElement`, `Text`, and `Document` (which are now allowed to be other types rather than the actual DOM elements).  These are the `<N, T, D>` that you will see everywhere in the differences below.  The other changes to existing files involve converting from calling `HTMLElement` methods to using the corresponding abstracted methods of the new library that mediates the DOM interaction.  This is the defined in the `core/DOMAdaptor.ts`, which gives the interface and abstract classes.

Specific instances of the DOMAdaptor are in the new `adaptors` directory.  The `adaptors/HTMLAdaptor.ts` file is the basic implementation for actual DOM elements, and it is used by `adaptors/jsdomAdaptor.ts` for the JSDOM version, and `adaptors/browserAdaptor.ts` for the browser-DOM version.  There is a legacy `adaptors/chooseAdaptor.js` that does what the current code does, which is select the browser or jsdom version on the fly, but this will probably not be used in production workflows; it just makes it easier to test both implementations without changing the code.

There is also a new `adaptor/liteAdaptor.ts` file that implements a non-DOM-based adaptor that uses minimal javascript elements rather than actual DOM elements.  This is the proof-of-concept for the fact that the abstraction from the DOM works as intended.  The adaptor needs to be able to do not only the HTMLElement actions (like appending and replacing child nodes, getting the next, previous, or parent node, setting attributes, classes, styles, and so on), but also needs to be able to handle some document-level actions (like the equivalent of `getElementsByTagName()`), and it also needs to be able to parse a serialized HTML file into its internal form.

For the latter, the `liteAdaptor.ts` file includes a `LiteParser` object that handles that.  It is pretty minimal, and I don't pretend to claim that it does all of HTML5 properly (particularly ill-formed HTML).  It is enough to do testing and processing of straight-forward HTML, but probably should not be used for production.  For production work, one would want to use something like `parse5` or `htmlparser2` instead; the `LiteAdaptor` object could be used as a template for producing adaptors for those tools (just like we have for jsdom).  I suspect the conversion would not be hard.

Because the output jax have relied in the past on an HTMLElement from the DOM to handle parsing of CSS styles, I needed to make a minimal replacement for that as well, for the case where there is no actual DOM to work with.  This is in the `util/Styles.ts` file, and is quite minimal (just enough to do what the CommonHTML output needs at the moment).  It could be extended to a more full-featured version, but I don't think it needs to be (and the would make it bigger and slower).  For now, this should be sufficient, but we can address it if the need arises in the future.

In the end, the changes needed for the page author are quite small.  In addition to the input and output jax, you now need to specify the DOM adaptor as well.  The handler needs to be passed the adaptor, and this is done when the handler is registered with MathJax.  So there is a new `RegisterHTMLHandler()` function that makes this easy to do.  I've modified the sample files to use this, so you can see how that is done.

Finally, the `util/DOM.ts` and `util/HTMLNodes.ts` files have been removed, since their functions are now handled via the DOM adaptor.

I suggest that you start by looking at the files in the following order:

1.  core/DOMAdaptor.ts
2. adaptors/HTMLAdaptor.ts
3. adaptors/browserAdaptor.ts
4. adaptors/jsdomAdaptor.ts
5. The other core, input, and output files
6. adaptors/liteAdaptor.ts
7. util/Styles.ts

I know it is a big PR, but I do think that majority of it is straight-forward conversion to use the types and adaptor, and should not pose too much of a problem.


dpvc:
PS, the `adaptors/liteAdaptor.ts` file may want to be broken into several pieces (certainly the parser can be broken out).  Also, the `input/mathml/MmlEntities.ts` file and its support files should probably be moved to utilities, since they are being used in the LiteParser as well.  I will do that as a second step (I didn't want to confuse things more by renaming those as part of this PR.)

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -491,7 +500,7 @@ export class CHTMLWrapper extends AbstractWrapper<MmlNode, CHTMLWrapper> {
         if (!this.styles) return;
         const styles = this.styles.cssText;
         if (styles) {
-            this.chtml.style.cssText = styles;
+            this.adaptor.setAttribute(this.chtml, 'style', styles);
```
zorkow: Why is that not setStyle? Or is this supposed to overwrite the existing style.

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -500,7 +509,8 @@ export class CHTMLWrapper extends AbstractWrapper<MmlNode, CHTMLWrapper> {
      */
     protected handleVariant() {
         if (this.node.isToken && this.variant !== '-explicitFont') {
-            this.chtml.className = (this.font.getVariant(this.variant) || this.font.getVariant('normal')).classes;
+            this.adaptor.setAttribute(this.chtml, 'class',
+                                    (this.font.getVariant(this.variant) || this.font.getVariant('normal')).classes);
```
zorkow: Will that interfere with classlist? Use addclass instead?

mathjax3-ts/core/DOMAdaptor.ts
```diff
@@ -0,0 +1,552 @@
...
+
+    /*
+     * @param{N | T} node   The HTML node whose tag or node name is to be obtained
+     * @return{string}      The tag or node name of the given node
+     */
+    kind(node: N | T): string;
+
+    /*
+     * @param{N|T} node  The HTML node whose value is to be obtained
+     * @return{string}   The value of the given node
```
zorkow: That could also be null. E.g., nodeValue of a non-text node.

mathjax3-ts/core/DOMAdaptor.ts
```diff
@@ -0,0 +1,552 @@
...
+
+/*****************************************************************/
+/*
+ *  The interface for the DOMAdaptor
+ *
+ *  N = HTMLElement node class
+ *  T = Text node class
+ *  D = Document class
+ */
+
```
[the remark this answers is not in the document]
dpvc: OK.  Do you mean just here, or in every file where a class where these are used?  (There are about 50 of them, all with the same meaning.)

mathjax3-ts/adaptors/browserAdaptor.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+import {HTMLAdaptor} from './HTMLAdaptor.js';
+
+//
+//  Let Typescript know about these
+//
+declare global {
+    interface Window {
```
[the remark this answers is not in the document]
dpvc: No, I don't think so.  This is adding missing properties that are part of the browser window.  I'm not sure why they aren't there already, but browser windows include these, and they are needed in order to be able to access these properties later on.  This is apparently the way you are supposed to make such additions (I looked it up, but can't find the citation now).

mathjax3-ts/adaptors/HTMLAdaptor.ts
```diff
@@ -0,0 +1,430 @@
...
+     */
+    public getElements(nodes: (string | N | N[])[], document: D) {
+        let containers: N[] = [];
+        for (const node of nodes) {
+            if (typeof(node) === 'string') {
+                containers = containers.concat(Array.from(this.document.querySelectorAll(node)));
+            } else if (Array.isArray(node)) {
+                containers = containers.concat(Array.from(node) as N[]);
+            } else if (node instanceof this.window.NodeList || node instanceof this.window.HTMLCollection) {
+                containers = containers.concat(Array.from(node as any as N[]));
```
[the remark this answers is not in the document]
dpvc: This is because Typescript apparently doesn't realize that NodeList and HTMLCollection are iterables (even though they are), so won't let me call `Array.from()` on them directly.  So I have to case them to something iterable, like an array.  Since I want the array to be `N[]` anyway (both `NodeList` and `HTMLCollection` will be arrays of `N`), I want to cast to that.  But Typescript won't let me cast to `N[]` directly (it doesn't realize the the types are compatible), and so I cast to `any` first, and then to `N[]`.  This is all just to get around the fact that Typescript doesn't properly know the types of the things involved.

mathjax3-ts/adaptors/liteAdaptor.ts
```diff
@@ -0,0 +1,1043 @@
...
+                    const n = this.elementById(body, node.slice(1));
+                    if (n) {
+                        containers.push(n);
+                    }
+                } else if (node.charAt(0) === '.') {
+                    containers = containers.concat(this.elementsByClass(body, node.slice(1)))
+                } else if (node.match(/^[-a-z][-a-z0-9]*$/i)) {
+                    containers = containers.concat(this.tags(body, node));
+                }
+            } else if (Array.isArray(node)) {
```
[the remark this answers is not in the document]
dpvc: I'm not sure what you have in mind, but `liteAdaptor` is a subclass of `AbastractDOMAdaptor`, and that has an abstract `getElements()` method, so there is no super method to call.  There is a similar method in the `HTMLAdaptor` class, so perhaps that is what you have in mind.  But `liteAdaptor` is not a subclass of that, and its `getElements()` method is not the same (it relies on `querySelector()`, for example, which we don't have in a lite document, and the handling of the arrays is slightly different).

mathjax3-ts/util/Styles.ts
```diff
@@ -0,0 +1,524 @@
...
+    }
+    this.styles[name] = value;
+}
+
+/*********************************************************/
+/*
+ * Patterns for the parts of a boarder
+ */
+const BORDER: {[name: string]: RegExp} = {
+    width: /^(?:[\d.]+(?:[a-z]+)|thin|medium|thick|inherti|initial|unset)$/,
```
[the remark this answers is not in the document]
dpvc: Thanks!

mathjax3-ts/input/mathml/MmlEntities.ts
```diff
@@ -517,7 +517,7 @@ export class MmlEntities {
      * @return {string}      The text with entiries replaced
```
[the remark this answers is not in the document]
dpvc: Fixed

mathjax3-ts/output/chtml/Wrapper.ts
dpvc: The `setStyle()` method sets a single style by passing the style selector and its new value.  But here we are setting the complete style attribute (e.g., `styles` might be `'border: none; padding: 3px 5px; font-family: Times, serif'` or something like that).  Setting the `'style'` attribute clears the existing styles and replaces them with the new ones (just as setting `style.cssText` did before).  So this is the correct usage.

mathjax3-ts/core/DOMAdaptor.ts
dpvc: It looks like `value()` is only being called on text nodes, so there are a couple of possible solutions:

* I could change the parameter to be just `T` rather than `N | T` (though I think that will require casting to T in several places).
* I could change the implementation in the `HTMLAdaptor` to return `node.nodeValue || ''` so that it is never null.
* I could change the return type to `@return{string|null}`.

Do you have a preference?

dpvc: I ended up making it always return a string (no matter whether it was a non-text node).  It turned out this was the most general solution, and the one that required the least fiddly typecasting.

dpvc:
I have made the changes to the branch that I think are needed.  Can you take another look?

mathjax3-ts/core/DOMAdaptor.ts
zorkow: Sorry for not replying to the initial remark. Too busy last week. But I like the string solution.

zorkow:
lgtm.

PR #94, opened 2018-04-08
Lowercases incorrect function names and uses new DOM adaptors in main.js

zorkow:


dpvc:
Thanks for the fixes.  I missed these during the update of the examples (I guess I don't use `main.js` myself anymore).

PR #95, opened 2018-04-08
Fixes legacy code to deal with new mathchoice node.

zorkow:


dpvc:
This certainly work, but I'm wondering if it wouldn't be better to change the `mathchoice` implementation to `TeXmathchoice` like it was before.  This is really a TeX extension for MathML, so having `TeX` in the name makes that clear.  I should have done that originally, but didn't look that closely at the original (sigh).

zorkow:
I don't really see a need for the tex prefix. But if you want to add it again, no problems. I would argue for Proper caml casing then however: TexMathchoice.

dpvc:
OK, I'll leave it as is.

PR #100, opened 2018-04-16
Tex to serialised mathml sample.

zorkow:
* Adds handling of TeXAtom nodes in the seralized mml visitor and by extension in the test visitor.  Nodes are handled equivalent to the old style by adding an mrow with the MJ-... texclass class.
* Adds a tex-mml.js processor in the samples files.
* Fixes some typos in comments.

I have not made any changes to the MathMLVisitor yet. But if the serialised version is fine, I can do that.


zorkow:
Changed script name as discussed yesterday.

PR #101, opened 2018-04-17
Fix import

ian-r-rose:
I got a build error when testing alpha-4 due to what appears to be a typo.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrapper.ts
+++ b/mathjax3-ts/output/chtml/Wrapper.ts
@@ -28,7 +28,7 @@ import {Property} from '../../core/Tree/Node.js';
 import {OptionList} from '../../util/Options.js';
 import {unicodeChars} from '../../util/string.js';
 import * as LENGTHS from '../../util/lengths.js';
-import {Styles} from '../../util/Styles.js';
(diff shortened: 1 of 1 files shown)


dpvc:
The file name *is* capitalized (see [here](https://github.com/mathjax/mathjax-v3/tree/alpha/mathjax3-ts/util)), so I don't know why you would get an error.  Can you check whether the file in your copy of the `util` folder has a lower-case name?

ian-r-rose:
Hm, yeah, I was just checking that. It seems that the released package that I am getting from npm has it lowercased. Is it possible something went amiss with the release?

dpvc:
I'll have to check into it.

dpvc:
PS, don't forget that the separation of MathJax from the DOM requires a slightly modified setup.  You need to create a DOM adaptor that indicates the DOM you will be using.  See the files in the `samples` directory, and in the v3-demos repository for examples.  Also, the webpack configuration in the v3-demos repo shows how to prevent the warning about dynamic loading of files.  See also [this comment](https://github.com/mathjax/mathjax-v3/issues/88#issuecomment-378257424).

PR #103, opened 2018-05-02
Fix problem with \xrightarrow not stretching.

dpvc:
This fixes the problem with `\xrightarrow` not stretching reported by Volker.  This was due to the `<mo>` being asked to stretchy both horizontally (by the `<munderover>`) and vertically (by the enclosing `<mrow>`, since it is an embellished operator in that row).

The resolution is to only allow stretching in one direction (since we don't support bi-directional stretching), that being the first direction for which stretching was requested and possible.

This also properly scales the maximum width computations in `munderover` so the size of the arrow is correct.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrapper.ts
+++ b/mathjax3-ts/output/chtml/Wrapper.ts
@@ -337,6 +337,18 @@ export class CHTMLWrapper<N, T, D> extends AbstractWrapper<MmlNode, CHTMLWrapper
         bbox.clean();
     }
 
+    /*
+     * Mark BBox to be computed again (e.g., when an mo has stretched)
+     */
+    public invalidateBBox() {
+        if (this.bboxComputed) {
+            this.bboxComputed = false;
+            if (this.parent) {
+                this.parent.invalidateBBox();
+            }
+        }
+    }
+
     /*******************************************************************/
 
     /*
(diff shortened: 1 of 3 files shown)


zorkow:
It doesn't seem to work for me. Still no stretch. This is the MathML that I get from my parser, which is the same as in legacy:
```XML
<math display="block">
  <mover>
    <mo>←</mo>
    <mpadded width="+10mu" lspace="7mu" voffset=".15em">
      <mi>a</mi>
      <mi>b</mi>
      <mi>c</mi>
      <mi>d</mi>
    </mpadded>
  </mover>
</math>
```

dpvc:
It works for me with my examples as well as the MathML that you give.  Did you try pasting that into the MathML lab?  That works for me.

What branch are you using?  Did you remember to compile it with typescript after checking out the branch (or merging it into some other branch)?

zorkow:
I did all of the above.
But I just found the problem: It works in Firefox but not in Chrome (dev channel)!

zorkow:
PS: I also always have to do two fixes to the Mml-Lab.js file when I want to use it. One is definitely due to Mac vs Linux. I'll submit a PR for them.

dpvc:
OK, do you want me to try to figure out the issue with Chrome (dev channel), or do you think it is not worth trying to debug in a development version?  I haven't used it in a while (is that "Chrome Canary"?)  I'm kind of curious what could be going on, though, as the stretched version is a completely different HTML structure from the unstretched version.  Strange.

zorkow:
Further to the above: 
It works fine in IE and Edge. 
But it also does not work in the latest stable Opera, so it seems to be a general blink problem.

zorkow:
It also does not work in Chrome stable (Version 66.0.3359.117 (Official Build) (64-bit)). I forgot I had one installed. So it's not a dev channel problem. Have you tested in Chrome on Mac?

dpvc:
I was just getting ready to check it there.  I will see what I can find out.

PR #104, opened 2018-05-04
Fixes filename and spacing problems.

zorkow:
* Captilisation in mathml.js filename.
* Trims input string that leads to an error for me.

### DIFF
--- a/lib/Mml-lab.js
+++ b/lib/Mml-lab.js
@@ -1,4 +1,4 @@
-import {MathML} from "mathjax3/input/MathML.js";
+import {MathML} from "mathjax3/input/mathml.js";
 import {CHTML} from "mathjax3/output/chtml.js";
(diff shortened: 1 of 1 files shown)


lib/Mml-lab.js
```diff
@@ -55,7 +55,7 @@ const Lab = window.Lab = {
 }
 
 if (window.location.search !== "") {
-  Lab.mml.value = decodeURIComponent(window.location.search.substr(1));
+  Lab.mml.value = decodeURIComponent(window.location.search.substr(1)).trim();
```
dpvc: Should a similar change be made in the `TeX-lab.js` file?

zorkow: No, that's not necessary as the texparser deals with whitespace surrounding an expression.
On the other hand, we could possibly move the whitespace trimming into the mathml.compile method.

PR #106, opened 2018-05-08
Add support for menclose

dpvc:
This PR adds support for the `menclose` element, including all the notations defined in the MathML spec.  (Version 2.x doesn't do all the arrows, but this version includes them all.)  It would be possible to add additional notations using `CHTMLmenclose.Notations.add()` if you wanted to create non-standard ones.  Also, the padding, line thickness, and arrowhead values can be modified using `data-padding`, `data-thickness` and `data-arrowhead` attributes on the `menclose` element (version 2.x used non-standard `padding` and `thickness` attributes, but these should be `data-*` attributes instead; I don't recall if the TeX input jax ever set these non-standard attributes).

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/menclose.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/menclose.ts
@@ -419,7 +419,7 @@ export class CHTMLmenclose<N, T, D> extends CHTMLWrapper<N, T, D> {
         Notation.Border('left'),
 
         Notation.Border2('actuarial', 'top', 'right'),
-        Notation.Border2('madrawb', 'bottom', 'right'),
+        Notation.Border2('madruwb', 'bottom', 'right'),
 
         Notation.DiagonalStrike('up', 1),
         Notation.DiagonalStrike('down', -1),
--- a/mathjax3-ts/output/chtml/Wrappers/menclose.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/menclose.ts
@@ -1,0 +1,951 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2018 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  Implements the CHTMLmenclose wrapper for the MmlMenclose object
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+import {CHTMLWrapper, StringMap} from '../Wrapper.js';
+import {CHTMLWrapperFactory} from '../WrapperFactory.js';
+import {CHTMLmsqrt} from './msqrt.js';
+import {BBox} from '../BBox.js';
+import {MmlMenclose} from '../../../core/MmlTree/MmlNodes/menclose.js';
+import {MmlNode, AbstractMmlNode, AttributeList} from '../../../core/MmlTree/MmlNode.js';
+import {OptionList} from '../../../util/Options.js';
+import {StyleList, StyleData} from '../CssStyles.js';
+import {split} from '../../../util/string.js';
+
+/*****************************************************************/
+
+const ARROWX = 4, ARROWDX = 1, ARROWY = 2;  // default relative arrowhead values
+
+const THICKNESS = .067;  // default rule thickness
+const PADDING = .2;      // default padding
+
+const SOLID = THICKNESS + 'em solid';  // a solid border
+
+/*
+ * The functions used for notation definitions
+ */
+export type Renderer<N, T, D> = (node: CHTMLmenclose<N, T, D>, child: N) => void;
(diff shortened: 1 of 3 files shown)


mathjax3-ts/output/chtml/Wrappers/menclose.ts
```diff
@@ -0,0 +1,951 @@
...
+     */
+    public static Notations: NotationDefList = new Map([
+
+        Notation.Border('top'),
+        Notation.Border('right'),
+        Notation.Border('bottom'),
+        Notation.Border('left'),
+
+        Notation.Border2('actuarial', 'top', 'right'),
+        Notation.Border2('madrawb', 'bottom', 'right'),
```
pkra: Typo -- `madruwb`.

dpvc: Thanks!  Fixed.

mathjax3-ts/output/chtml/Wrappers/menclose.ts
```diff
@@ -0,0 +1,951 @@
...
+export type BBoxExtender<N, T, D> = (node: CHTMLmenclose<N, T, D>) => number[];
+export type BBoxBorder<N, T, D> = (node: CHTMLmenclose<N, T, D>) => number[];
+export type Initializer<N, T, D> = (node: CHTMLmenclose<N, T, D>) => void;
+
+/*
+ * The definition of a notation
+ */
+export type NotationDef<N, T, D> = {
+    renderer: Renderer<N, T, D>;        // renders the HTML for the notation
+    bbox: BBoxExtender<N, T, D>;        // gives the offsets to the child bounidng box: [top, right, bottom, left]
```
zorkow: Typo

mathjax3-ts/output/chtml/Wrappers/menclose.ts
```diff
@@ -0,0 +1,951 @@
...
+     * The data for diagonal arrow notations
+     *   [neg, c, pi, double, origin, remove]
+     */
+    diagonalArrowDef: {
+        updiagonal: [ 1, -1, 0,       false, 'bottom left',  'updiagonalstrike'],
+        northeast:  [ 1, -1, 0,       false, 'bottom left',  'updiagonalstrike'],
+        southeast:  [-1,  1, 0,       false, 'top left',     'downdiagonalstrike'],
+        northwest:  [ 1,  1, Math.PI, false, 'bottom right', 'downdiagonalstrike'],
+        southwest:  [-1, -1, Math.PI, false, 'top right',    'updiagonalstrike'],
+        northeastsouthwest: [ 1, -1, 0, true, 'bottom left', 'updiagonalstrike northeastarrow updiagonalarrow southwestarrow'],
```
zorkow: linelength

mathjax3-ts/output/chtml/Wrappers/menclose.ts
```diff
@@ -0,0 +1,951 @@
...
+ * the spacing to leave for an arrowhead
+ */
+export const arrowHead = (node: CHTMLmenclose<any, any, any>) => {
+    return Math.max(node.padding, node.thickness * (node.arrowhead.x + node.arrowhead.dx + 1));
+};
+
+/*
+ * Service routines used for notations
+ */
+export const Notation = {
```
zorkow: Should that not just be a namespace with some of its functions exported?

mathjax3-ts/output/chtml/Wrappers/menclose.ts
```diff
@@ -0,0 +1,951 @@
...
+
+    /*
+     * Common BBox and Border functions
+     */
+    fullBBox: ((node) => new Array(4).fill(node.thickness + node.padding)) as BBoxExtender<any, any, any>,
+    fullPadding: ((node) => new Array(4).fill(node.padding)) as BBoxExtender<any, any, any>,
+    fullBorder: ((node) => new Array(4).fill(node.thickness)) as BBoxBorder<any, any, any>,
```
zorkow: Should all these really be exported? Or are they not just private to this module?

mathjax3-ts/output/chtml/Wrappers/menclose.ts
```diff
@@ -0,0 +1,951 @@
...
+     */
+    arrowDef: {
+        up:    [-Math.PI / 2,  1, false, 'bottom left', 'X', true,  'verticalstrike'],
+        down:  [ Math.PI / 2, -1, false, 'top left',    'X', true,  'verticakstrike'],
+        right: [  0,          -1, false, 'top left',    'Y', false, 'horizontalstrike'],
+        left:  [  Math.PI,    -1, false, 'top right',   'Y', false, 'horizontalstrike'],
+        updown:    [Math.PI / 2, -1, true, 'top left',  'X', true,  'verticalstrike uparrow downarrow'],
+        leftright: [0,           -1, true, 'top left',  'Y', false, 'horizontalstrike leftarrow rightarrow']
+    } as {[name: string]: [number, number, boolean, string, string, boolean, string]},
+
```
zorkow: ditto

mathjax3-ts/output/chtml/Wrappers/menclose.ts
dpvc: I suppose that might be better.  I will see about changing it.

dpvc:
> Do you envision the components ever going to be overwritten later?

Not overridden, but I do think that they could be augmented.

dpvc:
I've separated the `Notation` items out into a separate utility file, to make it easier to use if needed for subclasses or extensions.

I suspect that a number of the support files (like BBox, Font, Notation, etc) may be abstracted to ones that can be used by other output jax.  I've been thinking about the SVG output recently, and most of the CHTML wrappers would apply to that as well (with only the `toCHTML()` functions being replaced by a corresponding `toSVG` one).  So I suspect to abstract the Wrapper classes into ones that can be used as the basis for any output jax.  But that is for the future.

dpvc:
In any case, this is ready for re-review.

PR #107, opened 2018-05-08
Add support for mglyph

dpvc:
This PR adds preliminary support for mglyph elements.

Currently, you need to give explicit `width` and `height` attributes (they default to `1em` otherwise), because loading images may be asynchronous, and for node-based use, you might want to read from files rather than from the network to find the images, so that would require additional configuration and name mapping.  I'm going to leave those for future extensions.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/mglyph.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/mglyph.ts
@@ -93,7 +93,6 @@ export class CHTMLmglyph<N, T, D> extends CHTMLWrapper<N, T, D> {
         }
         const img = this.html('img', {src: src, style: styles, alt: alt, title: alt});
         this.adaptor.append(chtml, img);
-        this.drawBBox();
     }
 
     /*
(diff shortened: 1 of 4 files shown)


mathjax3-ts/output/chtml/Wrappers/mglyph.ts
```diff
@@ -0,0 +1,107 @@
...
+     * @override
+     * @constructor
+     */
+    constructor(factory: CHTMLWrapperFactory<N, T, D>, node: MmlNode, parent: CHTMLWrapper<N, T, D> = null) {
+        super(factory, node, parent);
+        this.getParameters();
+    }
+
+    /*
+     * Obtain the widt, height, and voffset.
```
zorkow: Typo

dpvc:
> I notice there is a remark re future extensions in the code. Do we have/need a central way to collate those?

I don't expect to have any others for now (all the MathML elements are now implemented in CHTML), but if there are some of yours that you want to keep track of, I suppose we could start a meta issue with the list of them.  Unless you have a better suggestion in mind?

PR #108, opened 2018-05-08
Add support for mathml spacing rules

dpvc:
Add support for mathml spacing rules.  This is now a configuration option to the CHTML output jax (`mathmlSpacing: true`).  You can also override the TeX spacing rules using explicit `lspace`, `rspace`, and `form` attributes on the MathML elements.


dpvc:
> does that affect my form computations in the TeX parser?

I don't think so, because you are creating the MathML yourself, not processing pre-existing MathML that might have an explicit `form` attribute.  Any `form` attribute will be one that you applied yourself, and if you don't apply `from` attributes, there will not be any difference.  

If you want me to check your usage, can you indicate the branch to check in and where it is taking place?

zorkow:
Got it, thanks. I thought that was meant as a way to explicitly remember the form attribute.

PR #110, opened 2018-05-09
Fix placement of table labels

dpvc:
Force label heights to size properly, and handle row spacing and row lines properly, as well as the placement of the baseline in the label itself.  The largest change is to move the code that sets the cell baseline into a separate routine so that it can be called on the label itself separately.

This should fix the problem the underlies the issue @zorkow raised F2F concerning a tagged split environment inside an align environment.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/mtable.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/mtable.ts
@@ -696,12 +696,11 @@ export class CHTMLmtable<N, T, D> extends CHTMLWrapper<N, T, D> {
         for (let i = 0; i < this.numRows; i++) {
             const row = this.childNodes[i];
             if (row.kind === 'mlabeledtr') {
-                if (h) {
-                    adaptor.insert(this.html('mjx-mtr', {style: {height: this.em(h)}}), current);
-                    adaptor.setStyle(current, 'height', this.em((equal ? HD : H[i] + D[i]) + space[i] + space[i+1]));
-                    current = adaptor.next(current) as N;
-                    h = 0;
-                }
+                h && adaptor.insert(this.html('mjx-mtr', {style: {height: this.em(h)}}), current);
+                const cell = adaptor.firstChild(current) as N;
+                adaptor.setStyle(cell, 'height', this.em(equal ? HD : H[i] + D[i]));
+                current = adaptor.next(current) as N;
+                h = this.rLines[i];
             } else {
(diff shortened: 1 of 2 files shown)


zorkow:
I found another problem, stemming from the multline environment. Here the cell containing _r_ is set too far on the right (or the label too far to the left).

```XML
<math display="block">
  <mtable columnwidth="100%" width="85%" rowspacing=".5em" displaystyle="true">
    <mtr>
      <mtd columnalign="left">
        <mi>l</mi>
      </mtd>
    </mtr>
    <mtr>
      <mtd>
        <mi>c</mi>
      </mtd>
    </mtr>
    <mlabeledtr>
      <mtd id="mjx-eqn-1">
        <mtext>(1)</mtext>
      </mtd>
      <mtd columnalign="right">
        <mi>r</mi>
      </mtd>
    </mlabeledtr>
  </mtable>
</math>
```

zorkow:
For ease of testing I have just pushed branch `tag_structure`. It does AMS labelling by default. (Referencing does not work yet). 

Also, I am finally very happy with my new tagging structures. For example, take this pathological case:
```LaTeX
\begin{multline}
  a\\
  \tag{B} 
  \begin{split} b\\ c\end{split}\\
 d
\end{multline}
```
v3 now gives `(B)` as a label, which is the same as LaTeX output. v2 gives the default numerical label.
I am very excited about that.

dpvc:
@zorkow:

> I am very excited about that.

It sounds like you have done some major cleanup of the wonky code from v2.  Thanks!

> I found another problem, stemming from the multline environment.

I've added a new commit that changes how the spacing is handled.  It changes how the space for the tags is reserved, and also changes how percentage width tables are handled.  In the past, the percentage was of the full width of the space available for the equation, but now it is of the space minus the space used by the tag.  That will prevent the table from overlapping the tags.  In the past, if you used `width="100%"`, that would take the full width, and overlap the tags.

This may mean that the current definition for the table produced by the multline environment should be changed.  It uses `width="85%"`, which was a hack to get the environment to leave a little space on the side, like it does in LaTeX, but it was always a hack.  The space should be a fixed amount, not a percentage, as it is this way.

One solution would be to use `<mtable width="100%" style="padding: 0 2.5em">`, since this PR makes handling of the padding CSS style adds that space around the table.  We have so far avoided using `style` in any important way, however, so this breaks the practice of using only native MathML attributes.  Alternatively, we could use a custom `data-padding` attribute (or something similar).  or we could just stick with the percentage, but that will mean that the spacing varies with the size of the container, as well as with the presence/absence of labels.

What do you think?

PR #114, opened 2018-05-13
Maction events

dpvc:
This PR adds support for the event handlers needed for some maction types.

As part of this, the `addEventHandlers()` method has been removed from the MathDocument (and associated methods in other objects that were to support it) in favor of a `postFilter` function list that gets run after the typesetting is finished.  That can be used to add more events (like the menu events, or zoom events, or explorer events), and seemed to be more general.  (Volker also complained at one point that this was not a reasonable method to have for the MathDocument in general, since it might not even support event handlers.)

There are a number of other smaller changes, like the addition of `MathItem.rerender()`, which will re-typeset a math item (this is used by the `toggle` action to serenader the expression when it is clicked).  Some fixes were need in the `updateDocument()` and `removeFromDocument()` methods to get this to work properly.

The MathML and TeX lab script files were updated to take advantage of these changes.

The DOMAdaptor had two new methods added (`setStatus()` and `getStatus()`) to set and get the document's status line (as needed for the `statusline` action type).  But since browsers no longer allow that to be  changed under javascript control (and many don't even should one in any case), the implementation of these methods set and get the document title instead.  This could be changed to use a separate overlay element, like the one in version 2.7, or it could be disabled entirely.

### DIFF
--- a/lib/TeX-lab.js
+++ b/lib/TeX-lab.js
@@ -32,7 +32,7 @@ const Lab = window.Lab = {
       math.compile();
       math.typeset(doc);
       math.updateDocument(doc);  
-    }).catch(err => {console.log("Error: "+err.message); console.log(err.stack)});
+    }).catch(err => {console.log("Error: " + err.message); console.log(err.stack)});
   },
   
   Keep() {
--- a/lib/TeX-lab.js
+++ b/lib/TeX-lab.js
@@ -29,7 +29,7 @@ const Lab = window.Lab = {
     handleRetriesFor(function () {
       math.compile();
       math.typeset(doc);
-    }).then(() => Lab.Update(math.typesetRoot.outerHTML))
+    }).then(() => Lab.Update(math.typesetRoot))
       .catch(err => {console.log("Error: "+err.message); console.log(err.stack)});
   },
   
@@ -38,7 +38,8 @@ const Lab = window.Lab = {
   },
   
   Update(html) {
-    this.output.innerHTML = html;
+    this.output.innerHTML = '';
+    this.output.appendChild(html);
   },
   
   setDisplay(checked) {
--- a/lib/TeX-lab.js
+++ b/lib/TeX-lab.js
@@ -1,17 +1,14 @@
 import {TeX} from "mathjax3/input/tex.js";
 import {CHTML} from "mathjax3/output/chtml.js";
-import {AbstractMathItem} from "mathjax3/core/MathItem.js";
-import {AbstractMathDocument} from "mathjax3/core/MathDocument.js";
+import {HTMLMathItem} from "mathjax3/handlers/html/HTMLMathItem.js";
+import {HTMLDocument} from "mathjax3/handlers/html/HTMLDocument.js";
 import {handleRetriesFor} from "mathjax3/util/Retries.js";
 import {browserAdaptor} from "mathjax3/adaptors/browserAdaptor.js";
-
-class LabMathItem extends AbstractMathItem {};
-class LabMathDocument extends AbstractMathDocument {};
 
 let tex = new TeX();
 let chtml = new CHTML();
 
-let doc = new LabMathDocument(document, browserAdaptor(), {InputJax: tex, OutputJax: chtml});
+let doc = new HTMLDocument(document, browserAdaptor(), {InputJax: tex, OutputJax: chtml});
 document.head.appendChild(chtml.styleSheet(doc));
 
 const Lab = window.Lab = {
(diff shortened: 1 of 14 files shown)


zorkow:
> But since browsers no longer allow that to be changed under javascript control (and many don't even
>  should one in any case),

Right, that answers my question in the PR #109 . One could hack something with a faked href link. But that'd be ugly.

lib/TeX-lab.js
```diff
@@ -20,27 +17,28 @@ const Lab = window.Lab = {
...
+    math.end = {node: text, n: 0, delim: ''};
     this.jax = math;
 
     handleRetriesFor(function () {
       math.compile();
       math.typeset(doc);
-    }).then(() => Lab.Update(math.typesetRoot.outerHTML))
-      .catch(err => {console.log("Error: "+err.message); console.log(err.stack)});
+      math.updateDocument(doc);  
+    }).catch(err => {console.log("Error: "+err.message); console.log(err.stack)});
```
zorkow: whitespace

mathjax3-ts/core/OutputJax.ts
```diff
@@ -154,4 +162,19 @@ export abstract class AbstractOutputJax<N, T, D> implements OutputJax<N, T, D> {
...
     }
 
+    /*
+     * Execute a set of filters, passing them the MathItem and any needed data,
+     *  and return the (possibly modified) data
+     *
+     * @param{FunctionList} filters  The list of functions to be performed
+     * @param{MathItem} math         The math item that is being processed
+     * @param{any} data              Whatever other data is needed
+     * @return{any}                  The (possibly modidied) data
```
zorkow: typo

zorkow:
> ... the implementation of these methods set and get the document title instead. 

I find that very peculiar when suddenly the title of the tab changes. On the other hand often the tab is too small to notice the change in the first place.
Do we still want to support this? This is part of a spec that simply has not kept up with time.
Personally I feel that the tooltip event serves the same purpose.

mathjax3-ts/core/DOMAdaptor.ts
```diff
@@ -87,6 +87,18 @@ export interface DOMAdaptor<N, T, D> {
      */
     root(doc: D): N;
 
+    /*
+     * @param{D} doc    The document whose statusline is to be obtained
+     * @return{string}  The document's status line (title)
```
zorkow: status line or statusline

mathjax3-ts/adaptors/liteAdaptor.ts
```diff
@@ -112,6 +112,20 @@ export class LiteAdaptor extends AbstractDOMAdaptor<LiteElement, LiteText, LiteD
...
+        return '';  // statusline is not maintained
+    }
+
+    /*
+     * @override
+     */
+    public setStatus(doc: LiteDocument, status: string) {
+        // statusline is not maintained
+    }
+
```
zorkow: Do we really want those? See my earlier comment.

mathjax3-ts/core/OutputJax.ts
```diff
@@ -154,4 +162,19 @@ export abstract class AbstractOutputJax<N, T, D> implements OutputJax<N, T, D> {
         return null as N;
     }
 
+    /*
+     * Execute a set of filters, passing them the MathItem and any needed data,
+     *  and return the (possibly modified) data
+     *
+     * @param{FunctionList} filters  The list of functions to be performed
+     * @param{MathItem} math         The math item that is being processed
+     * @param{any} data              Whatever other data is needed
```
zorkow: Should those be rest arguments ...data: any[]?

mathjax3-ts/output/chtml/Wrappers/maction.ts
```diff
@@ -78,38 +82,133 @@ export class CHTMLmaction<N, T, D> extends CHTMLWrapper<N, T, D> {
...
+            const document = node.factory.chtml.document;
+            const mml = node.node as MmlMaction;
+            //
+            // Add a click handler that changes the selection and rerenders the expression
+            //
+            node.setEventHandler('click', (event: Event) => {
+                if (!math.start.node) {
+                    //
+                    // If the MathItem was created by hand, it might not have a node
+                    // telling it where to replace the existing math, so set it.
```
zorkow: I don't understand that.

mathjax3-ts/core/OutputJax.ts
dpvc: No, the data is used as part of an object in the function below, and if it were `...data: any[]`, then that would always be an array, and the functions that are being executed would have to be changed to deal with the array (instead of just the data item as they currently are).  If you want an array of items, then you can pass an array as `data`.  I don't want to force it to be an array (in fact, at the moment, only one item is passed, so forcing an array would make things harder to use).

mathjax3-ts/output/chtml/Wrappers/maction.ts
dpvc: We discussed this face-to-face, but just or the record, there is a common use case where you create a `MathItem` by hand (for things like converting a TeX string to an HTML string), and that `MathItem` is not attached to a document directly.  This is done in the TeX and MathML labs, for instance, where the TeX or MathML is converted without it being part of a document, and the resulting HTML node is inserted into the page.  

This takes care of setting the node in those cases, so that a toggle action can remove the old HTML and replace it by the newly re-rendered HTML.

mathjax3-ts/adaptors/liteAdaptor.ts
dpvc: OK, I'll reimplement as a fixed-position display div, like version 2 uses.

PR #115, opened 2018-05-14
Update to work with node 10's esm module rather than system.js

dpvc:
This PR adds a `load.mjs` file that will trigger node's esm module (in version 10+) so that ES6 files can be loaded directly into node without the need of `system.js`.  That is, you can use `import` and `export` directly without down compiling.  I tested it with the `ES6` target for Typescript, and it works nicely, though I haven't changed that in this PR.  The only real changes needed was to make sure the imports use relative file names rather than root-level filenames.

### DIFF
--- a/mathjax2/legacy/extensions/TeX/AMSmath.js
+++ b/mathjax2/legacy/extensions/TeX/AMSmath.js
@@ -24,9 +24,9 @@
  *  limitations under the License.
  */
 
-let sm = require('mathjax3/input/tex/SymbolMap.js');
-let tc = require('mathjax3/input/tex/TexConstants.js');
-let BaseMethods = require('mathjax3/input/tex/BaseMethods.js').default;
+let sm = require('../../../../mathjax3/input/tex/SymbolMap.js');
+let tc = require('../../../../mathjax3/input/tex/TexConstants.js');
+let BaseMethods = require('../../../../mathjax3/input/tex/BaseMethods.js').default;
 
 
 MathJax.Extension["TeX/AMSmath"] = {
--- a/load.mjs
+++ b/load.mjs
@@ -1,0 +1,2 @@
+require = require("esm")(module/*, options*/)
+module.exports = require('./' + (process.argv[2] || 'main.js'));
--- a/mathjax2/legacy/extensions/TeX/AMSsymbols.js
+++ b/mathjax2/legacy/extensions/TeX/AMSsymbols.js
@@ -24,9 +24,9 @@
  *  limitations under the License.
  */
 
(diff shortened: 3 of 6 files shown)


dpvc:
Just to be clear, use

    node load.mjs samples/tex2html.js 'x-y'

to run the `tex2html` sample file using the `esm` loader rather than `system.js`, or use

    node -r esm samples/tex2html.js 'x-y'

to run the sample file directly (no need for a load file at all).

Note that all the samples can still be run in the browser using the `load.html` file as before.  E.g., open

    load.html?samples/tex2html?x-y

to run the same sample file in the browser.  (In this case, `system.js` is used, as before.)

zorkow:
>        node -r esm samples/tex2html.js 'x-y'

That does not quite work, because the expression is now the second argument and the scripts expect it at third argument position. E.g.,

         node -r esm samples/tex2html.js '' 'x-y'

works fine instead. Not sure what we can do, apart from making the position in `process.argv` conditional on whether or not `esm` is loaded.

mathjax2/legacy/extensions/TeX/AMSmath.js
```diff
@@ -24,9 +24,9 @@
  *  limitations under the License.
  */
 
-let sm = require('mathjax3/input/tex/SymbolMap.js');
-let tc = require('mathjax3/input/tex/TexConstants.js');
-let BaseMethods = require('mathjax3/input/tex/BaseMethods.js').default;
+let sm = require('../../../../mathjax3/input/tex/SymbolMap.js');
+let tc = require('../../../../mathjax3/input/tex/TexConstants.js');
+let BaseMethods = require('../../../../mathjax3/input/tex/BaseMethods.js').default;
```
jdalton: Hi @dpvc!

Can you explain the issue you had with the paths above a bit more?

dpvc: @jdalton,

Because v3 might be included into an application via `npm` (node package manager), the top-level directory might be `node_modules/mathjax-v3/mathjax3` or some such thing.  Using relative directories makes MathJax v3 able to be used from within whatever folder hierarchy the author is using, rather than expecting it to be in a particular location as the original form does.

dpvc: @jdalton, with `node -r esm samples/tex2html.js` perhaps the root directory is considered to be `./samples` since that is where `tex2html.js` is being loaded?  So since `mathjax3` is in the main directory (a sibling of `samples`) and not in `samples` itself, it isn't being found?

jdalton: I think this may be an issue on our end. Are you using symlinks?

PR #117, opened 2018-05-16
Scale math to match surrounding text

dpvc:
This PR implements the `getMetrics()` method that determines the metrics of the surrounding font, and computes the proper scaling for that.  It computes the `em` and `ex` sizes for the surroundings, though it does not yet compute the container width (which is needed for line breaking).

For the liteAdaptor, the font size is a fixed size set by a configuration parameter.  If the ex size can't be computed (e.g., in the lite and jsdom adaptors), a fixed percentage of the em size is used (it can be set in an option for the CHTML output jax).

### DIFF
--- a/mathjax3-ts/adaptors/liteAdaptor.ts
+++ b/mathjax3-ts/adaptors/liteAdaptor.ts
@@ -29,12 +29,25 @@ import {LiteList} from './lite/List.js';
 import {LiteWindow} from './lite/Window.js';
 import {LiteParser} from './lite/Parser.js';
 import {Styles} from '../util/Styles.js';
+import {userOptions, defaultOptions, OptionList} from '../util/Options.js';
 
 /************************************************************/
 /*
  * Implements a lightweight DOMAdaptor on liteweight HTML elements
  */
 export class LiteAdaptor extends AbstractDOMAdaptor<LiteElement, LiteText, LiteDocument> {
+    /*
+     * The default options
+     */
+    public static OPTIONS: OptionList = {
+        fontSize: 16,      // we can't compute the font size, so always use this
+    };
+
+    /*
+     * The options for the instance
+     */
+    public options: OptionList;
+
     /*
      * The document in which the HTML nodes will be created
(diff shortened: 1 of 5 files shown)


mathjax3-ts/adaptors/liteAdaptor.ts
```diff
@@ -499,12 +515,27 @@ export class LiteAdaptor extends AbstractDOMAdaptor<LiteElement, LiteText, LiteD
...
+        return [0, 0] as [number, number];
+    }
+
 }
 
 /************************************************************/
 /*
  * The function to call to obtain a LiteAdaptor
  */
-export function liteAdaptor() {
```
zorkow: @param comment missing

mathjax3-ts/core/DOMAdaptor.ts
```diff
@@ -293,6 +293,18 @@ export interface DOMAdaptor<N, T, D> {
...
+
+    /*
+     * @param{N} node        The HTML node whose font size is to be determined
+     * @return{number}       The font size (in pixels) of the node
+     */
+    fontSize(node: N): number;
+
+    /*
+     * @param{N} node        The HTML node whose dimensions are to be determined
+     * @return{number}       The width and height (in pixels) of the element
```
zorkow: [number, number]

mathjax3-ts/output/chtml.ts
```diff
@@ -147,9 +159,91 @@ export class CHTML<N, T, D> extends AbstractOutputJax<N, T, D> {
...
+            const parent = adaptor.parent(math.start.node);
+            const {em, ex, containerWidth, lineWidth, scale} = map.get(parent);
+            math.setMetrics(em, ex, containerWidth, lineWidth, scale);
+        }
+    }
+
+    /*
+     * Get a MetricMap for the math list
+     *
+     * @return{MetricMap}   The node-to-metrics map for all the containers that have math
```
zorkow: Missing @param

PR #118, opened 2018-05-16
Add support for indentalign and indentshift

dpvc:
This PR adds support for `indentalign` and `indentshift`, which control the alignment and indenting of displayed equations.  This is handled in the `math` wrapper, and also in the `mtable` wrapper since tables with labels are in full-width elements, and the alignment of the actual table must be inside the full-width container.  Stylesheet CSS handles the alignment, but the shifting requires explicit styles.  Because tables rely on the `indentalign` and `indentshift` values, they need to be added to the mtable inheritance.

This PR also includes fixes for the bubbling up of full-width element, like tables with labels.  So if a table is inside another element (e.g., `mstyle`, or `semantics`), that won't disrupt the label positioning.  This was being done by hand in `mrow`elements, but that has now been moved to the wrapper super-class.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrapper.ts
+++ b/mathjax3-ts/output/chtml/Wrapper.ts
@@ -23,7 +23,7 @@
 
 import {AbstractWrapper} from '../../core/Tree/Wrapper.js';
 import {Node, PropertyList} from '../../core/Tree/Node.js';
-import {MmlNode, TextNode, AbstractMmlNode, AttributeList} from '../../core/MmlTree/MmlNode.js';
+import {MmlNode, TextNode, AbstractMmlNode, AttributeList, indentAttributes} from '../../core/MmlTree/MmlNode.js';
 import {Property} from '../../core/Tree/Node.js';
 import {OptionList} from '../../util/Options.js';
 import {unicodeChars} from '../../util/string.js';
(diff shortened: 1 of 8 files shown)


mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -659,6 +678,43 @@ export class CHTMLWrapper<N, T, D> extends AbstractWrapper<MmlNode, CHTMLWrapper
...
+        }
+        if (indentshift === 'auto') {
+            indentshift = '0';
+        }
+        const shift = this.length2em(indentshift, this.metrics.containerWidth);
+        return [indentalign, shift] as [string, number];
+    }
+
+    /*
+     * @param{N} chtml       The HTML node whose indenting is to be adjusted
```
zorkow: indenting -> indentation ?

mathjax3-ts/output/chtml/Wrappers/math.ts
```diff
@@ -69,9 +75,20 @@ export class CHTMLmath<N, T, D> extends CHTMLWrapper<N, T, D> {
      */
     public toCHTML(parent: N) {
         super.toCHTML(parent);
-        if (this.node.attributes.get('display') === 'block') {
-            this.adaptor.setAttribute(this.chtml, 'display', 'true');
-            this.adaptor.addClass(parent, 'MJX-DISPLAY');
+        const chtml = this.chtml;
+        const adaptor = this.adaptor;
+        const attributes = this.node.attributes;
+        const display = (attributes.get('display') === 'block');
```
zorkow: Why parentheses?

dpvc: I've always preferred them when assigning a variable the result of a boolean operation.  It makes the order of operations explicit.  I can remove if you think it is important, but we discussed this once before, and you indicated that it was OK and that we could make it part of our style definition if I wanted.

PR #146, opened 2018-09-24
Implement SVG output

dpvc:
This PR adds support for the SVG output, plus support in CHTML for characters not in the MathJax fonts.  It also updates the labs to allow selecting of the output renderer, to allow visual testing.

The first few files are the lab updates.

Then there are changes to the DOM adaptors needed to allow measuring of DOM elements to support handling of unknown characters or characters from other fonts.  

Then are the changed CHTML files for the unknown character support, and a change to CHTML's maction to allow specifying the offset for math tooltips.

Then are the changes to the common files to support unknown characters (and the offsets for tooltips).

Then are the new SVG files for the output jax and all the MathML element wrappers and support files.  At the end are the font files that add the SVG path data to the common fonts.

It is a lot of files, but there are a lot of MathML element to implement, and quite a few font files (that you can ignore).


pkra:
> At the end are the font files that add the SVG path data to the common fonts.

FYI all modern browsers support `<use>` elements that are pointing to external SVGs (and polyfills are available for the rest, cf. https://github.com/jonathantneal/svg4everybody#implementation-status). It would be nice to have at least the option to easily compile webfont-like SVGs out of the source.

dpvc:
> It would be nice to have at least the option to easily compile webfont-like SVGs out of the source.

I'm not sure what you are asking for.  Are you using to be able to make an SVG web font from the MathJax fonts?  Or do you mean the typeset expressions into webfoot-like SVGs?  Or is it that you want to use external SVG fonts for the characters within the typeset expression itself?

pkra:
> I'm not sure what you are asking for.

I mean having one SVG per webfont files that contains a defs with all the paths corresponding to the glyphs in a matching "real" webfont. 

Then the SVG output could get passed a URL much like the CSS output gets passed webfont URL and the result is `<use xlink:href="path/to/MathJax_Main-Bold.svg#pathID"/>` -- but it's not an SVG font, just an SVG with paths that have IDs.

https://github.com/jonathantneal/svg4everybody#usage gives a basic overview over how this kind of SVG spriting works today.

pkra:
> but it's not an SVG font, just an SVG with paths that have IDs.

I realize this is a bit weird to differentiate. Just imagine the global SVG that's currently generated, but it's stored as an SVG, located on `https://example.com/global.svg` and the use elements pointing to `https://example.com/global.svg#pathID`.

hubgit:
I've tested this branch and (without having checked the output too precisely) it seems to work nicely.

dpvc:
@hubgit, thanks for the confirmation!

mathjax3-ts/output/common/OutputJax.ts
```diff
@@ -369,4 +381,103 @@ try {
...
+        if (bold) styles['font-style'] = 'bold';
+        return styles;
+    }
+
+    /**
+     * @param {Styles} styles   The style object to query
+     * @return {CssFontData}    The family, italic, and boolean values
+     */
+    public getFontData(styles: Styles) {
+        if (!styles) styles = new Styles();
```
zorkow: Missing braces.

mathjax3-ts/output/common/Wrappers/maction.ts
```diff
@@ -142,6 +156,20 @@ export function CommonMactionMixin<W extends AnyWrapper,
...
             this.action = handler;
             this.data = data;
+            this.getParameters();
+        }
+
+        /**
+         * Look up attribute parameters
+         */
+        public getParameters() {
+            let [dx, dy] = ['', ''];
```
zorkow: Remove.

mathjax3-ts/output/common/Wrappers/maction.ts
```diff
@@ -142,6 +156,20 @@ export function CommonMactionMixin<W extends AnyWrapper,
...
+            const offsets = this.node.attributes.get('data-offsets') as string;
+            if (offsets) {
+                [dx, dy] = split(offsets);
```
zorkow:
```suggestion
               let [dx, dy] = offsets ? split(offsets) : ['', ''];
```

mathjax3-ts/output/svg.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /**
+     * @param {number} m  A number to be shown in ex
+     * @return {string}   The number with units of ex
+     */
+    ex(m: number) {
+        m /= this.font.params.x_height;
+        if (Math.abs(m) < .001) return '0';
+        return (m.toFixed(3).replace(/\.?0+$/, '')) + 'ex';
```
zorkow: Combine into single return.

mathjax3-ts/output/svg/Notation.ts
```diff
@@ -0,0 +1,131 @@
...
+    return ({
+        top: [0, h - t, w, h - t],
+        right: [w - t, -d, w - t, h],
+        bottom: [0, t - d, w, t - d],
+        left: [t, -d, t, h],
+        vertical: [w / 2 - t, h, w / 2 - t, -d],
+        horizontal: [0, (h - d) / 2 - t, w, (h - d) / 2 - t],
+        up: [t, t - d, w - t, h - t],
+        down: [t, h - t, w - t, t - d]
+    } as {[name: string]: [number, number, number, number]})[kind];
```
zorkow: Here's a lot of computation being done that is not necessary. In every call you need exactly to compute the one for `kind`. I'd suggest a table mapping `kind` to methods that of type `(number,  number, number, number): [number, number, number, number]`. And then have `return table[kind](h, d, w, t)`.

This way you can also avoid duplicate computations. In particular, divisions are costly.

mathjax3-ts/output/svg/Wrapper.ts
```diff
@@ -0,0 +1,424 @@
...
+        }
+        if (align === 'center' || align === 'right') {
+            // FIXME
+        }
+    }
+
+    /**
+     * @param {number} x   The x-offset for the element
+     * @param {number} y   The y-offset for the element
+     * @param {N} eleemnt  The element to be placed
```
zorkow: typo

mathjax3-ts/output/svg/Wrapper.ts
```diff
@@ -0,0 +1,424 @@
...
+     */
+    public place(x: number, y: number, element: N = null) {
+        if (!element) {
+            element = this.element;
+        }
```
zorkow: This is only necessary if `x || y` holds. So why not have 
```
if (!(x || y)) return;
```
and then have the logic of the method?

mathjax3-ts/output/svg/Wrapper.ts
```diff
@@ -0,0 +1,424 @@
...
+     * @param {number} h        The height to be aligned
+     * @param {number} d        The depth to be aligned
+     * @param {string} align    How to align (top, bottom, middle, axis, baseline)
+     * @return {number}         The y position of the aligned baseline
+     */
+    protected getAlignY(H: number, D: number, h: number, d: number, align: string) {
+        if (align === 'top') return H - h ;
+        if (align === 'bottom') return d - D;
+        if (align === 'middle') return ((H - h) - (D - d)) / 2;
+        return 0; // baseline and axis
```
zorkow: Braces in the if statements.

mathjax3-ts/output/svg/Wrapper.ts
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    /*******************************************************************/
+    /**
+     * For debugging
```
zorkow: Maybe we should have a naming convention for methods that should eventually be removed.

mathjax3-ts/output/svg/Wrappers/menclose.ts
```diff
@@ -0,0 +1,343 @@
...
+            //
+            // Use a bottom border and an upward strike properly angled
+            //
+            renderer: (node, child) => {
+                const {w, h, d} = node.getBBox();
+                const [a, W] = node.getArgMod(1.75 * node.padding, h + d);
+                const t = node.thickness / 2;
+                const HD = h + d;
+                node.adaptor.append(node.element, node.path('mitre',
+                    'M', w, t - d,  'L', t + Math.cos(a) * t, t - d,  'L' , Math.cos(a) * HD + t, HD - d - t
```
zorkow: Compute `Math.cos(a)` only once.

mathjax3-ts/output/svg/Wrappers/menclose.ts
```diff
@@ -0,0 +1,343 @@
...
+                 'L', w + dw - x, m - t / 2,                // lower side of shaft
+                 'l', -dx, t / 2 - y, 'Z'                   // lower head
+             ) :
+             this.fill(
+                 'M', w + dw, m,                            // point of arrow
+                 'l', -(x + dx), y,  'l', dx, t / 2 - y,    // upper head
+                 'L', -dw, m + t / 2, 'l', 0, -t,           // upper side of shaft
+                 'L', w + dw - x, m - t / 2,                // lower side of shaft
+                 'l', -dx, t / 2 - y, 'Z'                   // lower head
+             ));
```
zorkow: Compute `t / 2` only once.

mathjax3-ts/output/svg/Wrappers/mtable.ts
```diff
@@ -0,0 +1,360 @@
...
+        const rLines = [this.fLine, ...this.rLines, this.fLine];
+        let y = this.getBBox().h - rLines[0];
+        for (let i = 0; i < this.numRows; i++) {
+            const row = this.childNodes[i];
+            [row.H, row.D] = this.getRowHD(equal, HD, H[i], D[i]);
+            [row.tSpace, row.bSpace] = [rSpace[i], rSpace[i+1]];
+            [row.tLine, row.bLine] = [rLines[i], rLines[i+1]];
+            row.toSVG(svg);
+            row.place(0, y - rSpace[i] - row.H);
+            y -= rSpace[i] + row.H + row.D + rSpace[i+1] + rLines[i+1];
```
zorkow: Spaces around `i + 1`.

mathjax3-ts/output/svg/Wrappers/mtable.ts
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    /******************************************************************/
+
+
+    /**
+     * Handle addition of labels to the table
+     *
+     * @param {N} svg     The container for the table contents
+     * @param {N} parent  The parent containing the the table
```
zorkow: Remove `the`

mathjax3-ts/output/svg/Wrappers/mtable.ts
```diff
@@ -0,0 +1,360 @@
...
+                y -= row.D + space[i + 1] + this.rLines[i];
+                current = adaptor.next(current) as N;
+            } else {
+                y -= space[i] + row.H + row.D + space[i + 1] + this.rLines[i];
+            }
+        }
+    }
+
+    /**
+     * @param {N} svg         The SVG container for the table
```
zorkow: What does the method do?

mathjax3-ts/output/svg/Wrappers/scriptbase.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    public static kind = 'scriptbase';
+
+    /**
+     * Set to true for munderover/munder/mover/msup (Appendix G 13)
+     */
+    public static useIC: boolean = false;
+
+    /**
+     * This gives the common output for msub and msup.  It is overriden
```
zorkow: `overridden`

dpvc:
@pkra, I'm sorry, I realized I never got back to you on your comments here.

> The same benefit that webfonts provide compared to the old PNG fonts -- instead of tracking 30,000 paths you only have a few dozen files.

But those few dozen files still need to have the 30,000 paths.  And I don't think the PNG fonts are an appropriate comparison, since those needed multiple png files for each glyph (one in each of 8 or so sizes), and that is not the case here.  Here, the 30,000 paths are still in only a dozen or so files (not thousands of separate files as for png fonts), they are just javascript files instead of SVG files.

> it aligns better with development workflows to have such resources separated, makes reasoning about them easier, simplifies caching in complex situations (e.g., across multiple document).

I'm willing to concede that separating resources is generally a good idea, but I'm not sure what "aligns better with development workflows" means in this case unless you are planning to re-use the SVG files in other contexts.  Even if we did put the paths into an SVG file, it would not be very reusable, as it would not be a full SVG font file (it wouldn't include the bounding box data, for example, which is part of an SVG font, as I recall).

Also, I'm not sure how it "makes reasoning about them easier", since you would not need to reason about them at all if they are handled by MathJax directly, as they are now, but would if they were separate files (that needed to be included in your production workflow).  Perhaps I'm missing the point.

Although you are right that external SVG files could improve caching of the paths, this would also make the URLs for the paths much longer (you need a full URL to the font file for each glyph), and so will make the page content longer.  I suspect that any savings in caching the font files themselves will be offset by the longer URLs, though of course it will depend on the content and how many equations there are.  But it will certainly make the non-cached page content longer.

Also, with the v3 components approach, where components can be loaded as needed (rather than packaging everything in one large webpack), much like v2 does now, you can regain the caching because the font pieces can be separate components that are cached separately.

> shipping large chunks of JS instead of simple SVG files does not seem ideal.

I'm not sure you have the comparison correct, as I don' think "large chunks of JS" and "simple SVG files" accurately describe the situation.  If you look at the SVG output jax data files like [this one](https://github.com/mathjax/mathjax-v3/blob/svg/mathjax3-ts/output/svg/fonts/tex/smallop.ts), you will see the main content of the file is made up of lines that look like

```
    0x5B: '202 -349V850H394V810H242V-309H394V-349H202',
```

consisting of a HEX number and (most of) the path for that number.  For the corresponding SVG path file, you would need

```
    <path id="5B" d="M202 -349V850H394V810H242V-309H394V-349H202Z"></path>
```

The extra data around the path in the JS file is `0x5B: '` and `',` (8 characters), whereas for the SVG, you need `<path id="5B" d="M` and `Z"></path>` (28 characters).  So that's a difference of 20 characters for each glyph.  If there are 30,000 glyphs, that means an additional 600K of data, so the SVG files are in general *larger* than the JS files.  And all that data needs to be shipped to the browser when these files are used.  In addition to the longer URLs to access the paths, that makes for a significant additional transfer load.

> As a comparison, the CSS output also doesn't inline its webfont data in the JS/TS code base

I've considered doing that as well (and had a version of this in an early v2 branch to try to get around the same-origin policy in Firefox for `file://` URLs).  One could certainly make an extension to do that, if desired.

> as I've argued elsewhere I think the CSS would be better to have outside the main JS/TS codebase

One of the reasons to do it in the code and not in separate CSS files is that it allows page authors to substitute characters or modify the font mappings more easily, without having to provide additional CSS files (or changed CSS files) or additional SVG files.  One can always export the CSS to a static file and disable the generation of the font CSS if you prefer that route; it is much harder to go the other way if MathJax used static CSS files by default.

But, with that being said, it is still a reasonable idea for an extension for those who want to have external SVG files, as I mentioned earlier.

pkra:
Thanks for the response. I disagree both from a philosophical and technical perspective.

I think it comes down to our old debates around making server-side rendering the priority and about making developer/author use casess a priority. 

Anyway, happy to agree to disagree.

pkra:
Sorry, accidentally hit close.

PR #150, opened 2018-10-03
Corrects the reset of tag structures.

zorkow:


mathjax3-ts/input/tex/Tags.ts
```diff
@@ -405,9 +405,10 @@ export class AbstractTags implements Tags {
   /**
    * @override
    */
-  public reset(n: number, keepLabels: boolean) {
-    this.offset = (n || 0);
+  public reset(n: number = 0, keepLabels: boolean = true) {
```
dpvc: Shouldn't `keepLabels` default to `false`?  I would think that `reset()` would reset both, and that if you want to keep things, you need to be explicit about it.

Issue #164, opened 2018-10-10
[SVG] tables with percentage width aren't properly scaled


dpvc:
Tables with no labels but with percentage widths are not properly scaled (due to the removal of the scaling transform that was done to get labeled tables to work).

Issue #163, opened 2018-10-10
[CHTML] percentage width tables get wrong width


dpvc:
In CHTML, a table with a percentage width has the percentage applied twice, so when it is not 100%, the width is wrong.

E.g.

```
<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">
  <mtable width="50%" mathbackground="red">
    <mtr>
      <mtd>
        <mi>a</mi>
      </mtd>
    </mtr>
  </mtable>
</math>
```

PR #167, opened 2018-10-23
Nested tables

### LINKED ISSUE #154: [SVG] table lines are not showing up in IE
It looks like the `outline` CSS that works everywhere else doesn't work in IE, so these will have to be handled explicitly by the SVG renderer rather than falling back on CSS styles.

dpvc:
This PR adds better support for nested tables, and in particular, nested tables with labels and with percentage widths.  The percentage-width tables are taken to be a percentage of the containing table cell (the line-wrap context, according to the MathML spec).  Nesting is also considered for fractions and `mpadded` elements.  An extra `data-align` attribute can be used to control the alignment in `mpadded` elements; the `columnalign` attribute controls alignment in tables nested inside tables, and `numalign`/`denomalign` control it in tables nested in fractions.

This PR also adds 60 test files (so more than half the files are just those tests).

Among other things, this resolves #154, #163, #164, #183, #184, and #185.

### DIFF
--- a/mathjax3-ts/output/common/Wrappers/math.ts
+++ b/mathjax3-ts/output/common/Wrappers/math.ts
@@ -1,6 +1,6 @@
 /*************************************************************
  *
- *  Copyright (c) 2017 The MathJax Consortium
+ *  Copyright (c) 2018 The MathJax Consortium
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
@@ -16,7 +16,7 @@
  */
 
 /**
- * @fileoverview  Implements the CommonMspace wrapper mixin for the MmlMspace object
+ * @fileoverview  Implements the CommonMath wrapper mixin for the MmlMath object
  *
  * @author dpvc@mathjax.org (Davide Cervone)
  */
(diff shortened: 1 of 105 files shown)


mathjax3-ts/output/svg/Wrappers/mtable.ts
```diff
@@ -191,118 +236,95 @@ CommonMtableMixin<SVGmtd<N, T, D>, SVGmtr<N, T, D>, SVGConstructor<N, T, D>>(SVG
...
-            'data-frame': true,
-            width: this.fixed(w), height: this.fixed(h+d), y: this.fixed(-d)
-        }
-        if (style !== 'solid') {
-            properties.style = {'outline-style': style};
-        }
-        return this.svg('rect', properties);
+        const t = this.fLine;
+        return this.svg('rect', this.setLineThickness(t, style, {
+            'data-frame': true, 'class': 'mjx-' + style,
```
zorkow: Is there no method for adding the prefix? Or a constant that defines the prefix? Just in the event we ever want to change it in the future.

samples/tables/lib/dual-tables.js
```diff
@@ -0,0 +1,100 @@
...
+const div = document.createElement('div');
+const template = document.getElementsByTagName('mjx-template')[0];
+
+function  CreateMathML() {
+    const mathml = template.innerHTML;
+    template.innerHTML = mathml.replace(/\&/g, '&amp;')
+                               .replace(/</g, '&lt;')
+                               .replace(/>/g, '&gt;')
+                               .replace(/\n/g, '<br/>')
+                               .replace(/\s/g, '\u00A0');
```
zorkow: Any chance this might ever change or grow? Then it might be easier to put it into a pair list that's processed.

samples/tables/lib/dual-tables.js
```diff
@@ -0,0 +1,100 @@
...
+    Substitute(mathml, window.variables, '');
+}
+
+function Substitute(mml, data, descr) {
+  if (data.length === 0) {
+    addMathML(mml, descr);
+  } else {
```
zorkow: Return after the corner case.

samples/tables/lib/dual-tables.js
dpvc: I don't think so.  This is to format the original HTML source from the template so that it is displayed as code in the browser.  These are the things that need to be changed to do that (HTML specials, newlines and spaces).   I suppose the spaces and newlines could be handed by setting the CSS for white-space, but otherwise, this should not need to be changed.

mathjax3-ts/output/common/Wrappers/math.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @template T  The Wrapper class constructor type
+ */
+export function CommonMathMixin<T extends WrapperConstructor>(Base: T): MathConstructor & T {
+    return class extends Base {
+
+        /**
+         * @override
+         */
+        public getWrapWidth(i: number) {
+            return (this.parent ? this.getBBox().w : this.metrics.containerWidth / this.jax.pxPerEm);
```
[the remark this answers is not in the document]
dpvc: We agreed that this could be allowed (some time ago) and it appears all over the place in the code.  If we want to remove them, let's add that to the updates issue and handle it in one edit.

PR #170, opened 2018-11-07
Mhchem package

zorkow:
Latest version of the mhchem package + tests.


abnerlee:
@zorkow Hi,
```
MhchemMethods.Machine = function(parser: TexParser, name: string, machine: string) {
  try {
    let arg = parser.GetArgument(name);
    let data = mhchemParser.go(arg, machine);
    // TODO: Harvest chemical information here from data, test looping through
    //       the array.
    let tex = texify.go(data);
    parser.string = tex + parser.string.substr(parser.i);
    parser.i = 0;
  } catch (err) {
    throw new TexError(err[0], err[1], err.slice(2));
  }
};
```

The err is a `Error` Object, so it does not have [0], [1], and `slice`, right?

dpvc:
> The err is a `Error` Object, so it does not have [0], [1], and slice, right?

No, you can `throw` any object you want, it doesn't have to be an `Error` object (see the [MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw) for example).  The mhchem package throws an array of strings.  See, for example:

https://github.com/mhchem/MathJax-mhchem/blob/f37e2a1c9bd887b60df6bd9cb552f8d7428d38e0/unpacked/mhchem.js#L164

so this error handling is correct as is.

Finally, mhchem is now a third-party extension (this is not the most recent mhchem PR), maintained by its author at its own GitHub repository.  You should report errors of this kind (if it were an error) on its [issue tracker](https://github.com/mhchem/MathJax-mhchem/issues).

PR #180, opened 2019-01-07
Definitions and utilities to create MathJax components

dpvc:
This implements a mechanism to break up MathJax3 into individual components that can be loaded separately as needed, so page authors can combine the individual pieces they need without having to make a custom webpack for their needs.

There is a dynamic loader the can be used to load the components based on a configuration (similar to how v2 loads pieces on demand), or they can be loaded by hand using separate scripts in the HTML file.

There is also a startup package that allows the components to register themselves when they are loaded, and that builds typeset and conversion methods automatically based on the components that are loaded.  That is, if you load the TeX input jax and the CommonHTML output jax, the startup module will build `MathJax.Typeset()` to typeset a page using TeX input and CHTML output, `MathJax.tex2chtml()` to convert a single TeX string to an HTML DOM tree, `MathJax.tex2mml()` to convert a tex string to serialized MathML, `MathJax.texReset()` to reset labels and equation numbers, and `MathJax.chtmlStylesheet()` to produce the stylesheet DOM element for CommonHTML.

The utilities needed to create the components are in the top-level `components` directory.  Its`bin` subdirectory has a `build` command to make the files needed for a component, and a `makeAll` command that will make all the components in the current directory and below.

Components are controlled by two files:  `build.json` and `webpack.config.js` in their source directories.  The first is used to define what files will be part of a component (these are the files that it will share with other components that build on it).  For example, the `core` component loads all the main MathJax files from the `core` and `util` directories, plus the HTML handler and browser adaptor files.   Other components can then use these files when they are built in reference to them.  The second file controls the webpack process, and in particular which other components it depends on.  These call on the `webpack.common.js` file in the `components` directory to do most of the work of configuring the webpack process.

See the `src` directory for a number of examples of these (though the `svg-output` package won't build unless you merge in the svg branch).  There are packages for TeX input, CHTML output, and several other pieces.  

There is also a component for the loader/startup modules that will dynamically load the needed components based on a configuration file (see the `samples/loader` directory for an example).  The components can also be loaded by hand or combined into larger components (see the `samples/preload` directory for several examples).  Finally, you can create your own custom components that link against existing ones (see the `samples/tex2mml` directory for an example).

I'm still putting together additional components (like for the TeX extensions), but those can be added as developed.

The `mathjax3-ts/components` directory contains the typescript files for the loader and startup modules, and a file for managing the global `MathJax` object that is used to configure MathJax and hold the library definitions and the typeset and conversion methods that are created automatically.

There are a lot of small files in this PR because the components each retire two or three control files, and there are a dozen or so components defined here.  It might be best to start with the `components/src` files that do those definitions just to see how those look, then the `components/samples` to see how the resulting packages are used.  Then the `mathjax3-ts/components` files to see how loader and startup work, and finally the `components/bin` files to see how the needed library files are created.

I hope there is enough information here to make it clear what is happening, but if not, let me know.


components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+ *
+ * Look for export commands within the file, and determine the objects
+ *   that they reference, then produce the library file that defines
+ *   them, and add the file to the package list.
+ *
+ * @param {string} base    The root directory for the .ts files
+ * @param {string} name    The path of the file relative to the base
+ */
+function processFile(base, name) {
+    console.log(name);
```
zorkow: Does that serve a purpose? If maybe have a meaningful string like `Processing file $(name)`.

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+    const file = fs.readFileSync(path.resolve(base, name)).toString();
+    const parts =
+        file.split(/(^export(?:\s+default)?(?:\s+abstract)?\s+[^ {*}]+\s+[a-zA-Z0-9_.$]+)/m);
```
zorkow: We should put these components into a `const` in the header file and assemble the RegExp programmatically, to allow for a future syntax changes/extensions of typescript.

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+        const name = words[words.length - 1];
+
+        if (words[1] === 'default') {
+
+            objects.push('default');
+
+        } else {
+
+            switch (type) {
+              case 'type':
```
zorkow: Same here: We need to allow for changes in syntax. Maybe via a dictionary, rather than a switch.

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+}
+
+/**
+ * @param {string} file     Path to a file
+ * @return {boolean}        True if the file exists, false if not
+ */
+function exists(file) {
+    if (!fs.existsSync(file)) return false;
+    //
+    // For case-insensitive filing systems (like Mac OS),
```
zorkow: s/filing/file

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * Read the configuration for the component
+ */
+const config = JSON.parse(fs.readFileSync(process.argv[2] || 'build.json'));
+
+/**
+ * Extract the configuration values
+ */
+const targets = config.targets || [];                                       // the files to include in the component
+const exclude = new Map((config.exclude || []).map(name => [name, true]));  // files to exclude from the component
```
zorkow: Why are these small and not allCaps as well?

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+    const dots = dir.replace(/[^\/]+/g, '..') || '.';
+    const relative = path.join(dots, '..', MATHJAX3, dir, path.basename(file)).replace(/\.ts$/, '.js');
+    const name = path.parse(file).name;
+    const lines = [
+        '"use strict";',
+        `Object.defineProperty(exports, '__esModule', {value: true});`
+    ];
+    let source = (dir.replace(/\//g, '.') + '.' + name).replace(/^\./, '')
+        .replace(/\.[^.]*/g, (x) => (x.substr(1).match(/[^a-zA-Z_]/) ? '[\'' + x.substr(1) + '\']' : x));
+    if (exists(path.resolve(MATHJAX3, file.replace(/\.ts$/, '')))) source += '_ts';
```
zorkow: Missing braces.

Why do we need a  `_ts`  on the filename?

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+/**
+ * Recursively make any needed directories
+ *
+ * @param {string} dir   The directory relative to the library directory
+ */
+function makeDir(dir) {
+    const fulldir = path.resolve(LIB, dir);
+    if (fs.existsSync(fulldir)) return;
+    makeDir(path.dirname(fulldir));
+    fs.mkdirSync(fulldir);
```
zorkow: Could we just use `shelljs` here?

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+    const lines = [
+        '"use strict";',
+        `Object.defineProperty(exports, '__esModule', {value: true});`,
+        `require('${GLOBAL}').combineWithMathJax({_: {`
+    ];
+    const packages = [];
+    PACKAGE = PACKAGE.sort(sortDir);
+    while (PACKAGE.length) {
+        const dir = path.dirname(PACKAGE[0]).split(path.sep)[0];
+        packages.push(processPackage(lines, '    ', dir));
```
zorkow: I assume that is meant to produce indentation? Maybe make this a configurable `const` in the light of issue ?

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+ * Sort file paths alphabetically
+ */
+function sortDir(a, b) {
+    const A = a.replace(/\//g, '|');          // Replace directory separator by one that is after the alphnumerics
+    const B = b.replace(/\//g, '|');
+    return (A === B ? 0 : A < B ? -1 : 1);
+ }
+
+/**
+ * Recursively process packages, collecting them into groups by subdirectory, properly indented
```
zorkow: subdirectories

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+            if (property.match(/[^a-zA-Z0-9_]/)) {
+                property = `"${property}"`;
+            }
+            packages.push(`${property}: ${component}`);
+        } else {
+            let subdir = path.dirname(PACKAGE[0]);
+            while (path.dirname(subdir) !== dir && subdir !== '.') {
+                subdir = path.dirname(subdir);
+            }
+            packages.push(processPackage(lines, space + (dir === '.' ? '' : '    '), subdir));
```
zorkow: ditto spaces.

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+        //   Get the subdirectory name
+        //   Process the subdirectory using an additional indentation
+        //
+        if (path.dirname(PACKAGE[0]) === dir) {
+            const file = PACKAGE.shift();
+            const name = path.basename(file);
+            const relativefile = path.join('..', MATHJAX3, dir, name).replace(/\.ts$/, '.js');
+            const component = `require('${relativefile}')`;
+            let property = name.replace(/\.ts$/, '');
+            if (property !== name && exists(path.resolve(MATHJAX3, file.replace(/\.ts$/, '')))) property += '_ts';
```
zorkow: Missing braces.

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+                subdir = path.dirname(subdir);
+            }
+            packages.push(processPackage(lines, space + (dir === '.' ? '' : '    '), subdir));
+        }
+    }
+    //
+    // Create the string defining the object that loads all the needed files into the proper places
+    //
+    if (dir === '.') return packages.join(',\n    ');
+    return path.basename(dir) + ': {\n    ' + space + packages.join(',\n    '+space) + '\n' + space + '}';
```
zorkow: I note there is a mix of string composition syntax here. Maybe we should just switch to the new one everywhere, starting with this file.

zorkow: Otherwise there missing spaces around `+`

components/bin/build
```diff
@@ -0,0 +1,318 @@
...
+}
+
+function rmDir(dir) {
```
zorkow: Missing JSDoc

components/bin/makeAll
```diff
@@ -0,0 +1,134 @@
...
+ */
+function buildLib(dir) {
+    const file = path.join(dir, 'build.json');
+    if (!fs.existsSync(file)) return;
+    console.log('Building ' + dir.replace(compRE, ''));
+    const wd = process.cwd();
+    try {
+        process.chdir(dir);
+        const result = execSync(build);
+        console.log('    ' + String(result).replace(/\n/g, '\n    '));
```
zorkow: Indentation?

components/bin/makeAll
```diff
@@ -0,0 +1,134 @@
...
+        const result = execSync('npx webpack --display-modules');
+        console.log('    ' + String(result).replace(/\n/g, '\n    ')
+                                           .replace(compRE, '[components]')
+                                           .replace(rootRE, '[mathjax3]'));
+    } catch (err) {}
+    process.chdir(wd);
+}
+
+
+/*
```
zorkow: `/**`

components/bin/makeAll
```diff
@@ -0,0 +1,134 @@
...
+function webpackLib(dir) {
+    const file = path.join(dir, 'webpack.config.js');
+    if (!fs.existsSync(file)) return;
+    console.log('Webpacking ' + dir.replace(compRE, ''));
+    const wd = process.cwd();
+    try {
+        process.chdir(dir);
+        const result = execSync('npx webpack --display-modules');
+        console.log('    ' + String(result).replace(/\n/g, '\n    ')
+                                           .replace(compRE, '[components]')
```
zorkow: ditto?

mathjax3-ts/components/loader.ts
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+import {MathJax as MJGlobal, MathJaxObject as MJObject, MathJaxLibrary,
+        MathJaxConfig as MJConfig, combineWithMathJax, combineDefaults} from './global.js';
+
+import {Package, PackageError, PackageReady, PackageFailed} from './package.js';
+export {Package, PackageError, PackageReady, PackageFailed} from './package.js';
+export {MathJaxLibrary} from './global.js';
+
+/*
```
zorkow: `/**`

components/webpack.common.js
```diff
@@ -0,0 +1,133 @@
...
+    return string.replace(/([\\.{}[\]()?*^$])/g, '\$1')
+}
+
+/**
+ * Creates the plugin needed for converting mathjax3 references to component/lib references
+ *
+ * @param {string} mathjax3    The location of the mathjax3 js files
+ * @param {string[]} lib       The component library directories to be linked against
+ * @param {string} dir         The directory of the component being built
+ * @return {any[]}             The plugin array (empty or with the conversion plugin)
```
zorkow: What exactly is the conversion plugin?

components/samples/a11y/config.js
```diff
@@ -0,0 +1,50 @@
...
+        input: ['tex'],
+        ready() {
+            MathJax.startup.defaultReady();
+//            console.log(MathJax.chtmlStylesheet());
+            console.log(MathJax.tex2mml('a+b+c+d+e'));
+//            console.log(MathJax.startup.adaptor.outerHTML(MathJax.tex2chtml('x+1')));
+//            MathJax.tex2mmlPromise('x+1').then((mml) => console.log(mml));
+//            MathJax.Typeset();
+//            const adaptor = MathJax.startup.adaptor;
+//            console.log(adaptor.outerHTML(adaptor.body(MathJax.startup.document.document)));
```
zorkow: Remove.

components/samples/a11y/test.html
```diff
@@ -0,0 +1,16 @@
...
+<script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
+<script src="config.js"></script>
+<script src="../../dist/startup.js" id="MathJax-script"></script>
+</head>
+<body>
+
+Check console output.
+
+</body>
+</html>
```
zorkow: Newline.

components/samples/loader/test.html
```diff
@@ -0,0 +1,17 @@
...
+<script src="config.js"></script>
+<script src="../../dist/startup.js" id="MathJax-script"></script>
+</head>
+<body>
+
+$$x+1\over x-1$$
+
+Done!
+</body>
+</html>
```
zorkow: Newline

components/samples/preload/combined.html
```diff
@@ -0,0 +1,18 @@
...
+<script src="combined.min.js" id="MathJax-script"></script>
+</head>
+<body>
+
+$$x+1\over x-1$$
+
+Done!
+<script src="run.js"></script>
+</body>
+</html>
```
zorkow: Newline

mathjax3-ts/components/global.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Combine configuration or data with the existing MathJax object
+ *
+ * @param {any} config   The data to be merged into the MathJax object
+ */
+export function combineWithMathJax(config: any): MathJaxObject {
+    return combineConfig(MathJax, config);
+}
+
+
+/*
```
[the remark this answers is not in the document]
dpvc: Since these weren't describing a specific `class`, `type`, `const`, etc., I didn't think they got `/**`.  But I've added the extra star.

components/bin/build
dpvc: As we discussed face-to-face, since there can be directories with the same base name as a `.ts` file, there needs to be a way to distinguish the two.  For example, there are both `output/svg.ts` and `output/svg` (which contains things like `output/svg/Wrapper.ts`).  Since `output/svg/Wrapper.ts` will be using `MathJax._.output.svg.Wrapper = ...`, we use `MathJax._.output.svg_ts = ...` for the contents of the `output/svg.ts` file.  It is unpleasant, but necessary.

components/bin/build
dpvc: I think this is simple enough not to require an additional dependency.

components/bin/build
dpvc: The singular is correct, here.  For example, you collect marbles into groups by color, not by colors.

components/bin/build
dpvc: I think this is one of those things that needs to be done as a global update.

components/bin/makeAll
dpvc: This is just for console output to tell you what is happening.  I don't think that needs to be configurable.

components/webpack.common.js
dpvc: It is the plugin described in the main comment five lines above:  the one that converts `mathjax3` references to `components/lib` references.

components/samples/a11y/config.js
dpvc: There are there for testing for me.  I'll remove them before the final version.  I'm not sure these samples will be kept in the distribution in the end at all.

dpvc:
> I think we should aim to make things a bit more robust, by extracting typescript syntax specific elements (and some others like indentation) into a meta-`config` file for the configuration.

I've moved these to constants at the top of the `build` file for now, and we can extract them into a separate configuration later if necessary.

Other than that, I think I've taken care of what you requested so far.

PR #182, opened 2019-01-09
Core update

dpvc:
This PR primarily updates the `Handler` structure so that it has references to the `MmlFactory` that it uses, as well as the `MathDocument` and `MathItem` classes that it will use.  That makes it possible to extend those classes by extending the `Handler`, and have the `Handler` know about the changes.  it also means that there is only one `MmlFactory` that is shared by all the components, rather than separate ones being created for each.  There are new `initialize()` methods for the jax so that they can do initialization after the `DOMadaptor` and `MmlFactory` are in place.

It also adds a new `BitField` class that is used to locate and store but flags.  This replaces the explicit fields used to record what actions have been processed on the math list, and can be used by extensions to tie into that record-keeping process.

There are also some fixes to other minor issues, like using entities for non-ascii text in serialized output,  some better handling of inherited attributes when creating new nodes after inheritance has already been performed, the ability to add namespace attributes (needed for SVG `xlink:href` attributes).  See the commit list for descriptions of some of the changes (and for a breakdown of the PD into it separate pieces, if that helps).


mathjax3-ts/core/InputJax.ts
```diff
@@ -63,11 +64,26 @@ export interface InputJax<N, T, D> {
      */
     adaptor: DOMAdaptor<N, T, D>;
 
+    /**
+     * The MmlFactory for this input jax
+     */
+    mmlFactory: MmlFactory;
```
zorkow: Do we ever need a getter for this?

mathjax3-ts/util/BitField.ts
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * @param {string} name    The name of the bit position to clear
+     */
+    public clear(name: string) {
+        this.bits &= ~this.getBit(name);
+    }
+
+    /**
+     * @param {string} name   The name of the bit to check if set
+     * @return {boolean}      True if the named but is set
```
zorkow: Typo: bit

mathjax3-ts/util/BitField.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * @param {string} name   The name whose bit position is needed (error if not defined)
+     * @return {number}       The position of the named bit
+     */
+    protected getBit(name: string) {
+        const bit = (this.constructor as typeof BitField).names.get(name);
+        if (!bit) {
+            throw new Error('Unknown bit-field name: ' + name);
```
zorkow: Do you really want to throw an exception here? Not just gracefully exit?

mathjax3-ts/core/InputJax.ts
dpvc: Not from outside the InputJax.  We need a setter because the MathDocument will pass its factory on to the input and output jax, but there is no need to get the factory from outside (the document already knows what it is).  Same for adaptors.

mathjax3-ts/util/BitField.ts
dpvc: Discusses face-to-face.  Yes, I think throwing is correct, here.

PR #192, opened 2019-02-01
Use output-specific FontData and mixins to add font-specific content

dpvc:
This moves common functions from the `fonts/tex.ts` files to output-jax-specific `FontData` classes, so that they can be shared among multiple fonts for an output jax.  It uses a mixin in `common/fonts/tex.js` so that the output-jax-specific `FontData` classes can be extended with the common information, and then `fonts/tex.js` in the specific output jax use that mixin to add the output-jax-specific tex information.

This necessitated a slight change to the data files for the different variants, so all those files are changed.  That makes it look like a lot of files are altered, but there is really nothing to those changes.

The large sections of what looks like new code in `FontData.ts` files are just chunks of existing code moved from `font/tex.js` into those files, so you should not need to spend much time on those.

Finally, the `CharOptions` object has output-specific information (e.g., paths for SVG, and css information for HTML), so I've abstracted that to have output-jax-specific versions (using templates).  This also allows these to be extended in possible future extensions.


mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -16,28 +16,377 @@
...
+     * @param {StyleList} styles  The style object to add styles to
+     * @param {string} vclass     The variant class string (e.g., .mjx-b) where this character is being defined
+     * @param {number} n          The unicode character being defined
+     * @param {CharData} data     The bounding box data and options for the character
+     */
+    protected addCharStyles(styles: StyleList, vclass: string, n: number, data: CHTMLCharData) {
+        const [h, d, w, options] = data as [number, number, number, CHTMLCharOptions];
+        const css: StyleData = {};
+        if (options.css) {
+            if (options.css & CSS.width) {
```
zorkow: At this point `options.css` exists.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -16,28 +16,377 @@
...
+                css.width = this.em(w);
+            }
+            if (options.css & CSS.padding) {
```
zorkow: ditto

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -16,28 +16,377 @@
...
+                css.padding = this.em0(h) + ' 0 ' + this.em0(d);
+            }
+            if (options.css & CSS.content) {
```
zorkow: ditto.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -16,28 +16,377 @@
...
+                css.content = '"' + (options.c || this.char(n, true)) + '"';
+            }
+        }
+        if (options.f !== undefined) css['font-family'] = 'MJXZERO, MJXTEX' + (options.f ? '-' + options.f : '');
```
zorkow: Missing braces.

mathjax3-ts/output/chtml/fonts/tex.ts
```diff
@@ -99,14 +91,30 @@ export class TeXFont extends CommonTeXFont {
     };
 
     /**
-     *  The stretchy delimiter data (incomplete at the moment)
+     * The data used for CSS for undefined characters each variant
```
zorkow: `of` each variant

mathjax3-ts/output/common/FontData.ts
```diff
@@ -428,9 +441,9 @@ export class FontData {
      */
     public createVariant(name: string, inherit: string = null, link: string = null) {
         let variant = {
-            linked: [] as CharMap[],
-            chars: (inherit ? Object.create(this.variant[inherit].chars) : {}) as CharMap
-        };
+            linked: [] as CharMap<C>[],
+            chars: (inherit ? Object.create(this.variant[inherit].chars) : {}) as CharMap<C>
+        } as V;
```
zorkow: Would that work by writing `variant: V = `. Not sure if that will coerce the type, but it would look cleaner.

mathjax3-ts/output/common/Wrapper.ts
```diff
@@ -482,6 +495,10 @@ AbstractWrapper<MmlNode, CommonWrapper<J, W, C>> {
         }
     }
 
+    /**
+     * @return {boolean}   True if this is the top-most container of an embellished operator that is
+     *                       itself an embellished operator (the maxiamal embellished operator for its core)
```
zorkow: typo: maximal

mathjax3-ts/output/chtml/FontData.ts
dpvc: Note that this is `&` not `&&`.  The `options.css` property is a bit field, and we are testing for the presence of a specific bit in that field, both here and the ones marked below.  The `CSS` object is a collection of named bit positions.

mathjax3-ts/output/common/FontData.ts
dpvc: No, it doesn't do the casting.  I'd still need the `as V`.

PR #194, opened 2019-02-06
Changes getColor to public.

zorkow:
Missed that one in my last review. Sorry!

### DIFF
--- a/mathjax3-ts/input/tex/color/ColorUtil.ts
+++ b/mathjax3-ts/input/tex/color/ColorUtil.ts
@@ -67,7 +67,7 @@ export class ColorModel {
      * @param {string} def The color definition: `red, `0.5,0,1`, `128,0,255`, `0.5`.
      * @return {string} The color definition in CSS format e.g. `#44ff00`.
      */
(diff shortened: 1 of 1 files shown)


zorkow:
@OmarIthawi Found a last little problem. Sorry for not spotting that earlier.

dpvc:
Approved

zorkow:
@OmarIthawi No, it simply does not compile otherwise, as we are trying to use `getColor` from within the `ColorMethods` module.

zorkow:
@dpvc  Thanks for the quick turnaround!

zorkow:
@OmarIthawi 
> I made it private, and it worked, so I didn't over think it.

You must be using different compiler settings, then. It did not work for me. 

> mathjax3-ts/input/tex/color/ColorMethods.ts(66,30): error TS2341: Property 'getColor' is private and only accessible within class 'ColorModel'.
> ...

OmarIthawi:
It might be just me overlooking it, and using a older version of the compiled code. Thanks for catching it @zorkow!

PR #199, opened 2019-02-22
Reduce the amount of CSS produced for CHTML output

dpvc:
This PR provides an option for "adaptive CSS", which produces only the CSS for the characters and MathML nodes that are used on the page.  This is handled by tracking the usage of characters as they are addd to the expressions, and only producing the CSS for those that are.  If multiple expressions are typeset, the CSS will include all the characters from all of them.  There is a clearing function that can be used to reset the cache (so that if this is being done server-side or via a web service, the cache can be reset).

Using the adaptive CSS improves responsiveness of the page during zooming, for example, since the number of CSS rules is vastly reduced.


mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -377,6 +412,36 @@ export class CHTMLFontData extends FontData<CHTMLCharOptions, CHTMLVariantData,
...
     }
 
+    /**
+     * @param {CHTMLCharData} data   The [h, d, w, options] data for the character
+     * @param {number} dw            The (optional) left offset of the glyph
+     * @param {number} ic            The (optional) italic correction value
+     * @return {string}              The padding string for the h, d, w.
+     */
+    public padding(data: CHTMLCharData, dw: number = 0, ic: number = 0) {
+        return [data[0], data[2] + ic, data[1], dw].map(n => this.em0(n)).join(' ');
```
zorkow: `map(this.em0).` should be sufficient.

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -377,6 +412,36 @@ export class CHTMLFontData extends FontData<CHTMLCharOptions, CHTMLVariantData,
         return em(Math.max(0, n));
     }
 
+    /**
+     * @param {CHTMLCharData} data   The [h, d, w, options] data for the character
+     * @param {number} dw            The (optional) left offset of the glyph
+     * @param {number} ic            The (optional) italic correction value
+     * @return {string}              The padding string for the h, d, w.
+     */
+    public padding(data: CHTMLCharData, dw: number = 0, ic: number = 0) {
```
zorkow: `[h, d, w, options]: CHTMLCharData`?

mathjax3-ts/output/chtml/Wrapper.ts
```diff
@@ -222,6 +178,13 @@ CommonWrapper<
         return chtml;
     }
 
+    /**
+     * Mark this class as having been typeset (so styles will be output
```
zorkow: Missing ')'

mathjax3-ts/output/chtml/FontData.ts
```diff
@@ -128,11 +129,42 @@ export class CHTMLFontData extends FontData<CHTMLCharOptions, CHTMLVariantData,
...
+     * @param {boolean} adapt   Whether to use adaptive CSS or not
+     */
+    public adaptiveCSS(adapt: boolean) {
+        this.options.adaptiveCSS = adapt;
+    }
+
+    /**
+     * Clear the cache of which characters have been used
+     */
+    public clearCache() {
```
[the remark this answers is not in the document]
dpvc: Perhaps, but they don't really do much of the same thing.  Something to consider for the future, however.

mathjax3-ts/output/chtml/FontData.ts
dpvc: Good point.  I didn't think of that.

mathjax3-ts/output/chtml/FontData.ts
dpvc: I didn't think of the, either.  I'm not sure I knew it could be done (or if I did, I forgot).

PR #200, opened 2019-02-26
Braket

zorkow:
The braket package.


mathjax3-ts/input/tex/braket/BraketItems.ts
```diff
@@ -0,0 +1,86 @@
...
+  public toMml() {
+    let inner = super.toMml();
+    let open = this.getProperty('open') as string;
+    let close = this.getProperty('close') as string;
+    if (this.getProperty('stretchy')) {
+      return ParseUtil.fenced(this.factory.configuration, open, inner, close);
+    }
+    let attrs = {fence: true, stretchy: false, symmetric: true, texClass: TEXCLASS.OPEN};
+    let openNode = this.create('token', 'mo', attrs, open);
+    let closeNode = this.create('token', 'mo', attrs, close);
```
dpvc: Shouldn't the `closeNode` have `texClass: TEXCLASS.CLOSE` rather than `TEXCLASS.OPEN`?

mathjax3-ts/input/tex/braket/BraketMappings.ts
```diff
@@ -0,0 +1,56 @@
...
+import {CommandMap, MacroMap} from '../SymbolMap.js';
+import BraketMethods from './BraketMethods.js';
+
+
+/**
+ * Macros for braket package.
+ */
+new CommandMap('Braket-macros', {
+  bra: ['Macro', '{\\langle {#1} \\vert}', 1],
+  ket: ['Macro', '{\\vert {#1} \\rangle}', 1],
```
dpvc: Are the outer braces necessary in these macros (and the ones below)?  They will produce an extra `<mrow>`, and I suppose from a semantic viewpoint, that is useful.

Are the inner braces needed?  I can see that they would be useful for `\bra{x\over y}` or `\bra{\bf x}`, but are those uses that are expected?

The spaces aren't needed (even if the inner braces are removed), but I don't mind if they are kept, but I'd prefer to see consistent spacing in the macros below (and I prefer `{#1}` to `{ #1}` or `{ #1 }`) as the extra spaces just cause the parser extra work in skipping over them.

zorkow: > Are the outer braces necessary in these macros (and the ones below)? They will produce an extra `<mrow>`, and I suppose from a semantic viewpoint, that is useful.

That's the reason.

> Are the inner braces needed?

Yes for `\bra` and `\ket` they are necessary, for the reasons you've mentioned and to stay true to the TeX behaviour. For `\Bra`, `\Ket` etc. they would not be strictly necessary (due to `\left` and `\right` treatment), however I would prefer the output of all commands to be similar.

 > The spaces aren't needed 

Fully agree. Also they varied and were unbalanced. Not sure where they came from. Maybe from the original package submitted to v2.

zorkow:
I've adaptations to tests and added the braket tests that were apparently never committed. Superficially this looks like a significant change since the review approval. But it's really only tests, and I am not trying to sneak some code in under the radar.

PR #203, opened 2019-03-21
Add ability to make blocks of options that are open-ended

dpvc:
Add ability to make blocks of options that are open-ended (expandable).  You need these for things like the `Macros` configuration for TeX input, for example.

Also, make the APPEND variable a string so that it can be included in the user configuration object without having to load MathJax modules to get access to this variable.

### DIFF
--- a/mathjax3-ts/util/Options.ts
+++ b/mathjax3-ts/util/Options.ts
@@ -44,14 +44,43 @@ export type OptionList = {[name: string]: any};
  *  E.g., an option of the form
  *
  *    {
- *      name: {APPEND: [1, 2, 3]}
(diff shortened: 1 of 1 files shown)


mathjax3-ts/util/Options.ts
```diff
@@ -44,14 +44,43 @@ export type OptionList = {[name: string]: any};
  *  E.g., an option of the form
  *
  *    {
- *      name: {APPEND: [1, 2, 3]}
+ *      name: {[APPEND]: [1, 2, 3]}
```
zorkow: I don't understand that construct. What exactly is its use?

zorkow: Got it now after f2f.

PR #211, opened 2019-04-11
Physics package

zorkow:
Adds full implementation of the physics package.

It looks more daunting than it is. The majority of the files are tests.


mathjax3-ts/input/tex/physics/PhysicsMethods.ts
```diff
@@ -0,0 +1,900 @@
...
+  if (++parser.macroCount > parser.configuration.options['maxMacros']) {
+    throw new TexError('MaxMacroSub1',
+                        'MathJax maximum macro substitution count exceeded; ' +
+                        'is there a recursive macro call?');
+  }
+};
+
+
+/**
+ * An operator that needs to be parsed (e.g., a gree letter or nabla) and
```
dpvc: Typo (gree => Greek)

mathjax3-ts/input/tex/physics/PhysicsMethods.ts
```diff
@@ -0,0 +1,900 @@
...
+  if (left === '{') {
+    arg = parser.GetArgument(name);
+    if (enlarge) {
+      left = '\\{';
+      right = '\\}';
+    }
+  } else {
+      parser.i++;
+      arg = parser.GetUpTo(name, right);
+    }
```
dpvc: Indentation isn't right?

mathjax3-ts/input/tex/physics/PhysicsMethods.ts
```diff
@@ -0,0 +1,900 @@
...
+    const sup = new TexParser(exp,
+                              parser.stack.env, parser.configuration).mml();
+    mml = parser.create('node', 'msup', [mml, sup]);
+  }
+  parser.Push(parser.itemFactory.create('fn', mml));
+  if (parser.GetNext() !== '(') {
+    return;
+  }
+  parser.i++;
+  let arg = parser.GetUpTo(name, ')');
```
dpvc: Does this need to take into account nested parens?  I'm not sure what the original package does, but this will not. So `\sin(\cos(x))` would end at the ')' after the 'x', and you would need to do `\sin({\cos(x)})` to get it to work.  Again, this might be in the original LaTeX package as well.

There are other places this same idea may apply.  I just noticed it here, though.

mathjax3-ts/input/tex/physics/PhysicsMethods.ts
```diff
@@ -0,0 +1,900 @@
...
+  }
+  if (parens) {
+    parser.i++;
+    macro += '{\\left(' + parser.GetUpTo(name, ')') + '\\right)}';
+  } else {
+    macro += parser.GetArgument(name);
+  }
+  const mml = new TexParser(macro, parser.stack.env,
+                            parser.configuration).mml();
+  // Ideally I would like to set this row to be a TEXCLASS.OP.
```
dpvc: Use a `TeXAtom` rather than an `mrow` to do that.

mathjax3-ts/input/tex/physics/PhysicsMethods.ts
```diff
@@ -0,0 +1,900 @@
...
+};
+
+
+/**
+ * The derivative macros. Its behaviour depends on the number
+ * of arguments provided. In case of
+ * 1 argument: will be part of the denominator.
+ * 2 arguments: argument one is numerator, argument two is denominator.
+ * 3+ arguments: arguments above 2 will be part of the denominator and the
+ *   exponent of the enumerator will depend on the number of denominator
```
dpvc: typo

mathjax3-ts/input/tex/physics/PhysicsMethods.ts
dpvc: I just checked, and the LaTeX physics package *does* process `\sin(\cos(x))` whereas this implementation doesn't, and I also checked the `xparse` documentation, and it does say that delimiters must be balanced (unless you use braces).  I think there may need to be a new function like `GetUpTo()` that skips over parentheses as well as braces in order to get the same results as the original.

zorkow: Nested automatic fences work now. I did it via a specialised stack item. 
The fully balanced fences as in `xparse` does not yet work. In other words an opening fence will greedily take the next corresponding closing fence as the one to stretch.
I have an idea how one can do this. However, this will require a change of treatment of delimiters for packages like `physics` following commands with fenced parameters, which I don't want to touch right now. I will postpone this for 3.1., as it will also be useful for SRE.

zorkow:
There's been quite a few changes since your original approval. Could you have another look at the code as sanity check, please?

dpvc:
> Could you have another look at the code as sanity check, please?

I've given it a look, and I'm OK with the changes, but I should point out that the new `AutoOpen` stack item makes this less compatible with the LaTeX version in some ways.  The LaTeX package is doing something very similar to the old version that used `GetUpTo()`, but where `GetUpTo()` would skip over parentheses (and brackets) in pairs.  This has several differences from how your implementation works:

*  The LaTeX version will skip pairs even when not part of a function application, as you know (e.g., `\sin((\frac{\cos(x)}{2}))` will enlarge the last `)`, while yours does the preceding one.  (But you know this.)
* The LaTeX version doesn't expand macros while looking for its matching close delimiter, while yours does.  So `\def\xyz{)} \sin(x\xyz)` will match the `\xyz` in your implementation, but the `)` in true LaTeX.

It might be better to make a new version of `GetUpTo()` that finds a token while skipping not only braces but also parentheses and brackets, and go back to your original design rather than use the `AutoOpen` stack item.  That would be more compatible with LaTeX.

I'm OK with going with this for now, but for the future, it might be good to be more compatible.

zorkow:
Thanks for taking another look. I had not yet considered the macro expansion. But that's a good point. I was thinking that in an overhaul of this package I should implement methods that more faithfully reproduce the  `xparse` behaviour. That should be useful for other packages as well. But only after the release!

PR #213, opened 2019-04-16
Fix maximum bits in BitFields

dpvc:
Fix maximum bits allowed in BitFields (since javascript bit operations are on signed 32-bit numbers).

### DIFF
--- a/mathjax3-ts/util/BitField.ts
+++ b/mathjax3-ts/util/BitField.ts
@@ -21,6 +21,8 @@
  * @author dpvc@mathjax.org (Davide Cervone)
  */
 
+const MAXBIT = 1 << 31;
+
 export class BitField {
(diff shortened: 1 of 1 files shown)


mathjax3-ts/util/BitField.ts
```diff
@@ -21,6 +21,8 @@
  * @author dpvc@mathjax.org (Davide Cervone)
  */
 
+const MAXBIT = 1 << 31;
```
zorkow: Should this be a static element of `BitField`, so it can be accessed in subclasses? Or is `BitField` considered final?

dpvc: I don't think there will be subclasses, but I've moved it to a static property, which makes sense.

PR #216, opened 2019-04-16
Add a REMOVE option for arrays, corresponding to APPEND

dpvc:
This PR adds a `REMOVE` option similar to the `APPEND` one so that if an option's default value is an array, you can use `prop: {[REMOVE]: [items]}` to remove the given items from the default values.  You can include both `APPEND` and `REMOVE` if you want, in order to remove some entries and then add others.

### DIFF
--- a/mathjax3-ts/util/Options.ts
+++ b/mathjax3-ts/util/Options.ts
@@ -154,35 +154,81 @@ export function copy(def: OptionList): OptionList {
  */
 export function insert(dst: OptionList, src: OptionList, warn: boolean = true) {
     for (let key of keys(src)) {
+        //
+        // Check if the key is valid (i.e., is in the defaults or in an expandable block)
+        //
         if (warn && dst[key] === undefined && !(dst instanceof Expandable)) {
             if (typeof key === 'symbol') {
                 key = key.toString();
             }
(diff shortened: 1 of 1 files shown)


mathjax3-ts/util/Options.ts
```diff
@@ -149,8 +163,17 @@ export function insert(dst: OptionList, src: OptionList, warn: boolean = true) {
         let sval = src[key], dval = dst[key];
         if (isObject(sval) && dval !== null &&
             (typeof dval === 'object' || typeof dval === 'function')) {
-            if (Array.isArray(dval) && Array.isArray(sval[APPEND]) && keys(sval).length === 1) {
-                dval.push(...sval[APPEND]);
+            const ids = keys(sval);
+            if (Array.isArray(dval) &&
+                ((ids.length === 1 && (ids[0] === APPEND || ids[0] === REMOVE) && Array.isArray(sval[ids[0]])) ||
+                  (ids.length === 2 && ids.sort().join(',') === APPEND + ',' + REMOVE &&
+                   Array.isArray(sval[APPEND]) && Array.isArray(sval[REMOVE])))) {
```
zorkow: Could you space out and comment these cases.

dpvc: Done.

PR #218, opened 2019-04-17
Update loader to add a "provides" property

dpvc:
This makes it possible to make modules that include other modules (e.g., combined modules that include both an input and an output module) and allow the loader to know that the sub-modules have been loaded.

### DIFF
--- a/mathjax3-ts/components/loader.ts
+++ b/mathjax3-ts/components/loader.ts
@@ -44,6 +44,7 @@ export interface MathJaxConfig extends MJConfig {
         paths?: {[name: string]: string};          // The path prefixes for use in locations
         source?: {[name: string]: string};         // The URLs for the extensions, e.g., tex: [mathjax]/input/tex.js
         dependencies?: {[name: string]: string[]}; // The dependencies for each package
+        provides?: {[name: string]: string[]};     // The sub-packges provided by each package
         load?: string[];                           // The packages to load (found in locations or [mathjax]/name])
         ready?: PackageReady;                      // A function to call when MathJax is ready
         failed?: PackageFailed;                    // A function to call when MathJax fails to load
@@ -106,6 +107,7 @@ export namespace Loader {
             let extension = Package.packages.get(name);
             if (!extension) {
                 extension = new Package(name);
+                addProvided(name);
             }
             extension.checkNoLoad();
             promises.push(extension.promise);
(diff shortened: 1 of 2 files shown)


mathjax3-ts/components/loader.ts
```diff
@@ -44,6 +44,7 @@ export interface MathJaxConfig extends MJConfig {
         paths?: {[name: string]: string};          // The path prefixes for use in locations
         source?: {[name: string]: string};         // The URLs for the extensions, e.g., tex: [mathjax]/input/tex.js
         dependencies?: {[name: string]: string[]}; // The dependencies for each package
+        provides?: {[name: string]: string[]};     // The sub-packges provided by each package
```
zorkow: typo

dpvc:
> It feels a bit as if we are writing our own package system.

Yes, it has grown a bit more full-featured than I had originally intended.

> I just wonder if there is not an existing pattern we could reuse.

Probably.  On the other hand, the loader is not a core component, and both doesn't have to be used and can be replaced by something else if needed.  It (or something like it) was needed for the CDN, but it can be avoided entirely if you are packaging MathJax with your own site.  It allows you to set the function used for loading components, in case you want to change that, but it also can be replaced as a whole.  For example, it would be possible I suspect, to replace the loader with react.js or some similar loader (though I admit I have only looked into that superficially); we might even be able to provide CDN-based versions of the components for that.  Someone wishing to go that route could use our component building code to package their own versions that contain the necessary glue for whatever loader they are using.  Most of the current components hook into the loader only conditionally (if it is there), and so many of the pieces might be able to be used without change (depending on the needs of the required loader).  The combined files (e.g., `tex-chtml` and the similar ones) do include the loader in order to allow additional packages to be loaded based on a configuration, but even those could be used as is if that was not needed, or if the extra components were to be loaded via another mechanism.  The startup component (`components/src/startup`) also uses the loader, but the startup code (`mathjax3-tx/components/startup.ts`) is independent of the loader, and so an alternative startup component could be built to use another package loader, if desired.

There are two core pieces that do refer to the loader:  the contextual menu does take advantage of the loader, if it is present, but will work without it (it just won't be able to load additional a11y features if they aren't present to begin with), and the require TeX extension (and the autoload extension that relies on it) uses it.  If a different package manager were used, the require extension would need to be modified.  It might be possible to abstract that to make such a substitution of loaders easier.

PR #220, opened 2019-04-17
Update TeX packages

dpvc:
This updates existing TeX packages to use the new configuration functionality, and to make them work with the `\require` macro (coming in a PR shortly).  This includes moving options into sub-blocks, for example, the `amsCd` options are moved into an `amsCd` block, and you would configure the TeX jax using

```
new TeX({
  packages: {[APPEND]: ['amsCd']},
  amsCd: {
    colspace: '10pt',
    rowspace: '10pt'
  }
});
```

for example.  This makes it clearer what those options refer to.  In addition to the `amsCd` options, the `color` extension's options are moved to a `color` block.

The `noerrors` package now adds a `data-mix-error` attribute that stores the actual error message (for easier retrieval), and the similar error attribute for "Math input error" messages `MathDocument` class has been moved to the `merror` element for consistency.  (Eventually, the context menu should retrieve this message and make it available to the user.)

### DIFF
--- a/mathjax3-ts/input/tex/ams_cd/AmsCdMethods.ts
+++ b/mathjax3-ts/input/tex/ams_cd/AmsCdMethods.ts
@@ -47,14 +47,15 @@ let AmsCdMethods: Record<string, ParseMethod> = {};
 AmsCdMethods.CD = function(parser: TexParser, begin: StackItem) {
   parser.Push(begin);
   let item = parser.itemFactory.create('array') as ArrayItem;
+  let options = parser.configuration.options.amsCd;
   item.setProperties({
-    minw: parser.stack.env.CD_minw || parser.configuration.options.harrowsize,
-    minh: parser.stack.env.CD_minh || parser.configuration.options.varrowsize
+    minw: parser.stack.env.CD_minw || options.harrowsize,
+    minh: parser.stack.env.CD_minh || options.varrowsize
   });
   item.arraydef = {
       columnalign: 'center',
-      columnspacing: parser.configuration.options.colspace,
-      rowspacing: parser.configuration.options.rowspace,
+      columnspacing: options.colspace,
+      rowspacing: options.rowspace,
       displaystyle: true
   };
   return item;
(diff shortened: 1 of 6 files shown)


mathjax3-ts/input/tex/ams_cd/AmsCdMethods.ts
```diff
@@ -122,14 +123,14 @@ AmsCdMethods.arrow = function(parser: TexParser, name: string) {
         if (a) {
           let nodeA = new TexParser(a, parser.stack.env, parser.configuration).mml();
           let mpadded = parser.create('node', 'mpadded', [nodeA], pad);
-          NodeUtil.setAttribute(mpadded, 'voffset', '1em');
+          NodeUtil.setAttribute(mpadded, 'voffset', '.1em');
```
zorkow: Good catch. That's the value of MJ2.

mathjax3-ts/input/tex/ams_cd/AmsCdMethods.ts
```diff
@@ -47,14 +47,15 @@ let AmsCdMethods: Record<string, ParseMethod> = {};
 AmsCdMethods.CD = function(parser: TexParser, begin: StackItem) {
   parser.Push(begin);
   let item = parser.itemFactory.create('array') as ArrayItem;
+  let options = parser.configuration.options.amsCd;
```
zorkow: Somehow I feel there should be a functional way of retrieving these options. Something like "global" vs "local" options of the package. But I think it's good enough as it is, for now.

E.g., each package should get a unique name, via the configurations, which should then automatically disambiguate names for options, etc.  This could then also be used in to avoid naming conflicts of `handlers`.

dpvc:
> Can you run them and confirm that these are as expected?

I've run them and carefully checked the first 6, but didn't go through all of them by hand.  The new output is correct.

mathjax3-ts/input/tex/ams_cd/AmsCdMethods.ts
dpvc: Sounds like a good plan.

I'd also like to see a configuration-specific data storage area (for things like the color model in the `color` extension) that can be used for data that is specific to the instance of the input jax holding the configuration.  Right now this is being added to the options, as really the only possibility at the moment, but that is a hack.

mathjax3-ts/input/tex/ams_cd/AmsCdMethods.ts
dpvc: It must have gotten edited somehow, because its not the value in the `extensions/AMScd.js` file.  Perhaps a global edit gone wrong.  No matter, we spotted it.

PR #222, opened 2019-04-17
Config macros package

dpvc:
This PR adds a package that allows the TeX input jax options to include a `macros` block that declares macros to be predefined within the TeX jax.  This is similar to the `Macros` option from v2.

### DIFF
--- a/mathjax3-ts/input/tex/config_macros/ConfigMacrosConfiguration.ts
+++ b/mathjax3-ts/input/tex/config_macros/ConfigMacrosConfiguration.ts
@@ -22,12 +22,11 @@
  * @author dpvc@mathjax.org (Davide P. Cervone)
  */
 
-import {Configuration, ConfigurationHandler} from '../Configuration.js';
+import {Configuration} from '../Configuration.js';
 import {expandable} from '../../../util/Options.js';
-import {NewcommandConfiguration} from '../newcommand/NewcommandConfiguration.js';
+import {CommandMap} from '../SymbolMap.js';
+import {Macro} from '../Symbol.js';
 import NewcommandMethods from '../newcommand/NewcommandMethods.js';
-import NewcommandUtil from '../newcommand/NewcommandUtil.js';
-import {ExtensionMaps} from '../MapHandler.js';
 import {TeX} from '../../tex.js';
 
 /**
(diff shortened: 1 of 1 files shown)


mathjax3-ts/input/tex/config_macros/ConfigMacrosConfiguration.ts
```diff
@@ -0,0 +1,63 @@
...
+ * The command map for the autoloaded macros
+ */
+const ConfigMacrosMap = new CommandMap('configMacros', {}, {});
+
+/**
+ * The configuration object for configMacros
+ */
+export const ConfigMacrosConfiguration = Configuration.create(
+    'configMacros', {
+        handler: {macro: ['configMacros']},
```
zorkow: If a Macro is defined in the preamble, will they be overwritten by those defined by `newcommand`? That should be expected behaviour.

dpvc: Yes, they should be, and they do work that way in my testing.  Because `newcommand` adds its macro (and other) user tables during the `init()` function, while the one for `configMacros` is put in place when it is appended, its table will come *after* the `newcommand` ones, so `newcommand` will override them.

I guess that the order in the packages array determines the order in which packages are appended, and so the order of the lookup tables that are part of the configurations for the various packages.  The components that include `configMacros` have it listed last, so that should mean the macros configuration would override anything in the other packages.  But with that in mind, I may need to reorder the list in `AllPackages.ts`, which currently is just alphabetical, as I recall.

I'm also wondering about `autoload` at this point, because if additional packages are loaded on the fly, it is possible that they will override the user-defined macros.  E.g., if a user re-defines `\textcolor` and then uses `\color` (which autoload the `color` package), their definition of `\textcolor` will be redefined to be the standard one.  Version 2 had a mechanism of marking macros as user-defined so that loaded packages would not overwrite them, but that isn't part of v3, which is using separate lookup tables to handle this sort of thing.

One solution would be to have a priority for the tables (like the ones for initialization), so that if a package is loaded later, it could insert its table earlier in the list, rather than just adding it at the front (like it is now).  That way, `configMacros` could put its tables at a priority that puts them at the front, and other packages that are loaded later would be inserted after it, meaning user macros would have precedence.  Similarly, `newcommand` could use a higher priority (lower number?) to put its tables in front of `configMacros` so that macros defined in the page override the ones in the configuration (and required or autoloaded packaged would not override them).

We may not ned to do that for the first release, but it is something that will probably need to be handled in the near future.

zorkow: I made  a note on the tracker.

PR #223, opened 2019-04-17
Autoload package

dpvc:
This PR provides a TeX package that defines macros to automatically load the packages that implements them.  This is similar to the `autoload-all` extension in v2.  The extensions that it will autoload are configurable in the TeX input jax options, so you could do

```
new TeX({
  packages: {[APPEND]: ['autoload'},
  autoload: {
    color: []
  }
});
```

to autoload everything except the color extension (thus leaving the original `\color` definition from v2), for example.  You could also add autoloading of other extensions, if desired.

All the current extensions are autoloaded except for `ams` and `physics`, the former because of the large number of macros (and because it is included in the standard tex component), and physics because it redefines standard macros, so should only be included explicitly by the page author.

### DIFF
--- a/mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
+++ b/mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
@@ -17,7 +17,7 @@
 
 
 /**
- * @fileoverview    Configuration file for the config-macros package.
+ * @fileoverview    Configuration file for the autoload package.
  *
  * @author dpvc@mathjax.org (Davide P. Cervone)
  */
@@ -112,7 +112,7 @@ export const AutoloadConfiguration = Configuration.create(
         handler: {macro: ['autoload']},
         options: {
             //
-            //  These are the extension names and the macros they contains.
+            //  These are the extension names and the macros they contain.
             //  You can prevent one from being autoloaded by setting
             //    it to [] in the options when the TeX input jax is created.
             //  You can include the prefix if it is not the default one from require
--- a/mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
+++ b/mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
@@ -31,6 +31,7 @@ import {TeX} from '../../tex.js';
 
 import {RequireLoad, TEXPATH} from '../require/RequireConfiguration.js';
 import {Package} from '../../../components/package.js';
+import {expandable} from '../../../util/Options.js';
 
 /**
  * A CommandMap class that allows removal of macros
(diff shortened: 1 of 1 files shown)


mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
```diff
@@ -0,0 +1,140 @@
...
+            //  These are the extension names and the macros they contains.
+            //  You can prevent one from being autoloaded by setting
+            //    it to [] in the options when the TeX input jax is created.
+            //  You can include the prefix if it is not the default one from require
+            //
+            autoload: expandable({
+                action: ['toggle', 'mathtip', 'texttip'],
+                bbox: ['bbox'],
+                boldsymbol: ['boldsymbol'],
+                braket: ['bra', 'ket', 'braket', 'set', 'Bra', 'Ket', 'Braket', 'Set', 'ketbra', 'Ketbra'],
```
zorkow: This will clash with the physics package.

mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * The configuration object for configMacros
+ */
+export const AutoloadConfiguration = Configuration.create(
+    'autoload', {
+        handler: {macro: ['autoload']},
+        options: {
+            //
+            //  These are the extension names and the macros they contains.
```
zorkow: typo

mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
```diff
@@ -0,0 +1,140 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+
+/**
+ * @fileoverview    Configuration file for the config-macros package.
```
zorkow: Update comment.

mathjax3-ts/input/tex/autoload/AutoloadConfiguration.ts
dpvc: If `physics` is loaded, either by `\require` or through including it in the configuration of the packages to be loaded, then its versions of these macros will supersede the ones from `autoload`.  If `bracket` has already been autoloader and `\require{physics}` is executed, the physics versions will supersede the `bracket` ones.

dpvc:
> We can keep the package for now, but I would like to register my philosophical objection

I understand the objections.  As per face-to-face meeting, we will go ahead with `autoload` as is for now.

PR #227, opened 2019-04-17
Add combined components

dpvc:
This PR adds components that combine an input and output jax, making a single file that can be loaded from the CDN that includes a complete MathJax configuration that is ready to use.  These are similar to the combined configuration files from v2.

Because these all include the startup component, you can use the `MathJax` global variable to configure them, and they all include the context menu, so assistive functions can be enabled by the user.

There are test files in each directory, but these can be removed, if you want.

### DIFF
--- a/components/src/mml-chtml/mml-chtml.js
+++ b/components/src/mml-chtml/mml-chtml.js
@@ -1,0 +1,8 @@
+import '../startup/lib/startup.js';
+import './preload.js';
+import '../core/core.js';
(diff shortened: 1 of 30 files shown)


dpvc:
Yes, all the test and sample files probably need to be removed.  That is part of the clean up for the final release that I'm in the process of doing.  That will be handled in a later PR.

zorkow:
Sounds good.

PR #228, opened 2019-04-17
Update labs

dpvc:
This removes the old TeX-lab and Mml-lab files, and updates v3-lab to use the new components hierarchy and the autoload extension to load other extensions automatically.


lib/v3-lab.js
```diff
@@ -93,7 +92,7 @@ function V3DocumentMixin(documentClass) {
 window.MathJax = {
     loader: {
         load: [
-            'input/tex',
+            'input/tex-full',
```
zorkow: See my previous comment in PR #226 on `tex-full` vs `tex`.

dpvc: The only comment there is the one about `autobold`.  Is that the one you are talking about?  I'm not sure it affects the lab, since its the lab expression is not in a bold setting.  

Is there maybe some comments that you haven't submitted?

zorkow: That's odd, I definitely made it. After submitting the review though and now it seems to have vanished.
Anyway the comment was: Since we have `tex-base` why do we need to call it `tex-full` instead of simply `tex`?

PR #229, opened 2019-04-18
Fixes to build v3.

zorkow:


mathjax3-ts/a11y/sre-node.ts
```diff
@@ -26,6 +26,8 @@ declare const require: (name: string) => any;
 
 const SRE = require('speech-rule-engine');
 
+declare const global: any;
+
```
dpvc: When I run `tsc` on this branch, I get

```
../context-menu/scripts/js/context_menu.d.ts(286,15): error TS2451: Cannot redeclare block-scoped variable 'global'.
mathjax3-ts/a11y/sre-node.ts(29,15): error TS2451: Cannot redeclare block-scoped variable 'global'.
```

which is why I removed those lines in a previous commit.  This seems to be a conflict that is generated by the inclusion of `context_menu.js` somewhere, though I don't see why it is affecting the `see-node.ts` file.  But If I change this to 

```
declare namespace global {};
```

it seems to work.

dpvc:
Also, the `AllPackages.ts` file is added in another PR, and adding this file will probably cause conflicts later on.  It is supposed to export an `AllPackages` variable that is an army consisting of the names of the packages that were loaded.

mathjax3-ts/a11y/sre-node.ts
zorkow: I've tried different branches on a new clone, but always get the same error:

     mathjax3-ts/a11y/sre-node.ts(32,1): error TS2304: Cannot find name 'global'.

I also can not see `global` in the `context-menu.d.ts` file. Maybe it is a typescript version difference. 

But your solution does indeed work.

zorkow:
The PR was originally meant as a basis for discussing the fixes I had to make for a successful build. 
But it seems the regexp and the `global` issue are useful. So I will remove the `AllPackages` file.

mathjax3-ts/a11y/sre-node.ts
zorkow: Not sure what changed to my previous compilation, but now I get the following error:
```
mathjax3-ts/a11y/sre-node.ts(34,1): error TS2708: Cannot use namespace 'global' as a value.
mathjax3-ts/a11y/sre-node.ts(35,1): error TS2708: Cannot use namespace 'global' as a value.
mathjax3-ts/a11y/sre-node.ts(36,1): error TS2708: Cannot use namespace 'global' as a value.
```
So it seems in your version it still sees the `global` constant and simply ignores the `namespace`.

dpvc: OK, I think I understand why.  I had made an additional PR to the context_menu code that added the `global` object to `context_menu.ts`, which you haven't yet acted on, and it is that change that is causing the issue here.  Since you haven't merged it yet, you don't have that change in your copy, and so are getting different results.

What confuses me is that there are two other files that declare `global` that don't seem to cause problems.

I'm still looking for a solution.

mathjax3-ts/a11y/sre-node.ts
```diff
@@ -26,7 +26,7 @@ declare const require: (name: string) => any;
 
 const SRE = require('speech-rule-engine');
 
-declare var global: any;
+declare let global: any;
```
dpvc: Is `let` to be preferred to the original `const`?

mathjax3-ts/a11y/sre-node.ts
dpvc: OK, I've worked out a solution but chasing my pull request for the `context-menu` repository to no longer try to use the `global` variable.  This was what was causing the problem, and I can work around the issue that I was trying to solve (the ability to load components that include the contextual menu into `node`) in a different way.

So this change can stay, and I will resolve the issue in a separate PR later (part of my changes to make `context-menu` and `speech-rule-engine` work as proper dependencies).

mathjax3-ts/a11y/sre-node.ts
zorkow: No. I played with `var`, forgot to take it out before I pushed and instead of resetting it to `const` I used `let`. Sorry about the confusion.

PR #230, opened 2019-04-23
Add ability to generate speech strings for enriched MathML

dpvc:
This PR implements preliminary speech support via the semantic-enrich extension.  It adds an option that enables SRE's `data-semantic-speech` tags, and adds support in the output jax to transfer that to the proper location in the output.  This does mean that the other `data-semantic-` attributes will also be present, but these can be removed easily using a tree walker prior to calling the `typeset()` method.  Perhaps an option to control that would be in order as well.  Note that the jax only process the `data-semantic-speech` attribute on the `math` tags, so only shallow speech is handled currently.  I think we can expand the support after v3 is released.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/math.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/math.ts
@@ -103,7 +103,7 @@ export class CHTMLmath<N, T, D> extends CommonMathMixin<CHTMLConstructor<N, T, D
             this.setIndent(chtml, align, shift);
         }
         //
-        // Transver speech to aria-label and hide child nodes
+        // Transfer speech to aria-label and hide child nodes
         //
         const speech = attributes.get('data-semantic-speech') as string;
         if (speech && !attributes.get('aria-label')) {


mathjax3-ts/output/chtml/Wrappers/math.ts
```diff
@@ -102,6 +102,18 @@ export class CHTMLmath<N, T, D> extends CommonMathMixin<CHTMLConstructor<N, T, D
         if (display && shift && !adaptor.hasAttribute(chtml, 'width')) {
             this.setIndent(chtml, align, shift);
         }
+        //
+        // Transver speech to aria-label and hide child nodes
```
zorkow: Typo

mathjax3-ts/a11y/semantic-enrich.ts
```diff
@@ -96,6 +104,10 @@ export function EnrichedMathItemMixin<N, T, D, B extends Constructor<AbstractMat
             if (typeof sre === 'undefined' || !sre.Engine.isReady()) {
                 MathJax.retryAfter(sreReady);
             }
+            if (document.options.enrochSpeech !== currentSpeech) {
```
zorkow: This should be 
```suggestion
            if (document.options.enrichSpeech !== currentSpeech) {
```

PR #232, opened 2019-04-30
Update dist directory

dpvc:
This PR is moving toward making the `components/dist` directory contain what will be delivered by the CDN, and makes changes so that SRE and context-menu will work when MathJax3 is loaded as an `npm` dependency, as well as when it is checked out as a `git` repository.
So it does the following:

1.  It includes a new `bin/copy` command to copy specific files to the `dist` hierarchy, and hooks that into the `bin/makeAll` command.  This is used to copy SRE files into pace for the CDN, and will be used to move the font `.woff` files into place in the next PR.

2.  It makes SRE a proper dependency.  This is because SRE needs to be loaded separately when used in the browser (it can't be webpacked), and also so that the path to it can be set so that if a component that uses it is loaded in `node`, you can use the node version rather than the browser version.  (This only applies to the components using the startup module; if you are importing the a11y code directly into your project, then `sre.js` should find the node version automatically.)

3.  It separates out the `context-menu` so that it is loaded in the component code, not the `ui/menu` code.  This is to avoid the issues we were having with `tsc` and the `global` variable.  I put a copy of `context_menu.d.ts` into the `ui/menu` directory so that the menu code knows about the types, but without having to know where the `context-menu` code itself is located.  The `component/ui/menu` component now loads the menu code so that it will be included in the webpack results, and the `webpack.common.js` file is modified to find the context-menu code when MathJax3 is an `npm` dependency or a checked out `git` repository (assuming `context-menu` becomes an `npm` package as well, though I'll have to update the name when it does).

4.  As part of 3 above, there are new files in `components/ui/menu` that can be used in `node` when `context-menu` becomes an `npm` package, or if you are preloading the `context-menu` code yourself.

5.  The v3-lab is modified to load the context menu code from the `node_modules` location (since we are not using the webpacked components that include it directly).

6.  Declarations for `global` are added to two files that need them (since they were getting the definition from the changes that I had made to `context-menu` but have removed to solve the conflict that we were seeing with `global`).

7.  Finally, the importing of `context-menu` are removed from the `ui/menu` code since that is now handled by the component code.


components/bin/copy
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Recursively make any needed directories
+ *
+ * @param {string} dir   The directory relative to the library directory
+ */
+function makeDir(dir) {
+    if (fs.existsSync(dir)) return;
+    makeDir(path.dirname(dir));
+    fs.mkdirSync(dir);
```
zorkow: Should we not just use the node facility for `mkdir -p`:
```suggestion
    fs.mkdirSync(dir, {recursive: true});
```

dpvc: I was not aware of this option (and I thought I had looked, but must have missed it).  That makes this function unnecessary, so I've modified the code to remove the function and use 1fs.mkdirSync(to, {recursive: true});` in place of `makeDir(to)` earlier.

dpvc:
I've changed `console.log()` to `console.info()` throughout the files in `bin`.  I also looked up the documentation on `console` (which I guess I never have done), and there are a number of fun things there that I will be able to use during debugging.  So thanks for the suggestion.

PR #235, opened 2019-04-30
Add ability to autoload based on an environment

dpvc:
This PR adds the ability to configure the `autoload` TeX package to autoload an extension when an environment is first used.  This allows `amsCd` to be autoloaded.


dpvc:
You re right, this would not provide for autoloading based on delimiters.  I'm not sure that is necessary, but I can look into it if you wish.

I understand your philosophy on autoload.  We can discuss this further, if you wish.  There is still time to have it removed before the final 3.0.0 release, if we come to that conclusion.  I'm going to merge this for now, and we can remove the autoload extension later (but before 3.0.0) if that is the decision.

dpvc:
OOPS, forgot to push the changes before merging.  I've added the commit to update-labs by hand.  Sorry about the confusion.

PR #236, opened 2019-05-01
Fix \vec and stretchy horizontal characters in CHTML

dpvc:
Fix problem where `\vec` did not produce the right character (the font CSS was being overriden), and where stretchy horizontal characters could be too long if the extender had to be shorter than its natural length.

The change in `TextNode.ts` was to remove a redundant CSS specification.

The change in `mo.ts` was to get horizontal stretchy characters to be the right length when the extender is shorter than the character's natural length (the older version would cause the extender to have a minimum width of the character's natural width, and so the length of the stretched character as a whole would be too long).

The change to `fonts/tex.js` is to prevent those CSS rules from overriding the ones set for the individual characters (it was preventing the `\vec` from specifying an alternate font).

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/TextNode.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/TextNode.ts
@@ -42,7 +42,7 @@ export class CHTMLTextNode<N, T, D> extends CommonTextNodeMixin<CHTMLConstructor
     public static autoStyle = false;
 
     public static styles: StyleList = {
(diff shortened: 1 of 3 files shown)


dpvc:
I added a commit that fixes a similar problem with vertical stretchy characters that are all extender (like vertical bars, e.g. `\vert`.  I had checked other vertical stretchy characters, but didn't try one with only extenders and no end characters.

PR #241, opened 2019-05-04
Rewrites some TeX sample files to remove TestMath.

zorkow:
Removes no longer existing TestMath from some of the files.
Before I continue, we should discuss, which of these are still useful. 
For example, it seems to me that `tex-nodes.js` does the same as `tex2mml.js`.

### DIFF
--- a/samples/tex-string.js
+++ b/samples/tex-string.js
@@ -3,6 +3,7 @@ import {MathJax} from '../mathjax3/mathjax.js';
 import {TeX} from '../mathjax3/input/tex.js';
 import {RegisterHTMLHandler} from '../mathjax3/handlers/html.js';
 import {chooseAdaptor} from '../mathjax3/adaptors/chooseAdaptor.js';
+import {STATE} from '../mathjax3/core/MathItem.js';
 
 RegisterHTMLHandler(chooseAdaptor());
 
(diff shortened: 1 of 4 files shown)


dpvc:
All of the samples need to be gone through in light of the changes that have occurred over time, but particularly the recent ones.  But they are all going to be moved to the dev repository (that I will be making once the code settles down), so I put that off for now.

>  it seems to me that `tex-nodes.js` does the same as `tex2mml.js`.

They are close, but not quite: `tex-nodes.js` output includes all inherited attributes as well as the explicit ones, plus a could other computed values for debugging purposes, whereas `tex2mml.js` just puts out the serialized MathML itself (no inherited or other values).

samples/tex-string.js
```diff
@@ -12,8 +13,7 @@ let html = MathJax.document('<html></html>', {
 
 MathJax.handleRetriesFor(() => {
 
-    html.TestMath(process.argv[3] || '').compile();
-    let math = html.math.pop().root;
+    let math = html.convert(process.argv[3] || '', {end: STATE.TYPESET});
```
dpvc: I think the state here should be `STATE.COMPILED` (or `STATE.CONVERT` to be consistent with the other ones).

samples/tex2html.js
```diff
@@ -15,8 +16,7 @@ let html = MathJax.document('<html></html>', {
 
 MathJax.handleRetriesFor(() => {
 
-    html.TestMath(process.argv[3] || '').compile().typeset();
-    let math = html.math.pop();
+    let math = html.convert(process.argv[3] || '', {end: STATE.CONVERT});
```
dpvc: I think this should be `STATE.TYPESET`, as `STATE.CONVERT` stops before the `typeset()` call.

zorkow:
I rewrote the state in the wrong file. Thanks for spotting this.
Should be fine now.

PR #242, opened 2019-05-06
Give tooltips a z-index, and track tooltip show/hide separately.

dpvc:
This PR fixes some problems with tools tips in CHTML and SVG output.  In CHTML, the tooltips could fall underneath later mathematics, so this adds a `z-index` to raise the tooltip above the rest of the math.  There was also a problem in both CHTML and SVG where an expression that has multiple tooltips could leave an earlier tooltip open when the mouse moved from its trigger to another tooltip trigger.  This was because the timers for the show/hide operations were stored in a global variable; this fix tracks them by the element they are attached to, so multiple ones can operate at once, and you don't lose the hide from an earlier one when you start a show on a new one.

### DIFF
--- a/mathjax3-ts/output/chtml/Wrappers/maction.ts
+++ b/mathjax3-ts/output/chtml/Wrappers/maction.ts
@@ -50,7 +50,8 @@ CommonMactionMixin<CHTMLWrapper<N, T, D>, CHTMLConstructor<N, T, D>>(CHTMLWrappe
             display: 'none',
             position: 'absolute',
             bottom: 0, right: 0,
-            Width: 0, height: 0
+            width: 0, height: 0,
+            'z-index': 500
         },
         'mjx-tool > mjx-tip': {
             display: 'inline-block',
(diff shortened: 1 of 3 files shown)


dpvc:
There are some changes from other branches that probably need to be merged into master in order to get the lab to work.  Because many of these pull requests are branched off other (non-master) branches, the interconnections are pretty complicated.  The base branch off which those other branches were created is one that had the TeX packages included (in order to be able to create the components for TeX and its extensions), and I didn't want to merge that until all the individual TeX packages were actually ready and merged into master.  So those base branches haven't been merged into master, even though the individual changes have been approved.  So not everything that has been approved is actually in master (sigh).  Once the remaining TeX packages are finalized and merged, I can start straightening out the mess of other branches and get master more up to date.  I think I've reviewed everything that is there, so it is just a matter of addressing any comments (or rejecting them) and then merging the TeX packages, and I can get the rest taken care of.

dpvc:
About the delay:  there is a delay after you move over an element before a tooltip is opened, but it closes immediately after you leave the element.  So I suspect that is what you are seeing, and so yes, it is by design in that sense.

PR #243, opened 2019-05-06
AMScd package

zorkow:
The AMS CD package ported from MJ2.


mathjax3-ts/input/tex/ams_cd/AmsCdMappings.ts
```diff
@@ -0,0 +1,38 @@
...
+import AmsCdMethods from './AmsCdMethods.js';
+
+
+new sm.EnvironmentMap('amsCd_environment', ParseMethods.environment,
+                      {CD: 'CD'}, AmsCdMethods);
+
+new sm.CommandMap('amsCd_macros', {
+  minwidth: 'minwidth',
+  minheight: 'minheight',
+}, AmsCdMethods);
```
dpvc: These are non-standard commands, right?  I checked the documentation, and it seems there is a standard one for the with: `\minCDarrowwidth`, but not one for height.  Since these commands are global, it might be best to have them CD-specific like this, rather than too generic.

zorkow: I've renamed the commands. PTAL.

zorkow:
PTAL.
(Added PTAL at the wrong position.)

dpvc:
Sorry, missed it earlier.  Looks good.

PR #246, opened 2019-05-07
Cancel package

zorkow:
Port of the MJ2 cancel package.


mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,30 @@ namespace ParseUtil {
...
+   * @param {{[key: string]: number}?} allowed A list of allowed options.
+   * @return {EnvList} The attribute list.
+   */
+  export function splitPackageOptions(attrib: string, allowed: {[key: string]: number} = {}): EnvList {
+    let def: EnvList = {};
+    if (attrib !== '') {
+      const attr = attrib.replace(/ /g, '').split(/,/);
+      for (let i = 0, m = attr.length; i < m; i++) {
+        const keyvalue = attr[i].split(/[:=]/);
+        if (allowed[keyvalue[0]]) {
```
dpvc: This should probably be `allowed.hasOwnProperty(keyvalue[0])`, since otherwise something like `[toString=1]` would be allowed.

mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,30 @@ namespace ParseUtil {
...
+   * Splits a package option list of the form [x=y,z=1] into an attribute list
+   * of the form {x: y, z: 1}.
+   * @param {string} attrib The attributes of the package.
+   * @param {{[key: string]: number}?} allowed A list of allowed options.
+   * @return {EnvList} The attribute list.
+   */
+  export function splitPackageOptions(attrib: string, allowed: {[key: string]: number} = {}): EnvList {
+    let def: EnvList = {};
+    if (attrib !== '') {
+      const attr = attrib.replace(/ /g, '').split(/,/);
```
dpvc: I had occasion recently to need a function like this in a v2 extension, but the values that can be assigned could contain braced groups (so that you could include commas and equal signs, for example).  It might be good to handle that setup (which unfortunately means you can't just use splitting).  Alternatively, we could do another function when that is needed.

mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,30 @@ namespace ParseUtil {
...
+      for (let i = 0, m = attr.length; i < m; i++) {
+        const keyvalue = attr[i].split(/[:=]/);
```
dpvc: If you had `key=x:y`, then you will end up with `{key: 'x'}` in the end, and the `:y` will be lost.  I'm not sure what the best alternative is.

mathjax3-ts/input/tex/cancel/CancelConfiguration.ts
```diff
@@ -0,0 +1,98 @@
...
+
+CancelMethods.CancelTo = function(parser: TexParser, name: string) {
+  const attr = parser.GetBrackets(name, '');
+  let value = parser.ParseArg(name);
+  const math = parser.ParseArg(name);
+  const def = ParseUtil.splitPackageOptions(attr, ALLOWED);
+  def ['notation'] = TexConstant.Notation.UPDIAGONALSTRIKE + ' ' +
+    TexConstant.Notation.UPDIAGONALARROW;
+  value = parser.create('node', 'mpadded', [value],
+                        {depth: '-.1em', height: ' + .1em', voffset: '.1em'});
```
dpvc: I don't think you need the spaces in the `height` (and I think they might cause problems).

mathjax3-ts/input/tex/cancel/CancelConfiguration.ts
```diff
@@ -0,0 +1,98 @@
...
+import {Configuration} from '../Configuration.js';
+import TexParser from '../TexParser.js';
+import {TexConstant} from '../TexConstants.js';
+import {CommandMap} from '../SymbolMap.js';
+import {ParseMethod} from '../Types.js';
+import ParseUtil from '../ParseUtil.js';
+
+
+/**
+ * The attributes allowed in \enclose{notation}[attributes]{math}
```
dpvc: Is this supposed to refer to `\canel` rather than `\enclose`?

mathjax3-ts/input/tex/cancel/CancelConfiguration.ts
```diff
@@ -0,0 +1,98 @@
...
+  thickness: 1
+};
+
+
+// Namespace
+export let CancelMethods: Record<string, ParseMethod> = {};
+
+
+/**
+ * Parse function for cancel macros of the form \(b|x)cancel[attributes]{math}
```
dpvc: Should `\(b|x)` be `\(b|x)?` or `\(b|x|)`?

mathjax3-ts/input/tex/enclose/EncloseConfiguration.ts
```diff
@@ -0,0 +1,77 @@
...
+ * The attributes allowed in \enclose{notation}[attributes]{math}
+ * @type {{[key: string]: number}}
+ */
+const ALLOWED: {[key: string]: number} = {
+  arrow: 1,
+  color: 1, mathcolor: 1,
+  background: 1, mathbackground: 1,
+  padding: 1,
+  thickness: 1
+};
```
dpvc: Can these be exported and used in the cancel file above, rather than having two copies?

mathjax3-ts/input/tex/enclose/EncloseConfiguration.ts
```diff
@@ -0,0 +1,77 @@
...
+  def.notation = notation;
+  if (def.arrow) {
+    def.notation += ' updiagonalarrow';
+    delete def.arrow;
+  }
+  parser.Push(parser.create('node', 'menclose', [math], def));
+};
+
+
+new CommandMap('enclose', {enclose:   'Enclose'}, EncloseMethods);
```
dpvc: Remove extra spacing before `'Enclose'`?

mathjax3-ts/input/tex/cancel/CancelConfiguration.ts
zorkow: Space resulted from query - replace operation. Removed.

mathjax3-ts/input/tex/cancel/CancelConfiguration.ts
zorkow: Actually no, as it accepts the same parameters as the `\enclose` macro. But the order is wrong.

mathjax3-ts/input/tex/enclose/EncloseConfiguration.ts
zorkow: Done.

mathjax3-ts/input/tex/ParseUtil.ts
zorkow: I've now implemented the full functionality of the LaTeX `keyval` function (https://www.ctan.org/pkg/keyval), with the exception that we remove arbitrary pairs of outer braces, while `keyval` only removes two pairs. But that should not make a difference.

mathjax3-ts/input/tex/ParseUtil.ts
zorkow: Outdated.

mathjax3-ts/input/tex/ParseUtil.ts
zorkow: Outdated.

zorkow:
* Actually the PR contains both the `cancel` and the `enclose` package. The former depends on the latter.
* I've now implemented the full functionality of the LaTeX `keyval` function (https://www.ctan.org/pkg/keyval), in `ParseUtil` to retrieve options. This should help with the `empheq` package as well.
* Added tests for `enclose` package.
* Added tests for attributes in `\cancel` and `\cancelto` macros.

dpvc:
Let me know when you want me to review this again.

zorkow:
Yes, please review again.

mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,110 @@ namespace ParseUtil {
...
+   *     are found.
+   * @return {EnvList} The attribute list.
+   */
+  export function keyvalOptions(attrib: string,
+                                allowed: {[key: string]: number} = null,
+                                error: boolean = false): EnvList {
+    let def: EnvList = readKeyval(attrib);
+    if (allowed) {
+      for (let key of Object.keys(def)) {
+        if (!allowed[key]) {
```
dpvc: This still needs to be 

```suggestion
        if (!allowed.hasOwnProperty(key)) {
```

since otherwise it always allows `toString`, `toSource`, valueOf` or any of the other standard properties of an `Object` to be specified, which we don't want.

mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,110 @@ namespace ParseUtil {
...
+          if (error) {
+            throw new TexError('InvalidOption',
+                               'Invalid optional argument: ', key);
```
dpvc: I think this wants to be

```suggestion
                               'Invalid optional argument: %1', key);
```

so that the key's name will be included in the message.

mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,110 @@ namespace ParseUtil {
...
+  }
+
+
+  /**
+   * Removes pairs of outer braces.
+   * @param {string} text The string to clean.
+   * @return {string} The cleaned string.
+   */
+  function removeBraces(text: string): string {
+    while (text.match(/^{.*}$/)) {
```
dpvc: I think there is a slight problem with this match.  For example, it will remove the braces from `a={x}and{y}` to give `a` the value `x}and{y`, which I don't think is what is wanted.  The braces should only be removed if they actually match.  (I haven't checked this with actual LaTeX, but I certainly *hop* it doesn't remove those braces.)

mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,110 @@ namespace ParseUtil {
...
+   * @return {[string, string, string]} The collected value, the actual end
+   *     character, and the rest of the string still to parse.
+   */
+  function readValue(text: string, end: string[]): [string, string, string] {
+    let rest = text;
+    let braces = 0;
+    let value = '';
+    while (rest) {
+      let c = rest[0];
+      rest = rest.slice(1);
```
dpvc: Making all these new copies off the string (one for each character) seems a bit inefficient.  Can you just use an index to move through the string rather than creating new strings each time you move to a new character?

mathjax3-ts/input/tex/cancel/CancelConfiguration.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @param {string} notation The type of cancel notation to use.
+ */
+
+CancelMethods.CancelTo = function(parser: TexParser, name: string) {
+  const attr = parser.GetBrackets(name, '');
+  let value = parser.ParseArg(name);
+  const math = parser.ParseArg(name);
+  const def = ParseUtil.keyvalOptions(attr, ENCLOSE_OPTIONS);
+  def ['notation'] = TexConstant.Notation.UPDIAGONALSTRIKE + ' ' +
+    TexConstant.Notation.UPDIAGONALARROW;
```
dpvc: In the old days, MathML didn't have a notation for the arrow used by `\cancel`, and so MathJax used the (non-standard) `updiagonalarrow` here (and the output jax implemented it).  but since other renderers didn't know about it, we added `updiagonalstrike` as well so that if used in native MathML, there would be at least some indication of the arrow.

Since then, MathML has added a number of arrows, and I believe the correct one is `northeastarrow`, so we should change that here (it may need to be added to `TexConstant.Notation`).  It has been long enough that I would be happy to remove the `updiagonalstrike` and just use `northeastarrow`, but I'm OK if you want to leave it for backward compatibility.

mathjax3-ts/input/tex/enclose/EncloseConfiguration.ts
```diff
@@ -0,0 +1,79 @@
...
+EncloseMethods.Enclose = function(parser: TexParser, name: string) {
+  let notation = parser.GetArgument(name).replace(/,/g, ' ');
+  const attr = parser.GetBrackets(name, '');
+  const math = parser.ParseArg(name);
+  const def = ParseUtil.keyvalOptions(attr, ENCLOSE_OPTIONS);
+  def.notation = notation;
+  if (def.arrow) {
+    def.notation += ' updiagonalarrow';
+    delete def.arrow;
+  }
```
dpvc: I think this is the same as my comments about `updiagonalarrow` as above.  It should be `northeastarrow`, and since `arrow` is not an allowed option (it's not in `ENCLOSE_OPTIONS`), I don't think this `if-then` will ever be triggered, in any case.  I think you can safely remove it.

We should probably keep a list of these changes from v2 so that they can be documented later.  I've started one, so if you have any others that you know of, let me know.

mathjax3-ts/input/tex/ParseUtil.ts
zorkow: This is now handled correctly. I've added unit tests for the method.

mathjax3-ts/input/tex/ParseUtil.ts
zorkow: Uses `index` counter now.

mathjax3-ts/input/tex/cancel/CancelConfiguration.ts
zorkow: For the record, official MathML notation is `longdiv, actuarial, phasorangle, radical, box, roundedbox, circle, left, right, top, bottom, updiagonalstrike, downdiagonalstrike, verticalstrike, horizontalstrike, northeastarrow, madruwb, text`.

zorkow: OK, total confusion.... 
https://www.w3.org/TR/MathML2/chapter3.html#presm.menclose
gives has `northeastarrow` but no `updiagonalarrow`.

This site
https://www.tutorialspoint.com/mathml/mathml_enclosing.htm
(first hit on DuckDuckGo when including `northeastarrow` in search) has `northeastarrow` but no `updiagonalarrow` and a peculiar `text` element.

Finally the Moziall docs (first hit on DDG with only menclose:
https://developer.mozilla.org/en-US/docs/Web/MathML/Element/menclose
has `updiagonalarrow` but not `northeastarrow`.

Once again MathML is not a standard but a mess. 

We should retain `updiagonalarrow`.

mathjax3-ts/input/tex/enclose/EncloseConfiguration.ts
zorkow: Removed.

zorkow:
PTAL.

Made the requested changes. In particular:
* Uses index for parsing in `keyval`.
* Correctly (but still greedily) removes outer braces and trims spaces.
* Changes to `notation` to `northeastarrow` and updates the corresponding enumerate in `TexConstants.ts`.

mathjax3-ts/input/tex/ParseUtil.ts
```diff
@@ -488,6 +488,123 @@ namespace ParseUtil {
...
+    let startCount = true;
+    while (index < length) {
+      let c = text[index++];
+      switch (c) {
+      case '{':
+        if (startCount) {
+          start++;
+        } else if (start > braces) {
+          start = braces;
+        }
```
dpvc: There is still a small problem with this function.  It will return `}` when the input is `{{}x}` because `start` will be 2 so two pairs of terminal characters are removed, leaving only the `}`.  That is because the `}` does end up setting `startCount` to `false`, but `start is never reduced to compensate for the `}` (since the `{` is never run again).

Instead, I think you want

```
      case ' ':
        break;
      case '{':
        braces++;
        if (startCount) {
          start = braces;
        }
        break;
      case '}':
        if (braces) {
          braces--;
        }
        if (startCount) {
            start = braces;
        }
      default:
        startCount = false;
      }
```

for the cases.  This make `start` be based on the brace count when `startCount` is turned off.  (I also like that it is a bit more symmetric.)

zorkow: And I thought I had caught all corner cases... Thanks! Will correct and merge.

zorkow: The problem I see with your solution is that you again only get the count for the balanced outer braces. Thus we get the same result as my initial, incorrect solutions.
But I feel we are nearly there.

dpvc: Ah, right.  Sorry about that.  Perhaps something that include the `if-then` in the `}` case, but keeps the `start > braces` check in the `{` case?

zorkow: I think I have a solution now. It's a bit ugly, though.

zorkow: It uses two flags now: `startCount` to note that we are counting left outer brackets. `stopCount` to note that we are counting initial closing brackets. `startCount` flips on the first non-left-brace. `stopCount` is set after the first `}` and unset again after the first non-`}` after the initial `}`.
(If that makes sense.)

zorkow: I'll intersperse comments into the code to explain my thinking.

Also, this is a typical automata theory problem. I'll run it by some colleagues one of these days. There must be a more elegant solution!

zorkow: Still one case missing: `{}` is not removed.

dpvc: I think if you just include

```
if (start > braces) {
  start = braces
}
```

in the `default` case of your original version, that would do it.  (I haven't looked at your new version yet).

dpvc: Also, you would need to add

```
startCount = false;
break;
```
after the '}' case, since you don't want it to fall through to `default` after this change.

zorkow: Thanks, but I think I got it now. Should at least work for now for all non-pathological cases. 
Unless you want to have another look, I'll merge.

PR #253, opened 2019-05-14
Update TeX dependencies (based on package PR reviews)

dpvc:
Update TeX extension dependencies, and remove `color` dependency from `tex` component, since it isn't included there.


dpvc:
I'm holding off on this, as I will need to update the dependencies again since some of them weren't needed.  I'm waiting until all the packages are merged, and I'll go back and verify that everything is correct.

PR #256, opened 2019-05-17
Fix a wrong css property name in TextNode.ts

KBone12:
I found a wrong css property name in mathjax3-ts/output/chtml/Wrappers/TextNode.ts. I changed the property name from `'white space'` to `'white-space'` at `mjx-measure-text`.


dpvc:
Thanks for catching that!

PR #257, opened 2019-05-20
Fix missing support for pageReady() in startup module

dpvc:
This PR fixes the startup module to perfrom the `pageReady()` function when the page is ready, which was defined in the configuration, but never actually executed.  It also makes the typeset promise always be defined, even if no typesetting occurs (this makes it easier to use, as you don't have to test for a null value).

### DIFF
--- a/mathjax3-ts/components/startup.ts
+++ b/mathjax3-ts/components/startup.ts
@@ -258,9 +258,11 @@ export namespace Startup {
     export function defaultReady() {
         getComponents();
         makeMethods();
-        if (CONFIG.typeset && MathJax.TypesetPromise) {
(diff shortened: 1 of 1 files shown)


mathjax3-ts/components/startup.ts
```diff
@@ -258,9 +258,11 @@ export namespace Startup {
     export function defaultReady() {
         getComponents();
         makeMethods();
-        if (CONFIG.typeset && MathJax.TypesetPromise) {
-            promise = pagePromise.then(() => MathJax.TypesetPromise());
+        if (CONFIG.pageReady) {
```
zorkow: Please add a comment on the use of `pageReady`.

Issue #166, opened 2018-10-16
Tracker for discussed future PR


zorkow:
* Unify indentation to 2 spaces.
* Add initialisation methods for packages to their configuration.
     * Generalise color package, newcommand etc structures.

PR #263, opened 2019-05-26
Unicode package

zorkow:
Port of the MJ2 unicode package.
I added a property in [`UnicodeConfiguration`](https://github.com/mathjax/mathjax-v3/pull/263/files#diff-542d0143534984cdd3d86875ddf33b0fR89) to retain information for later semantic processing


mathjax3-ts/input/tex/unicode/UnicodeConfiguration.ts
```diff
@@ -0,0 +1,101 @@
...
+import {CommandMap} from '../SymbolMap.js';
+import {ParseMethod} from '../Types.js';
+import ParseUtil from '../ParseUtil.js';
+import NodeUtil from '../NodeUtil.js';
+import {numeric} from '../../../util/Entities.js';
+
+// Namespace
+export let UnicodeMethods: Record<string, ParseMethod> = {};
+
+let UnicodeCache: {[key: number]: [number, number, string, number]} = {};
```
dpvc: This should probably by a TeX-input-jax-specific object rather than a global object.  (Since it is set based on user input, if you had two TeX input jax for different documents, the information would bleed from one to the other through this global.)  Unfortunately, this is not currently very easy, as the configuration objects don't have a means of specifying parser-specific data.  Other packages (like color and require) use the parser `options` object to store these, but it is a hack.

It's probably OK to leave this for now, but we need a more general solution to the parser-specific data problem for the future.

mathjax3-ts/input/tex/unicode/UnicodeConfiguration.ts
```diff
@@ -0,0 +1,101 @@
...
+ * @fileoverview Configuration file for the unicode package.
+ *
+ * @author v.sorge@mathjax.org (Volker Sorge)
+ */
+
+import {Configuration} from '../Configuration.js';
+import {EnvList} from '../StackItem.js';
+import TexParser from '../TexParser.js';
+import TexError from '../TexError.js';
+import {TexConstant} from '../TexConstants.js';
```
dpvc: This doesn't appear to be used anywhere.

mathjax3-ts/input/tex/unicode/UnicodeConfiguration.ts
zorkow: Alternative, we could make the cache a `Map<[number, string], [number, number, string, number]>` wrt. `[key, font]` pairs. Then bleed through should not matter.
That would, however mean that we would have use define the pairs as objects and give them a `Symbol.iterator`, which might be overkill.
Anyway, for now added to issue #166.

mathjax3-ts/input/tex/unicode/UnicodeConfiguration.ts
zorkow: Removed.

zorkow:
@dpvc I screwed up the merge by not rebasing to master first. I'll redo the PR and merge.

Issue #121, opened 2018-05-18
spacing of \colon


Debenben:
Using the current demo https://mathjax.github.io/mj3-demos/mj3-tex2html-global.html with Firefox/Linux and input `\begin{align}x \colon y \to z \\ x : y\to z \end{align}` results in:

![screenshot from 2018-05-18 22-29-51](https://user-images.githubusercontent.com/35615783/40256509-0447828e-5aeb-11e8-8dfa-ad298912d2e8.png)

where in the first line the colon should be closer to the x than to the y like in LaTeX:

![screenshot from 2018-05-18 22-30-45](https://user-images.githubusercontent.com/35615783/40256549-23cf5000-5aeb-11e8-997e-134b30f812d6.png) 

or MathJax on math.stackexchange: 

![screenshot from 2018-05-18 22-32-49](https://user-images.githubusercontent.com/35615783/40256643-6d170082-5aeb-11e8-9b25-a8a0e77e2b71.png)

see https://tex.stackexchange.com/questions/37789 for further discussion of the difference.

dpvc:
Thanks for the bug report.

@Volker, the `\colon` macro is supposed to set the tex class to `PUNCT`, but it does seem to be.  Can you check into this?

zorkow:
Corrected to @zorkow. I'll have a look.

dpvc:
@zorkow, any word on this issue? It seems to still be in beta.3, and we will need to get it fixed for the production release.

PR #271, opened 2019-05-31
Fix problem with texClass being overwritten.  #121

dpvc:
This PR preserves the `texClass` that has been set by the TeX input jax (it was being ovewritten by the data from the operator dictionary).  To tell when TeX has set it, we set a `texClass` property on the node, otherwise we don't know if it was the value set by the constructor or the TeX input jax.

This resolves issue #121.


dpvc:
Rats!  I referenced the wrong issue for this, so the branch is incorrectly named.  I am closing this PR and will rename the branch and open a new one.

PR #272, opened 2019-05-31
Prevent texClass from being overwritten.  #121

dpvc:
This PR preserves the `texClass` that has been set by the TeX input jax (it was being ovewritten by the data from the operator dictionary).  To tell when TeX has set it, we set a `texClass` property on the node, otherwise we don't know if it was the value set by the constructor or the TeX input jax.

This resolves issue #121.


dpvc:
OK, my fault.  The comparisons should have been either `== null` (not `=== null`) or `=== undefined`.  I have pushed a new commit that does the latter.  See if that doesn't resolve the issue.

Note that, in order to tell if the `texClass` has been set explicitly, the code adds a `texClass` property to the node, so some tests will be different because off that additional property.

zorkow:
OK, I've run the tests and only 546 fail now. Obviously I did not do a diff for all of them, but quite a number and the vast majority of changes are simply the additional `texclass` property.
The only other changes of note are that invisible left/right fences now retain their open/close `texclass`.
Similarly, explicitly declared embellished operators stay `texclass=1` now.

PR #275, opened 2019-06-05
Set inherited attributes in an explicit filter.

zorkow:
As discussed in the meeting:
* Adds filter for setting inherited attributes.
* Removes the special case for errors.


dpvc:
Looks good.  Thanks for the PR.

lgtm

PR #281, opened 2019-06-08
Allow forward references for \ref and \eqref macros

dpvc:
This PR includes fixes to allow forward references in `\ref` and `\eqref` macros.  This is handled by keeping track of equations that have references to undefined labels, and then recompiling them after all the other expressions have been processed.  That is accomplished in the `compile()` method of the `MathDocument`, which uses the `inputData` object to pass information back to the document that a `MathItem` needs to be recompiled (the `inputData.recompile` item, which includes the MathItem's previous state and the tag counter).

In order to get the labels and equation numbers to be correct, the Tags object stores the ids, labels, and counter for the current equation separately from the ids, labels, and counter for the previously processed equations.  That way, if an equation is restarted due to a `\require` or autoloaded extension, for example, and the equation is reprocessed, the id's, labels, and counter for the partially processed equation can be removed (so there will not be duplicate label warnings, and the counter will start at the right place).  This is the reason for the `allIds`, `allLabels`, and `allCounter` properties to be separate from `ids`, `labels`, and `counter`.

When an expression is reprocessed because of missing label references, the counter is reset to the proper one for when it was originally processed so it gets the proper auto tags when recompiled.

### DIFF
--- a/mathjax3-ts/input/tex/Tags.ts
+++ b/mathjax3-ts/input/tex/Tags.ts
@@ -469,7 +469,7 @@ export class AbstractTags implements Tags {
    */
   public finishEquation(math: MathItem<any, any, any>) {
     if (this.redo) {
-      math.inputData.redo = {
+      math.inputData.recompile = {
         state: math.state(),
         counter: this.allCounter
       };
--- a/mathjax3-ts/input/tex/Tags.ts
+++ b/mathjax3-ts/input/tex/Tags.ts
@@ -24,6 +24,7 @@
 
 import TexParser from './TexParser.js';
 import {MmlNode} from '../../core/MmlTree/MmlNode.js';
+import {MathItem} from '../../core/MathItem.js';
 import {EnvList} from './StackItem.js';
 import ParseOptions from './ParseOptions.js';
 import {OptionList} from '../../util/Options.js';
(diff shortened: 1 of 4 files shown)


mathjax3-ts/input/tex/Tags.ts
```diff
@@ -174,9 +179,23 @@ export interface Tags {
...
+   * Initialise tagging for a MathItem
+   * (clear equation-specific labels and ids, set counter
+   * and check for recompile)
+   * @param {MathItem} math   The MathItem for the current equation
+   */
+    startEquation(math: MathItem<any, any, any>): void;
+
+  /**
+   * Move equation-specific labels and ids to global ones,
+   * save the counter, and marth the MathItem for redos
```
zorkow: marth -> mark?

mathjax3-ts/input/tex/Tags.ts
```diff
@@ -552,19 +607,20 @@ export namespace TagsFactory {
   export let OPTIONS: OptionList = {
     // Tagging style, used to be autonumber in v2.
     tags: defaultTags,
-    //  This specifies the side on which \tag{} macros will place the tags.
-    //  Set to 'left' to place on the left-hand side.
+    // This specifies the side on which \tag{} macros will place the tags.
+    // Set to 'left' to place on the left-hand side.
     TagSide: 'right',
-    //  This is the amound of indentation (from right or left) for the tags.
+    // This is the amound of indentation (from right or left) for the tags.
```
zorkow: amound -> amount

mathjax3-ts/input/tex/Tags.ts
dpvc: Fixed

mathjax3-ts/input/tex/Tags.ts
dpvc: Fixed

PR #290, opened 2019-06-15
Adds a second output region to the v3 lab.

zorkow:
Adds an optional second output region to the v3 lab. It renders the same input but with a different mathitem. This is useful for testing against interaction of stateful tools like the explorer.


dpvc:
lgtm

PR #294, opened 2019-06-16
A11y keyboard magnifier

zorkow:
Correctly implements the goal directed keyboard magnifier.
Adds options for alignment: `top`, bottom, center`.

### DIFF
--- a/mathjax3-ts/a11y/explorer/Region.ts
+++ b/mathjax3-ts/a11y/explorer/Region.ts
@@ -313,20 +313,19 @@ export class HoverRegion extends AbstractRegion {
    * @override
    */
   protected static style: CssStyles =
-    new CssStyles({['.' + HoverRegion.className]: {
-      position: 'absolute', top: '0', height: '1px', width: '1px',
-      padding: '1px', overflow: 'hidden'
-    },
-                   ['.' + HoverRegion.className + '_Show']:
-                   {
-                     top: '0', position: 'absolute', width: 'max-content', height: 'auto',
-                     padding: '0px 0px', opacity: 1, 'z-index': '202',
-                     left: 0, right: 0, 'margin': '0 auto',
-                     'background-color': 'rgba(0, 0, 255, 0.2)',
-                     'box-shadow': '0px 10px 20px #888',
-                     border: '2px solid #CCCCCC'
-                   }
-                  });
+    new CssStyles({
+      ['.' + HoverRegion.className]: {
+        position: 'absolute', height: '1px', width: '1px',
+        padding: '1px', overflow: 'hidden'
+      },
+      ['.' + HoverRegion.className + '_Show']:
+      {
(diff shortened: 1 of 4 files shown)


mathjax3-ts/a11y/explorer/Region.ts
```diff
@@ -335,36 +334,67 @@ export class HoverRegion extends AbstractRegion {
...
+
+  /**
+   * Adds a clone of the given node to the hover region.
+   * @param {HTMLElement} node The current HTML node.
+   */
+  public AddNode(node: HTMLElement) {
     this.Clear();
-    this.inner.appendChild(node);
+    let mjx = node.cloneNode(true) as HTMLElement;
+    this.inner.appendChild(mjx);
```
dpvc: Unfortunately, simply cloning the element without the proper containers will not get the spacing and sizing correct.  A number of the CSS rules rely on the `mjx-container` and also on the `mjx-math` element being a parent of the element being displayed.  So you will probably need to include these (when they aren't already present).  These sometimes have attributes that you want to include as well (like font size, and display settings), so you probably need to find the actual elements and clone them (shallowly).

Here is an example that does it, though it is not pretty:

```
    public AddNode(node: HTMLElement) {
        this.Clear();
        let mjx = node.cloneNode(true) as HTMLElement;
        if (mjx.nodeName !== 'MJX-CONTAINER') {
            mjx.style.marginLeft = max-style.marginRight = '0';                // remove element spacing (could be done in CSS)
            let container = node;
            while (container && container.nodeName !== 'MJX-CONTAINER') {
                container = container.parentNode as HTMLElement;
            }
            if (mjx.nodeName !== 'MJX-MATH') {
                mjx = container.firstChild.cloneNode(false).appendChild(mjx).parentNode as HTMLElement;
            }
            mjx = container.cloneNode(false).appendChild(mjx).parentNode as HTMLElement;
            mjx.style.margin = '0';                 //  remove displayed math margins (could be done in CSS)
        }
        this.inner.appendChild(mjx);
        this.inner.style.lineHeight = '0';               // could be moved to when this.inner is created
    }
```

I've also removed the spacing that is added between elements, and above and below displayed expressions.  Setting the `line-height` to 0 avoids unwanted extra space below the expression.  Finally, you can remove the `MJX-TEX` class from the class list at line 368, as that is handled by the `mjx-math` element.

zorkow: I've integrated your suggestions and added some support for SVG. The SVG area is still too large, but at least displays the correct content. I am not going to spend any more time on  that right now, though, and leave it as a future issue.

dpvc: I've made a PR to this branch that handles the SVG bounding box, and noted some other problems.  The main one is that tagged equations in SVG don't get processed properly, and that might want to be addressed for v3.  The other issues can wait.

PR #297, opened 2019-06-18
Fix size of SVG viewBox for magnifier.

dpvc:
This PR fixes the magnifier for SVG elements to have the proper bounding box.

Both the SVG and HTML magnifier doesn't get the scaling factor correct when there is scaling caused by container elements that have been removed (e.g., contents of superscripts).  Perhaps that is OK, but it does mean the magnified result may be too big (or too small in some cases) compared to the original size.

It might be possible to walk up the tree and take transformations into account.  Alternatively, the output could be modified to include scaling information (which is already being computed by both output jax) as attributes so that the explorer could pick that up.  The explorer could turn on/off an output option that controls that, for example.

Also, SVG expressions with tags don't work properly.  This is probably due to the fact that the SVG structure for these in v3 is quite different from the structure for v2, and that isn't being taken into account by the explorer.

Despite these limitations, I think this is an improvement over the current version.

### DIFF
--- a/mathjax3-ts/a11y/explorer/Region.ts
+++ b/mathjax3-ts/a11y/explorer/Region.ts
@@ -404,19 +404,6 @@ export class HoverRegion extends AbstractRegion {
    * @override
    */
   protected highlight(highlighter: sre.Highlighter) {
-    let child = this.inner;
-    while (child) {
-      if (child.hasAttribute('sre-highlight')) {
-        return;
-      }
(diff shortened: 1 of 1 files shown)


zorkow:
Thanks! That looks certainly better. 
There are indeed a couple of issues with the new SVG output and the old highlighter. I have implemented a v3 specific one in the mean time, although I notice I have not pushed yet.

PR #298, opened 2019-06-18
Fixes mispelled typesetPromise method.

zorkow:
Replaces `TypesetPromise` by `typesetPromise` in two places:
* Fixes the spelling in the lab.
* Fixes spelling in `startup.ts`, although I could not produce an error with the old cap spelling.


dpvc:
Thanks for catching these.

PR #299, opened 2019-06-18
Fixes problem with double highlighting in CHTML magnifier.

zorkow:
Fixes the problem that the magnification region is highlighted twice in CHTML that was reintroduced in PR #297.


dpvc:
Sorry about causing the problem.

PR #302, opened 2019-06-21
A11y explorer refactor

zorkow:
Major refactoring of `Explorer` and `Region` module:
* Introduces type template for regions of explorers.
* Separates mouse and keyboard explorers into their own files.
* Introduces a number of explorers (including the old hover highlighter).
* Cleans up use of `sre` methods and the type file.
* Adds all missing comments. 
* SVG/CHTML works well together without extra cases with the exception of the magnification. 

Missing bits are now listed in Issue #303


zorkow:
@dpvc 
Have you started with this already? If not, I would have one more commit before you review.

dpvc:
I've looked at it, but haven't made any comments.  So go ahead and make your change.  I hope to get it done today.

zorkow:
OK, I've made the changes. Nothing major, just a better separation of key and mouse explorers.

PR #305, opened 2019-06-24
Support for loading latest version of a component

dpvc:
This PR adds a `latest.js` file that can be used to load the lastest version of a component from one of the supported CDNs.  it would be used as

```
<script src="path-to-cdn/mathjax/3.0.0/latest.js?tex-chtml.js"><script>
```

to get the latest (3.x.y) version of `tex-chtml.js`.  It will not load 4.x.y, but will use the highest 3.x.y version.  If it is not loaded from one of the known CDNs, or if anything goes wrong during the current-version lookup (e.g., can't reach the CDNs API for getting the version), it will try to load the file from the same location as `latest.js` itself.


dpvc:
@zorkow, hold off on this for now.  I thought of something else that needs to be done with it first.  Sorry for pulling the trigger too soon.

dpvc:
OK, I've finished my updates, including translating to Typescript.

Just FYI, this is based on the `latest.js` file from v2, but I've refactored some of it, and added some improvements, including several new CDN.  One important improvement is the following:  the program will only load the latest version within the current major version (so the v2 version only loads the latest 2.x.y version); but if the current version is a higher major version, the v2 `latest.js` will just load the version that `latest.js` came from.  That means it may drop back to an earlier version when 3.0.0 is the current version (i.e., 2.7.3's `latest.js` currently loads 2.7.5, but once 3.0.0 is released, it will fall back to 2.7.3 because the current version isn't 2.x.y).  This new 3.0.0 `latest.js` will check *all* the versions (from GitHub) in order to find the latest 3.x.y, even if the current version is 4.x.y, so you should always get the latest 3.x.y regardless of the current version (that's not the case with the v2 version, unfortunately).

Also, if `latest.js` detects a newer version, it will actually load `latest.js` from the new version, just in case anything was modified.  That way, if we need to make any changes to `latest.js` in a later version, you will get the changes even if you loaded `latest.js` from an earlier version.

So the upshot is that this should be more reliable that the v2 `latest.js`.

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+ *   Retrieve the cached version (if any)
+ *   Load the given version of the file, if the version is cached,
+ *   Otherwise find the latest version and load that.
+ * Otherwise,
+ *   Load using the version where latest.js was loaded.
+ */
+export function loadLatest() {
+    script = getScript();
+    if (script && script.cdn) {
+        var version = getSavedVersion();
```
zorkow:
```suggestion
        const version = getSavedVersion();
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+ * Find the script that loaded latest.js
+ * If the script is from a known CDN:
+ *   Retrieve the cached version (if any)
+ *   Load the given version of the file, if the version is cached,
+ *   Otherwise find the latest version and load that.
+ * Otherwise,
+ *   Load using the version where latest.js was loaded.
+ */
+export function loadLatest() {
+    script = getScript();
```
zorkow:
```suggestion
    const script = getScript();
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+ */
+const MJX_LATEST = 'mjx-latest-version';
+
+/**
+ * The amount of time a cached version number is valid
+ */
+const SAVE_TIME = 1000 * 60 * 60 * 24 * 7;   // one week
+
+/**
+ *  data for the script that loaded latest.js
```
zorkow:
```suggestion
 *  Data for the script that loaded latest.js
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+
+/*=====================================================================*/
+
+/**
+ * Produce an error message on the console
+ *
+ * @param {string} message   The error message to display
+ */
+function Error(message: string) {
+    if (console && console.error) console.error('MathJax(latest.js): ' + message);
```
zorkow:
```suggestion
    if (console && console.error) {
        console.error('MathJax(latest.js): ' + message);
    }
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+}
+
+/**
+ * @return {ScriptData}   The data for the script tag that loaded latest.js
+ */
+function getScript() {
+    if (document.currentScript) {
+        return scriptData(document.currentScript as HTMLScriptElement);
+    }
+    const script = document.getElementById('MathJax-script') as HTMLScriptElement;
```
zorkow: Does that mean `MathJax-script` can be given in any tag, even non-`script` ones?

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+    try {
+        const data = version + ' ' + Date.now();
+        localStorage.setItem(MJX_LATEST, data);
+    } catch (err) {}
+}
+
+/**
+ * Get the version from localStorage, and make sure it is fresh enough to use
+ *
+ * return {string|null}   The version string (if one has been saved) or null (if not)
```
zorkow:
```suggestion
 * @return {string|null}   The version string (if one has been saved) or null (if not)
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+ */
+function loadVersion(version: string) {
+    if (script.version && script.version !== version) {
+        script.file = 'latest.js?' + script.file;
+    }
+    loadMathJax(script.cdn.base + version + '/' + script.file, script.id);
+}
+
+/**
+ * Check if the given version acceptable and load it if it is.
```
zorkow:
```suggestion
 * Check if the given version is acceptable and load it if it is.
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+
+/*=====================================================================*/
+
+/**
+ * Create an XMLHttpRequest object, if possible
+ *
+ * @return {XMLHttpRequest}   The XMLHttpRequest instance
+ */
+function getXMLHttpRequest() {
+    if (window.XMLHttpRequest) return new XMLHttpRequest();
```
zorkow:
```suggestion
    if (window.XMLHttpRequest) {
        return new XMLHttpRequest();
    }
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+        try {return new window.ActiveXObject('Microsoft.XMLHTTP')} catch (err) {}
+    }
+}
+
+/**
+ * Request JSON data from a CDN.  If it loads OK, call the action() function
+ * on the data.  If not, or if the action returns false, run the failure() function.
+ *
+ * @param {CdnData} cdn        The CDN whose API will be used
+ * @param {Function} action    The function to perfrom when the data are received
```
zorkow:
```suggestion
 * @param {Function} action    The function to perform when the data are received
```

mathjax3-ts/components/latest.ts
```diff
@@ -0,0 +1,398 @@
...
+ * @param {Function} failure   The function to perform is that data can't be obtained,
```
zorkow:
```suggestion
 * @param {Function} failure   The function to perform if data can't be obtained,
```

mathjax3-ts/components/latest.ts
dpvc: Right.  I originally had this file as a plain javascript one (where I didn't want to use ES6 for backward compatibility), then decided to convert to typescript.  Missed changing that one.

mathjax3-ts/components/latest.ts
dpvc: This one should be left as is, as there is a `let script = ...` at line 122 that is being set by this command.  That variable is used elsewhere and so can't be localized to this block.

mathjax3-ts/components/latest.ts
dpvc: OK.

mathjax3-ts/components/latest.ts
dpvc: I will add a check that it is on a `script` tag.  Note that this is only for compatibility with old browsers (namely IE), since modern browsers would have already found the script via `currentScript` above).

dpvc:
> From line 150 I see that betas can never be latest. I assume that is intended. But in line 287 we check for beta versions/release candidates. I am not sure I fully understand that.

There are two versions being tracked:  the version where `latest.js` is being loaded, and the latest version actually available on the CDN.  Line 150 finds the first of these (and we do assume that this is not a beta version; I could make it look for that, too, though I think it would be odd to use `latest.js` from a beta version except for our testing purposes).  Line 287 is analyzing the versions available on the CDN to find the latest release version.  This rules out beta versions, since we don't want `latest.js` to direct you to a beta version, only official release versions.

The reason we need to know both versions is so that we can tell if `latest.js` is directing us to a higher version than it originated from.  If so, it actually calls the `latest.js` in that version, so that if there are changes in `latest.js` in the new version, we will get those.  (Had the current v2 .x`latest.js` done that, we would not have the issue we have discussed about needing to have people update to 2.7.6 to get the fixed `latest.js` that must be used once version 3.0 is out.)

zorkow:
Thanks for the explanation. I agree that latest should be default __not__ look for beta versions.

PR #312, opened 2019-07-16
Startup document

dpvc:
This PR makes `getDocument()` return the document (so that it can be used to create additional documents, if needed, e.g., for multiple iframes or web components).  It also moves the creation of the original document into the `getComponents()` function rather than the `makeMethods()` function, since that seems more natural.

### DIFF
--- a/mathjax3-ts/components/startup.ts
+++ b/mathjax3-ts/components/startup.ts
@@ -480,7 +480,7 @@ export namespace Startup {
     /**
      * Create the document with the given input and output jax
      *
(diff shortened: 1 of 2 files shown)


mathjax3-ts/components/startup.ts
```diff
@@ -476,11 +480,15 @@ export namespace Startup {
     /**
      * Create the document with the given input and output jax
      *
+     * @param {any=} root        The Document to use as the root document (or null to use the cofigured document)
```
zorkow:
```suggestion
     * @param {any=} root        The Document to use as the root document (or null to use the configured document)
```

PR #322, opened 2019-08-02
Support for additional output configuration options

dpvc:
Add support for missing `minScale` and `mtextInheritFont` configuration options, and add new `merrorInheritFont` option.  This required adding new `mtext` wrappers (since the default wrapper actions are no longer sufficient).

### DIFF
--- a/mathjax3-ts/output/svg/Wrappers/mtext.ts
+++ b/mathjax3-ts/output/svg/Wrappers/mtext.ts
@@ -22,7 +22,7 @@
  */
 
 import {SVGWrapper, SVGConstructor} from '../Wrapper.js';
-import {CommonMtext, CommonMtextMixin} from '../../common/Wrappers/mtext.js';
+import {CommonMtextMixin} from '../../common/Wrappers/mtext.js';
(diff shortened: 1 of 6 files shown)


mathjax3-ts/output/svg/Wrappers/mtext.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+/**
+ * @fileoverview  Implements the SVGmtext wrapper for the MmlMtext object
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+import {SVGWrapper, SVGConstructor} from '../Wrapper.js';
+import {CommonMtext, CommonMtextMixin} from '../../common/Wrappers/mtext.js';
```
zorkow: `CommonMtext` does not seem to be used.

dpvc: Ok, removed it.  It turns out that many of the wrappers have similar unused includes.  Should I make a PR to remove them as well?

PR #323, opened 2019-08-14
Slim down the startup component

dpvc:
This PR updates `mathjax3-ts/components/startup.js` to have fewer dependencies (i.e., not rely on `output/common/OutputJax.js`), and to link to the core library for `core/MathItem.js` (since it is not needed until after core will be loaded).  This reduces the component size from 47KB to 18KB.


zorkow:
> This reduces the component size from 47KB to 18KB.
Nice!

PR #329, opened 2019-08-30
A11y updates

zorkow:
PR makes a number of fixes and improvements for the A11Y module:
* Improved synchronisation between between menu and a11y module. Option values are now directly retrieved from the a11y document via `getA11y`. `a11yVar` is similar to `variable` method in menu to generate the JSON with handling of setting and getting of options in the a11y document.
* Since speech rule options can not only be changed in the menu but also via keys in the walker, some more syncing is needed in the `KeyExplorer`. This is currently restricted to `speech` modality. Eventually that needs to be generalised once we have more than one type of Braille, for example.
* `speechRules` options are simplified.
* `magnification` options are simplified and corrected.
* I've introduced a general way to handle dynamic submenu changes in the `MJContextMenu` class.
Currently this is only used for the ClearSpeak preferences menu. The current menu option is effectively an placeholder that is updated from SRE.


zorkow:
And of course it only works properly when SRE `v3_clearspeak_rule_cycling` branch.

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -207,6 +207,10 @@ export class SpeechExplorer extends AbstractKeyExplorer<string> {
   public Update(force: boolean = false) {
     super.Update(force);
     this.region.Update(this.walker.speech());
+    let options = this.speechGenerator.getOptions();
```
dpvc: Should this be just `this.getOptions()` like line 109 above?  Or does this not need the extra walker update from that function?

ts/ui/menu/MJContextMenu.ts
```diff
@@ -78,6 +86,7 @@ export class MJContextMenu extends ContextMenu.ContextMenu {
                 const semantics = this.findID('Settings', 'semantics');
                 input === 'MathML' ? semantics.disable() : semantics.enable();
                 this.getAnnotationMenu();
+                this.dynamicSubmenus();
```
dpvc: I'm wondering if `getAnnotationMenu()` could be folded into `dynamicSubmenus()` eventually (future release).

ts/ui/menu/Menu.ts
```diff
@@ -981,6 +974,29 @@ export class Menu {
         };
     }
 
+    /**
+     * Create JSON for an a11y specific variable.
+     *
+     * @param {keyof MenuSettings} name   The setting for which to make a variable
+     * @param {(T) => void} action        Optional function to perform after setting the value
```
dpvc: This argument doesn't seem to be there any longer.

ts/a11y/explorer/KeyExplorer.ts
zorkow: No that part is necessary in case the speech options have changed by key pressing while walking the expression.  Merging in the `document` options would destroy the effect. I've added a comment to that extent.

ts/ui/menu/MJContextMenu.ts
zorkow: I had been thinking exactly the same thing, but did not want to play with it for the time being. 
I was also thinking that these types of dynamic menu updates could eventually move into the `contextmenu` package altogether. Added it to the futures list.

ts/ui/menu/Menu.ts
zorkow: Mindless copy and pasting...

PR #330, opened 2019-08-31
Add node-main component

dpvc:
This PR adds a `node-main` component and makes it the default for node `require('mathjax-full')`.

This lets you do

```
require('mathjax-full').init({
  ...configuration...
}).then((MathJax) => {
  ...your MathJax code...
}).catch((err) => console.error(err.message));
```

to load, configure, and run MathJax code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "3.0.0",
   "description": "Beautiful math in all browsers. MathJax is an open-source JavaScript display engine for LaTeX, MathML, and AsciiMath notation that works in all browsers. This package includes the source code as well as the packaged components.",
   "license": "Apache-2.0",
-  "main": "es5/node-main.js",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "3.0.0",
   "description": "Beautiful math in all browsers. MathJax is an open-source JavaScript display engine for LaTeX, MathML, and AsciiMath notation that works in all browsers. This package includes the source code as well as the packaged components.",
   "license": "Apache-2.0",
-  "main": "es5/node-main.js",
+  "main": "components/src/node-main.js",
```
zorkow: Should that not be 
```suggestion
  "main": "components/src/node-main/node-main.js",
```
?

dpvc: Right, thanks!

PR #331, opened 2019-08-31
Bussproofs package

zorkow:
The bussproofs package.
* Includes components to build it.
* Fixes the bbox problem for incomplete tables elements.
* Note, that sequent balancing does not fully work yet, in terms of the spatial layout.


dpvc:
I've taken a quick look here, but haven't done a detailed review.  It looks fine, except that the Utils includes two adopters and the RegisterHTMLHandler function that are no longer used.  That will increase the size of the component.  I will make a PR to fix that if you are done for the day.

Also, because this relies on the CHTML output, it needs to have that as a dependency, and should include that in the components library so that the component doesn't duplicate the CHTML code and font data.  In fact, I'm thinking we can get away with using the a dummy subclass of the common output jax rather than CHTML in order to void loading all the CHTML code that is not necessary (the bounding box information is all in common).  This will mean you only have to have a dependency on the common output, which should be able to be resolved by either output jax.  That should make the component smaller.  I'm a little worried about including it in all-packages if it has to include all of the CHTML output as part of it.

I'll have to get back to you on the details, but wanted to say what I was doing before you ended your day.

zorkow:
Thanks for the quick turnaround.  I wanted to cleanup dependencies but then forgot in the rush. Sorry about that.
I was also thinking that we don't need to be specific wrt. output jax. But `CommonOutputJax` is an abstract class so can't be instantiated. How should I do that?

dpvc:
Just make a subclass of it and instantiate that:

```
class BBoxJax extends CommonOutputJax {}

const jax = new BBoxJax();
```

I'm working on a PR to handle these issues.

zorkow:
I see. I've just removed the obsolete imports. I hope we are not running into merge conflicts.

dpvc:
I'll rebase my changes when I'm done.

dpvc:
Can you give me a sample proof that I can use for testing?

zorkow:
I'll send an email.

zorkow:
There is still one more issue I had not remembered: The background highlighting only works properly with style
```HTML
<style>
mjx-mtable * {
background-color: inherit;
}
</style>
```
and with tweaking the opacity in the explorer. Otherwise the columns are too "thin". 

For the time being I think the current version is fine and I file an issue for later consideration.

PR #332, opened 2019-09-02
Fix Bussproofs handing of document and output jax

dpvc:
This PR changes how `bussproofs` obtains a MathDocument and OutputJax.  It should use the actual MathDocument and OutputJax in use, since (in the future) these will be able to specify the font in use, and that will affect the spacing.  So we don't want to use a generic math document and output jax, but the specific one in use for the document.

Unfortunately, the InputJax doesn't have access to the MathDocument when it is created (since the InputJax is instantiated before the MathDocument), so the TeX extensions don't have access to it while they are initialized.

In order to overcome this, I've added the MathDocument to the arguments to `InputJax.compile()` function (it probably should have had that to begin with), and pass it along to the pre- and post-filters via `InputJax.executeFilters()` (and `OutputJax.executeFilter()` for consistency).  Because the TeX extensions can supply pre-and post-filters, this gives them access to the document.

So the bussproofs extension grabs the document using a pre-filter and sets the `doc` variable using that, rather than creating a new document.  It gets the output jax from the MathDocument rather than creating a new instance of one.

Because of the addition of the document to the `compile()` and `executeFilters()` functions, there are a bunch of non-bussproof files that are being adjusted, here.

Also, there was a typo in the component file for the SVG output that prevented it from loading a font file when the compressed component file was used.  Glad I caught that one!

### DIFF
--- a/ts/input/tex/bussproofs/BussproofsUtil.ts
+++ b/ts/input/tex/bussproofs/BussproofsUtil.ts
@@ -28,31 +28,24 @@ import NodeUtil from '../NodeUtil.js';
 import ParseUtil from '../ParseUtil.js';
 
 import {CHTML} from '../../../output/chtml.js';
-import {HTMLMathItem} from '../../../handlers/html/HTMLMathItem.js';
 import {MmlNode} from '../../../core/MmlTree/MmlNode.js';
 import {mathjax} from '../../../mathjax.js';
 import {Property, PropertyList} from '../../../core/Tree/Node.js';
+import {MathItem} from '../../../core/MathItem.js';
+import {MathDocument} from '../../../core/MathDocument.js';
 
+
+type MATHITEM = MathItem<any, any, any>;
+type MATHDOCUMENT = MathDocument<any, any, any>;
+
+type FilterData = {math: MATHITEM, document: MATHDOCUMENT, data: ParseOptions};
 
 /**
  *  Global constants local to the module. They instantiate an output jax for
  *  bounding box computation.
  */
-let jax: any = null;
-let doc: any = null;
-let item: any = null;
-
-
-/**
- * Initialisation method for setting the jax and document needed to compute the
- * bounding box information.
- */
-export let initUtil = function() {
-  jax = new CHTML();
-  doc = mathjax.document('<html></html>', {
-    OutputJax: jax
-  });
-};
+let doc: MATHDOCUMENT = null;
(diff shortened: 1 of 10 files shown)


ts/input/tex/bussproofs/BussproofsUtil.ts
```diff
@@ -28,31 +28,24 @@ import NodeUtil from '../NodeUtil.js';
 import ParseUtil from '../ParseUtil.js';
 
 import {CHTML} from '../../../output/chtml.js';
```
zorkow: Can this be removed now?

dpvc: Yes, sorry, I missed that.

dpvc:
> So if I understand correctly, we are now using a fully instantiated outputJax instead of the abstract one that only does the Bbox computation. 

Correct.

> Since the major concern was that the component would become too large having to include the CHTML jax, does this version reduce that size, given that it now has to import MathDocument?

Yes, this takes care of the size concern.  The MathDocument is used only in Typescript for the typing, and isn't actually referenced from the web packed copy, so it does not contribute to the size.  Also, it is in the core component, which this module links to (so doesn't load the full file and its dependencies), so even if we did use MathDocument in a fundamental way, it is in the shared core component and would be obtained from there.

So the upshot is, the component is 17eKB, as opposed to 373KB originally, so this does take care of the size issue.  Even if we had linked against the CHTML output component (which would have reduced the size), that would still have a dependency issue.  So if someone was using SVG output, you would still load CHTML as well, making a large download.  So being able to access the user's document and output jax is a big win ion size, as well as not having to worry about the font that the user has specified.

> I do feel a bit uneasy about having to drag the document around. In many ways it goes against the idea that the InputJax work on single items. I suppose it is all right for now, but after the release we should think again if there is not an alternative way to support packages like bussproofs that need (abstract) bounding boxes.

I understand the concern, but if we are going to be allowing the measuring of sizes, we need access to the output jax, since the size of things will depend on the font that is selected in the output.  While it would be possible to make only the font information available to the input jax, you then run into the issue of which output jax is being used, and potentially having a dependency on two output jax.

We can certainly think about it further, but if the input jax is going to get output sizes, it is probably going to need to have access to the output jax for that.

dpvc:
I fixed the import.

zorkow:
Thanks for the explanations. I shall merge directly.

PR #340, opened 2019-09-11
Fix multiple issues with handling of classlist.  mathjax/MathJax#2185

dpvc:
Fix problem with handling of multiple classes being added to the output, and fix problem with multiple classes being added to TeXAtom output during serialization of the internal MathML.

Resolves mathjax/MathJax#2185.  See that issue for examples.


dpvc:
@zorkow — Ok, I've adjusted to avoid the empty class names.  Can you re-review?

PR #344, opened 2019-09-15
Fix spacing around super- and subscripts.  mathjax/MathJax#2187

dpvc:
Correct the default `texClass` for `AbstractMmlBaseNode` elements. It should be `TEXCLASS.ORD`, not `TEXCLASS.NONE`.  This affects the situation when the base of the exponent is not a simple symbol, like when it is `\mathbb{...}` or even just `{...}`.

 Resolves mathjax/MathJax#2187


zorkow:
Looks OK, but I would like to check if this breaks my tests first.

PR #346, opened 2019-09-15
Fix problem with math following <br>.  mathjax/MathJax#2202

dpvc:
Make sure we use text nodes for the start and end locations (there can be `<br>` nodes in the list, since they are allowed inside math).  We want to make sure we skip over them.

Resolves issue mathjax/MathJax#2202


NicolasCARPi:
I was running into this issue! Thanks for the fix. Now waiting for this fix to be released in stable! :)

PR #350, opened 2019-09-22
Check that parent is defined before getting its metrics.  mathjax/MathJax#2191

dpvc:
If a typeset expression has been removed from the page, typesetting again can cause an error when the metrics are being determined.  This PR checks that the location where the metrics are being taken has a parent element before trying to use it.

Resolves issue mathjax/MathJax#2191.

### DIFF
--- a/ts/output/common/OutputJax.ts
+++ b/ts/output/common/OutputJax.ts
@@ -23,7 +23,7 @@
 
 import {AbstractOutputJax} from '../../core/OutputJax.js';
 import {MathDocument} from '../../core/MathDocument.js';
-import {MathItem, Metrics} from '../../core/MathItem.js';
+import {MathItem, Metrics, STATE} from '../../core/MathItem.js';
 import {MmlNode} from '../../core/MmlTree/MmlNode.js';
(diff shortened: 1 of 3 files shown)


dpvc:
I added code to check that math item metrics are not being remeasured when they have already been measured, since it is not efficient to remeasure old math when new math is being processed.

dpvc:
I've also added code that prevents typeset math from attempting to be removed when the container is no longer in the document (e.g., the container had `innerHTML` set and the typeset math is now orphaned).

ts/output/common/OutputJax.ts
```diff
@@ -296,9 +300,11 @@ export abstract class CommonOutputJax<
         //
         for (const math of html.math) {
             const node = adaptor.parent(math.start.node);
-            const map = domMaps[math.display? 1 : 0];
-            if (!map.has(node)) {
-                map.set(node, this.getTestElement(node, math.display));
+            if (node && math.state() < STATE.METRICS) {
+                const map = domMaps[math.display? 1 : 0];
```
zorkow: Missing space before ?

dpvc: Fixed.

PR #353, opened 2019-09-23
Add attribute to identify texClass for TeXAtom output. mathjax/MathJax#2193

dpvc:
This PR adds an attribute to CHTML and SVG output for the TeXAtom elements that identifies the texClass of the element.  This was prompted by the discussion in mathjax/MathJax#2193


dpvc:
> I am a bit confused why we use mjx-data for chtml but not for svg.

It's the other way `round:  no `mjx-data` in html because those are custom elements, and we get to decide what the attributes are (no need for `data-` there, as with all the other attributes we've used), whereas in `svg`, the attributes are determined by the SVG spec, and so we need to use `data-mjx-` for those.

PR #355, opened 2019-09-27
A11y aria labels computation

zorkow:
Fixes the aria-label computation in `semantic-enrich` by recursing to the semantic root node of the node if necessary.


dpvc:
Thanks for doing this one.  I hadn't gotten around to it.

LGTM.

PR #356, opened 2019-09-27
Moves speech generation to first initialisation of walker.

zorkow:
__Old behaviour:__
* speech attributes were computed during walker initialisation.
* explorer could immediately access pre-computed speech

SRE's newest version will lazily compute the elements of a walker, which in particular reduces the initial parse time for the enriched nodes that are passed as strings. That's moot if the explorer immediately computes the speech on initialisation (at page load).

__New behaviour:__
* speech is only computed when explorer is started.
* we add a init flag to see if the explorer has already initialised
* while this slightly reduces the startup speed for the individual walkers (only noticable with very large equations) it reduces the load time on pages.


dpvc:
Looks good.  LGTM.

PR #357, opened 2019-09-27
Override for enrichSpeech option to general aria labels in A11y.

zorkow:
Sets the `enrichspeech` option by default to `shallow` so aria labels will be computed when a11y is active.


dpvc:
LGTM.

Issue #306, opened 2019-06-27
[SVG] Safari compatibility


uetchy:
Safari does not display SVG generated from MathJax v3 while Chrome and Firefox can handle it.

## Reproducible example
[![Edit MathJax v3](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/mathjax-v3-vwffw?fontsize=14)

## Version
- Safari 12.1.1
- macOS Mojave

dpvc:
It turns out that WebKit has problems serializing namespaces attributes like `xlink:href`, and so the links from the `use` elements to their actual paths are not being made when you serialize the SVG element and then repairs it with `innerHTML`.

I have a patch that fixes the issue, so it will be fixed in the official 3.0.0 release.

In the meantime, however, there are several possible ways to work around this.  The easiest is to use

```
      OutputJax: new SVG({fontCache: 'none'})
```

which turns off the font cache, so explicit paths will be used rather than `use` elements.

Alternatively, depending on what you are actually trying to accomplish, you could avoid the serialization and simply append the SVG element into the page directly.  Simply use

```
      const math = html.math.pop();
      resolve(math);
```

in your `tex2svg()` function, and then use 

```
  document.querySelector("#app").appendChild(math);
```

to add it to the page.  This is more efficient than serializing the SVG element and then re-parsing that serialized string again.

-----

I also have several suggestions about your code example.  First, as of beta.4, there is a `convert()` function that does the compile/typeset operation for you, so you don't have to make a new document containing the math, and call the processing functions by hand.  So you can use

```
function tex2svg(equation) {
  return new Promise((resolve, reject) => {
    const adaptor = chooseAdaptor();
    RegisterHTMLHandler(adaptor);

    const html = MathJax.document('', {
      InputJax: new TeX(),
      OutputJax: new SVG({fontCache: 'none'})
    });

    MathJax.handleRetriesFor(() => {
      const math = html.convert(equation, {display: false});
      resolve(adaptor.innerHTML(math));
    }).catch(err => reject(err));
  });
}
```

Also, you don't really need to create a new document and adaptor for each equation, and you should **not** register the HTML handler more than once, so you can move that code outside your `tex2svg()` function:

```
const adaptor = chooseAdaptor();
RegisterHTMLHandler(adaptor);

const html = MathJax.document('', {
  InputJax: new TeX(),
  OutputJax: new SVG({fontCache: 'none'})
});

function tex2svg(equation) {
[...]

PR #366, opened 2019-10-10
Fix duplicative namespace on href attribute in SVG wrapper

### LINKED ISSUE #365: SVG output contains garbled `xlink:xlink:href` attribute
While trying to generate an SVG, with some relatively straightforward code:

```ts
const adaptor = liteAdaptor();
RegisterHTMLHandler(adaptor);

const Tex = new TeX({packages: AllPackages});
const Svg = new SVG({});
const MathJaxDoc = mathjax.document('', {InputJax: Tex, OutputJax: Svg});

const node = MathJaxDoc.convert(source, {display: true});
const svgContent = adaptor.innerHTML(node);
```

However, this is outputting the following broken SVG for a simple "1+2+3" formula:

```svg
<svg style="vertical-align: -0.186ex" 
  xmlns="http://www.w3.org/2000/svg" width="8.925ex" height="1.692ex" role="img" focusable="false" viewBox="0 -666 3944.9 748" 
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs>
    <path id="MJX-2-TEX-N-31" d="M213 578L200 573Q186 568 160 563T102 556H83V602H102Q149 604 189 617T245 641T273 663Q275 666 285 666Q294 666 302 660V361L303 61Q310 54 315 52T339 48T401 46H427V0H416Q395 3 257 3Q121 3 100 0H88V46H114Q136 46 152 46T177 47T193 50T201 52T207 57T213 61V578Z"></path>
    <path id="MJX-2-TEX-N-2B" d="M56 237T56 250T70 270H369V420L370 570Q380 583 389 583Q402 583 409 568V270H707Q722 262 722 250T707 230H409V-68Q401 -82 391 -82H389H387Q375 -82 369 -68V230H70Q56 237 56 250Z"></path>
    <path id="MJX-2-TEX-N-32" d="M109 429Q82 429 66 447T50 491Q50 562 103 614T235 666Q326 666 387 610T449 465Q449 422 429 383T381 315T301 241Q265 210 201 149L142 93L218 92Q375 92 385 97Q392 99 409 186V189H449V186Q448 183 436 95T421 3V0H50V19V31Q50 38 56 46T86 81Q115 113 136 137Q145 147 170 174T204 211T233 244T261 278T284 308T305 340T320 369T333 401T340 431T343 464Q343 527 309 573T212 619Q179 619 154 602T119 569T109 550Q109 549 114 549Q132 549 151 535T170 489Q170 464 154 447T109 429Z"></path>
    <path id="MJX-2-TEX-N-33" d="M127 463Q100 463 85 480T69 524Q69 579 117 622T233 665Q268 665 277 664Q351 652 390 611T430 522Q430 470 396 421T302 350L299 348Q299 347 308 345T337 336T375 315Q457 262 457 175Q457 96 395 37T238 -22Q158 -22 100 21T42 130Q42 158 60 175T105 193Q133 193 151 175T169 130Q169 119 166 110T159 94T148 82T136 74T126 70T118 67L114 66Q165 21 238 21Q293 21 321 74Q338 107 338 175V195Q338 290 274 322Q259 328 213 329L171 330L168 332Q166 335 166 348Q166 366 174 366Q202 366 232 371Q266 376 294 413T322 525V533Q322 590 287 612Q265 626 240 626Q208 626 181 615T143 592T132 580H135Q138 579 143 578T153 573T165 566T175 555T183 540T186 520Q186 498 172 481T127 463Z"></path>
  </defs>
  <g stroke="currentColor" fill="currentColor" stroke-width="0" transform="matrix(1 0 0 -1 0 0)">
    <g data-mml-node="math">
      <g data-mml-node="mn">
        <use xlink:xlink:href="#MJX-2-TEX-N-31"></use>
      </g>
      <g data-mml-node="mo" transform="translate(722.2, 0)">
        <use xlink:xlink:href="#MJX-2-TEX-N-2B"></use>
      </g>
      <g data-mml-node="mn" transform="translate(1722.4, 0)">
        <use xlink:xlink:href="#MJX-2-TEX-N-32"></use>
      </g>
      <g data-mml-node="mo" transform="translate(2444.7, 0)">
        <use xlink:xlink:href="#MJX-2-TEX-N-2B"></use>
      </g>
      <g data-mml-node="mn" transform="translate(3444.9, 0)">
        <use xlink:xlink:href="#MJX-2-TEX-N-33"></use>
      </g>
    </g>
  </g>
</svg>
```

Specifically, notice the `<use xlink:xlink:href="#MJX-2-TEX-N-33"></use>`. From what I can tell, the issue is that `xlink` is being added by both the SVG wrapper and the LiteAdaptor:

https://github.com/mathjax/MathJax-src/blob/070448bc97b97fe83e9ec8027665373d347f1032/ts/adaptors/liteAdaptor.ts#L426-L428

https://github.com/mathjax/MathJax-src/blob/070448bc97b97fe83e9ec8027665373d347f1032/ts/output/svg/Wrapper.ts#L319

vaskevich:
Fixes #365. Not sure if there is any applicable test coverage.

### DIFF
--- a/ts/output/svg/Wrapper.ts
+++ b/ts/output/svg/Wrapper.ts
@@ -316,7 +316,7 @@ CommonWrapper<
     protected useNode(variant: string, C: string, path: string) {
         const use = this.svg('use');
         const id = '#' + this.jax.fontCache.cachePath(variant, C, path);
-        this.adaptor.setAttribute(use, 'xlink:href', id, XLINKNS);
+        this.adaptor.setAttribute(use, 'href', id, XLINKNS);
         return use;
     }
 


dpvc:
Thanks for the proposed fix.  This actually undoes a fix for an earlier issue (#306), and I'd like to provide a solution that works for both the LiteDOM and the browser DOM.  Because HTML5 deprecated namespaces, the serialization of the SVG elements by the browser adaptor no longer includes the `xlink:` before the `href`, and that is why it was added in the attribute name itself.  But as you point out, this causes duplication in the LiteDOM output.

We had a discussion about that in a recent developer's meeting, and what we would like to propose is a modification to your PR.  In addition to removing the `xlink:` where you have, we would like to add the lines

https://github.com/mathjax/MathJax-src/blob/1e1a30f0a89e52fa7764dbb510763145a4da49f1/ts/adaptors/liteAdaptor.ts#L426-L428

in the corresponding place in the HTMLadaptor.  This will mean that, instead of putting the namespace in the namespace in the attribute explicitly (in the place where you have removed it), it will be added automatically in (both) the adaptors.  It would also be good to make it

```
        if (ns) {
            name = ns.replace(/.*\//, '') + ':' + name.replace(/^.*:/, '');
        }
```

so that any explicit namespaces in the attributes are not duplicated.

Could you modify the PR to do those things?

dpvc:
I've made a new PR that includes the changes that I requested above.  I would have used your PR as the starting place, except that we use the git-flow model, which means that changes should be in a branch from the `develop` branch (not the `master` branch), and targeted to the `develop` branch.  Merges are made to `master` only when a release is made.  Rather than have you rebase your branch, since this was a small change, I just made a new branch with your change plus the ones I requested.  So I'm closing this PR.

We appreciate your interest in contributing to MathJax.  Thanks!

vaskevich:
Ah, this fell off my radar. Thanks for implementing the proper fix!

Issue #328, opened 2019-08-28
LiteDOM doesn't parse self-closing tags


dpvc:
Lite DOM doesn't parse self-closing tags.  E.g., parsing

    <math><mi/></math>

will fail.  See mathjax/mj3-demos-node#5 for original report.

PR #370, opened 2019-10-20
Handle self-closing tags in liteDOM.  #328

dpvc:
This PR fixes the liteDOM to handle self-closing tags (in particular, for MathML that has things like `<mspace width="1em" />` or `<mrow/>`).

Resolves issue #328

### DIFF
--- a/ts/adaptors/lite/Parser.ts
+++ b/ts/adaptors/lite/Parser.ts
@@ -189,7 +189,7 @@ export class LiteParser implements MinDOMParser<LiteDocument> {
         //
         // Get the child to be added to the node
         //
-        const kind = tag.match(/<(.*?)[\s\n>]/)[1].toLowerCase();
+        const kind = tag.match(/<(.*?)[\s\n>/]/)[1].toLowerCase();
         const child = adaptor.node(kind) as LiteElement;
         //
         // Split out the tag attributes as an array of space, name, value1, value3, value3,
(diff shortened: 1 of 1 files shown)


ts/adaptors/lite/Parser.ts
```diff
@@ -211,7 +211,7 @@ export class LiteParser implements MinDOMParser<LiteDocument> {
             //   Otherwise, the child tag becames the parent node to which
```
zorkow: Typo

ts/adaptors/lite/Parser.ts
```diff
@@ -189,7 +189,7 @@ export class LiteParser implements MinDOMParser<LiteDocument> {
         //
         // Get the child to be added to the node
         //
-        const kind = tag.match(/<(.*?)[\s\n>]/)[1].toLowerCase();
+        const kind = tag.match(/<(.*?)[\s\n>/]/)[1].toLowerCase();
```
zorkow: Should that not be 
```suggestion
        const kind = tag.match(/<(.*?)[\s\n/>]/)[1].toLowerCase();
```
?

dpvc: The order of the characters in the `[...]` doesn't matter.  It just matches up to any one of the enclosed characters.  So it is correct as is.

zorkow: Right. I now realise that it is only any of the four characters.

PR #372, opened 2019-10-20
Add support for serializing annotation-xml contents

dpvc:
This PR modifies the `XMLnode` object (which holds the XML content of an `annotation-xml` element) so that its contents can be serialized.  To do this, we store the adaptor from the input jax in the XML node so that it can be used to serialize the xml content when needed.  The `SerializeMmlVisitor` uses the new function of the `XMLnode` to serialize it.  The MathML input jax is modified to pass the adaptor along, but it looks like more because I made a local `adaptor` variable to simplify `this.adaptor` calls.

### DIFF
--- a/ts/input/mathml/MathMLCompile.ts
+++ b/ts/input/mathml/MathMLCompile.ts
@@ -174,18 +174,19 @@ export class MathMLCompile<N, T, D> {
         if (mml.arity === 0) {
             return;
         }
-        for (const child of this.adaptor.childNodes((node)) as N[]) {
-            const name = this.adaptor.kind(child);
+        const adaptor = this.adaptor;
+        for (const child of adaptor.childNodes((node)) as N[]) {
+            const name = adaptor.kind(child);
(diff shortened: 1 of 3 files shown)


ts/input/mathml/MathMLCompile.ts
```diff
@@ -174,18 +174,19 @@ export class MathMLCompile<N, T, D> {
         if (mml.arity === 0) {
             return;
         }
-        for (const child of this.adaptor.childNodes((node)) as N[]) {
-            const name = this.adaptor.kind(child);
+        const adaptor = this.adaptor;
+        for (const child of adaptor.childNodes((node)) as N[]) {
```
zorkow: Is there a reason for the double parenthesised `((node))`?

dpvc: I think they are probably left over from some older `as N` cast or some such thing.  I'll remove them.

PR #374, opened 2019-10-21
Improve movablelimits support

dpvc:
This PR adds a new post-filter to the TeX input jax that converts in-line `munderover`, `munder`, and `mover` elements with the `movablelimits` property that have a base that is not an `mo` (e.g., `\mathop{xyz}` into an explicit `msubsup`, `msub`, or `msup` element.  This means the internal MathML will be correct MathML (rather than relying on the TeX-supplied `movablelimits` property), so the output jax don't have to look for that property any more and can just go by the attribute on the core mo.  It also means the `SerializeMmlVisitor` won't have to be modified to take the property into account (it currently outputs incorrect MathML for these situations).

### DIFF
--- a/ts/input/tex/FilterUtil.ts
+++ b/ts/input/tex/FilterUtil.ts
@@ -235,18 +235,18 @@ namespace FilterUtil {
    */
   let _moveLimits = function (options: ParseOptions, underover: string, subsup: string) {
     for (const mml of options.getList(underover)) {
-      const display = mml.attributes.get('displaystyle');
-      if (!display) {
-        const base = mml.childNodes[(mml as any).base] as MmlNode;
-        const mo = base.coreMO();
-        if (base.getProperty('movablelimits') && !mo.attributes.getExplicit('movablelimits')) {
-          let node = options.nodeFactory.create('node', subsup, mml.childNodes);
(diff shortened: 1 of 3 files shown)


ts/input/tex/FilterUtil.ts
```diff
@@ -225,6 +225,47 @@ namespace FilterUtil {
...
+   * movablelimits and bases that are not mo's, and creates new msubsup
+   * elements to replace them if they aren't in displaystyle.
+   *
+   * @param {MmlNode} ath The node to rewrite.
+   * @param {ParseOptions} data The parse options.
+   */
+  let _moveLimits = function (options: ParseOptions, underover: string, subsup: string) {
+    for (const mml of options.getList(underover)) {
+      const display = mml.attributes.get('displaystyle');
+      if (!display) {
```
zorkow: Can we have a 
```suggestion
      if (display) {
         continue;
      } 
```
here? It is already very nested.

dpvc:
> I assume that depends on running `setInherited` first.

Right.  That is why I added a separate filter.

ts/input/tex/FilterUtil.ts
dpvc: OK, sure.

PR #378, opened 2019-10-30
Allow applying multiple CSS classes via parser.create method

gcalmettes:
Allow addition of multiple classes to nodes when multiple classes are provided (DOMTokenList or array of classes) with the `parser.create` method.

See [this issue](https://github.com/mathjax/MathJax/issues/2230)

Previously, this would fail:

```
const def = { class: 'class1 class2' } // or const def = { class: ['class1', 'class2'] }
const math = parser.ParseArg(name)
const node = parser.create('node', 'mrow', [math], def)
```

This PR allows for

```
const def = { class: 'class1 class2' } // or const def = { class: ['class1', 'class2'] }
const math = parser.ParseArg(name)
const node = parser.create('node', 'mrow', [math], def)
```

### DIFF
--- a/ts/adaptors/HTMLAdaptor.ts
+++ b/ts/adaptors/HTMLAdaptor.ts
@@ -419,8 +419,11 @@ extends AbstractDOMAdaptor<N, T, D> implements MinHTMLAdaptor<N, T, D> {
     /**
      * @override
      */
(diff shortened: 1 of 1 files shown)


gcalmettes:
closing this as it is a duplicate of #340

PR #388, opened 2019-11-11
Issue 2239

zorkow:
Solves https://github.com/mathjax/MathJax/issues/2239

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -67,10 +67,25 @@ namespace ParseUtil {
    */
   export function matchDimen(
     dim: string, rest: boolean = false): [string, string, number] {
-    let match = dim.match(rest ? dimenRest : dimenEnd);
-    return match ? [match[1].replace(/,/, '.'), match[4], match[0].length] :
-      [null, null, 0];
+      let match = dim.match(rest ? dimenRest : dimenEnd);
+      return match ?
+        muReplace([match[1].replace(/,/, '.'), match[4], match[0].length]) :
+        [null, null, 0];
   }
+
+
+  /**
+   * Transforms mu dimension to em if necessary.
+   * @param {[string, string, number]} [value, unit, length] The dimension triple.
+   * @return {[string, string, number]} [value, unit, length] The transformed triple.
+   */
+  function muReplace([value, unit, length]: [string, string, number]): [string, string, number] {
+    if (unit !== 'mu') {
+      return [value, unit, length];
+    }
+    let em = Em(UNIT_CASES[unit](parseFloat(value || '1')));
+    return [em.slice(0, -2), 'em', em.length];
+  };
 
 
   /**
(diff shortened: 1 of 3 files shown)


dpvc:
I've been thinking about this problem more since we talked.  In TeX, the `mu` is meant to be relative to the size of the material being typeset (so in `\scriptstyle`, the `mu` is smaller than in `\textstyle`).  All other measurements in TeX are unaffected by the style (including `em` and `ex`).  In version 2, the HTML-CSS and SVG output jax went to great pains to mirror this behavior; but since the output jax doesn't know where the math came from, this means it also affected MathML input in the same way, so that `em`s in superscripts were the same size as `em`s in the main text, which is counter to how MathML should work.  In v2 CommonHTML and in both output formats in v3, the font-relative units (`em` and `ex`) are interpreted as relative to the size, as they should be in MathML, and not as in TeX.  It is unusual to find units of `em` or `ex` in actual TeX/LaTeX, so this inconsistency with TeX is unlikely to be a problem, and it makes creating macros that work in both text and superscripts easier (without having to use `\mathchoice`).

I had originally thought that the conversion from `mu` to `em would need to take the `scriptlevel` into account, since `mu` are relative to that, but since the `em` are already relative to the script level, there is no need (which is why your code doesn't use the script level in its conversion).  

Since the script level is not used, that means that the conversion can be done when the dimension is first created, rather than waiting for the post filters.  That is, you could do the `mu` to `em` conversion in `TexParser.getDimension()`:

https://github.com/mathjax/MathJax-src/blob/1e1a30f0a89e52fa7764dbb510763145a4da49f1/ts/input/tex/TexParser.ts#L385-L408

probably at lines 394 and 402 (to check for `mu` and do the conversion).  This would mean you don't have to scan every attribute for `mu`s, and only need to check the dimensions actually entered by the user.  There is one other line that would need to be changed:

https://github.com/mathjax/MathJax-src/blob/1e1a30f0a89e52fa7764dbb510763145a4da49f1/ts/input/tex/ams_cd/AmsCdMethods.ts#L121

that uses `mu` in attributes, which could be changed to `em` directly.

Because `em` are relative to script level now, I think this is the better way to go.  Do you agree?

zorkow:
I agree, converting the dimension immediately will make life easier and the code overall cleaner. 
When I implemented this I also thought that the methods in `ParseUtil` should really be combined/replaced with those in `Util`.  But that's work for a different PR.

zorkow:
Incorporates the discussion. PTAL.

ts/input/tex/ParseUtil.ts
```diff
@@ -67,12 +67,27 @@ namespace ParseUtil {
...
+   * Transforms mu dimension to em if necessary.
+   * @param {[string, string, number]} [value, unit, length] The dimension triple.
+   * @return {[string, string, number]} [value, unit, length] The transformed triple.
+   */
+  function muReplace([value, unit, length]: [string, string, number]): [string, string, number] {
+    if (unit !== 'mu') {
+      return [value, unit, length];
+    }
+    let em = Em(UNIT_CASES[unit](parseFloat(value || '1')));
+    return [em.slice(0, -2), 'em', em.length];
```
dpvc: Should `em.length` be `length`, the length of the original match?  The transformed string's length has nothing to do with the original length, and if the length is supposed to indicate a position in the original string, the original `length` is the one that tells you that.

zorkow: Good catch. Changed.

PR #392, opened 2019-11-18
Remove mi2mo action

dpvc:
Remove the `mi2mo()` action for `munderover` and related constructs (since a TeX input post-filter now handles non-`mo` elements that have movable limits).  In addition, it fixes a problem where an explicitly set `texClass` could be lost if `lspace` or `rspace` was set (this was the case with `mi2mo()`, but now that that is removed).  This was a cause for some test failures from mathjax/MathJax-dev#30.  But these changes will require additional update to those tests, since the `mi` are no longer changed to `mo`.

Finally, it also adds missing jsdoc comments to one function.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -196,7 +196,7 @@ namespace ParseUtil {
    * empty braces (TeXAtom with child being an empty inferred row),
    * is an <mo>, precede it by an empty <mi> to force the <mo> to
    * be infix.
-   * @param {ParseOptions} configuration The current parse otpions.
+   * @param {ParseOptions} configuration The current parse options.
    * @param {MmlNodep[]} nodes The row of nodes to scan for an initial <mo>
    */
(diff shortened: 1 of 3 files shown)


ts/input/tex/ParseUtil.ts
```diff
@@ -192,10 +192,12 @@ namespace ParseUtil {
...
   /**
-   *  If the initial child, skipping any initial space or
-   *  empty braces (TeXAtom with child being an empty inferred row),
-   *  is an <mo>, preceed it by an empty <mi> to force the <mo> to
-   *  be infix.
+   * If the initial child, skipping any initial space or
+   * empty braces (TeXAtom with child being an empty inferred row),
+   * is an <mo>, precede it by an empty <mi> to force the <mo> to
+   * be infix.
+   * @param {ParseOptions} configuration The current parse otpions.
```
zorkow: typo.

dpvc: Fixed.

PR #393, opened 2019-11-18
Add more data-mjx-* attributes to MathML serialization 

dpvc:
This PR exposes more internal properties via `data-mjx-*` attributes, including the `variantForm` property (mathjax/MathJax#2237) and the internal `mathvariant` values, like `-tex-calligraphic` (mathjax/MathJax#2238).  This also moves the data for TeXAtom nodes to `data-mjx-texclass` rather than using the `class` attribute for these.  Also, the xmlns attribute is added to math elements, if needed.

The MathML input jax is modified to read these and transfer the data back to the internal elements it creates.  It still looks for the corresponding values in the `class`, however, for backward compatibility with v2 output.  It also corrects the `caligraphic` misspelling.

Resolves issues mathjax/MathJax#2237, mathjax/MathJax#2238, and mathjax/MathJax#2214.
Also resolves issue from mathjax/MathJax#2264

### DIFF
--- a/ts/core/MmlTree/SerializedMmlVisitor.ts
+++ b/ts/core/MmlTree/SerializedMmlVisitor.ts
@@ -98,9 +98,8 @@ export class SerializedMmlVisitor extends MmlVisitor {
      * @return {string}       The serialized contents of the mrow, properly indented.
      */
     public visitTeXAtomNode(node: MmlNode, space: string) {
-        let texclass = node.texClass < 0 ? 'NONE' : TEXCLASSNAMES[node.texClass];
         let children = this.childNodeMml(node, space + '  ', '\n');
-        let mml = space + '<mrow' + this.getAttributes(node) + ' data-mjx-texclass="' + texclass + '">' +
+        let mml = space + '<mrow' + this.getAttributes(node) + '>' +
             (children.match(/\S/) ? '\n' + children + space : '') + '</mrow>';
         return mml;
     }
@@ -181,6 +180,9 @@ export class SerializedMmlVisitor extends MmlVisitor {
         const variants = (this.constructor as typeof SerializedMmlVisitor).variants;
         variant && variants.hasOwnProperty(variant) && data.push(' data-mjx-variant="' + variant + '"');
         node.getProperty('variantForm') && data.push(' data-mjx-alternate="1"');
+        const texclass = node.getProperty('texClass') as number;
+        texclass !== undefined &&
+            data.push(' data-mjx-texclass="' + (texclass < 0 ? 'NONE' : TEXCLASSNAMES[texclass] + '"');
         return data.join('');
     }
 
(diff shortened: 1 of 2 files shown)


ts/core/MmlTree/SerializedMmlVisitor.ts
```diff
@@ -147,13 +155,42 @@ export class SerializedMmlVisitor extends MmlVisitor {
...
+     * Create the list of data-mjx-* attributes
+     *
+     * @param {MmlNode} node        The node whose data list is to be generated
+     * @return {string}             The final class attribute (or empty string)
+     */
+    protected getDataAttributes(node: MmlNode) {
+        const data = [];
+        const variant = node.attributes.getExplicit('mathvariant') as string;
+        const variants = (this.constructor as typeof SerializedMmlVisitor).variants;
+        variant && variants.hasOwnProperty(variant) && data.push(' data-mjx-variant="' + variant + '"');
```
zorkow: This should go into the aux method.

ts/core/MmlTree/SerializedMmlVisitor.ts
```diff
@@ -147,13 +155,42 @@ export class SerializedMmlVisitor extends MmlVisitor {
...
+        node.getProperty('variantForm') && data.push(' data-mjx-alternate="1"');
```
zorkow: ditto.

ts/core/MmlTree/SerializedMmlVisitor.ts
```diff
@@ -147,13 +155,42 @@ export class SerializedMmlVisitor extends MmlVisitor {
...
+        const texclass = node.getProperty('texClass') as number;
+        if (texclass !== undefined) {
+            let setclass = true;
+            if (texclass === TEXCLASS.OP && node.isKind('mi')) {
+                const name = (node as MmlMi).getText();
+                setclass = !(name.length > 1 && name.match(MmlMi.operatorName));
+            }
+            setclass && data.push(' data-mjx-texclass="' + (texclass < 0 ? 'NONE' : TEXCLASSNAMES[texclass] + '"'));
```
zorkow: ditto.

dpvc:
OK, I think I've done what you asked.  But I also reorganized things a bit to make creation of the attributes strings more consistent.  I also added the missing `xmlns` attribute on the `math` element that was reported in mathjax/MathJax#2264, and some of the changes were to make that easier.

dpvc:
On second thought, maybe the `xmlns` attribute should be added when the `math` element is *created* rather than in the output here?

zorkow:
> On second thought, maybe the `xmlns` attribute should be added when the `math` element is _created_ rather than in the output here?

If I understand correctly that would mean we make our main data structure dependent on a namespace. I do not think that's a good idea.

dpvc:
> If I understand correctly that would mean we make our main data structure dependent on a namespace. I do not think that's a good idea.

Not really dependent on it, but it would create that attribute automatically when the `math` node is created.  I'm happy to leave it in the serializer.

Issue #385, opened 2019-11-08
Output of invisible characters


zorkow:
Edge (and to some extent IE) displays invisible unicode characters: invisible times, invisible separator, function application. Remove them from CSS so they are simply not "displayed" in any browser.

dpvc:
It looks like the Cambria font has visible glyphs for these positions (which seems like a bad idea to me).  I'm looking into not including them in the `content` CSS properties so they won't show up at all.

PR #397, opened 2019-11-18
Don't use CSS content for invisible characters.  #385

dpvc:
This PR prevents invisible characters from being inserted via the `content` CSS rules.  It also corrects the range of invisible characters (from U+2061 through U+2064 to U+2060 through U+2063).

Resolves issue #385


ts/output/chtml/fonts/tex/normal.ts
```diff
@@ -153,10 +153,10 @@ export const normal: CHTMLCharMap = AddCSS(font, {
     0x203E: {c: '\\2C9'},
     0x2044: {c: '/'},
     0x2057: {c: '\\2032\\2032\\2032\\2032'},
+    0x2060: {c: ''},
     0x2061: {c: ''},
     0x2062: {c: ''},
     0x2063: {c: ''},
-    0x2064: {c: ''},
```
zorkow: What's the reason for removing invisible plus?

ts/output/common/fonts/tex/normal.ts
```diff
@@ -215,10 +215,10 @@ export const normal: CharMap<CharOptions> = {
     0x203E: [.59, -0.544, .5],
     0x2044: [.75, .25, .5],
     0x2057: [.56, 0, 1.1],
+    0x2060: [0, 0, 0],
     0x2061: [0, 0, 0],
     0x2062: [0, 0, 0],
     0x2063: [0, 0, 0],
-    0x2064: [0, 0, 0],
```
zorkow: ditto.

ts/output/chtml/fonts/tex/normal.ts
dpvc: Sorry, I misread my Unicode chart.  I'll put it back.

PR #401, opened 2019-11-23
Don't leak fonts and other properties to initial array entry.  mathjax/MathJax#2247

dpvc:
The local environment object should not be copied when pushing an array element on the stack.  This prevents fonts and other properties from being incorrectly copied into the first cell of the array.  E.g.

```
\bf \begin{array}{c} a & a\end{array}
```

should show two italic a's, not a bold one and an italic one.

 Resolves issue mathjax/MathJax#2247.

### DIFF
--- a/ts/input/tex/StackItem.ts
+++ b/ts/input/tex/StackItem.ts
@@ -254,7 +254,7 @@ export interface StackItem extends NodeStack {
 
   /**
    * Copy local properties when pushed to stack?
-   * @type {boolena}
+   * @type {boolean}
    */
    copyEnv: boolean;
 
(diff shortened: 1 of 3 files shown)


dpvc:
PS, this implements v2 code that was not included in the v3 version of the `StackItem` class and `ArrayItem`.

ts/input/tex/StackItem.ts
```diff
@@ -252,6 +252,12 @@ export interface StackItem extends NodeStack {
    */
    env: EnvList;
 
+  /**
+   * Copy local properties when pushed to stack?
+   * @type {boolena}
```
zorkow: typo

PR #410, opened 2019-12-18
Add AssistiveMML support to v3

dpvc:
This PR implements an assistive-mml extension that corresponds to the version 2 AssistiveMML extension.  It is implemented as an a11y extension similar to how `semantic-enrich` and the other ones operate (via extensions to the MathDocument and MathItem classes to add methods for creating the hidden MathML).  Because the extension needed to be able to generate CSS styles to hide the MathML, this required extending the HTMLDocument class to allow extensions to add CSS to the document.

This also adds a menu item to to the accessibility menu to control whether assistive MathML is inserted into the page or not (default is yes), so that those who don't need it will be able to turn it off for faster rendering.

It would probably be a good idea to move the output/common/CssStyles.ts file to utilities at some point, but I didn't do that here since it touches a number of files.

Resolves issue mathjax/MathJax#2260

### DIFF
--- a/components/src/a11y/assistive-mml/assistive-mml.js
+++ b/components/src/a11y/assistive-mml/assistive-mml.js
@@ -1,0 +1,7 @@
+import './lib/assistive-mml.js';
+
+import {AssistiveMmlHandler} from '../../../../js/a11y/assistive-mml.js';
+
+if (MathJax.startup) {
+    MathJax.startup.extendHandler(handler => AssistiveMmlHandler(handler));
+}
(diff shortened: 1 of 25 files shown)


zorkow:
OK, the webpack was probably my fault for not explicitly typescripting before building. 
Nevertheless, the typo in `preload.js` should be fixed. That has no influence on the build.
But what's even more odd is that it does not seem to have any influence on the behaviour.

dpvc:
> the typo in `preload.js` should be fixed.

Indeed.

> But what's even more odd is that it does not seem to have any influence on the behaviour.

That string is used to tell MathJax that `a11y/assisitve-mml` was preloaded (even though there is no such thing).  That is only to prevent MathJax from trying to load it dynamically.  So because of the misspelling, MathJax would simply reload the (correctly spelled) `a11y/assistive-mml` when it is requested, so there is no harm to the functionality, but an extra file would be loaded that didn't need to be.

dpvc:
OK, I've made the update.

PR #413, opened 2019-12-26
Skip TeX comments correctly.  mathjax/MathJax#2271.

dpvc:
Update `FindTeX` to properly skip comments, even if they include unbalanced braces.

Resolves issue mathjax/MathJax#2271

### DIFF
--- a/ts/input/tex/FindTeX.ts
+++ b/ts/input/tex/FindTeX.ts
@@ -154,7 +154,7 @@ export class FindTeX<N, T, D> extends AbstractFindMath<N, T, D> {
      * @return {RegExp}     The regular expression for the end delimiter
      */
     protected endPattern(end: string) {
-        return new RegExp(quotePattern(end) + '|\\\\(?:[a-zA-Z]|.)|[{}]', 'g');
+        return new RegExp(quotePattern(end) + '|\\\\(?:[a-zA-Z]|.)|[{}%]', 'g');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


zorkow (approved):
lgtm.

dpvc:
This is potentially a breaking change, so I am not going to include it in this version.  The problem is that

```
... $$ x + y % comment $$ ...
```

would be processed by the current version (and version 2), but would not be processed by this PR, since the second `$$` is now part of a comment.  Although the behavior of this PR is the correct one for LaTeX, it might break existing content that takes advantage of this error in processing.

dpvc:
There should also be additional changes to actually remove all comments before processing the rest of the expression, as currently that is not being done.  For example, 

```
\def\xyz#1%#2{[#1][#2]}
```
is valid, and can be used via `\xyz a%b` to get `[a][b]` as output.  This is not how actual LaTeX works, and so this should be fixed as well.  But I know that this is used on the Puzzling StackExchange site in some MathJax-based puzzles, and it will break those.

So I'm wondering if it is better to leave it as is, or break those incorrect usages?

ts/input/tex/FindTeX.ts
```diff
@@ -180,6 +180,8 @@ export class FindTeX<N, T, D> extends AbstractFindMath<N, T, D> {
                 braces++;
             } else if (match[0] === '}' && braces) {
                 braces--;
+            } else if (match[0].substr(0) === '%') {
```
zorkow, line 183: `substr` is deprecated
```suggestion
            } else if (match[0].substring(0) === '%') {
```

zorkow (changes requested):
Replace `substr`.

PR #417, opened 2019-12-31
Fix duplicate xlink namespace in SVG output.  #365

dpvc:
This PR manages the `xlink` namespace in all adaptors so that the namespace appears in the adaptor's `outerHTML` and `innerHTML`, and also in the browser's `outerHTML` and `innerHTML`.  It also works with XMLSerializer, though the `<use>` elements will have `xmlns:xlink` attributes that are probably redundant since that is on the outer `<svg>` element.

Resolves issue #365, and replaces PR #366.

### DIFF
--- a/ts/adaptors/HTMLAdaptor.ts
+++ b/ts/adaptors/HTMLAdaptor.ts
@@ -381,7 +381,11 @@ extends AbstractDOMAdaptor<N, T, D> implements MinHTMLAdaptor<N, T, D> {
      * @override
      */
     public setAttribute(node: N, name: string, value: string, ns: string = null) {
-        return (ns ? node.setAttributeNS(ns, name, value) : node.setAttribute(name, value));
+        if (!ns) {
+            return node.setAttribute(name, value);
+        }
+        name = ns.replace(/.*\//, '') + ':' + name.replace(/^.*:/, '');
+        return node.setAttributeNS(ns, name, value);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ts/adaptors/HTMLAdaptor.ts
```diff
@@ -381,7 +381,11 @@ extends AbstractDOMAdaptor<N, T, D> implements MinHTMLAdaptor<N, T, D> {
      * @override
      */
     public setAttribute(node: N, name: string, value: string, ns: string = null) {
-        return (ns ? node.setAttributeNS(ns, name, value) : node.setAttribute(name, value));
+        if (!ns) {
+            return node.setAttribute(name, value);
+        }
+        name = ns.replace(/.*\//, '') + ':' + name.replace(/^.*:/, '');
+        return node.setAttributeNS(ns, name, value);
```
zorkow: I am a bit confused: when there is a namespace, we add it to the name, but then still provide the namespace explicitly?

dpvc: Yes, that is correct.  Because HTML5 removes namespaces, if you take `outerHTML` for an svg element in the browser, the `xlink:href` references will be listed as just `href` (no `xlink` regardless of whether the namespace.  This means that svg's that people save for use as a stand-alone will not load properly.  Doing it this way preserves the name correctly in `outerHTML` output.

zorkow: OK, thanks for the explanation.

PR #424, opened 2020-01-10
Update font data in several ways.  mathjax/MathJax#2286

dpvc:
This PR fixes several problems with the font data used in both CHTML and SVG.  

* Missing italic correction values are added (there was a cut-off for what was considered "significant" that was too high).

* Remove initial 0 from skew values (not really a bug, but reduces the file size).

* Correct delimiter HDW (height-depth-width) values (the wrong font was used to get these values originally).

There's really not much to see here, as it is just changes to the font data.

Resolves issue mathjax/MathJax#2286 (and more).


zorkow:
I can't see a difference, neither in CHTML nor SVG. I still looks too close for me, at least for the \partial^2 example.

dpvc:
> I can't see a difference

This is what I see without the patch:

develop

and this is with it:

patched

Can you verify that you are not seeing a difference with this branch?

zorkow:
Yes and no: I can not see it on Firefox under Linux.
It works on Windows 10: Firefox, Chrome, Edge
Linux: Chrome, Opera.

So I would say: It works.

dpvc:
Peculiar about Linux Firefox.  Something to look into in the future.

Issue #420, opened 2020-01-05
ESM Distribution


brettz9:
I'm looking to upgrade our MathJax 2.0 into svgedit. Our source is also ESM based (using Rollup), with a working ESM-in-the-browser version for easier live debugging as well as a compiled version.

While I was able to get a basic ESM import of MathJax working (after making a [fix to a TS file](https://github.com/mathjax/MathJax-src/issues/414#issuecomment-570899428)), the module that was returned had some properties like `handlers` but was bare of the `startup` property that I understand it is supposed to have. (And without specific guidance in the docs about startup for ESM, I wasn't sure what may be the issue there.)

When I tried importing from the components source (since the components, such as `tex-mml-svg` that we were using previously, wouldn't seem to be able to give ESM output if using webpack as it doesn't [natively support ESM output](https://github.com/webpack/webpack/issues/2933)), I got an error because two files are trying to import `components/src/dependencies.js` which is a CJS file.

I am guessing that these files should really be in ESM format, and that perhaps webpack may overlook this inconsistency, but Rollup will not (unless using a CJS plugin, and I wouldn't think going that route should really be necessary since you've already gone through the trouble of converting most of source source with resolvable paths anyways, and I'd expect webpack should still work in *processing* ESM as it does the rest of your codebase).

Anyways, I haven't dug in too deeply, but just wondering if I can report these as bugs (or better yet, if I can ask for an already-built ESM distribution for mathjax-full), or if you are settled on the formats of the files as is.

dpvc:
Can you tell me a little bit more about how you are using and importing MathJax in you project (as there are several ways that can be done)?   The `startup` module is very closely tied to MathJax's dynamic loading mechanism, and if you are trying to package MathJax into a single-file build, you may not want the dynamic loading.  The `components/src/dependencies.js` provides information about that loading process, and would not be needed if you are going to preload everything that is needed.  For example, you could probably take `components/src/tex-mml-svg/tex-mml-svg.js` as a template and replace the final `import '../startup/startup.js';` with just a portion of that file that doesn't do any loading.  Something like

```
import '../startup/lib/startup.js';
import './preload.js';
import '../core/core.js';
import '../input/tex/tex.js';
import '../input/mml/mml.js';
import '../output/svg/svg.js';
import '../output/svg/fonts/tex/tex.js';
import '../ui/menu/menu.js';
import {CONFIG} from '../../../js/components/loader.js';
CONFIG.ready();
```

[...]

PR #429, opened 2020-01-21
Make source.js and dependencies.js into ESM modules.  #420

dpvc:
Convert `source.js` and `dependencies.js` to ESM modules.  This is a little cleaner, and still works with web pack, but also should allow it to work with rollup (see #420).

Resolves (part of) issue #420.

### DIFF
--- a/components/src/dependencies.js
+++ b/components/src/dependencies.js
@@ -1,7 +1,4 @@
-"use strict";
-Object.defineProperty(exports, '__esModule', {value: true});
-
-exports.dependencies = {
+export const dependencies = {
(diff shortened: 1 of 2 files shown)


zorkow:
On second thought... I worked fine in my test page, but I now see a `__dirname is not defined` in the lab. Is there anything that needs to be changed?

dpvc:
> Is there anything that needs to be changed?

Yes, I thought I had mentioned it somewhere, but can't find it, so I guess I forgot.  You need to add

```
<script>
  __dirname = new URL('components/src', location).href;
</script>
```

just before the `<script>` with the `System.import()` in it.

zorkow:
Ah, thanks for the info.

PR #431, opened 2020-01-24
Don't include extenders in SVG output if they have negative width. mathjax/MathJax#2300

dpvc:
This PR fixes a problem with the SVG output that is exhibited in the Chrome browser for the example

    \overleftrightarrow{abc}

which produces errors in the console about a negative width, and displays the dash that is supposed to form the extender between the two arrows in the wrong location.

Resolves issue mathjax/MathJax#2300

### DIFF
--- a/ts/output/svg/Wrappers/mo.ts
+++ b/ts/output/svg/Wrappers/mo.ts
@@ -178,6 +178,7 @@ export class SVGmo<N, T, D> extends CommonMoMixin<SVGConstructor<any, any, any>>
                                                  //   (glyphs with rounded or anti-aliased ends don't stretch well,
                                                  //    so this makes for sharper ends)
         const y = (s * (h - d) - Y) / 2;         // The bottom point to clip the extender
+        if (Y <= 0) return;
         const svg = this.svg('svg', {
             width: this.fixed(w), height: this.fixed(Y),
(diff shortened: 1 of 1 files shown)


zorkow:
I still get two negative value errors; for attribute `width` and `viewBox`.

dpvc:
@zorkow, I just checked again and with the patch I get no error but without I do.  Can you check again that you have run typescript after checking out the branch?

zorkow:
I can still see it on all my Linux browsers + Chrome on Win10. Edge and Firefox on Win10 work fine.

dpvc:
It works fine for me in Windows10 Chrome.  Here it is with this patch:

Chomr-W10-OK

Note that there are only two `<use>` elements within the selected group.

Here it is **without** the patch:

Crhome-W10-bad

Note that there is an additional `<svg>` element within this group.  It is the source of the errors.

Can you check the output if your version and see which it looks like?

dpvc:
I just tried it on Ubuntu 19.10 (fresh install), and it works for me.  But it does appear that you have to explicitly clear the cache when you change the branch.  (I also restarted Chrome, but am not sure that is necessary.  It was for a time on the Mac, and I didn't go back to see if it was for Ubuntu).

dpvc:
So I think it is OK, and I'm login to merge it.

PR #445, opened 2020-02-26
Use em rather than mu units for extensible arrows.  (mathjax/MathJax#2357)

dpvc:
The `mu` units were not being translated to `em`, and so were being left verbatim in the attributes of the internal MathML.  These would then be used verbatim in the assistive MathML, causing errors in the console.

Resolves issue mathjax/MathJax#2357.

### DIFF
--- a/ts/input/tex/ams/AmsMethods.ts
+++ b/ts/input/tex/ams/AmsMethods.ts
@@ -46,7 +46,7 @@ let AmsMethods: Record<string, ParseMethod> = {};
 
 
 /**
- * Handle AMS array environments. 
+ * Handle AMS array environments.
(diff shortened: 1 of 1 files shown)


ts/input/tex/ams/AmsMethods.ts
```diff
@@ -46,7 +46,7 @@ let AmsMethods: Record<string, ParseMethod> = {};
 
 
 /**
- * Handle AMS array environments. 
+ * Handle AMS array environments.
```
zorkow: That's been in the previous PR already?

dpvc: It may be.  It should not cause a problem, however.

PR #449, opened 2020-03-04
Add ability to include MJXZERO to font-family used for undefined characters

dpvc:
This PR fixes a placement issue for CHTML output when undefined characters are used.  (Unfortunately, it is an old branch that I forgot to make a PR for back in October, and I don't remember the exact situation that caused the problem.)

### DIFF
--- a/ts/output/chtml/fonts/tex.ts
+++ b/ts/output/chtml/fonts/tex.ts
@@ -60,6 +60,8 @@ import {delimiters} from '../../common/fonts/tex/delimiters.js';
 export class TeXFont extends
 CommonTeXFontMixin<CHTMLCharOptions, CHTMLVariantData, CHTMLDelimiterData, CHTMLFontDataClass>(CHTMLFontData) {
 
+    protected static defaultCssFamilyPrefix = 'MJXZERO, ';
+
     /**
      * The classes to use for each variant
      */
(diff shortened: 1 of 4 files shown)


ts/output/svg.ts
```diff
@@ -326,7 +326,7 @@ CommonOutputJax<N, T, D, SVGWrapper<N, T, D>, SVGWrapperFactory<N, T, D>, SVGFon
         const adaptor = this.adaptor;
         if (variant !== '-explicitFont') {
             const [family, italic, bold] = this.font.getCssFont(variant);
-            adaptor.setAttribute(svg, 'font-family', family);
+            adaptor.setAttribute(svg, 'font-family', this.font.cssFamilyPrefix + family);
```
zorkow: Put `,` concatenation here.

ts/output/chtml/fonts/tex.ts
```diff
@@ -60,6 +60,8 @@ import {delimiters} from '../../common/fonts/tex/delimiters.js';
 export class TeXFont extends
 CommonTeXFontMixin<CHTMLCharOptions, CHTMLVariantData, CHTMLDelimiterData, CHTMLFontDataClass>(CHTMLFontData) {
 
+    protected static defaultCssFamilyPrefix = 'MJXZERO, ';
```
zorkow: I'm not a big fan of putting the concatenation in the string. That presumes we always know that it will only ever be used with a some postfix.
Just in case the string is inadvertently used on its own somewhere, could that lead to a CSS error?

dpvc: I don't think so, but I've changed the usage as you requested.

PR #450, opened 2020-03-04
Fix problem with infinite loop with SRE component in IE.

dpvc:
This resolves the issue in IE where you get a stack overflow when starting up the explorer (or other tools that rely on SRE).  The issue was that the Promise polyfill would cause `combineConfig()` to loop infinitely (it apparently has a recursive reference in it).

This also lengthens the timeout from 10 to 20 seconds, since IE is slow enough to sometimes need it.


zorkow:
Lgtm

zorkow:
Not sure how I can approve the PR in this app.

dpvc:
No problem, I merged without explicit approval.

PR #453, opened 2020-03-05
Add tagFormat to AllPackes, and update the preloaded list.  (mathjax/MathJax#2328)

dpvc:
This PR adds `tagFormat` to the AllPackages file, and updates the preloaded list to include it (and and `color_v2`, which was loaded but missing from the list.  Also, it adds `bussproofs` to the package list.

Resolves issue mathjax/MathJax#2328.


dpvc:
OK, I'll take care of it.  Sorry for the inconsistency.

dpvc:
OK, I've made the change.  Can you check that this is what you wanted?

PR #455, opened 2020-03-16
Miscellaneous font updates

dpvc:
This PR updates the font handling to resolve a number of issues, including:

* Map `mathvariant` values to their Math Alphanumeric equivalents, when possible.
* Don't substitute characters for variants that we don't have (e.g., bold-script, sans-serif-bold-italic).
* Don't substitute italic Greek for upright Greek (e.g., in `mathvariant="normal"`), but rely on Math Alphanumeric range in system fonts instead.
* Properly handle Plane1 unknown characters (mathjax/MathJax#2369).
* Change font handling to give fonts explicitly (through a class) rather than a list of fonts to check through (making it easier and more reliable to modify the font data for customization).
* Include all CSS rules when `adaptiveCSS` is false (mathjax/MathJax#2378).
* Fix incorrect internal name for `sans-serif-bold-italics`.
* Make script variants default to italics rather than upright, when possible.
* Remove redundancy in the font data files.
* Fix a number of mis-mapped characters and substitutions.
* Fix problems with italics being taken from the wrong font for some glyphs.
* Fix some tslint issues (not all of them, though).

Most of the files here are the font data files, and there isn't much point in looking through them (they are generated programmatically, and I've checked them visually using a font-table generator).


ts/output/common/Wrapper.ts
```diff
@@ -701,11 +700,20 @@ export class CommonWrapper<
...
-    protected unicodeChars(text: string) {
-        return unicodeChars(text);
+    protected unicodeChars(text: string, variant: string = '') {
+        let chars = unicodeChars(text);
+        //
+        //  Remap to Math Alphanumerics block
+        //
+        const map = this.font.getVariant(variant).chars;
+        if (map) {
+            chars = chars.map((n) => ((map[n] || {})[3] || {}).smp || n);
```
zorkow, line 714: This line took me a while to work out. Maybe you could add a comment.

dpvc: Done

dpvc: Done.

PR #464, opened 2020-03-26
A11y locale option

zorkow:
* Integrates new SRE with dynamic locale loading. This will only work with SRE v3.0.0-beta.10 and above.
* Moves type declaration for SRE from `semantic-enrich.ts` into the `sre.ts` module.


ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -25,7 +25,7 @@
 
 import {A11yDocument, Region} from './Region.js';
 import {Explorer, AbstractExplorer} from './Explorer.js';
-import {sreReady} from '../sre.js';
+import {sreReady, SRE} from '../sre.js';
```
dpvc, line 28: I was a little confused by this, as `SRE` is a type in `sre.js`, but is being used as an object below.  Typescript should complain about that.

It turns out that `sre-node.ts` is the culprit.  Because it doesn't include an `import` or `export` command, it is not compiled as a module (in strict mode), so the `const SRE = require(...)` that it includes ends up being a global definition, not a local one, and the `sre-node.d.ts` ends up including a `declare const SRE` command.  Since Typescript processes all the `.d.ts` files (whether the `.js` file is included or not), that makes `SRE` a global even when `sre-node.js` isn't used.  That is why the use of `SRE` as a global is OK below.

This really needs to be fixed.  That could be done by adding

    export {};

to the bottom of `sre-node.ts` to force it to be treated as a module.  Then you could change

    export type SRE = {

below to

    export declare const SRE: {

in `sre.ts`.  That will make this `import` command import that `SRE` object, not the type, and all will be well, without the need for `sre-node.ts` to inadvertently create global variables.

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -188,16 +188,19 @@ export class SpeechExplorer extends AbstractKeyExplorer<string> {
    * @override
    */
   public Start() {
+    let options = this.getOptions();
+    if (SRE.engineSetup().locale !== options.locale) {
+      SRE.setupEngine({locale: options.locale});
+    }
```
dpvc, line 194: There is a potential problem, here, as `SRE` may not be available yet.  Importing `sre.js` may start an asynchronous load of SRE, and that may not have completed when this is called.  So you may need to do this in an `sreReady()` promise as well (or within an `srePromise.then()` call, where `srePromise` is imported from `sre.js`, and is resolved when SRE has been loaded, though perhaps is not yet ready).

ts/a11y/sre.ts
```diff
@@ -61,5 +63,15 @@ export const sreReady = new Promise<void>((resolve, reject) => {
...
 
+
+/**
+ * The functions we need from SRE.
+ */
+export type SRE = {
+  toEnriched(mml: string): Element;
+  setupEngine(options: OptionList): void;
+  engineSetup(): OptionList;
+};
```
dpvc, line 77: See my comment above about changing this to `export declare const SRE: {...}` instead.

I can make a PR to this PR that includes the change here and to `sre-node.ts` if that is easier.

zorkow:
I tried your suggestions, unfortunately, it does not work. The `const` declaration in `sre.ts` overwrites the `SRE` variable that is set in `sre_browser.js`. And that is really what we rely on when call it `SRE.setupEngine` in `semantic-enrich.ts`. In other words, while `sreReady` resolves, `SRE` is still undefined.
But I have another idea, to move the `sre` functionality into 'sre.ts`, so we should be able to avoid all this messing about. Once I've worked it out (and it works as I think it should), I'll push it to this PR.

dpvc:
> The const declaration in sre.ts overwrites the `SRE` variable that is set in `sre_browser.js`. 

I don't think that is right, as `declare const` should not produce any output in the .js files.  It is precisely for handling outside libraries like this.  Also, I tried it and it worked fine for me.  Did you change the `=` to a `:` as in the code I gave above?  You are supposed to be declaring a constant of a particular type, not giving that constant a value.  Perhaps that is why your version overwrote the global `SRE`?

zorkow:
That's what I've got:
```
/**
 * The functions we need from SRE.
 */
export declare const SRE: {
  toEnriched(mml: string): Element;
  setupEngine(options: OptionList): void;
  engineSetup(): OptionList;
};
```

zorkow:
That's the error I get:
```
Error: Cannot read property 'toEnriched' of undefined
```

zorkow:
It works fine if I declare the `const` locally in each file. But not if I export it. 

But if it works for you, maybe my system is just screwed up somewhere... Like for the error we were chasing last Thursday.

zorkow:
The problem is when calling `sre_js_1.SRE.setupEngine` in js, `SRE` is of course undefined. 
Previously it would just call `SRE.setupEngine`, calling the one of the API functions.

dpvc:
Let me check again and make sure I did what I said I did...

dpvc:
OK, you are right.  I must have not tested properly (maybe didn't `tsc` or something).  My apologies.

What was your other idea for this?

zorkow:
The first was to move all the calls to sre into `sre.ts` and then either a single method calling sre API methods (which is ugly due to different return types) or to export a different method for each call, similary to sreReady. While this would keep everything cleanly in the `sre.ts` file only, I am not sure if we would need to duplicate this for `sre-node.ts`. I have not tested that yet.

dpvc:
I don't think you would need to change `sre-node.ts`, which is just a replacement for `sre_browser.js` when SRE is used in node.  Basically, all it does is make sure that both the`SRE` and `sre` objects are available globally, as they are in the browser, so that the other a11y modules can call them no matter if they are in a browser or node.  The `sre.ts` file loads either `sre_browser.js` or `sre-node.ts`, depending on whether we are in a browser or not.  So other modules only call on `sre.ts`.

I suppose it would be possible to make a common interface in `sre.ts` so that only `sre.ts` has to know about the actual SRE library (kind of like the DOMadaptor approach).  This is what it was trying to do by making `SRE` and `sre` globals available, which were intended to make the common interface available in both browser and node.  Perhaps there is a better way to do that.

zorkow:
> I suppose it would be possible to make a common interface in sre.ts so that only sre.ts has to know about the actual SRE library (kind of like the DOMadaptor approach). 

Is that not the role of `sre_browser.d.ts`? Maybe we should have everything except for the ready method in there.

dpvc:
My understanding is that `sre_browser.js` only works in the browser.  MathJax needs to work in both the browser and node, so so it has to get SRE one way for the browser and another way for node.  These are not equivalent, as the browse has two global variables, `SRE` and `sre`, while node has a local variable from the `require()` call that loads it.  The MathJax `sre.ts` file is meant to normalize the two usages.

dpvc:
The `sre_browser.d.ts` file gives the types for the `sre` global object, but my recollection is that the value returned by `require('speech-rule-engine')` is not the same as this `sre` global from `sre_browser.js`  That is why `sre-node.ts` has to "fake" the `sre` object.

zorkow:
That's correct. `SRE` in the browser is a way to bind all the API functionality that you get with `require` in node. I was more thinking along the line of providing a `.d.ts` file for `SRE`.
Something along the lines of this:

```
import {OptionList} from '../util/Options.js';

declare namespace SRE {
  export function toEnriched(mml: string);
  export function setupEngine(obj: OptionList);
  export function engineSetup(): OptionList;
}
```

dpvc:
I'm OK with that.

zorkow:
The problem is really that sre is inconsistent between 

At some point I was tempted to insert
```
  global.SRE = module.exports;
```
in its [API definitions](https://github.com/zorkow/speech-rule-engine/blob/847b041271597ab6d57753f1a240f8cec1017292/src/common/api.js#L149).
But then I thought, that might be too intrusive, for someone who wants to use the node library.

But that's effectively what we do now in MathJax. I could change that if you think it is a good idea.

dpvc:
> The problem is really that sre is inconsistent between

Something seems to be missing at the end of that sentence.  

> I could change that if you think it is a good idea.

I would avoid additional globals in node if they can be avoided.  I did it in `sre-node.ts` as an expedient, but it might be better to have `sreReady` pass the `SRE` and `sre` values to its `then()` functions in order to avoid globals.  These could be cached via local variables in the various modules, or they could just use `sreReady().then()` whenever they needed one of them.  (I haven't looked at explorer to see how complicated that would be).

zorkow:
> > The problem is really that sre is inconsistent between
> 
> Something seems to be missing at the end of that sentence.

Sorry, I meant _discrepancy between node and browser_. 
In the sense: In the former you have the API methods only, in the latter you have the entire library.

> 
> > I could change that if you think it is a good idea.
> 
> I would avoid additional globals in node if they can be avoided. I did it in `sre-node.ts` as an expedient, but it might be better to have `sreReady` pass the `SRE` and `sre` values to its `then()` functions in order to avoid globals. These could be cached via local variables in the various modules, or they could just use `sreReady().then()` whenever they needed one of them. (I haven't looked at explorer to see how complicated that would be).

I was playing with that earlier, but that will make it very unwieldy. And major rewrite of the explorer methods.

dpvc:
> Sorry, I meant inconsistent between node and browser.

That's what I thought.  I did find that inconsistency somewhat challenging when doing the `semantics-enrich` extension.

There are a number of places where MathJax needs to do something different between browsers and node, and all of them are hacks just to make it work.  It would ben good to go back and normalize some of those situations so that they are easier to maintain and more consistent with each other.

ts/a11y/sre.ts
zorkow: This is now in `sre_browser.d.ts`.

ts/a11y/explorer/KeyExplorer.ts
zorkow: This is now in called only after the promise resolves. Then creates another promise, in case a new locale file needs to be loaded. This also fixes the problem of order for the `KeyExplorers` that we had discussed.

zorkow:
Following our discussions, I have now
* moved the `SRE` declaration into `sre_browser.d.ts`. I feel this is where it really belongs. 
* rendered the type for SRE configurations more precise (we cannot `import {OptionList}` in the `d.ts`).  
* used `sreReady` for locale setting, which also provides an elegant solution to the ordering issue of `KeyExplorers`.

PTAL.

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -188,16 +188,21 @@ export class SpeechExplorer extends AbstractKeyExplorer<string> {
...
-      }).catch((error: Error) => console.log(error.message));
+      sreReady().then(() => {
+        if (SRE.engineSetup().locale !== options.locale) {
+          SRE.setupEngine({locale: options.locale});
+        };
+        sreReady().then(() => {
+          this.Speech(this.walker);
+          this.Start();
+        }).catch((error: Error) => console.log(error.message));
+      });
```
dpvc, line 202: I'm wondering if the `catch()` here should be moved down to the outer `sreReady()` promise?  That way, it would catch errors not only ion the `Speech()` and `Start()` calls, but also the `setupEngine()` call.

Also, in the earlier version, the locale was checked every time the KeyExplorer's `Start()` was called, and now it is only done when `this.init` is false (so only once).  Does that matter?  Do the options ever change after initialization?

dpvc:
This all looks good, though I had two small questions (listed above).

ts/a11y/explorer/KeyExplorer.ts
zorkow: I moved it inside, as `setupEngine` does it's own error handling. But it is a good idea to have a catch for every outer promise. 

Currently, the only way to change locale is by adding `a11y` option. We can't do it from the menu yet. So it will only happen on init. But I have a think about how it could be done more generally.

dpvc:
OK, if you are ready to go with it now, I will merge.  Let me know.

zorkow:
I've move the catch and added a remark that locale should always be updated.
For now it is ready to merge.

dpvc:
OK, great!  Thanks.

PR #469, opened 2020-04-03
Revert "Add ability to include MJXZERO to font-family used for undefined characters"

dpvc:
Reverts mathjax/MathJax-src#449


dpvc:
This should have been targeted at develop.  Argh!

PR #471, opened 2020-04-04
Fix positions of superscripts in "cramped" style.  (mathjax/MathJax#2395)

dpvc:
Get `texprimestyle` from the property list, not the attribute list.  Also, fix the handling of D' style (display prime), which should also be cramped.

Resolves issue mathjax/MathJax#2395.

### DIFF
--- a/ts/output/common/Wrappers/scriptbase.ts
+++ b/ts/output/common/Wrappers/scriptbase.ts
@@ -325,8 +325,9 @@ export function CommonScriptbaseMixin<W extends AnyWrapper,
          */
         public getU(bbox: BBox, sbox: BBox) {
             const tex = this.font.params;
-            const attr = this.node.attributes.getList('displaystyle', 'texprimestyle', 'superscriptshift');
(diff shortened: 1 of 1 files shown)


ts/output/common/Wrappers/scriptbase.ts
```diff
@@ -325,8 +325,9 @@ export function CommonScriptbaseMixin<W extends AnyWrapper,
          */
         public getU(bbox: BBox, sbox: BBox) {
             const tex = this.font.params;
-            const attr = this.node.attributes.getList('displaystyle', 'texprimestyle', 'superscriptshift');
-            const p = (attr.displaystyle ? tex.sup1 : attr.texprimestyle ? tex.sup3 : tex.sup2);
+            const attr = this.node.attributes.getList('displaystyle', 'superscriptshift');
+            const prime = this.node.getProperty('texprimestyle');
+            const p = (prime ? tex.sup3 : attr.displaystyle ? tex.sup1 : tex.sup2);
```
zorkow, line 330: Can we move the outer parenthese to the inner ` ? : ` ?, that will make it clearer.

dpvc: Done.

PR #474, opened 2020-04-08
Adapts Menu entry to new reduced number of ChromeVox styles in SRE.

zorkow:
SRE now only offers two styles for ChromeVox speech rules:
* `default` (here called Standard) which used to be `short`, and was always the default setting.
* `alternative` that gives alternative readings to some characters and rules. As before.


dpvc:
OK, I will merge.

PR #477, opened 2020-04-13
Fix issue with \overbrace in SVG output (mathjax/MathJax#2402)

dpvc:
This fixes a problem with not passing the variant name to `unicodeChars()` in the SVG `Wrapper.ts`, and also improves error checking in `unicodeChars()` to avoid similar problems in the future.

Resolves issue mathjax/MathJax#2402.

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -699,17 +699,18 @@ export class CommonWrapper<
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dpvc:
Now that I have merged the `pretty-code` (which changes indentation), there are conflicts, and they are more complicated than I want to try to resolve here.  So I'm making a new branch with the identical changes (with the new indentation), and will merge that instead.

PR #483, opened 2020-05-08
Make formatError a configuration option for tex input jax

dpvc:
This PR makes the tex input jax's `formatError()` function configurable, in line with the `compileError()` and `typesetError()` functions already in place for the `MathDocument` object.  This allows easy trapping of TeX run-time errors (like missing braces) without having to subclass the TeX input jax itself.

It would be nice to be able to pass the `MathItem` that is being processed at the time of the error, but that is not available.  We could save it in the TeX instance so that it can be retrieved in the `formatError()` function.

### DIFF
--- a/ts/input/tex.ts
+++ b/ts/input/tex.ts
@@ -70,7 +70,8 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
     // Digit pattern to match numbers.
     digits: /^(?:[0-9]+(?:\{,\}[0-9]{3})*(?:\.[0-9]*)?|\.[0-9]+)/,
     // Maximum size of TeX string to process.
-    maxBuffer: 5 * 1024
+    maxBuffer: 5 * 1024,
+    formatError: (jax: TeX<any, any, any>, err: TexError) => jax.defaultFormatError(err)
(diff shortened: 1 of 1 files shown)


ts/input/tex.ts
```diff
@@ -217,6 +218,15 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
    * @return {Node} The merror node.
    */
   protected formatError(err: TexError): MmlNode {
```
zorkow, line 220: Should that be `public`?

dpvc: No, this is the function that is internal to the TeX input jax.  It calls the format function in the options list, and that one calls the (public) `defaultFormatError()` by default.

Since this protected `formatError()` is only called from the `compile()` function (when an error is thrown), I suppose its contents could be moved to there, and then rename `defaultFormatError()` as (public) `formatError()` and reduce the need for an extra function.  I should have through too that originally.

dpvc:
OK, I've made the changes I suggested above.  See if you like that better.

zorkow:
Thanks for the changes. That has already been approved.

PR #485, opened 2020-05-09
Rename extensions

dpvc:
This PR implements the renaming of tex extensions to normalize the names of the packages and their directories.

* amsCd, ams_cd => amscd
* colorV2, color_v2 => colorv2
* configMacros, config_macros => configmacros
* tagFormat, tag_format => tagformat

Note that git does not actual track renaming of files, so GitHub tries to work out the renaming that was done, but doesn't always get it right, here, since a number of the files are very similar.  So there are some differences that may seem weird due to that.

I added code to the renamed components and to the dependencies file in order to make the current names still work in the new version.  That involved not just making the loader load the new names using the old ones, but also fixing the package list when an old name appeared, and adjusting the preferences if the old name was used (e.g., `tex.tagFormat` preferences need to me moved to `tex.tagformat`).


components/src/input/tex/extensions/rename.js
```diff
@@ -0,0 +1,18 @@
+import {combineConfig} from '../../../../../js/components/global.js';
+
+//
+// Look for a package name in the package list and change it to a new name
+//   rhen rename tex options for it, if there are any.
```
zorkow, line 5: Typo

PR #491, opened 2020-05-13
General improvements to input/output errors

dpvc:
This PR improves the formatting of 'Math output error' messages (to make them display like other errors, and to allow the contextual menu to be used, and to not cause errors with assistive-mml).  The message is also stored in the `inputData`/`outputData` object for easier retrieval, if needed.

It also adds tooltips to the input and output errors to give the message for the error that occurred, and adds a similar tooltip to the `noerrors` output so that it is easier to get the TeX error that caused the problem.  The SVG output jax's `merror` wrapper is modified to move the message to a `<title>` element where it will be displayed by the browser as a tool tip.

For `noerrors` output, it would be nice to be able to save the message in the MathItem's `inputData` object, but since the TeX input jax's `formatError()` method doesn't have access to the MathItem, that can't be done.  It would be nice to change that (see #483).

### DIFF
--- a/ts/input/tex.ts
+++ b/ts/input/tex.ts
@@ -173,7 +173,6 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
    * @override
    */
   public compile(math: MathItem<N, T, D>, document: MathDocument<N, T, D>): MmlNode {
-xxx();
     this.parseOptions.clear();
     this.executeFilters(this.preFilters, math, document, this.parseOptions);
     let display = math.display;
--- a/ts/input/tex.ts
+++ b/ts/input/tex.ts
@@ -173,6 +173,7 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
    * @override
    */
   public compile(math: MathItem<N, T, D>, document: MathDocument<N, T, D>): MmlNode {
+xxx();
     this.parseOptions.clear();
     this.executeFilters(this.preFilters, math, document, this.parseOptions);
     let display = math.display;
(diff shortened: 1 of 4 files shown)


dpvc:
> We could have it as a proper title element, i.e., not just an attribute, but then it would need to be above the aria-hidden attribute.

It actually already *is* a `<title>` element, it was just in the wrong element.  This was written before the font-inheritance PR landed, and the error message in SVG was made up of the character paths.  I put the `<title>` element in the background (yellow) rectangle, since it was more likely to be hovered over than the letters themselves.  When the font-inheritance PR was merged, that made the error message into a `<text>` element, and that was on top of the background rectangle, so it blocked the rectangle (from a mouse-hovering standpoint) and the title wasn't displayed.

It turns out that moving the `<title>` to the `<g>` that represents the error itself will let it work both with and without font inheritance.

In CHTML is the tooltip read by screen readers?  You say it doesn't interfere, but does that mean it is read or not read?  I hadn't intended for it to be read.  If you think it needs to be, then we can look into how to address that.

I've made a new merge to fix the SVG title issue.  See if that works better for you.

ts/input/tex.ts
```diff
@@ -173,6 +173,7 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
    * @override
    */
   public compile(math: MathItem<N, T, D>, document: MathDocument<N, T, D>): MmlNode {
+xxx();
```
zorkow, line 176: That should probably be removed.

dpvc: OOPS!  That was my testing code to force a compile error.

dpvc: Done.

PR #496, opened 2020-05-14
Jsdoc lint

zorkow:
Lints the JSDoc to some extent. This is based on a crude emacs lisp function and eyeballing.
Things it definitely will not catch:
* Functions that take a parameter but have no `@param` element in JSDoc 
* Missing `return` statements.

Things that might be partially missed: 
* Incorrect or incomplete types (in particular if they contain templates).
* Missing parameters when parameter declaration block was broken over multiple lines.

Currently done for all modules expect `input` and `output`.

### DIFF
--- a/ts/input/tex/Symbol.ts
+++ b/ts/input/tex/Symbol.ts
@@ -33,7 +33,7 @@ export class Symbol {
   /**
    * @constructor
    * @param {string} _symbol The symbol parsed.
-   * @param {string} char The corresponding translation.
+   * @param {string} _char The corresponding translation.
    * @param {Attributes} _attributes The attributes for the translation.
    */
   constructor(private _symbol: string, private _char: string,
--- a/ts/input/tex/Symbol.ts
+++ b/ts/input/tex/Symbol.ts
@@ -32,9 +32,9 @@ export class Symbol {
 
   /**
    * @constructor
-   * @param {string} _symbol The symbol parsed.
-   * @param {string} _char The corresponding translation.
-   * @param {Attributes} _attributes The attributes for the translation.
+   * @param {string} symbol The symbol parsed.
+   * @param {string} char The corresponding translation.
+   * @param {Attributes} attributes The attributes for the translation.
    */
   constructor(private _symbol: string, private _char: string,
               private _attributes: Attributes) {
@@ -59,9 +59,9 @@ export class Macro {
 
   /**
    * @constructor
-   * @param {string} _symbol The symbol parsed
-   * @param {ParseMethod} _func The parsing function for that symbol.
-   * @param {Args[]} _args Additional arguments for the function.
+   * @param {string} symbol The symbol parsed
+   * @param {ParseMethod} func The parsing function for that symbol.
+   * @param {Args[]} args Additional arguments for the function.
    */
   constructor(private _symbol: string, private _func: ParseMethod,
               private _args: Args[] = []) {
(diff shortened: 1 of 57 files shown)


zorkow:
This is now finished (modulo the above caveats).
Some additional comments:

* In `FilterUtil, BoldsymbolConfiguration`: The `arg` elements are given in
  destructured manner (i.e. every field separately).
* In `ColorUtil`: There are JSdoc for `ColorModelProcessors.set`, which is not
  really a function definition. I left them as they are.
* Files where I added parameter comments but am not 100% certain that they are
  correct: `FontData, common/Wrapper`

ts/input/tex/TexError.ts
```diff
@@ -36,7 +36,8 @@ export default class TexError {
 
   /**
    * The old MathJax processing function.
-   * @param {Array.<string>} input The input message.
+   * @param {string} str The basic error message.
+   * @param {Array[]} args The arguments to be replaced in the error message.
```
dpvc, line 40: Should this be `string[]`?

ts/input/tex/Symbol.ts
```diff
@@ -33,7 +33,7 @@ export class Symbol {
   /**
    * @constructor
    * @param {string} _symbol The symbol parsed.
-   * @param {string} char The corresponding translation.
+   * @param {string} _char The corresponding translation.
```
dpvc, line 36: Since the underscore is to prevent Typescript complaining about unused arguments, I'm wondering if we want to include the `_` in the variable name or not.  I think of this as not part of the name, but a mark to indicate we know it is not used (similar to a trailing `?` or `=`).  What do you think?  I notice that it isn't added in other places that have underscores.

ts/output/chtml/FontData.ts
```diff
@@ -370,9 +372,9 @@ export class CHTMLFontData extends FontData<CHTMLCharOptions, CHTMLVariantData,
 
   /**
    * @param {StyleList} styles  The style object to add styles to
-   * @param {string} vclass     The variant class string (e.g., .mjx-b) where this character is being defined
+   * @param {string} vletter    The variant class string (e.g., .mjx-b) where this character is being defined
```
dpvc, line 375: The description is no longer accurate. This is now the variant class letter (e.g., 'B' for bold, 'SS for sans-serif).

ts/output/common/OutputJax.ts
```diff
@@ -170,7 +170,8 @@ export abstract class CommonOutputJax<
    * Get the cssStyle and font objects
    *
    * @param {OptionList} options         The configuration options
-   * @param {FontDataClass} defaultFont  The default FontData constructor
+   * @param {CommonWrapperFactory} defaultFactory  The default wrapper factory
```
dpvc, line 173: This should be "The default wrapper factory class" (missing "class")

ts/output/common/OutputJax.ts
```diff
@@ -545,6 +545,8 @@ export abstract class CommonOutputJax<
    *
    * @param {N} text         The text element to measure
    * @param {string} chars   The string contained in the text node
+   * @param {string} variant     The variant for the text
+   * @param {CssFontData} font   The style cssText string containing the font information
```
dpvc, line 549: This is not a string, but a triple `[string, boolean, boolean]` indicating the CSS font to use for the given variant, and wither the font is italic or bold (the two booleans).  Perhaps `The family, italic, and bold data for explicit fonts` would be best.

ts/input/tex/Symbol.ts
zorkow: I agree, in particular we have not been consistent in having the underscore in the JSDoc.
I removed all preceding underscores from the JSDoc now.

Issue #351, opened 2019-09-22
API needed for clearing math items for a given container


dpvc:
We have a call that will clear the math list, but it would be good to have a function (on the HTMLDocument class) that accepts a container and removes the math items that are within that container.  That way, if a container's contents are replaced (e.g., via `innerHTML`), then it would be easy to fix up the make list to take that into account.

PR #502, opened 2020-05-27
Add functions to manage math items within a given container. (#351)

dpvc:
This PR adds API calls to allow managing of math items with container elements in the DOM.  For example, if you are updating a portion of the DOM (by setting its `innerHTML` for example) and typeset math is removed, you want to be able to remove the associated MathItems from the list of MathItems held in the MathDocument.  Previously there was no API for that.

There two new main calls added to the base MathDocument class:  `clearMathItemsWithin()` and `getMathItemsWithin()`.  Both take a container element; the first removes the MathItems from the MathList that are found within the container, and the second returns an array of MathItems found within the container.

There is also a new LinkedList method, `remove()` for removing an item from the list.

Finally, there is a new DOMAdaptor method (`contains()`) to determine if an element is contained in another element.

This PR also fixes a bug in the liteAdaptor's `replace()` call where the parent elements weren't being set for the new nodes being inserted into the LiteDOM.

Resolves issue #351.

### DIFF
--- a/ts/core/DOMAdaptor.ts
+++ b/ts/core/DOMAdaptor.ts
@@ -115,6 +115,15 @@ export interface DOMAdaptor<N, T, D> {
   getElements(nodes: (string | N | N[])[], document: D): N[];
 
   /**
+   * Determine if a container node contains a given node in somewhere in its DOM tree
+   *
+   * @param {N} container  The container to search
+   * @param {N|T} node     The node to look for
+   * @return {boolean}     True if the node is in the container's DOM tree
+   */
+  contains(container: N, node: N | T): boolean;
+
+  /**
    * @param {N|T} node  The HTML node whose parent is to be obtained
    * @return {N}        The parent node of the given one
    */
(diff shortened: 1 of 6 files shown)


ts/core/DOMAdaptor.ts
```diff
@@ -114,6 +114,15 @@ export interface DOMAdaptor<N, T, D> {
    */
   getElements(nodes: (string | N | N[])[], document: D): N[];
 
+  /**
+   * Determine if a container node contains a given node in somewhere in its DOM tree
```
zorkow, line 118: remove first `in`

ts/core/MathDocument.ts
```diff
@@ -857,6 +884,34 @@ export abstract class AbstractMathDocument<N, T, D> implements MathDocument<N, T
...
+   * @override
+   */
+  public getMathItemsWithin(elements: ContainerList<N>) {
+    if (!Array.isArray(elements)) {
+      elements = [elements];
+    }
+    const adaptor = this.adaptor;
+    const items = [] as MathItem<N, T, D>[];
+    const containers = adaptor.getElements(elements, this.document);
+    ITEMS: for (const item of this.math) {
```
zorkow, line 904: Newline after `ITEMS:`.

ts/core/DOMAdaptor.ts
dpvc: done

ts/core/MathDocument.ts
dpvc: done

PR #503, opened 2020-05-29
Fix processing of texClass for multi-character <mi> elements.

dpvc:
When porting of `mo.adjustTeXClass()` to v3, I messed up these lines.  I used the wrong property, (it is capitalized in the two other places it is used), so the if-then was never satisfied.  But even if it had been, I set the wrong object's TeX class.  This fixes both issues and makes the results consistent with v2.


dpvc:
> Maybe at some point we should have a standard way to spell: options, properties, attributes, etc.

Agreed.  In fact, this typo was because I thought we were going to camel case, but the TeX input jax retained the original capitalization.  Something for the next round of global updates.

PR #504, opened 2020-05-29
Update string handling to handle SMP unicode blocks better.

dpvc:
This PR updates the string-handling functions to take advantage of some ES6 string functions (`fromCodePoint()`, `codePointAt()`, and the fact that `Array.from()` splits a string into an array of unicode characters).  It also enhances the TeX parser string handling to handle unicode characters better.  E.g., you can use a (single) unicode character as a macro name (e.g., `\α`) or as an argument (e.g. `\sqrt α`) or in macro definition templates (e.g., `\def\x #1α{[#1]}` so that `\x abcα` will produce `[abc]`), and so on.

### DIFF
--- a/ts/input/tex/TexParser.ts
+++ b/ts/input/tex/TexParser.ts
@@ -183,11 +183,9 @@ export default class TexParser {
     let c: string;
     let n: number;
     while (this.i < this.string.length) {
-      c = this.string.charAt(this.i++);
-      n = c.charCodeAt(0);
-      if (n >= 0xD800 && n < 0xDC00) {
-        c += this.string.charAt(this.i++);
-      }
+      n = this.string.codePointAt(this.i);
+      c = String.fromCodePoint(n);
+      this.i += c.length;
       this.parse('character', [this, c]);
     }
   }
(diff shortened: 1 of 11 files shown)


ts/input/tex/TexParser.ts
```diff
@@ -183,11 +183,9 @@ export default class TexParser {
     let c: string;
     let n: number;
     while (this.i < this.string.length) {
-      c = this.string.charAt(this.i++);
-      n = c.charCodeAt(0);
-      if (n >= 0xD800 && n < 0xDC00) {
-        c += this.string.charAt(this.i++);
-      }
+      n = this.string.codePointAt(this.i);
+      c = String.fromCodePoint(n);
```
zorkow, line 187: Collapse into one and add a utility method for
`String.fromCodePoint(this.string.codePointAt`

ts/input/tex/TexParser.ts
```diff
@@ -317,7 +315,9 @@ export default class TexParser {
       // @test MissingCloseBrace
       throw new TexError('MissingCloseBrace', 'Missing close brace');
     }
-    return this.string.charAt(this.i++);
+    const c = String.fromCodePoint(this.string.codePointAt(this.i));
```
zorkow, line 318: Ditto.

ts/input/tex/TexParser.ts
```diff
@@ -259,14 +257,14 @@ export default class TexParser {
     while (this.nextIsSpace()) {
       this.i++;
     }
-    return this.string.charAt(this.i);
+    return String.fromCodePoint(this.string.codePointAt(this.i));
```
zorkow, line 260: Replace with utility method.

ts/input/tex/TexParser.ts
dpvc: Done (and for the ones below)

PR #505, opened 2020-06-01
Update new context menu

zorkow:
__Not yet ready for review__


zorkow:
This is now ready for review. There are still two issues:

1. A layout/display issue of the selection box that needs to be sorted out on the context menu side.
1. A coordination issue between selection box settings and smart settings for clearspeak. This needs to be sorted at the SRE end (for now). 

Both are minor issues. So it is ready modulo version numbers of dependencies.

dpvc:
I've looked through it and tried it out, and it looks good.  There is one change to the `ui/menu` component that is needed, but I have made a PR for that, which should be merged first, and then this one can be merged.  I made a corresponding PR for the MathJax-dev repository that allows the lab to work with the new menu code.

dpvc:
I should mention that I'm seeing debug messages from the `mj-context-menu` code (in `selection_box.ts` lines 134 and 135) when the explorer is active and you select the expression after exiting the explorer.

I also had trouble getting the highlight options to work (hover and flame).  I just may jot be doing it right, but they didn't seem to do anything.  If I enabled flame and then reload, I could sometimes get the flame to work, but not always.

zorkow:
> I should mention that I'm seeing debug messages from the `mj-context-menu` code (in `selection_box.ts` lines 134 and 135) when the explorer is active and you select the expression after exiting the explorer.
> 

I wanted to make one more clean release of the package, after fixing one more layout issue, which will remove the logging, too.

> I also had trouble getting the highlight options to work (hover and flame). I just may jot be doing it right, but they didn't seem to do anything. If I enabled flame and then reload, I could sometimes get the flame to work, but not always.

Thanks for spotting this. I can see that, too. What's odd is that the semantic info hoverer that use the same principle work fine. I'll look into that.

zorkow:
> > I also had trouble getting the highlight options to work (hover and flame). I just may jot be doing it right, but they didn't seem to do anything. If I enabled flame and then reload, I could sometimes get the flame to work, but not always.
> 
> Thanks for spotting this. I can see that, too. What's odd is that the semantic info hoverer that use the same principle work fine. I'll look into that.

I figured out the problem: It is an ordering issue. If `hover/flame` is on before `collapsible` is active, the respective explorers contain the old base node and hence don't see the `maction` elements
(Switching them off/on again solves the problem.)

Switching on `collapsible` would have to reinitialise the respective hoverers.

PR #508, opened 2020-06-01
Allow TeX extensions to store data that is local to the TeX input jax.

dpvc:
This PR provides a storage area for TeX extensions to keep information that is local to the TeX input jax instance in which they are operating.  The `color` extension uses that to handle information about defined colors, and the `require` and `autoload` extensions use it to track the extensions that are loaded and configured.

In the past, this data was stored in the configurations `options` object, as the only available location, but that muddles the meaning of the `options` object, which was the default options with the user-supplied options merged in.  This PR separates out the data storage from the option specifications.

### DIFF
--- a/ts/input/tex/ParseOptions.ts
+++ b/ts/input/tex/ParseOptions.ts
@@ -67,6 +67,12 @@ export default class ParseOptions {
    */
   public tags: Tags;
 
+  /**
+   * Storage area for parser-specific package data (indexed by package name)
+   * @type {{[name: string]: any}}
+   */
+  public packageData: {[name: string]: any} = {};
+
   // Fields for ephemeral options, i.e., options that will be cleared for each
   // run of the parser.
   /**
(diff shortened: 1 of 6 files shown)


ts/input/tex/ParseOptions.ts
```diff
@@ -67,6 +67,12 @@ export default class ParseOptions {
    */
   public tags: Tags;
 
+  /**
+   * Storage area for parser-specific package data (indexed by package name)
+   * @type {{[name: string]: any}}
+   */
+  public packageData: {[name: string]: any} = {};
```
zorkow, line 74: I think this should be a map.

zorkow:
You might want to hold off reviewing PR #518 until this has been merged.

zorkow:
I've just seen #518  is already approved. No worries. I will merge once this PR is merged.

ts/input/tex/ParseOptions.ts
dpvc: Done.  Can you re-review?

PR #509, opened 2020-06-04
Add new textmacros extension for handling some macros in text-mode

dpvc:
This PR ports the `text-macros` extension from v2 to v3, and makes a component for it.

This includes the changes to Configuration.ts to add `Configuration.createUnNamed()` that produces an un-registered configuration, as well as the change to `ParserUtil.internalMath` to allow that function to be overridden.  

There is a fix to `GetCS()` to remove a space following the control sequence (this makes no difference in math mode, but does in text mode, which is why we never noticed it earlier).

This PR relies on the `extension-data` branch, which is why it branches from there and targets it for merging.


ts/input/tex/textmacros/TextMacrosMethods.ts
```diff
@@ -0,0 +1,284 @@
...
+    parser.saveText();
+    parser.stack.env.mathsize = size;
+  },
+
+  /**
+   * @param {TextParser} parser   The text-mode parser
+   * @param {string} name         The control sequence that called this function
+   */
+  CheckAutoload(parser: TextParser, name: string) {
+    const autoload = parser.configuration.packageData.autoload;
```
zorkow, line 257: Could those macros not be included into the `Autoload` package? Or is that meant to work even if `Autoload` is not improrted?

ts/input/tex/textmacros/TextParser.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * Add mathsize, mathcolor, and mathvariant to token nodes,
+   *   if they are specified in the environment
+   *
+   * @param {MmlNode} mml   The node to adjust
+   */
+  public addAttributes(mml: MmlNode) {
+    const env = this.stack.env;
+    if (mml.isToken) {
```
zorkow, line 157: Swap order and start with `if (!mml.isToken) return;`

ts/input/tex/textmacros/TextMacrosMethods.ts
```diff
@@ -0,0 +1,284 @@
...
+        const cs = parser.GetCS();
+        if (cs === ')') c = '\\(';  // \( is the opening delimiter for \)
+      case '$':
+        //
+        //  If there are no unbalanced braces and we have found the close delimiter,
+        //    process the contents of the delimiters in math mode (using the original TeX parser)
+        //
+        if (braces === 0 && open === c) {
+          const config = parser.texParser.configuration;
+          const mml = (new TexParser(parser.string.substr(i, j - i), parser.stack.env, config)).mml();
```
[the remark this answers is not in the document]
dpvc: I don't think that is the case.  The value of `i` is the initial `parser.i` value, while `j` is the current `parser.i` within the loop, which is always at least as large as the initial `parser.i`.  If you have a situation where `j - i` is negative, can you give the details?

ts/input/tex/textmacros/TextMacrosMethods.ts
dpvc: This function checks if a macro is set up to be autoloaded, which will only be true if the `autoload` package has been included in the TeX parser, and if the macro is included in the configuration for `autoload`.  

When it is set to be autoloaded, we need to perform that autoloading in the original TeX parser, not the parser that is handling the text macros.  That is why we check for it being an autoloaded macro.  (The original TeX parser is also used when the macro is not defined so that the proper error message or fallback action is performed.)  If it was autoloaded, the expression is restarted, so that the proper macro definition will be used.  Otherwise, the text parser is used (line 267 uses `[texParser, name]` while line 271 uses `[parser, name]`).

The upshot is, this will still work if `autoload` extension is not loaded, or not included in the packages for the TeX parser, or if the macro is not configured to autoload.

dpvc: I've improved the comments to try to make this clearer.  But I believe the code is correct as is.

ts/input/tex/textmacros/TextParser.ts
dpvc: Done

ts/input/tex/textmacros/TextMacrosMethods.ts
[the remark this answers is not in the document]
zorkow: You're right. I believe what I was thinking at the time was that `i` is actually `parser.i` element that can be increased by any parse method, which is outside the control of the loop. But as it is the capture `parser.i` is should remain constant. My bad.

ts/input/tex/textmacros/TextMacrosMethods.ts
zorkow: I think the new comment clarifies my confusion: If `autoload` is there it will have filled the `packageData` and we ask `TexParser` to handle it. Thanks.

ts/input/tex/textmacros/TextMacrosMethods.ts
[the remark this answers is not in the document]
dpvc: I thought that might have been the confusion.  No problem.

ts/input/tex/textmacros/TextMacrosMethods.ts
dpvc: Glad that helped out.  It is a bit complicated how that works, and it took me a bit to remember the details.

PR #514, opened 2020-06-11
Add Safe extension and component

dpvc:
This PR ports the Safe extension from v2 to v3.  It hooks into the MathML input jax via the `filterAttribute()`method that was set up for that, and via a new `filterClassList()` for handling the class attribute.  Some adjustments were made to make handling the filters a bit easier.

It hooks into the TeX input jax as a post filter.  In v2, it hooked into the TeX parser itself so that it only filtered user-supplied values, but that is hard to maintain, as it must be aware of all the extensions and when user input is allowed.  The post-filter avoids that, at the expense of having to look at all the attributes of all the elements, so it is less efficient, but more reliable.

This PR includes a component for the safe extension, so it can be loaded from the CDN.  it also corrects an error in handling `fontsize` in the `common/Wrapper.ts` file.

### DIFF
--- a/ts/ui/safe/SafeMethods.ts
+++ b/ts/ui/safe/SafeMethods.ts
@@ -1,0 +1,288 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2020 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  Support functions for the safe extension
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+import {length2em} from '../../util/lengths.js';
+import {Safe, FilterFunction} from './safe.js';
+
+/**
+ * The default attribute-filtering functions
+ */
+export const SafeMethods: {[name: string]: FilterFunction<any, any, any>} = {
+
+  /**
+   * Filter HREF URL's
+   *
+   * @param {Safe<N,T,D>} safe  The Safe object being used
+   * @param {string} url        The URL being tested
+   * @return {string|null}      The URL if OK and null if not
+   *
+   * @template N  The HTMLElement node class
+   * @template T  The Text node class
+   * @template D  The Document class
+   */
+  filterURL<N, T, D>(safe: Safe<N, T, D>, url: string): string | null {
+    const protocol = (url.match(/^\s*([a-z]+):/i) || [null, ''])[1].toLowerCase();
+    const allow = safe.allow.URLs;
+    return (allow === 'all' || (allow === 'safe' &&
+                                (safe.options.safeProtocols[protocol] || !protocol))) ? url : null;
+  },
+
+  /**
+   * Filter a class list
+   *
+   * @param {Safe<N,T,D>} safe  The Safe object being used
+   * @param {string} list       The class list being tested
+   * @return {string|null}      The class list if OK and null if not
+   *
+   * @template N  The HTMLElement node class
+   * @template T  The Text node class
+   * @template D  The Document class
+   */
+  filterClassList<N, T, D>(safe: Safe<N, T, D>, list: string): string | null {
+    const classes = list.trim().replace(/\s\s+/g, ' ').split(/ /);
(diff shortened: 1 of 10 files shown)


ts/ui/safe/safe.ts
```diff
@@ -0,0 +1,267 @@
...
+   *
+   * @param {string} id      The name of the attribute
+   * @param {string} value   The value of the attribute
+   * @return {string|null}   The sanitized value
+   */
+  public mmlAttribute(id: string, value: string): string | null {
+    if (id === 'class') return null;
+    const filter = (this.filterAttributes.hasOwnProperty(id) ? this.filterAttributes[id] :
+                    (id.substr(0, 5) === 'data-' ? this.filterAttributes['data-'] : null));
+    if (!filter) return value;
```
zorkow, line 251: I thought we only allowed that for empty `return`.

ts/ui/safe/safe.ts
```diff
@@ -0,0 +1,267 @@
...
+  /**
+   * Sanitize a node's attributes
+   *
+   * @param {MmlNode} node      The node to sanitize
+   */
+  protected sanitizeNode(node: MmlNode) {
+    const attributes = node.attributes.getAllAttributes();
+    for (const id of Object.keys(attributes)) {
+      if (this.filterAttributes.hasOwnProperty(id)) {
+        const value = this.filterMethods[this.filterAttributes[id]](this, attributes[id]);
```
zorkow, line 228: Is it certain that for every `filterAttribute` there will be a `filterMethod`?

ts/ui/safe/SafeMethods.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+    return styles;
+  },
+
+  /**
+   * Filter an individual name:value style pair
+   *
+   * @param {Safe<N,T,D>} safe  The Safe object being used
+   * @param {string} style      The style name being tested
+   * @param {N} div             The temp DIV node containing the style object to tbe tested
```
zorkow, line 155: typo

ts/ui/safe/SafeMethods.ts
```diff
@@ -0,0 +1,288 @@
...
+    return this.filterStyleValue(safe, style, value, div);
+  },
+
+  /**
+   * Filter a style's value, handling compount values (e.g., borders that have widths as well as styles and colors)
+   *
+   * @param {Safe<N,T,D>} safe  The Safe object being used
+   * @param {string} style      The style name being tested
+   * @param {string} value      The value of the style to test
+   * @param {N} div             The temp DIV node containing the style object to tbe tested
```
zorkow, line 182: typo

ts/ui/safe/SafeMethods.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+    const name = style.replace(/Top|Right|Left|Bottom/, '');
+    if (!safe.options.safeStyles[style] && !safe.options.safeStyles[name]) {
+      return null;
+    }
+    return this.filterStyleValue(safe, style, value, div);
+  },
+
+  /**
+   * Filter a style's value, handling compount values (e.g., borders that have widths as well as styles and colors)
```
zorkow, line 177: typo

ts/ui/safe/safe.ts
```diff
@@ -0,0 +1,267 @@
...
+      classes: 'safe',   // safe start with MJX-
+      cssIDs:  'safe',   // safe start with MJX-
+      styles:  'safe'    // safe are in safeStyles below
+    },
+    //
+    // Largest padding/border/margin, etc. in em's
+    //
+    lengthMax: 3,
+    //
+    // Valid range for scriptsisemultiplier
```
zorkow, line 69: typo

ts/ui/safe/safe.ts
```diff
@@ -0,0 +1,267 @@
...
+      outlineBottom: true,
+      outlineLeft: true,
+      fontSize: [.707, 1.44]
+    })
+  };
+
+  /**
+   * The attribute-to-filter-method mapping
+   */
+  public filterAttributes: {[name: string]: string} = {
```
zorkow, line 155: I think this should be a `Map`. This would avoid all the `hasOwnProperty` tests later.

ts/ui/safe/safe.ts
dpvc: I didn't think so, but have changed it.

ts/ui/safe/safe.ts
dpvc: Nothing is certain.  What are you suggesting?  Checking and reporting an error if not?  Isn't that what will happen anyway if not?  Because this will be called on every attribute on every node, it should be kept as slim as possible.  If someone adds a `filterAttribute` without creating the corresponding `filterMethod`, then they will get an error message at this point.

dpvc:
I've made the fixes you requested.

ts/ui/safe/safe.ts
dpvc: Changed to a Map

ts/ui/safe/safe.ts
zorkow: I was wondering if we should catch that error at this point, i.e., if the `filterMethod` does not exist, then we could simply continue in the loop, or throw an explicit error, rather than get a `undefined is not a function` error.

PR #517, opened 2020-06-19
Fixes code point error when string is empty.

zorkow:
Corrects problem that was introduced when moving from `charAt` to using `codePointAt` in the TeX parser. Sometimes `getNext` is called when the string is already exhausted, which leads to a problem 
when `undefined` is passed to `String.fromCodePoint`. 

Simplest example: `x'`

I am aware that the method now will also return the empty string if the string contains unicode 0 (`NULL`?). My questions:
1. Can code point 0 ever occur in our strings?
1. Even if it does, does it matter when we return `''`?


dpvc:
I don't think we will ever see char 0, but you could use

```
    return code === undefined ? '' : String.fromCodePoint(code);
```

if you want to cover that case.

PR #518, opened 2020-06-22
New TeX configuration

zorkow:
Implements the new TeX configuration.
* The main change is that the combined configurations of the TeX parser are held in a new class `ParserConfiguration`. It is created new for every time the parser is called. It maintains a priority list of currently active configurations, their init methods, config methods, etc. In addition it administers the `SubHandlers` object. So they never have to be recreated, just added to. 
* Configurations can now be ordered by priority. While each configuration gets a default priority, the order can be changed by providing explicit priorities for packages when the tex parser is configured. 
* In addition `config` and `init` methods of each Configuration can have explicit priorities. Otherwise the configuration's priority is taken.
* The `SubHandlers` also maintain a priority list of `MapHandlers`. This makes it easier to keep extension maps (e.g., for user macros) at top priority. 
* We now offer two creator patterns: `create` for a named configuration. `temp` for an ephemeral configuration. In other words we can have fixed configurations for the packages, while adding temporary ones when we need them.
* Extension mappings for packages like `ams` (for `DeclareMathOperator`) and `newcommand` are now created locally in the packages and ephemeral configurations are added with the `temp` pattern.
* Likewise `configmacro` package now uses an ephemeral configuration 

Ultimately, I think `ParserConfiguration` and `ParseOptions` should be collated into a single class. I held off doing that for now. But it should be a pretty quick adjustment if we want to have that in 3.1.


ts/input/tex/Configuration.ts
```diff
@@ -278,3 +233,169 @@ export namespace ConfigurationHandler {
...
+   * The collated node creators.
+   * @type {{[key: string]: any}}
+   */
+  public nodes: {[key: string]: any}  = {};
+
+
+  /**
+   * @constructor
+   */
+  constructor(packages: (string | [string, number])[]) {
```
dpvc, line 297: Missing jsdoc for the `packages` parameter.

ts/input/tex/Configuration.ts
```diff
@@ -278,3 +233,169 @@ export namespace ConfigurationHandler {
...
+   * @param {Configuration} configuration   The configuration where this one is being initialized
+   */
+  public init() {
+    this.initMethod.execute(this);
+  }
+
+  /**
+   * Init method for when the jax is ready
+   *
+   * @param {Configuration} configuration   The configuration where this one is being initialized
```
dpvc, line 318: This parameter isn't there any more.

ts/input/tex/Configuration.ts
```diff
@@ -278,3 +233,169 @@ export namespace ConfigurationHandler {
...
+   * Retrieves and adds configuration for a pacakge with priority.
+   * @param {(string | [string, number]} pkg Package with priority.
+   */
+  public addPackage(pkg: (string | [string, number])) {
+    const name = typeof pkg === 'string' ? pkg : pkg[0];
+    let conf = ConfigurationHandler.get(name);
+    if (conf) {
+      this.configurations.add(
+        conf, typeof pkg === 'string' ? conf.priority : pkg[1]);
+    }
```
dpvc, line 338: Would it be simpler to make this

```
const [name, priority] = typeof pkg === 'string' ? [pkg, config.priority] : pkg;
const conf = ConfigurationHandler.get(name);
if (conf) {
      this.configurations.add(conf, priority);
}
```

ts/input/tex/Configuration.ts
```diff
@@ -278,3 +233,169 @@ export namespace ConfigurationHandler {
...
+    this.append(config);
+    this.configurations.add(config, config.priority);
+    this.init();
+    const parser = jax.parseOptions;
+    parser.nodeFactory.setCreators(config.nodes);
+    for (const kind of Object.keys(config.items)) {
+      parser.itemFactory.setNodeClass(kind, config.items[kind]);
+    }
+    TagsFactory.addTags(config.tags);
+    userOptions(parser.options, options);
```
dpvc, line 359: You need to add

```
    defaultOptions(parser.options, config.options);
```

before this line otherwise packages that have configuration options (like `amscd`, `color`, etc.) will not get their default configurations, and will throw an error trying to set their user-supplied options at this line.  That can be tested using

```
node -r esm
require('./components/src/node-main/node-main.js').init({loader: {load: ['input/tex', '[tex]/require']}}).then(() => MathJax.tex2mmlPromise('\\require{color} {\\color{red} x}')).then((x) => console.log(x)).catch((e) => console.log(e.message))
```

for example.

ts/input/tex/Configuration.ts
```diff
@@ -22,42 +22,87 @@
...
+import {TagsFactory} from './Tags.js';
 
 
-export type HandlerConfig = {[P in HandlerType]?: string[]};
-export type FallbackConfig = {[P in HandlerType]?: ParseMethod};
 export type StackItemConfig = {[kind: string]: StackItemClass};
 export type TagsConfig = {[kind: string]: TagsClass};
-export type ProcessorList = (Function | [Function, number])[];
+export type Processor<T> = [T, number];
+export type ProcessorList = Processor<Function>[];
```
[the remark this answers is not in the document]
zorkow: Added the type.

ts/input/tex/Configuration.ts
zorkow: The problem is that in the first line you do not yet have the configuration `config`. So you need the name first to look up the configuration. Then you add it either with the given priority or the default configuration if none is given. I know, it looks ugly. I reduce it to only use `typeof pkg === 'string'` once.

zorkow: Now I recall why it was like this: Typescript complains if the test is done elsewhere. So we currently avoid some casting.
So I'll leave it as is.

ts/input/tex/Configuration.ts
```diff
@@ -82,140 +128,39 @@ export class Configuration {
...
-    return Configuration.create(
-      'extension',
-      {handler: {character: [],
-                 delimiter: [ExtensionMaps.NEW_DELIMITER],
-                 macro: [ExtensionMaps.NEW_DELIMITER,
-                         ExtensionMaps.NEW_COMMAND,
-                         ExtensionMaps.NEW_MACRO],
-                 environment: [ExtensionMaps.NEW_ENVIRONMENT]
-                }});
+  public static temp(config: {handler?: HandlerConfig,
```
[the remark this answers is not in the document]
zorkow: I renamed it to `local`. I think that describe it best.

ts/input/tex/Configuration.ts
zorkow: Thanks for spotting this.

ts/input/tex/Configuration.ts
dpvc: Got it.  Makes sense.

PR #523, opened 2020-07-06
Fix LiteDOM comments and add support for DOCTYPE

dpvc:
I ran into this while updating the node demos.  It turns out that comments were not being output properly from the LiteDOM.  I also wanted to be able to preserve the `DOCTYPE` comment so that the full-page node commands can retain it in their output.  So this PR fixes the comment problem, and adds the `DOCTYPE` support.

### DIFF
--- a/ts/core/DOMAdaptor.ts
+++ b/ts/core/DOMAdaptor.ts
@@ -97,6 +97,12 @@ export interface DOMAdaptor<N, T, D> {
   root(doc: D): N;
 
   /**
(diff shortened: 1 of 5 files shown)


ts/core/DOMAdaptor.ts
```diff
@@ -96,6 +96,12 @@ export interface DOMAdaptor<N, T, D> {
    */
   root(doc: D): N;
 
+  /**
+   * @paramn {D} doc    The document whose doctype is to be obtained
```
zorkow, line 100: typo `@param`

PR #528, opened 2020-07-22
fix: 527

qq240814476:
#527 
just use  [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -708,7 +708,7 @@ export class CommonWrapper<
         //
         //  Remap to Math Alphanumerics block
         //
-        const map = this.font.getVariant(variant).chars;
+        const map = this.font.getVariant(variant)?.chars;
         if (map) {
             //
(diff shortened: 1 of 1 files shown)


qq240814476:
> #527
> just use [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

by the way， I feel so honored to be a MathJax contributor～ LOL🤣

dpvc:
Thanks for your interest (and excitement!) in contributing to MathJax.

This bug has already been fixed in #498 (through a slightly different resolution), so I'm going to close this PR, though the enthusiasm is appreciated.

Just FYI, we follow the git-flow approach in this repository, which means that all changes should be made on a branch from the `develop` branch, and PRs should be targeted to `develop`, not `master`.  So if you make future pull requests, it would  be best to work within that paradigm.  Thanks!

PR #532, opened 2020-08-10
Fix processing of root directory for components to work under Windows (mathjax/MathJax#2486)

dpvc:
It turns out that the implementation of node's `path` module that is included by webpack doesn't know about Windows backslashes as directory separators, so the `node-main` module fails to set the MathJax root directory properly when the `mathjax` package is used under Windows.

This PR handles that situation, and simplifies the lookup a bit, also making sure that the root is properly set no matter where the `node-main.js` file was teen from.  In the past, the `es5` directory in `mathjax-full` or `mathjax` would be taken in preference to the directory where `node-main` came from, so the wrong MathJax tree could be used.  This PR makes sure that the files used for MathJax come from the same location as `node-main.js`.

Resolves issue mathjax/MathJax#2486.

### DIFF
--- a/components/src/node-main/node-main.js
+++ b/components/src/node-main/node-main.js
@@ -21,16 +21,28 @@ combineDefaults(MathJax.config.loader, 'provides', provides);
 
 MathJax.config.loader.paths.mathjax = (function () {
   //
-  // Try to locate the mathjax-full or mathjax package
-  // If neither is found, try to use the directory where this file is located
+  // Convert a windows path to a unix path (when needed)
+  //
+  const convertWindows = (name) => (path.win32 || !name.match(/^[a-z]:\\/i)) ? name : name.replace(/\\/g, '/');
+  //
+  // Locate the directory for this file:
+  //   Note that __dirname is not effective in webpacked files,
+  //   but the complete path is listed in an error message's requireStack when require.resolve() fails.
   //
   try {
-    return path.resolve(MathJax.config.loader.require.resolve('mathjax-full/package.json'), '../es5');
-  } catch (err) {}
-  try {
-    return path.resolve(MathJax.config.loader.require.resolve('mathjax/package.json'), '../es5');
+    //
+    //  Try to locate a non-existing file in order to throw an error
+    //
+    const dir = MathJax.config.loader.require.resolve('mathjax/es5/non-existing-file');
+    //
+    //  (in case it ever exists, use its directory)
+    //
+    return path.dirname(convertWindows(dir));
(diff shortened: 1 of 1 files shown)


components/src/node-main/node-main.js
```diff
@@ -21,16 +21,28 @@ combineDefaults(MathJax.config.loader, 'provides', provides);
 
 MathJax.config.loader.paths.mathjax = (function () {
   //
-  // Try to locate the mathjax-full or mathjax package
-  // If neither is found, try to use the directory where this file is located
+  // Convert a windows path to a unix path (when needed)
+  //
+  const convertWindows = (name) => (path.win32 || !name.match(/^[a-z]:\\/i)) ? name : name.replace(/\\/g, '/');
```
zorkow, line 26: I believe the `path.win32` object exists on all platforms, so that will always be true.
Could that not be easier (and maybe more robustly) checked with 
```process.platform === 'win32'```?
For example,
```
process.platform === 'win32' ? path.win32.normalize(name) : name;
```

dpvc: > I believe the `path.win32` object exists on all platforms

Note in the web packed version.  That is the source of the issue.  Webpack apparently provides its own replacement for the `path` library and it is effetely only the posix implementation.  There is no `path.win32` in the web packed components.  This will affect users how install the `mathjax` rather than the `mathjax-full` npm package, as they will be using the web packed version (since the source versions are not available).  In order to test this, you have to make the components and `require('./es5/node-main')` rather than `require('mathjax-full')` (or temporarily change the 'main' entry in the `package.json` file.)

Similarly, the `path.normalize` approach doesn't work in the webpacked version (and `path.win32.normalize()` isn't there) since it only knows about posix directory separators.

I'll have to check whether `process.platform` is correct in webpack (I assume it will be, but the issues with `path` make me worried).

dpvc: Just to be clear, the `path.win32` teat was to determine if we are in the webpacked version or not.  Outside the web packed version, everything works properly in the original code.  It is the webpacked one that is being addressed, here.

zorkow: I did not know that. I always thought webpack works on node and has access to all its built-in libraries. The current version works, so we should leave it as is. I just thought we could short-cut.

dpvc: I was surprised, too.  I had wanted to use `path.win32.normalize` as well, but it wasn't there, and `path.normalize` didn't work.  So I was forced to use this hack.

PR #538, opened 2020-08-14
Don't remap stretchy characters (mathjax/MathJax#2497)

dpvc:
The PR fixes a problem where a stretchy delimiter that turns out to be rendered as a single character might be remapped to another character inadvertently.  Stretchy characters should not be remapped, which is handled here by moving the test for the stretchy character to outside the remapping.

Resolves the second part of mathjax/MathJax#2497.


dpvc:
Good call.  I've added the service routine.

PR #541, opened 2020-08-17
Better handling of root location and async loading in node-main

dpvc:
This PR improves the mechanism for getting the root directory for the version of `node-main` that is running.  It uses `eval()` to get access to the actual node version of the `path` library and the `__dirname` variable rather than the webpack versions of them, which was what was causing problems before.  This means we don't have to do all the hacks to handle Windows or use the error object to look up the path; we can use use `eval('__dirname')` and `path` as we would expect to.  This works since `node-main.js` is only for use in node, not sin the browser itself, where those values don't exist.

### DIFF
--- a/components/src/node-main/node-main.js
+++ b/components/src/node-main/node-main.js
@@ -1,4 +1,4 @@
-const path = require('path');
+const path = eval("require('path')");  // use actual node version, not webpack's version
 
 /*
  * Load the needed MathJax components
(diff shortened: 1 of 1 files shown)


zorkow (approved):
This is much nicer and cleaner. I've tested it and it works.

dpvc:
> This is much nicer and cleaner.

Yes, I thought so, too.  I think this eliminates all the hacks.

> I've tested it and it works.

Great, thanks!

PR #542, opened 2020-08-27
Update convert() to accept 'family' parameter (mathjax/MathJax#2512)

dpvc:
Update `convert()` to accept `family` parameter, and use it for the inherited font, when requested.  This fixes an error message produced when you pass the result of `getMetricsFor()` as the options for `MathJax.tex2svg()` and related functions.

Resolves issue mathjax/MathJax#2512.


zorkow (approved):
lgtm.

PR #543, opened 2020-08-27
Fix autoload to handle macro or \begin followed by spaces.  (mathjax/MathJax#2511)

dpvc:
This PR fixes a problem with the `autoload` extension when the auto-loaded macro or `\begin` is followed by spaces.

Resolves issue mathjax/MathJax#2511.


zorkow (approved):
lgtm.

PR #544, opened 2020-08-27
Convert mu to em in CD environment.  (mathjax/MathJax#2513)

dpvc:
This PR changes two uses of `mu` units for their corresponding `em` units, so that `mu` units don't end up in the internal MathML.


zorkow (approved):
lgtm.

PR #546, opened 2020-09-04
Fix issue with \textit, etc., not working with textmacros extension. (mathjax/MathJax#2514)

dpvc:
The `textmacros` extension resets the font to normal when `\text{}` is used (as it does in actual LaTeX), since you can change fonts internally within `\text{}` with `textmacros`.  (This differs from `\text{}` without `textmacros`, since the only way to set the font within `\text{}` previously was to set the font outside it, a conscious decision to differ from LaTeX to allow that functionality.)  But because `\textit{}` and the other similar macros were defined as macros that used `{\it\text{#1}}`, the change in behavior for `\text{}` caused these macros to fail with the `textmacros` extension.

This PR changes the implementations of `\textit{}` and the other similar macros to use javascript functions (rather than replacement macros) that get passed the needed `mathvariant` for initializing the font.  The `\text{}` macro retains its original behavior without `textmacros`, but has its new behavior with the extension, and now all the related ones now work with the extension as well.

Resolves issue mathjax/MathJax#2514


ts/input/tex/base/BaseMethods.ts
```diff
@@ -1003,10 +1003,11 @@ BaseMethods.BuildRel = function(parser: TexParser, name: string) {
  * @param {TexParser} parser The calling parser.
  * @param {string} name The macro name.
  * @param {string} style Box style.
+ * @param {srting} font the mathvariant to use
```
zorkow, line 1006: Capitalise.

zorkow (approved):
That change makes the implementation of the `\textXX` macros a lot cleaner.
Just one trivial change.

zorkow:
I think once 3.1.1. is out I should spend some time to extend our tests. In particular

1. Systematically add test cases for every single macro that we have. While most extension packages have tests for every macro, for the ones in Base and AMS I primarily constructed tests for all the cases in the original v2 methods only.

2. Run tests multiple time, adding other extensions successively so we can detect unwanted interactions automatically in the future.

3. Automate all that via some CI tool.

dpvc:
> I think once 3.1.1. is out I should spend some time to extend our tests.

That sounds good.

PR #547, opened 2020-09-04
Handle AsciiMath configuration properly, in particular, displaystyle.  (mathjax/MathJax#2520)

dpvc:
This PR fixes two issues with AsciiMath:

1.  The AsciiMath component configuration was not properly handled (it was being processed too late, after AsciiMath had already been configured, and the AsciiMath parameters were being flagged as having no default value (they are only defined in the legacy code, not the new Typescript code).

2.  The conversion from the legacy internal format was not properly handling boolean values, which caused the `displaystyle` configuration option to be always treated as false.

Resolves issue mathjax/MathJax#2520.


zorkow (approved):
I did not really understand the code very much. But I can see that it is now handled earlier. I assume early enough.

PR #549, opened 2020-09-06
 Fix issue with textmacros in SVG mode. (mathjax/MathJax#2524)

dpvc:
This PR fixes a problem where inferred mrows were being improperly included in the contexts of `text{}` when mathematics is also within the text.  This issue affected SVG output primarily.  The fix is to convert the inferred mrow into a real mrow.

Resolves issue mathjax/MathJax#2524.


zorkow (approved):
For some reason I could not reproduce the reported error, so I can only assume that is the right fix.
It looks exactly the same for me in both `develop` and `issue2524` branch (or when pulling `tex-svg-full` from the cdn for that matter).

PR #550, opened 2020-09-08
Only ignore space following alphabetic control sequences.  (mathjax/MathJax#2527)

dpvc:
In `textmacros`, the spacing counts, so we need to follow LaTeX rules poperly:  only skip a space after an alphabetic control sequence, not ones that are a single special character.

Resolves issue mathjax/MathJax#2527.


zorkow (approved):
lgtm.

PR #551, opened 2020-09-08
Add missing support for \ref and \eqref in textmacros.  (mathjax/MathJax#2528)

dpvc:
This PR adds missing support for `\ref` and `\eqref` in the `textmacros` extension.  The macros were mapped, but the method for them was missing, and the `TextParser` needed to use the `tags` object from the original `TexParser` in order to be able access the correct tagging data.

Resolves issue mathjax/MathJax#2528.


zorkow (approved):
lgtm.

PR #552, opened 2020-09-08
Add 'smart quote' version of acute and grave accents. (mathjax/MathJax#2526)

dpvc:
This PR adds two macros to `textmacros`: `\’` and `\‘` that mirror the `\'` and `` \` `` macros, but with "smart quote" characters (U+2019 and U+2018) rather than the usual ascii characters.  The is because some content-management systems do automatic substitution of smart quotes, and would convert `\'` to `\’`, perhaps without the user's knowledge.

Resolves issue mathjax/MathJax#2526.


zorkow (approved):
lgtm.

PR #554, opened 2020-09-14
Allow to disable font loading from an URL. (mathjax/MathJax#2539)

Mogztter:
resolves mathjax/MathJax#2539

### DIFF
--- a/ts/output/chtml/FontData.ts
+++ b/ts/output/chtml/FontData.ts
@@ -200,7 +200,9 @@ export class CHTMLFontData extends FontData<CHTMLCharOptions, CHTMLVariantData, 
     //
     //  Add fonts with proper URL
     //
-    this.addFontURLs(styles, CLASS.defaultFonts, this.options.fontURL);
+    if (this.options.fontURL !== false) {
+      this.addFontURLs(styles, CLASS.defaultFonts, this.options.fontURL);
+    }
     //
     //  Create styles needed for the delimiters
     //


Mogztter:
@dpvc I'm willing to implement an alternative solution if you think that my proposal is not acceptable. Or maybe there's another way to disable `@font-face` generation?

dpvc:
I know it has been a long time responding to this, and I apologize for that.  Things never seem to get any less busy these days.  One reason for the delay is that I have concerns about the PR, and wasn't sure how to voice them properly.

My main issue with this is that I am hesitant to provide a configuration option that would essentially break MathJax output unless you made a fairly sophisticated setup externally.  That is, unless you provide the CSS for the fonts yourself in some way, MathJax's output would be broken.  This seems like a dangerous outcome for a simple configuration setting, especially when the use-case is relatively limited.

I think there is a more appropriate way to handle this, which is to override the `addFontURLs()` by replacing the method in the existing one (in the next feature release of MathJax, you will be able to subclass the `FontData` object and tell the CHTML output jax to use your subclass, but that isn't available in the current release).  Since dealing with this use-case already requires a relatively sophisticated understanding of MathJax internals, this should not be that hard for those who need it.

For example, 

``` js
MathJax = {
  startup: {
    ready() {
      const {CHTMLFontData} = MathJax._.output.chtml.FontData;
      CHTMLFontData.prototype.addFontURLs = () => {};
    }
  }
}
```

would disable the generation of `@font-face` declarations.   A more sophisticated version would be

``` js
MathJax = {
  startup: {
    ready() {
      const {CHTMLFontData} = MathJax._.output.chtml.FontData;
      const addFontURLs = CHTMLFontData.prototype.addFontURLs;
      CHTMLFontData.prototype.addFontURLs = (styles, fonts, url) => {
        // add whatever CSS you need to the style obejct.
        //   The fonts variable is the original @font-face definitions 
        //   with %%URL%% where the URL shoud go which is in the url 
        //   argument.  You can change the fonts in any way that you 
        //   need, or ignore them entirely.
        //   See
        //
        //      https://github.com/mathjax/MathJax-src/blob/2dd53ce6c8af3c9cceba0baf014ea9b065130774/ts/output/chtml/FontData.ts#L217-L223
        //
        //   for the original code.
        //
        if (url !== false) {
          addFontURLs.call(this, styles, fonts, url);
        }
      }
    }
  }
}
```

where you use `addFontURLs()` to do whatever you need in order to set up the CSS the way you want it to be.  This keeps the CSS creation in the same place as MathJax's original code, and so there is not the disconnect between MathJax and your external CSS.

I'm sorry it has taken me so long to write this up for you.  It has been guilting me in my inbox for quite a while, and I know you have been waiting a long time.

dpvc:
PS, for example, the second configuration above could be used to replace the src for the fonts with data URLs.

Mogztter:
Hey @dpvc, no worries, you don't need to apologize. I'm extremely grateful for all the work you've done and the time you took to help me with a detailed explanation 🤗 


> I think there is a more appropriate way to handle this, which is to override the addFontURLs() by replacing the method in the existing one 

I didn't know I could do that and I completely agree that this is a better solution 👍🏻 

> (in the next feature release of MathJax, you will be able to subclass the FontData object and tell the CHTML output jax to use your subclass, but that isn't available in the current release)

Just to be clear, it's currently possible to override the `addFontURLs` by replacing the method but it's not yet possible to extend/subclass the FontData?

> This keeps the CSS creation in the same place as MathJax's original code, and so there is not the disconnect between MathJax and your external CSS.

That sounds good, I will give it a try, thanks again for your time 🙌🏻

dpvc:
> no worries, you don't need to apologize.

Thank you; you are very kind.  I have been feeling bad about this PR for some time, and am glad to have finally been able to give you some feedback.

> Just to be clear, it's currently possible to override the addFontURLs by replacing the method but it's not yet possible to extend/subclass the FontData?

It is possible to subclass it, but you can't currently tell the CHTML output jax to use the subclass (it always uses the `TeXFont` subclass of `FontData`).  So in the future, you will be able to do this the "right" way by subclassing rather than modifying the `FontData` prototype.  You can extend `FontData` by adding new methods to the prototype, if you need them in your updated `addFontURLs()`.  You can say the old method and call that from your new one in order to accomplish the equivalent of `super` calls, as illustrated in the second example above.

But again, in the next feature release, you will be able to use a proper subclass.

dpvc:
> I completely agree that this is a better solution

Given that, I'm going to close this PR, then.  Thanks for the willingness to contribute!

deflexable:
@dpvc does mathjax now support this feature, or is there any work around to this

dpvc:
> does mathjax now support this feature

Not directly.

> is there any work around to this

I gave two mechanisms that could be used to handle it in the comments above.  The second one allows you to configure `fontURL: false` in your MathJax configuration.

PR #558, opened 2020-10-16
Fixing SCRIPT mapping for 'I'

jfoxx77:
Fixes issue #2555:  \mathscr{I} renders incorrectly @mathjax /MathJax


dpvc:
Thanks for being willing to contribute to the MathJax project!  We appreciate it.

There are two issues, however:  first, the SVG output also needs to be adjusted (and you don't have the necessary data for that), and second, these files are generated by a script (not currently public), and so the fix really needs to be made in the script so that your edits won't be lost if the files are regenerated later for any reason.

I'll have to fix the script, which will take care of the SVG output, and we already have an `issue2555` branch for the fix (waiting on the SVG fix).  So I'm afraid I need to close this PR for now.  But thank you!

PR #561, opened 2020-11-03
Moves a11y options to enriched.

zorkow:
* Turns a11y options into an expandable element that is initially defined on `EnrichedDocument`.
* Ensures that there is a default for basic SRE parameters `domain`, `style`, `locale` set in enriched. As a consequence our node demos for speech enrichment will get Mathspeak as default speech output.

I am not certain whether the reference to `BaseDocument` in  [Line 250](https://github.com/mathjax/MathJax-src/compare/update_dep...a11y_options_for_enrich?expand=1#diff-6a4d02d0afbecccda42cdbd221dc69cf4c268bc08ba502ee9699b9801f24a3b2R250)
is the correct way to construct the expandable option.


ts/a11y/semantic-enrich.ts
```diff
@@ -245,7 +248,11 @@ export function EnrichedMathDocumentMixin<N, T, D, B extends MathDocumentConstru
         ...BaseDocument.OPTIONS.renderActions,
         enrich:       [STATE.ENRICHED],
         attachSpeech: [STATE.ATTACHSPEECH]
-      })
+      }),
+      a11y: {
```
dpvc, line 252: I think you want this to be `expandable()` as well.  What that does is prevents the error checking when options are given in the configuration, so that the options for the explorer will not cause an error if they are included in the configuration but the explorer isn't loaded yet.  In fact, we may need to add an expandable `a11y` option in one of the parent classes so that even if `semantic-enrich` is not loaded, the `a11y` options can be included in the configuration without error.

This does mean, however, that there will be no errors reported for misspelled options, for example.  it might be worth removing `expandable()` from the explorer options, since that is at the top level.  I think the MathDocument is recreated when explorer is loaded, and that would cause the error checking to occur at that point, at least.

dpvc:
> I am not certain whether the reference to BaseDocument in Line 250 is the correct way to construct the expandable option.

Yes, that is the correct way to do it.

zorkow:
I tested with the node demos and 
1. a11y options can now be passed to the enriched document, and 
2. incorrectly spelled options will still be flagged.

ts/a11y/explorer.ts
```diff
@@ -246,8 +246,7 @@ export function ExplorerMathDocumentMixin<B extends MathDocumentConstructor<HTML
         ...BaseDocument.OPTIONS.renderActions,
         explorable: [STATE.EXPLORER]
       }),
-      a11y: expandable({
-        ...BaseDocument.OPTIONS.a11y,
+      a11y: {
```
dpvc: The `locale` and `speechRules` properties are already in the `BaseDocument.OPTIONS.a11y` since they are in the `EnrichedMathDocument` document class created by the `semantic-enrich` extension, that is the `BaseDocument` here (via `HTMLDOCUMENT`). So I think you want to keep the `...BaseDocument.OPTIONS.a11y` that was old line 250 and not include those two properties here.

Also, the `a11y` property in the `OPTIONS` in `semantic-enrich` should also be made `expandable()` so that if enrichment is enabled before the explorer, you don't get errors about these other explorer options.

dpvc (changes requested):
I think there is still one more issue to deal with, here, since some of the a11y options are needed in the `semantic-enrich` extension, so need to be defined there, not here, and passed on.  That is one of the difficulties of using `a11y` for the options for several extensions.

ts/a11y/explorer.ts
zorkow: I see what you mean. But the more I think about it, the more I think all this should be done differently as to not conflate concerns. Here's an alternative proposal:
Let's have two set of options: `sre` and `a11y`.
* `sre` are defined expandable on both `EnrichedMathDocument` and `ExplorerMathDocument`. They are directly passed to SRE, which does its own sifting anyway, so no need for syntax checking or defaults (we might want to set `domain`, though). 
* `a11y` would be options only for the explorer, only defined on `ExplorerMathDocument`, non-expandable.

This should simplify the places where we call `setupEngine`. It would also allow us to make use of additional SRE features. E.g, at the moment enrichment can only use normal speech, but I could see a use case for enriching with SSML markup, which will need the `markup` parameter to be set. `enrichSpeech` would become an `sre.speech` option, no conflict with the `a11y.speech` counterpart. We also ensure that the SRE options in MathJax are the same as in SRE itself and thereby avoid confusion.

Does that make sense?

The only thing we need to be mindful about is the connection between `a11y.speechRules` and `sre.domain`, `sre.style`. But that should be doable.

zorkow: > `...`(we might want to set `domain`, though).

And `speech: 'none'` of course. I did not think that far when writing that line.

dpvc: I think the separation you describe is a good one.  Can you make the needed changes, or do you need me to do anything?

It does make a possible breaking change to the options, but I doubt many are setting these in their pages, so I'm not too worried about that.

zorkow: > I think the separation you describe is a good one. Can you make the needed changes, or do you need me to do anything?
>

I'll do that.

> It does make a possible breaking change to the options, but I doubt many are setting these in their pages, so I'm not too worried about that.

That is true. We can keep `a11y` expandable for this patch release and then make the breaking change for the next point release.

dpvc: OK, sounds good.

zorkow:
Rewrote as discussed.
* Splits options into fixed `a11y` and expandable `sre`.
* Renames `speech` and `braille` menu variables to avoid potential clashes with names in `sre` options. (I know, I thought that would not happen, but does due to the current registration of the explorers in `attachExplorers`.
* Handling of `domain/style` vs `speechRules` options is still a bit awkward.

__Switching locales currently does not work fault free with SRE v3.2.0-beta.1__ due to rule sets now being loaded dynamically. This will be addressed in a subsequent PR.

ts/ui/menu/Menu.ts
```diff
@@ -473,9 +473,9 @@ export class Menu {
         this.submenu('Accessibility', 'Accessibility', [
           this.checkbox('Activate', 'Activate', 'explorer'),
           this.submenu('Speech', 'Speech', [
-            this.checkbox('Speech', 'Speech Output', 'speech'),
-            this.checkbox('Subtitles', 'Speech Subtitles', 'subtitles'),
-            this.checkbox('Braille', 'Braille Output', 'braille'),
+            this.checkbox('Speech', 'Speech Output', 'speechExp'),
+            this.checkbox('Subtitles', 'Speech Subtities', 'subtitles'),
```
dpvc, line 477: You've put back the typo that was fixed in an earlier PR.

ts/a11y/explorer.ts
dpvc: OK, this looks pretty good.

Question:  Are there other SRE options that can appear in the `sre` block?  If not, then you don't need an `expandable()` for that.  If there are, could we give their values here, or are they too esoteric (or too many)?  As it stands, the `expandable()` will mean that no errors will be reported for misspelled or undefined options.  You can add to the options later as you do in `setA11yOptions()` even if they aren't in the initial `OPTIONS`.  You only need `expandable()` if the user can include arbitrary options, or if more options will be added by another handler later.  I thought the reason to separate `sre` from `a11y` was to avoid that situation.

ts/a11y/explorer.ts
```diff
@@ -240,17 +240,22 @@ export function ExplorerMathDocumentMixin<B extends MathDocumentConstructor<HTML
...
+        domain: 'mathspeak',               // speech rules domain
+        style: 'default',                  // speech rules style
+        locale: 'en'                       // switch the locale
+      }),
       a11y: {
         align: 'top',                      // placement of magnified expression
         backgroundColor: 'Blue',           // color for background of selected sub-expression
         backgroundOpacity: 20,             // opacity for background of selected sub-expression
-        braille: false,                    // switch on Braille output
+        brailleExp: false,                 // switch on Braille output
```
dpvc, line 258: Note that these changes in configuration options mean that some people's configurations will no longer be valid, and will begin to report errors when this update is made.  Since that is a breaking change, it should require a different version number than 3.1.3.  Of course, I doubt that many people have such options in their pages, but it might be nice to make this more backward compatible.  One possibility would be to use the `constructor()` below to do 

``` javascript
const a11y = (args[2] || {}).a11y;
if (a11y) {
  for (const option of ['braille', 'speech']) {
     if (a11y.hasOwnProperty(option)) {
       a11y[option + 'Exp'] = a11y[option];
       delete a11y[option];
     }
  }
}
```

just before the `super(...args)` so that old options will be converted to the new ones for now.  We can remove this in a future release.

ts/a11y/explorer.ts
```diff
@@ -260,11 +265,10 @@ export function ExplorerMathDocumentMixin<B extends MathDocumentConstructor<HTML
         infoRole: false,                   // show semantic role on mouse hovering
         infoType: false,                   // show semantic type on mouse hovering
         keyMagnifier: false,               // switch on magnification via key exploration
-        locale: 'en',                      // switch the locale
```
dpvc, line 263: Something similar to the above, but setting `sre.locale` would be good, too.

ts/a11y/explorer.ts
```diff
@@ -240,17 +240,22 @@ export function ExplorerMathDocumentMixin<B extends MathDocumentConstructor<HTML
...
       renderActions: expandable({
         ...BaseDocument.OPTIONS.renderActions,
         explorable: [STATE.EXPLORER]
       }),
+      sre: expandable({
+        speech: 'shallow',                 // overrides option in EnrichedMathDocument
+        domain: 'mathspeak',               // speech rules domain
+        style: 'default',                  // speech rules style
+        locale: 'en'                       // switch the locale
+      }),
```
dpvc, line 253: Since the base document will be an `EnrichDocument`, you should be able to do

``` javascript
      sre: expandable({
        ...BaseDocument.OPTIONS.sre,
        speech: 'shallow'
      }),
```

in order to avoid having to repeat all the options that stay the same.  That way, if any are ever added to the semantic-enrich extension, they will be included here automatically.

ts/a11y/semantic-enrich.ts
```diff
@@ -240,12 +240,17 @@ export function EnrichedMathDocumentMixin<N, T, D, B extends MathDocumentConstru
...
         enrich:       [STATE.ENRICHED],
         attachSpeech: [STATE.ATTACHSPEECH]
-      })
+      }),
+      sre: expandable({
+        speech: 'none',                    // by default no speech is included
+        domain: 'mathspeak',               // speech rules domain
+        style: 'default',                  // speech rules style
+        locale: 'en'                       // switch the locale
+      }),
```
dpvc, line 253: See the discussion in `explorer.ts` about whether this should be expandable or not.

dpvc (changes requested):
Still a couple of questions that I have, and an issue for backward compatibility.

ts/ui/menu/Menu.ts
zorkow: Well spotted! How did that happen?

ts/a11y/explorer.ts
zorkow: You are right, we should not introduce a breaking change. It is only necessary for the `speech` option anyway, so we can get away with a simple `if`.

ts/a11y/explorer.ts
zorkow: I'd rather keep it `expandable` for a few reasons:

* There are about a dozen options (despite the imminent removal of the `cache` option). Not all are exposed and some are rather esoteric, e.g., `pitch` and `rate` set baselines for systems using `ssml` or `acss` markup, like, mathlive, emacspeak, chromevox. They still might make sense server-side.
* Some options are simply ignored if not set, but can influence behaviour even if set to `0`, `null`, or `false`.
* We don't have to keep options in sync: E.g., if I were to use MathJax-full with a newer version of SRE and pass on a new option I can do that without MathJax complaining that there is no default.

dpvc: OK, just was checking to make sure it was OK that unknown options were not being flagged.

ts/ui/menu/Menu.ts
dpvc: Probably in the merge of develop into this branch.  Sometimes It may think this file is more current and so took its changes over the ones from develop.  Don't know for sure.

ts/a11y/explorer.ts
dpvc: OK, sounds good.  (Is that because Braille is new in this version?)

zorkow: No, the `braille` option was always the switch for the Braille explorer. SRE does not have a `braille` option, but a `modality` option that takes `braille` as a value. I just wanted to make it uniform to `speechExp`.

ts/a11y/explorer.ts
zorkow: `locale` is not explicitly set in `sre` whenever `a11y` value changes. I.e., `a11y` value dominates `sre` value.

ts/a11y/explorer.ts
zorkow: Done.

ts/a11y/explorer.ts
zorkow: This was entirely unnecessary as it did not solve the actual problem anyway. That problem is related to the order of attached key explorers: the last stops event propagation. However, detaching and re-attaching explorers does not always keep the order of the events on the node in sync. So sometimes events are stopped too early.
A subsequent PR will implement the following:
* Explorers attach themselves only once by adding events on update of the document.
* They are only detached "virtually", i.e., just don't run.
* The last "virtually" attached explorer is `stoppable`.

zorkow:
I think that should now be pretty much ready to go. But it needs three follow up PRs: 
1. Introduce "virtual" detachment for explorers (see comments).
2. Take care of menu entries for SRE's new rule loading mechanism (see comments).
3. Restrict options for explorers that can be changed. E.g., the braille explorer should never update it's `locale` (as we currently only have nemeth anyway) nor respond to updates for `domain`.

ts/ui/menu/Menu.ts
```diff
@@ -474,7 +474,7 @@ export class Menu {
           this.checkbox('Activate', 'Activate', 'explorer'),
           this.submenu('Speech', 'Speech', [
             this.checkbox('Speech', 'Speech Output', 'speech'),
-            this.checkbox('Subtitles', 'Speech Subtitles', 'subtitles'),
+            this.checkbox('Subtitles', 'Speech Subtities', 'subtitles'),
```
dpvc, line 477: This typo still needs to be corrected.

ts/a11y/semantic-enrich.ts
```diff
@@ -240,12 +240,17 @@ export function EnrichedMathDocumentMixin<N, T, D, B extends MathDocumentConstru
     public static OPTIONS: OptionList = {
       ...BaseDocument.OPTIONS,
       enableEnrichment: true,
-      enrichSpeech: 'none',                   // or 'shallow', or 'deep'
```
dpvc, line 243: This the `enrichSpeech` option has been moved to the `sre` block, but `enrichSpeech` is [documented](https://docs.mathjax.org/en/latest/options/accessibility.html?highlight=enrichSpeech#the-configuration-block), it would be a breaking change to remove it.  For this version, it would be good to transfer it to `sre` if it is set in the user options.  Something like

``` javascript
      //
      //  For now, move enrichSpeech to sre.speech (will remove enrichSpeech in a later version)
      //
      const options = args[2] || {};
      if (options.enrichSpeech) {
        if (!options.sre) {
          options.sre = {};
        }
        options.sre.speech = options.enrichSpeech;
        delete options.enrichSpeech;
      }
```

before the `super(..args)` should do it.

dpvc (changes requested):
Two more small items.

dpvc:
@zorkow, (ping) there is still the issue with the type in the subtitles, and I made a comment about transferring an old option to the new location above.  These still need to be addressed before merging (sorry).

dpvc:
I noticed while updating the documentation that `locale` is still in the `a11y` block in the explorer.  Now that it is in `sre` in semantic-enrich, is that still correct?

zorkow:
> I noticed while updating the documentation that `locale` is still in the `a11y` block in the explorer. Now that it is in `sre` in semantic-enrich, is that still correct?

No, we can move this. I really wanted to get back to this, but then more SRE issues kept popping up. 
Really sorry about all the delays. I'll finish all this off next week.

dpvc:
> Really sorry about all the delays. I'll finish all this off next week.

No problem.  I know how to goes.  I had expected to be finished with my part several weeks ago!

zorkow:
Takes care of the remaining issues. PTAL.

dpvc:
It looks like `locale` is still in the `a11y` list

https://github.com/mathjax/MathJax-src/blob/effceb60dc6136150781393a418803a49b871fc3/ts/a11y/explorer.ts#L266

Should that be removed, since it is now in the `sre` block?

dpvc:
Or is that for a later version?

dpvc:
And do we need more backward-compatibility code to move the value from `a11y` to `sre`, if it is specified?

zorkow:
I think I got a bit confused after merging `develop` that added the locale handing for the cs preference menu.
And you are right: We need more backward compatibility code for that. But also we need to ensure that setup is called in the enrichment code so that the given locale is actually loaded. Only doing this on change of speech value is no longer enough.

I have another (hopefully final look) at it tomorrow after my lecture. Sorry for trying to rush it today.

zorkow:
* Adds backward compatibility code in
     *  `semantic-enrich` for `enrchSpeech` option
     *  `explorer` for `a11y` options `speechRules` and `locale`
* Removes all old options
* The two auxiliary options are clearly marked for removal.
     *  As both have to happen before the `super` call there is unfortunately no inheritance.

This should be the last revision (but I've said that before...) In hindsight it would have probably been easier to do this change as breaking change for 3.2.

PTAL.

zorkow:
[1a14eb5](https://github.com/mathjax/MathJax-src/pull/561/commits/1a14eb5) pushes the changes that did not make it last night due to some rebasing issue.
[e5ec488](https://github.com/mathjax/MathJax-src/pull/561/commits/e5ec488878a53253dcd67692dedf2571b6d3771b) is the final change we discussed today.

dpvc:
Since I can't think of anything more to hold up this PR, I guess I'll have to merge it!  👍 🥇

dpvc (approved):

PR #562, opened 2020-11-12
Fix prime display issue for AsciiMath input, svg output

xieyubo:
For AsciiMath notation:
```
h'x'''(x) = g'(h(x)')
```
Before fix:

image

After fix:

image


dpvc:
Thanks for interest in MathJax, and willingness to contribute.  The AsciiMath processor is not something that we maintain ourselves.  It is maintained at the https://github.com/asciimath/asciimathml site, and we incorporate their parser into MathJax pretty much verbatim.  So changes in AsciiMath must be submitted to their site, not here (or they will be lost during the next update).

For that reason, I am going to close this PR.

PR #563, opened 2020-12-01
Fix webpack.common.js to handle windows paths correctly (mathjax/MathJax-demos-web#30)

dpvc:
The webpack configuration didn't handle Windows path separators properly, which caused the building of components to fail on windows.  This PR resolves the issues by using `path.sep` and fixing `quoteRE()` to properly quote the special characters like `\`.

Resolves issue mathjax/MathJax-demos-web#30.


zorkow (approved):
lgtm.

PR #564, opened 2020-12-03
Make mspace elements have TEXCLASS.NONE (mathjax/MathJax#2576)

dpvc:
Currently, `mspace` elements have `TEXCLASS.ORD`, but this causes spacing issues in some instances, for example `A=\!=B`.  This PR sets the `texClass` to `TEXCLASS.NONE` and overrides `setTeXclass()` so that the spacing will be based on the elements surrounding the space, as it is in actual TeX.

Resolves issue mathjax/MathJax#2576.


zorkow (approved):
lgtm.

PR #566, opened 2020-12-09
Fix autoloading when macro name is followed by a space.  (mathjax/MathJax#2580)

dpvc:
If an autoloading macro is followed by a space, the space is lost when the macro name is reprocessed, and so the macro can become incorrect if it is followed by other tex.  For example, `\ce C` ends up as `\ceC` after the autoload.  This PR replaces the missing space (that was removed during the initial parsing for the macro name, and will be removed again when it is re-parsed).

Resolves issue mathjax/MathJax#2580.


zorkow (approved):
lgtm.

PR #571, opened 2020-12-30
Make sure strechy character extenders have non-zero height in CHTML for Safari (mathjax/MathJax#2598)

dpvc:
It turns out that the zero height boxes used for the characters in the stretchy horizontal extenders in CHTML are not placed properly in Safari (but are in Firefox and Chrome).  If the height is non-zero, they they are, so this PR forces a non-zero height.

Resolves issue mathjax/MathJax#2598.


zorkow (approved):
lgtm.

PR #575, opened 2021-01-18
Update LiteAdaptor to handle full-width characters better. (mathjax/MathJax-demos-node#29)

dpvc:
This PR improves the LiteAdaptor's handling of full-width characters.  There is now a separate width used for CJK characters from normal characters, and these are now customizable via options to the adaptor itself.

### DIFF
--- a/ts/adaptors/liteAdaptor.ts
+++ b/ts/adaptors/liteAdaptor.ts
@@ -64,11 +64,11 @@ export class LiteAdaptor extends AbstractDOMAdaptor<LiteElement, LiteText, LiteD
     '\uFE10-\uFE19', // Vertical Forms
     '\uFE30-\uFE6B', // CJK Compatibility Forms ... Small Form Variants
     '\uFF01-\uFF60\uFFE0-\uFFE6', // Halfwidth and Fullwidth Forms
-    '\u1B000-\u1B001', // Kana Supplement
-    '\u1F200-\u1F251', // Enclosed Ideographic Supplement
-    '\u20000-\u3FFFD', // CJK Unified Ideographs Extension B ... Tertiary Ideographic Plane
+    '\u{1B000}-\u{1B001}', // Kana Supplement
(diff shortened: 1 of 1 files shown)


ts/adaptors/liteAdaptor.ts
```diff
@@ -568,7 +594,12 @@ export class LiteAdaptor extends AbstractDOMAdaptor<LiteElement, LiteText, LiteD
    */
   public nodeSize(node: LiteElement, _em: number = 1, _local: boolean = null) {
     const text = this.textContent(node);
-    return [.6 * text.length, 0] as [number, number];
+    const non = text.replace(LiteAdaptor.cjkPattern, '').split('').length; // # of non-CJK chars
```
zorkow, line 597: Counting with `length` could give the wrong number of characters if there are other characters from an upper Unicode plane in the string.

ts/adaptors/liteAdaptor.ts
```diff
@@ -568,7 +594,12 @@ export class LiteAdaptor extends AbstractDOMAdaptor<LiteElement, LiteText, LiteD
...
+    const CJK = text.split('').length - non;                               // # of cjk chars
+    return [
+      CJK * this.options.cjkCharWidth + non * this.options.unknownCharWidth,
```
zorkow, line 600: ditto.

zorkow (commented):
One concern about counting characters using `length` of the string.

ts/adaptors/liteAdaptor.ts
dpvc: Right.  I thought it was unicode safe, but I was wrong.  I should have used `Array.from()`, which handles unicode character properly.  I will make another commit for it.

dpvc:
I have made the correction, and also fixed a problem with the Plane 1 characters in the regular expression for CJK characters.

zorkow (approved):
lgtm.

PR #576, opened 2021-01-18
Add support for setting the unknown character family for output jax

dpvc:
This PR provides a new output jax parameter that specifies the font family to use for unknown characters (it defaults to `serif` as int eh past, but makes it easier to change the default to `monospace` when the LiteAdaptor is used.

This required moving support for `FontData` options from `chtml/FontData.ts` into `common/FontData.ts`, and making sure all the constructors passed the options on properly.


zorkow (approved):
lgtm.

PR #577, opened 2021-01-19
Don't let height of final SVG be 0 (mathjax/MathJax#2605)

dpvc:
SVG elements are not displayed if their height (or width) is 0.  We already had a test for 0 width, but not for zero height.  So `\smash{X}` would no display properly.  This PR resolves that issue.


zorkow (approved):
lgtm.

PR #579, opened 2021-01-26
Show all modules during packing

dpvc:
Not sure why this was changed, but I used the full list to check that everything is packing from the proper locations.  If that is a problem, perhaps we need to make a command-line option to control it.


zorkow (approved):
lgtm.

PR #580, opened 2021-01-26
Make sure bevelled fraction is symmertically placed (needed for STIX2 fonts)

dpvc:
This PR makes sure the `/` used for bevelled fractions is centered on the math axis.  This didn't matter with the MathJax TeX fonts, whose glyphs are centered already, but for STIX2, the slash isn't centered, so this makes a difference.


zorkow (approved):
lgtm.

PR #582, opened 2021-01-27
Updates to handle smallmatrix properly

dpvc:
This PR implements the proper table layout for the `smallmatrix` and `subarray` environments.  These use the `useHeight` property that is supposed to to change how the matrix is laid out (by not adding extra height and depth to make sure rows are full height, as usual).  That feature was not implemented, so this PR adds support for `useHeight`.

These environments are typeset in script style, which had been handled by adding an `mstyle` around the table with `scriptlevel="1"`, but this had the side-effect of making the math axis be the script-size math axis (too low).  So the script level is now set through an internal property instead (which causes the table to scale down its children).  That required handling the relative size change of children in the bounding box computations.

The MathML serializer was modified to expose the small matrix internal properties as data attributes, and the MathML input to read those and restore the internal properties.

Finally, the `mrow` stretchy operator heights didn't take scaled children into proper account when figuring the heights, so there is a fix there as well.  There should probably be a `getChildBBox()` function that computes a properly scaled version of the BBox, since that ends up being a common issue that has caused problems before.  (That will also be needed in order to handle borders, margins, and padding CSS, which v2 does, but v3 doesn't.)  I will make that a separate PR, as it will touch on lots of files.

### DIFF
--- a/ts/core/MmlTree/MmlNodes/mtable.ts
+++ b/ts/core/MmlTree/MmlNodes/mtable.ts
@@ -61,7 +61,7 @@ export class MmlMtable extends AbstractMmlNode {
    * Extra properties for this node
    */
   public properties = {
-    useHeight: 1
+    useHeight: true
   };
 
   /**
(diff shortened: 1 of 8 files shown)


dpvc:
Note, this will affect some of the tests, as the MathML is modified for `smallmatrix` and `subarray` environments.

zorkow (approved):
Looks good.
There's not effect on the current tests as this does not change the JSON output structure.

PR #584, opened 2021-01-31
Use the correct parser for noundefined (mathjax/MathJax#2622)

dpvc:
This PR corrects the problem mentioned in mathjax/MathJax#2622:  that any undefined macros used in text mode when textmacros is in effect will end up being placed before the text material, regardless of where the undefined macro appears.


zorkow (approved):
lgtm.

PR #585, opened 2021-01-31
Handle inferred mrows properly (mathjax/MathJax#2624)

dpvc:
This PR fixes issue mathjax/MathJax#2624, which is caused by an inferred mrow being included in the internal MathML.


zorkow (approved):
lgtm.

dpvc:
I'm looking at making a more general solution (since there are other places this same issue occurs), so I'm going to put this one on hold for the moment.

dpvc:
It turns out that #601 resolves this (and other similar issues like mathjax/MathJax#2554) in a more general and straight-forward way, so this PR is no longer needed.  I'm closing it for now, but will re-open if #601 is not accepted.

PR #586, opened 2021-01-31
Use mrow rather than inferredMrow if text mode produces more than one. (mathjax/MathJax#2577)

dpvc:
This PR prevents an unwanted inferredMrow from being used for the content of text-mode material that produces more than one element.  This prevents overlaps in the rendering in SVG.

Resolves issue mathjax/MathJax#2577.


zorkow (approved):
lgtm.

PR #588, opened 2021-02-03
fix: don't override (SVG) anchor styles

pkra:
Matches CHTML output.

### DIFF
--- a/ts/output/svg.ts
+++ b/ts/output/svg.ts
@@ -74,9 +74,6 @@ CommonOutputJax<N, T, D, SVGWrapper<N, T, D>, SVGWrapperFactory<N, T, D>, SVGFon
     },
     'mjx-container[jax="SVG"] > svg': {
       overflow: 'visible'
-    },
-    'mjx-container[jax="SVG"] > svg a': {
-      fill: 'blue', stroke: 'blue'
     }
   };
 


pkra:
Background:

https://github.com/mathjax/MathJax-src/blob/fbd71f3f7ca06dc71304e8a2c03f13be233ccac5/ts/output/svg.ts#L235-L239

sets g's up to inherit currentColor. Since one can assume that links are styled (via UA or author stylesheet), there's really no reason to change it. This matches the CSS for the CHTML output which also doesn't override link color.

dpvc:
> Since one can assume that links are styled (via UA or author stylesheet), there's really no reason to change it.

Unfortunately, the assumption isn't true, in general.  It is for Firefox, but for Chrome and Safari (and I assume blink-based Edge), it is not.  Without these styles, the text of the link is black.  It may be possible to find another set of CSS that provides for system-colored links, but as of now, removing this is a problem for Chrome and Safari.

pkra:
> Unfortunately, the assumption isn't true, in general. It is for Firefox, but for Chrome and Safari (and I assume blink-based Edge), it is not. Without these styles, the text of the link is black. It may be possible to find another set of CSS that provides for system-colored links, but as of now, removing this is a problem for Chrome and Safari.

I disagree in so far as IIRC this is all intentional UA behavior. So users are seeing what the authors / their UAs want. Why change it? (Technically, this also extends to underline, focus and hover styles for in non-svg anchors which MathJax does not copy either.)

Anyway, the practical problem I see is that the current styles break page designs. E.g., on mathjax.org, there _is_ styling and it doesn't show because the chosen selector is less specific than the one used by MathJax's. For me, that's a bug in MathJax, not the site's CSS. But I can see how one might think it's the other way around.

dpvc:
> So users are seeing what the authors / their UAs want. Why change it?

Because the CSS to style HTML `<a>` elements doesn't affect SVG `<a>` elements except in Firefox.  You cite MathJx.org, but that uses `a {color: #008200}` to style the anchors, and that won't affect the anchors in SVG for non-Firefox browsers, so even if we leave this out, you don't get the inheritance from the surrounding that you are looking for.  I don't see a way to get WebKit to inherit the color property as the fill and stroke colors (without using CSS variables to set both).

While it may be appropriate for general SVG to not have the anchor link color affect the SVG elements in the `<a>`, for mathematics, I think most authors would expect for the links to be identified by a color.  I can't really get the page-specified link color (and other styling) very readily (thought it could probably be obtained with `getComputedStyle()` at the cost of a page reflow), so thought it was better to at least provide *SOME* styling.  Personally, I doubt that the "no indication of a link" is an intentional author choice, and that most will not have thought about setting SVG anchor styles at all.  Page authors who do are free to override the MathJax styling with their own, and yes, they would have to set a MathJax-specific style selector to do that, so would need to be aware of that.  They would have to make specific settings for SVG links already.

The best I can offer is better documentation about needing to do this, but simply styling HTML anchors will not affect SVG anchors for most browsers.   

> there *is* styling and it doesn't show because the chosen selector is less specific than the one used by MathJax's

But that styling doesn't affect SVG links, except in Firefox.  So there is *no* styling from the page for the SVG anchors, other than the default "no styling at all".  So the MathJax styling "breaks the page styling" only in that it overrides "no styling", as it does not override the green styling, which doesn't apply to SVG links, in general.

> this also extends to underline, focus and hover styles which MathJax does not copy from non-svg anchor styles either.

Indeed, that is correct.  These are not copied to SVG anchors in general, as the styling of HTML anchors doesn't apply to the contents of SVG anchors, except in Firefox.  And things like underline or other text decoration don't apply to paths.

MathJax doesn't honor most CSS styling.  In CHTML output, it can have an effect, but if it changes the size (like borders and padding), MathJax won't know about it (and won't modify the layout to accommodate that), unless it is explicit in the style attribute of the MathML elements, and even then, MathJax only understands a few CSS attributes.  For SVG, most won't have any effect.  This is certainly true of CSS controlling the anchors.

I'm sorry that we can't do better with that, but not everything can be automatic.  Since you *can* use CSS to override the MathJax styling, if it bothers a page author, they can certainly add CSS To resolve it.  Styling MathJax output does require additional CSS, in general.

PR #589, opened 2021-02-09
Make sure hitbox is translated along with the hit-able element. (mathjax/MathJax#2530)

dpvc:
This PR resolves the issue with hit-boxes for the SVG output when an element in the middle of an `mrow` has an `href` attribute.  The hotbox was not placed properly, so this patch causes the box to be translated properly.

(I had fixed this some time ago, but apparently didn't push the branch or make the PR).

Resolves issue mathjax/MathJax#2530.


zorkow (approved):
Tested and works.

PR #591, opened 2021-02-18
Fix pack and makeAll to display the modules

dpvc:
OK, I see now why you changed the `--display-modules` option, since it has been removed in web pack v5.  This PR adjusts the `pack` and `makeAll` commands to display the modules again using information from `webpack --json` (which was the only way I could find to get the information reliably).  Because the data is so large, I had to change from `execSync` to `spawn` and use promises to handle the IO.


zorkow (commented):
lgtm.

zorkow (approved):

PR #592, opened 2021-02-23
Improve build tools to make it earlier create custom components

dpvc:
This PR adjusts the component building tools to make it easier to create custom components (like tex extensions) outside of the `MathJax-src/components` directory.  The main changes are in the `webpack.common.js` file, and allow for better control over the libraries that are linked.  It also now supports imports like

```
import {MathItem} from 'mathjax-full/js/core/MathItem';
```

where `mathjax-full` will be found in `node_modules`, and the `.js` is omitted.  (Our v3-lab doesn't handle that, but node and webpack do, so it is useful to allow it.)

Most of the other changes are to make the output easier to read (more directory prefix collapsing, and separate indicators for MathJax's components and js code as compared to local components and local js code).


zorkow (approved):
lgtm.

PR #593, opened 2021-02-24
Fixes table walking by setting modifier key.

zorkow:
Pass the shift key to the table walker as modifier. This allows table walking similar to v2.


dpvc:
Looks good.

PR #595, opened 2021-02-24
numcases: export Configuration.create

pkra:


pkra:
Maybe it should be capitalized (NumcasesConfiguration) but empheq's export also starts lowercase while usually these kinds of exports seem to start with an upper case letter.

dpvc:
OOPS, sorry, left that out.  I'll upper case both of them as well in a separate commit.

PR #601, opened 2021-02-25
Handle appending of inferred rows in MmlNode.appendChild() (mathjax/MathJax#2554)

dpvc:
Inferred mrows are used to hold multiple MathML nodes in situations where a single node is expected (e.g., they act as MathML fragments).  For example, the TexParser's `ParseArg()` function can return an inferred mrow.  These should not appear as children of normal MathML elements, and so historically the code had to be careful to check the return value of such functions in order to handle inferred mrows as a special case.  While this was done in many places, some situations still slipped through.  An unexpected inferred mrow did not affect CHTML output, but it does cause problems for SVG output, with mathjax/MathJax#2554, mathjax/MathJax#2624, and mathjax/MathJax#2577 being examples.

Rather than try to patch each new occurrence as it appears, this PR resolves it at the lowest lever:  the MmlNode's `appendChild()` function.  In the event that an inferred mrow is appended to an element, the mrow's children are appended instead.  This catches the problem in `setChildren()` as well, since that uses `appendChild()` internally.

This PR removes some of the special-casing that was done in NodeFactory and ParseUtil, since appending the children now handles the inferred mrows directly.  (There is one difference, which is that NodeFactory would turn the inferred mrows into actual mrows, but that should not be necessary.  It will mean some tests will probably have to be updated, though.)

This PR also moves the `texclass` property to below the static properties to make `tslint` happy.

Resolves issues mathjax/MathJax#2554 and mathjax/MathJax#2624.

Note that this means that PR #585 is no longer needed, as this PR is a better solution.

### DIFF
--- a/ts/core/MmlTree/MmlNode.ts
+++ b/ts/core/MmlTree/MmlNode.ts
@@ -448,9 +448,7 @@ export abstract class AbstractMmlNode extends AbstractNode implements MmlNode {
       //  (just add its children).
       //
       if (this.arity === Infinity) {
-        for (const node of child.childNodes) {
-          super.appendChild(node);
-        }
+        child.childNodes.forEach((node) => super.appendChild(node));
         return child;
       }
       //
--- a/ts/core/MmlTree/MmlNode.ts
+++ b/ts/core/MmlTree/MmlNode.ts
@@ -443,10 +443,26 @@ export abstract class AbstractMmlNode extends AbstractNode implements MmlNode {
       return child;
     }
     if (child.isInferred) {
-      for (const node of child.childNodes) {
-        super.appendChild(node);
+      //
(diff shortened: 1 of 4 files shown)


ts/core/MmlTree/MmlNode.ts
```diff
@@ -436,6 +442,28 @@ export abstract class AbstractMmlNode extends AbstractNode implements MmlNode {
...
     }
+    if (child.isInferred) {
+      //
+      //  If we can have arbitrary children, remove the inferred mrow
+      //  (just add its children).
+      //
+      if (this.arity === Infinity) {
+        for (const node of child.childNodes) {
+          super.appendChild(node);
+        }
```
zorkow, line 453: This could be done with a `forEach`.

zorkow (approved):
Only one minor thing.

PR #603, opened 2021-03-01
Use TeX algorithm for stretchy accents

dpvc:
This PR adjusts the algorithm used to stretch accents like those used in `\widehat` to better match the one used by TeX.  Prior to this, the accents used the horizontal stretchy character algorithm, which takes the smallest version that is larger than the needed size, but TeX's algorithm for accents uses the largest that is smaller than the width of the base.

Here, we mark math accents with a new `mathaccent` property, and when one is stretched, it uses the largest version that is smaller than the base.  We use the fudge-factors (`delimitershortfall` and `delimiterfactor`) as for other stretchy characters because we don't really have the proper italic corrections, and so the sizes turn out to not produce the desired result for the MathJax TeX fonts.  This fudge-factor allows the algorithm to work appropriately for both the MathJax fonts and the STIX2 fonts.


dpvc:
I will be replacing this PR with a new one, so don't review.

dpvc:
Closing in favor of #618.

PR #604, opened 2021-03-02
Handle variant forms for pseudo-scripts better.  (mathjax/MathJax#2519)

dpvc:
This PR improves the handling of pseudo scripts so that both the super-script version (using `msup`) and the non-superscript versions will render properly (with a font that has the proper glyphs available, like STIX2).  In the past, MathJax used the larger sized primes in the `normal` mathvariant, and this lead to incorrect sizes being used when the primes were not in `msup`.  This PR now expects the regular size to be in the `normal` variant and the larger size in the `-tex-variant` font, and the `mo` element sets the correct variant for the position it is used in.  It also remaps common characters that might be used for primes to their appropriate prime character (e.g., `'` maps to U+2032 (prime)).

That means `<msup><mi>x</mi><mo>'</mo></msup>`, `<mi>x</mi><mo>'</mo>` and `<mtext>x&#x2032</mtext>` will all render properly (when the font contains the right characters, which the MathJax TeX fonts don't, but the STIX2 fonts do).

This is intended as a step toward making MathJax work better with STIX2, as well as setting the stage for modifications to the MathJax fonts later this summer.


zorkow (changes requested):
I remembered the discussion we had on that:
We wanted to handle pseudoprimes in `core` and giving the resulting node a `pseudoprime` property instead.
Firstly, we could expose the property (e.g., for SRE). And secondly, it would not need to be handled explicitly in the `Wrapper`.

dpvc:
Right you are!  Thanks for coming up with it again.  It had totally slipped my mind.

dpvc:
OK, I've moved the pseudo-script and prime checking to the `MmlMo` class and marked them with properties.  The pseudo-prime is exposed on the serialized MathML output for your convenience in SRE.

ts/core/MmlTree/MmlNodes/mo.ts
```diff
@@ -384,4 +440,35 @@ export class MmlMo extends AbstractMmlTokenNode {
...
+
+  /**
+   * Determine of the mo consists of primes, and remap them if so.
+   *
+   * @param {string} mo   The test of the mo element
+   */
+  protected checkPrimes(mo: string) {
+    const PRIMES = (this.constructor as typeof MmlMo).primes;
+    const REMAP = (this.constructor as typeof MmlMo).remapPrimes;
+    if (!mo.match(PRIMES)) return;
```
zorkow, line 469: Swap lines 468 and 469.

ts/output/common/Wrappers/mo.ts
```diff
@@ -361,7 +367,10 @@ export function CommonMoMixin<T extends WrapperConstructor>(Base: T): MoConstruc
      * @override
      */
     public remapChars(chars: number[]) {
-      if (chars.length === 1) {
+      const primes = this.node.getProperty('primes') as string;
+      if (primes) {
+        chars = unicodeChars(primes);
```
zorkow, line 372: Just have `return unicodeChars(primes);` here and save an `else`.

ts/core/MmlTree/MmlNodes/mo.ts
```diff
@@ -384,4 +440,35 @@ export class MmlMo extends AbstractMmlTokenNode {
...
+    const isPseudo = !parent || !(parent.isKind('msubsup') && !parent.isKind('msub'));
+    this.setProperty('pseudoscript', isPseudo);
+    if (isPseudo) {
+      this.attributes.setInherited('lspace', 0);
+      this.attributes.setInherited('rspace', 0);
+    }
+  }
+
+  /**
+   * Determine of the mo consists of primes, and remap them if so.
```
zorkow, line 462: typo.

zorkow (approved):
Only some minor things.

PR #605, opened 2021-03-02
Sre issue 462

zorkow:
This is a step towards a solution of the rendering discrepancies caused by enrichment (see issue SRE https://github.com/zorkow/speech-rule-engine/issues/462). The basic idea is to follow the `fencepointer` to the node with the corresponding semantic `id` and take that to compute the base bound box.

* It currently only works for the `chtml` renderer.  
* Not implemented for `msubsup` or `mmultiscript` nodes.

__This is a draft PR. Do not merge!__

### DIFF
--- a/ts/output/common/Wrappers/scriptbase.ts
+++ b/ts/output/common/Wrappers/scriptbase.ts
@@ -266,8 +266,7 @@ export function CommonScriptbaseMixin<
      * @override
      */
     public computeBBox(bbox: BBox, recompute: boolean = false) {
-      const baseChild = this.getRealBaseChild();
-      const basebox = baseChild.getBBox();
+      const basebox = this.baseChild.getBBox();
       const scriptbox = this.script.getBBox();
       const [x, y] = this.getOffset(basebox, scriptbox);
       bbox.append(basebox);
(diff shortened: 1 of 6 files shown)


zorkow:
Also works for `svg` now.

@pkra's example renders now pretty much as expected. The only minor difference is that the subscripts on the right jump slightly. They did not do that before.

ts/output/common/Wrappers/scriptbase.ts
```diff
@@ -277,7 +302,7 @@ export function CommonScriptbaseMixin<
      * @return {boolean}  True if the base is an mi, mn, or mo (not a largeop) consisting of a single character
      */
     public isCharBase(): boolean {
-      let base = this.baseChild;
+      let base = this.getRealBaseChild();
```
zorkow, line 305: This is ruining the placement in sub/sup expressions. So should be moved to a specialist method in the `mmultiscripts` class.

zorkow: Done

pkra:
More test cases.

Too high/low (sup/sub):

\begin{equation*} N(k,s) = \Big \|\psi (\cdot )\frac {\partial ^k\hat{f}}{\partial \xi _d^k}(\cdot ,h(\cdot ))\Big \|_{\dot{H}^{-s}}^2 \end{equation*}

Now too low (sub sigma):

\big \|\phi (\hat{f}-\hat{f}_n)\big |_{\Sigma }\big \|_{H^\ell (\Sigma )}

dpvc:
I'm closing this in favor of #617.

PR #607, opened 2021-03-03
Updates needed to support STIX2 fonts

dpvc:
This PR includes changes needed to support the STIX2 fonts.  (It is cherry-picked from the first commit on the `stix2` branch, but is needed for other changes that will be made in a separate PR.  The `stix2` branch can be rebased off this branch, if needed, when we make the PR for stix2.)

Most of the changes here are to allow specifying the variant to use for stretchy characters, so there are a lot of places where a variant needs to be passed into a function that was not the case before.  There is also some refactoring of some duplicate code into `getCenterOffset()`, and simplification of of scaling transforms (e.g., `matrix(1 0 0 0 -1 0)` is replaced by `scale(1,-1)` which does the same thing more transparently).

### DIFF
--- a/ts/output/svg.ts
+++ b/ts/output/svg.ts
@@ -236,7 +236,7 @@ CommonOutputJax<N, T, D, SVGWrapper<N, T, D>, SVGWrapperFactory<N, T, D>, SVGFon
     //
     const g = this.svg('g', {
       stroke: 'currentColor', fill: 'currentColor',
-      'stroke-width': 0, transform: 'matrix(1 0 0 -1 0 0)'
+      'stroke-width': 0, transform: 'scale(1,-1)'
     }) as N;
     //
     //  The svg element with its viewBox, size and alignment
@@ -262,9 +262,9 @@ CommonOutputJax<N, T, D, SVGWrapper<N, T, D>, SVGWrapperFactory<N, T, D>, SVGFon
       adaptor.setStyle(svg, 'min-width', this.ex(W));
       adaptor.setAttribute(svg, 'width', pwidth);
       adaptor.removeAttribute(svg, 'viewBox');
-      const scale = wrapper.metrics.ex / (this.font.params.x_height * 1000);
-      adaptor.setAttribute(g, 'transform', 'matrix(1 0 0 -1 0 0) scale(' +
-                           this.fixed(scale, 6) + ') translate(0, ' + this.fixed(-h * 1000, 1) + ')');
+      const scale = this.fixed(wrapper.metrics.ex / (this.font.params.x_height * 1000), 6);
+      adaptor.setAttribute(g, 'transform', 'scale(' + scale + ',-' + scale + ') '
+                            + 'translate(0, ' + this.fixed(-h * 1000, 1) + ')');
     }
     if (this.options.fontCache !== 'none') {
       adaptor.setAttribute(svg, 'xmlns:xlink', XLINKNS);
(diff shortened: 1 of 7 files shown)


ts/output/svg.ts
```diff
@@ -262,9 +262,9 @@ CommonOutputJax<N, T, D, SVGWrapper<N, T, D>, SVGWrapperFactory<N, T, D>, SVGFon
       adaptor.setStyle(svg, 'min-width', this.ex(W));
       adaptor.setAttribute(svg, 'width', pwidth);
       adaptor.removeAttribute(svg, 'viewBox');
-      const scale = wrapper.metrics.ex / (this.font.params.x_height * 1000);
-      adaptor.setAttribute(g, 'transform', 'matrix(1 0 0 -1 0 0) scale(' +
-                           this.fixed(scale, 6) + ') translate(0, ' + this.fixed(-h * 1000, 1) + ')');
+      const scale = this.fixed(wrapper.metrics.ex / (this.font.params.x_height * 1000), 6);
+      adaptor.setAttribute(g, 'transform', 'scale(' + scale + ',-' + scale + ') '
+                            + 'translate(0, ' + this.fixed(-h * 1000, 1) + ')');
```
zorkow, line 267: Can you rewrite that into a string literal.

It appears that string literals are now faster than concatenation on all browsers, so we should try to prefer that.

ts/output/svg/Wrappers/mo.ts
```diff
@@ -186,60 +203,65 @@ CommonMoMixin<SVGConstructor<any, any, any>>(SVGWrapper) {
       y: this.fixed(B - D), x: this.fixed((W - w) / 2),
       viewBox: [0, y, w, Y].map(x => this.fixed(x)).join(' ')
     });
-    this.addGlyph(n, 0, 0, svg);
+    this.addGlyph(n, v, 0, 0, svg);
     const glyph = adaptor.lastChild(svg);
-    adaptor.setAttribute(glyph, 'transform', 'scale(1, ' + this.jax.fixed(s) + ')');
+    adaptor.setAttribute(glyph, 'transform', 'scale(1,' + this.jax.fixed(s) + ')');
```
zorkow, line 208: ditto.

ts/output/svg/Wrappers/mo.ts
```diff
@@ -81,100 +79,119 @@ CommonMoMixin<SVGConstructor<any, any, any>>(SVGWrapper) {
...
-    return this.placeChar(n, x, y, parent || this.element, '-size4');
+  protected addGlyph(n: number, variant: string, x: number, y: number, parent: N = null): number {
+    return this.placeChar(n, x, y, parent || this.element, variant);
   }
 
   /***********************************************************/
 
   /**
    * @param {number} n    The character number for the top glyph
+   * @param {srting} v    The variant for the top glyph
```
zorkow, line 168: typo (and multiple times below).

zorkow (approved):
Some minor issues. O/w fine.

ts/output/svg.ts
dpvc: OK, will do.  We will probably need to do a pass over the whole codebase to update that.

ts/output/svg/Wrappers/mo.ts
dpvc: done.

ts/output/svg/Wrappers/mo.ts
dpvc: done.

PR #608, opened 2021-03-03
Handle math accents more like TeX.

dpvc:
This PR handles math accents more like TeX does:  when the base is a single character, the italic correction is not added, and the width of the accent does not affect the resulting width (so if the accent is too large, it extends beyond the bounding box of the character below it).  The latter is done by making the accent zero width, and offset to the left by half its natural width.

In addition, the handling of the skew values is updated (to work with the STIX2 font).  In the past, MathJax did not have proper skew values, so it uses the italic correction to compute a skew value, which is not needed for STIX2, which has proper skew values.


dpvc:
I am going to replace this PR with a new one, so don't review.

dpvc:
Closing in favor of #618.

PR #609, opened 2021-03-04
Improve xarrow spacing, and fix CHTML output for short extensible arrows.

dpvc:
This PR improves the positioning of the content above and below stretchy arrows generated by xArrows to make it match TeX positioning better.  It also adjusts the bounding box to accommodate the `voffset` amounts (in the past, elements could fall outside the bounding box).

This also fixes a regression with short stretchy arrows in CHTML caused by #571 (the extensible piece ends up with a minimum size of the size of the extension character, so would not "overlap" the end piece to make short extensions).  For example, without this fix, `\xrightarrow{xx}` would be too long in CHTML (on the develop branch).  This puts back the pre-571 handling of the size of the extension.

### DIFF
--- a/ts/input/tex/ams/AmsMappings.ts
+++ b/ts/input/tex/ams/AmsMappings.ts
@@ -1,3 +1,4 @@
+
 /*************************************************************
  *
  *  Copyright (c) 2017 The MathJax Consortium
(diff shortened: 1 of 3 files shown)


zorkow (approved):
lgtm.

PR #610, opened 2021-03-04
Handle automatic tex class of OP better.

dpvc:
This PR improves the assignment of TeX class OP to multi-character `mi` elements, and to elements followed by invisible apply.  This improves spacing of enriched content where invisible apply has been inserted after a single-character identifier (e.g., `fR(s)`).

It also refactors the MathML input to not need the `texAtom()` method, and to remove `xmlns` attributes that appear on serialized MathML from SRE.


zorkow:
The superfluous `xmlns` elements only seem to appear during browser usage of SRE. There seems to be some issue with the namespace resolver.  Do you think it is worthwhile investigating?

dpvc:
> Do you think it is worthwhile investigating?

Probably not, as it is not hard to remove it at this point in the code, so it should not cause a problem.  Unless you are concerned about other uses outside of MathJax.

PR #611, opened 2021-03-05
Use consistent characters for \overline and \underline, and treat them both as accents

dpvc:
This PR changes the character used for `\overline` and `\underline` to be U+2015 (HORIZTONAL LINE) and makes them both have `accent="true"` (only `\overline` had that originally) so that the results for both are consistent both in spacing and thickness.


dpvc:
I will be replacing this PR with a new one, so don't review.

dpvc:
Closing this in favor of #619.

PR #612, opened 2021-03-07
[packge.json] Upgrade all packages except typescript

artagnon:
Typescript 4.2.3 is a little more strict, and requires some code cleanup
before we can bump up its version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,237 +1,4584 @@
 {
   "name": "mathjax-full",
   "version": "3.1.2",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "mathjax-full",
+      "version": "3.1.2",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "esm": "^3.2.25",
+        "mj-context-menu": "^0.6.1",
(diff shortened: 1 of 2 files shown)


dpvc:
First of all, thank you for your interest in MathJax, and willingness to contribute (I see that you have made several pull requests).  We appreciate your efforts.

There are a couple of issues, however.  First, we follow a [Git-Flow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) model for MathJax-src, which means that changes to master should only be for releases or hot-fixes (or corrections to README, or security updates from dependably, etc.).  All development is done on the `develop` branch, and that includes updates to the development tools, e.g., the `package.json` file.  For example, #560 updated `package.json` and also updated the code to work in Typescript v4.  your updates would conflict with that when we merge `develop` into `master` when we make the next release.

So PRs for MathJax-src should be on a branch that originates from the `develop` branch and is targeted at `develop`.

We normally do an update of the dependencies when we do a release, and will be doing so for the 3.1.3 release that is currently being worked on, so there is no need to submit PRs for that, unless there is a serious problem with one of the dependencies (like a security issue).  So I am going to close this PR for that reason.  But thank you!

artagnon:
Thanks for the explanation! I've rebased my code.

PR #617, opened 2021-03-10
Refactor handling of baseCore, and take enrichment into account (zorkow/speech-rule-engine#462)

dpvc:
This PR refactors the handling of the base core element for munderover and msubsup elements so that super- and subscripts can be placed properly after SRE enrichment.  It also lays the groundwork for better TeX emulation for accents and over- and underlines (which will be in separate PRs).  

The base core element, along its italic correction and relative scaling factor, are now found during the `constructor()` rather than calling them (multiple times) when needed.  The base core is used for vertical positioning of super- and subscripts, but the actual base for width and bounding box computations.  The handling of the base core BBox and script BBoxes have been moved to the common wrappers, where possible.

For consistency, the `script` property has been change to `scriptChild`, to be comparable to the other child pointers.

In `common/Wrappers/scriptbase.ts`, because the `coreIC()` and `coreScale()` methods have been removed and new function added in that location, the diff gets a little mixed up about the changes, so it's a bit complicated to look at.  You might view the actual file for that bit in order to read it easier.

Resolves issue zorkow/speech-rule-engine#462.

### DIFF
--- a/ts/output/common/Wrappers/scriptbase.ts
+++ b/ts/output/common/Wrappers/scriptbase.ts
@@ -321,8 +321,8 @@ export function CommonScriptbaseMixin<
       if (fence.node.attributes.getExplicit('data-semantic-id') === id) {
         return fence;
       }
-      for (let child of fence.childNodes) {
-        let result = this.getBaseFence(child, id);
+      for (const child of fence.childNodes) {
+        const result = this.getBaseFence(child, id);
         if (result) {
           return result;
         }
--- a/ts/output/common/Wrappers/scriptbase.ts
+++ b/ts/output/common/Wrappers/scriptbase.ts
@@ -46,7 +46,7 @@ export interface CommonScriptbase<W extends AnyWrapper> extends AnyWrapper {
   /**
    * The core mi or mo of the base (or the base itself if there isn't one)
    */
-  baseCore: W;
+  readonly baseCore: W;
 
   /**
    * The base element's wrapper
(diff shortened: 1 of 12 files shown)


dpvc:
PS, this supersedes #605.

ts/output/common/Wrappers/scriptbase.ts
```diff
@@ -215,77 +264,119 @@ export function CommonScriptbaseMixin<
...
+     * @param {W} fence The potential fence.
+     * @param {string} id The fencepointer id.
+     * @return {W} The original fence the scripts belong to.
      */
-    public coreIC(): number {
-      const corebox = this.baseCore.getBBox();
-      return (corebox.ic ? 1.05 * corebox.ic + .05 : 0);
+    public getBaseFence(fence: W, id: string): W {
+      if (!fence || !fence.node.attributes || !id) {
+        return null;
```
zorkow, line 319: Should that not simply return `fence`? This is related to my question above.

ts/output/common/Wrappers/scriptbase.ts
```diff
@@ -215,77 +264,119 @@ export function CommonScriptbaseMixin<
...
+    /***************************************************************************/
+    /*
+     *  Methods for information about the core element for the base
+     */
+
+    /**
+     * @return {W}    The wrapper for the base core mi or mo (or whatever)
+     */
+    public getBaseCore(): W {
+      let core = this.getSemanticBase() || this.childNodes[0];
```
zorkow, line 292: This confuses me: If the semantic base exists, would that not be a wrapper and therefore I would have take the first child?

zorkow (commented):
Can you please clarify my two questions? Then I think I can finish it off quickly. 
It's certainly a much cleaner effort than what I attempted.

ts/output/common/Wrappers/scriptbase.ts
zorkow: Actually in this view it is below.

ts/output/common/Wrappers/scriptbase.ts
dpvc: The wrapper is the output-jax-specific shell around the actual (internal) MathML node, node an enclosing MathML node.  If the fence is an `mo` (an `MmlMo` node), for example, then `getSemanticBase()` will return the `CHTMLmo` object that is the wrapper for that `MmlMo` node.  You don't need to take its children, because it is already the wrapper that you want.

ts/output/common/Wrappers/scriptbase.ts
dpvc: No.  The three cases here are cases where the fence has not been found, but only the first one has `fence` being `null`, so if you returned `fence` in the other two cases, that would indicate that you found a semantic fence when you really didn't.  You only want to return the `fence` when it exists, is a node with attributes (e.g., not a text node), and the `data-semantic-id` attribute matches the (non-empty) `id` that you are looking for.  That is what line 322 below does.  

Note that this was [your code](https://github.com/mathjax/MathJax-src/pull/605/files#diff-827f9299f8e5cea1db7b2c0162aebec7f45941a24f73e6254d532b88156e0678) from your original PR (with the addition of the check for the attributes).

zorkow (approved):
OK thanks for the explanation. It makes sense. I just thought we could avoid the `return null` part, but it does not look like it.

dpvc:
No problem.  Thanks for the original PR that prompted this updated version!

PR #618, opened 2021-03-10
Handle accents more like TeX does (better sizes and placement) (mathjax/MathJax#712, mathjax/MathJax#2474)

dpvc:
This PR combines what was the `tex-accents` and `stretchy-accents` into one, based on the `sre-fences` branch.

It adjusts the algorithm used to stretch accents like those used in `\widehat` to better match the one used by TeX. Prior to this, the accents used the horizontal stretchy character algorithm, which takes the smallest version that is larger than the needed size, but TeX's algorithm for accents uses the largest that is smaller than the width of the base.

Here, we mark math accents with a new `mathaccent` property, and when one is stretched, it uses the largest version that is smaller than the base. We use the fudge-factors (`delimitershortfall` and `delimiterfactor`) as for other stretchy characters because we don't really have the proper italic corrections, and so the sizes turn out to not produce the desired result for the MathJax TeX fonts (it also turns out that the wide hat and wide tilde delimiter data is wrong in the files, but that will be fixed in a separate PR). This fudge-factor allows the algorithm to work appropriately for both the MathJax fonts and the STIX2 fonts.

Also, when the base is a single character, TeX does not add the italic correction, and the width of the accent does not affect the resulting width (so if the accent is too large, it extends beyond the bounding box of the character below it). This PR implements those two features in MathJax as well.  The latter is done by making the accent zero width, and offset to the left by half its natural width.

In addition, the handling of the skew values is updated (to work with the STIX2 font). In the past, MathJax did not have proper skew values, so it uses the italic correction to compute a skew value, which is not needed for STIX2, which has proper skew values.

### DIFF
--- a/ts/output/common/FontData.ts
+++ b/ts/output/common/FontData.ts
@@ -519,6 +519,11 @@ export class FontData<C extends CharOptions, V extends VariantData<C>, D extends
   public params: FontParameters;
 
   /**
+   * Factor to multiply italic correction by for delta computations for munderover
+   */
+  public skewIcFactor: number = .75;
+
+  /**
    * Any styles needed for the font
    */
   protected _styles: StyleList;
--- a/ts/output/common/FontData.ts
+++ b/ts/output/common/FontData.ts
@@ -519,7 +519,7 @@ export class FontData<C extends CharOptions, V extends VariantData<C>, D extends
   public params: FontParameters;
 
   /**
-   * Factor to multiply italic correction by for delta computations for munderover
+   * Factor by which to multiply italic correction for computation of delta in munderover
    */
   public skewIcFactor: number = .75;
 
--- a/ts/output/svg/Wrappers/mo.ts
+++ b/ts/output/svg/Wrappers/mo.ts
@@ -64,11 +64,10 @@ CommonMoMixin<SVGConstructor<any, any, any>>(SVGWrapper) {
     if (stretchy && this.size < 0) {
       this.stretchSVG();
     } else {
-      if (symmetric || attributes.get('largeop')) {
-        const u = this.fixed(this.getCenterOffset());
-        if (u !== '0') {
-          this.adaptor.setAttribute(svg, 'transform', 'translate(0 ' + u + ')');
-        }
+      const u = (symmetric || attributes.get('largeop') ? this.fixed(this.getCenterOffset()) : '0');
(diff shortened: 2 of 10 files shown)


dpvc:
PS, this replaces #603 and #608.

dpvc:
This PR resolves issues mathjax/MathJax#712 and mathjax/MathJax#2472.

ts/output/common/FontData.ts
```diff
@@ -518,6 +518,11 @@ export class FontData<C extends CharOptions, V extends VariantData<C>, D extends
    */
   public params: FontParameters;
 
+  /**
+   * Factor to multiply italic correction by for delta computations for munderover
```
zorkow, line 522: typo

ts/output/svg/Wrappers/mo.ts
```diff
@@ -64,11 +64,10 @@ CommonMoMixin<SVGConstructor<any, any, any>>(SVGWrapper) {
...
     } else {
-      if (symmetric || attributes.get('largeop')) {
-        const u = this.fixed(this.getCenterOffset());
-        if (u !== '0') {
-          this.adaptor.setAttribute(svg, 'transform', 'translate(0 ' + u + ')');
-        }
+      const u = (symmetric || attributes.get('largeop') ? this.fixed(this.getCenterOffset()) : '0');
+      const v = (this.node.getProperty('mathaccent') ? this.fixed(this.getAccentOffset()) : '0');
+      if (u !== '0' || v !== '0') {
+        this.adaptor.setAttribute(svg, 'transform', 'translate(' + v + ' ' + u + ')');
```
zorkow, line 70: Use a template literal.

zorkow (approved):
Two minor things, only.

ts/output/common/FontData.ts
dpvc: Sorry, I don't see it.  Can you point it out?

ts/output/svg/Wrappers/mo.ts
dpvc: Done.

ts/output/common/FontData.ts
dpvc: I'm going to merge this for now, and will fix the typo later (since other approved branches depend on this and I'd like to merge them).

dpvc: I'll also change to "Factor by which to multiply italic correction for delta computations for munderover".

PR #619, opened 2021-03-10
Improve spacing between accents and their bases, and handle under- and overlines more like TeX.

dpvc:
This PR changes the character used for `\overline` and `\underline` to be U+2015 (HORIZTONAL LINE) and makes them both have `accent="true"` (only `\overline` had that originally) so that the results for both are consistent both in spacing and thickness.  It also adds detection of these accents so that the spacing between the line and the base as well as the horizontal placements will match TeX's algorithm, and it adds a bit more space between math accents and the base (since on screen they were getting a bit tight).

This resolves issue mathjax/MathJax#2633.

### DIFF
--- a/ts/output/common/Wrappers/scriptbase.ts
+++ b/ts/output/common/Wrappers/scriptbase.ts
@@ -312,18 +312,19 @@ export function CommonScriptbaseMixin<
       const core = this.baseCore = this.getBaseCore();
       if (!core) return;
       //
-      //  Check if the base is a mi or mo that needs italic correction removed
-      //
-      if (('noIC' in core) && (this.constructor as CommonScriptbaseClass).useIC) {
-        (core as unknown as CommonMo).noIC = true;
-      }
-      //
       // Get information about the base element
       //
       this.setBaseAccentsFor(core);
       this.baseScale = this.getBaseScale();
       this.baseIc = this.getBaseIc();
       this.baseIsChar = this.isCharBase();
+      //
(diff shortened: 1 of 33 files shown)


dpvc:
PS, this PR replaces #611.

pkra:
> dpvc wants to merge 1 commit into tex-stretchy-accents from underline-overline

That seems odd since underline-overline's PR #611 was closed.

pkra:
> That seems odd since underline-overline's PR #611 was closed.

Oops, read this the wrong way around. Sorry for the noise.

ts/output/common/Wrappers/scriptbase.ts
```diff
@@ -276,17 +312,23 @@ export function CommonScriptbaseMixin<
...
+      //
+      this.setBaseAccentsFor(core);
+      this.baseScale = this.getBaseScale();
+      this.baseIc = this.getBaseIc();
+      this.baseIsChar = this.isCharBase();
+      //
       //  Check if the base is a mi or mo that needs italic correction removed
       //
-      if (('noIC' in core) && !(this.constructor as CommonScriptbaseClass).useIC) {
+      if (this.baseIsChar && (this.isMathAccent || this.node.isKind('msubsup'))) {
```
dpvc, line 324: Note:  This test isn't right, but is fixed in the mathaccent-ic branch.

zorkow (approved):
lgtm.

PR #622, opened 2021-03-12
Fix problem with SVG output crashing when \not is used with a letter

dpvc:
This PR fixes a regression due to #589 that affects the SVG output when `\not` is followed by a letter, as in `\not a`.  This generates `<mi>a&#x338;</mi>`, and the contents of the `<mi>` is actually two text items, due to how the TeX input jax adds the U+0038.  The second text item needs to be placed with a non-zero *x* value, and causes the change from #589 to attempt to access the node's `attributes` value, which doesn't exist on a text node, causing an error.  This PR prevents that.


zorkow (approved):
lgtm.

PR #624, opened 2021-03-12
[es6] Switch builds to es6

artagnon:


dpvc:
Again, thanks for your willingness to contribute to MathJax!  I appreciate your rebasing the PR to the correct branch.

We have been discussing the possibility of adding an `es6` directory with compilation to ES2015, but are net yet ready to do so, and it would have to be in parallel with `es5` for the time being (not replace it).  We still support IE11 and other older platforms (e.g., some e-book readers) that do not support es6 yet, and so can not simply drop es5 completely, as your PR would do.  When we do move to ES6, it will be in parallel with ES5, and so some of the tools in your PR will need to be modified to allow control over which version is to be targeted.  That will take some additional work.  

Furthermore, since this would be a breaking change, it could not be in v 3.1.3, but would need to wait for 3.2.0 at the earliest.

Finally, your PR includes changes to the `LinkedList.ts` that are not really related to the change to ES6.  The changes to that file should be a separate PR.  The reason that `END` is used rather then `null` is that `null` could be a legitimate item to push onto the list, and your changed code would not be able to distinguish that from the end (or beginning) of the list.  The `END` symbol is something that can't be valid data for the list, so can't be accidentally misunderstood.  I would not want to change that to null.  I'm happy to consider updating the iterator, if more recent versions of Typescript offer something better, but we have discussed it in the last developer meeting, and would like to say with `END` as it is.

artagnon:
Okay, let's close this, and have a look at #625 instead. I've modernized the iterators preserving `END`.

PR #625, opened 2021-03-12
Bump package.json

artagnon:


dpvc:
As I mentioned in your previous version of this PR, updates to `package.json` are done when we prepare a release, and there is no need to continually change the `package.json` in the repo before then.  Developers are free to update their local versions as they see fit.  It also may not be appropriate to put a beta version of SRE in there.  We expect to have the official 3.2 for the next release.

This PR seems to include a number of other updates as well, which should probably be separate PRs so that we could accept or deny each of those distinct categories of change separately on its own merits.  Here, we are not interested in the changes to `package.json`, so can't accept your other changes as they currently stand, since all off them are in one PR.

I'm OK with removing the `toArray()` method, but note that there is a similar one in `PrioritizedList` that should also be removed, and that its one invocation in `input/tex/MapHandler.ts` should be changed to `Array.from()` as well.

The change to `TextMacrosMethods.ts` is incorrect.  The underscore is used to prevent `tslint` from flagging the variable as an unused variable.  This is done in many places through the codebase, and should not be changed here.

The changes to the iterator look reasonable, but I would need to do some testing to check the functionality.  Also, you would need to add the jsDoc comments.  So if you want to separate that change out into a PR by itself, that would be one that we would consider.

artagnon:
Thanks for the detailed feedback! :)

PR #626, opened 2021-03-12
Modernize iterators

artagnon:
Clean up some legacy code in the first patch, and modernize the LinkedList iterator in the second.


ts/util/LinkedList.ts
```diff
@@ -224,48 +211,49 @@ export class LinkedList<DataClass> {
...
+      current = current.next;
+    }
+  }
+
+  /**
+   * Reverse iterator for the list
+   *
+   * @yield {DataClass} The previous item in the iteration sequence
+   */
+   protected *reverse_iterator(): IterableIterator<DataClass> {
```
dpvc, line 232: In MathJax, we use camelCase for function names, not underscores.  Can you change this to `reverseIterator`?

I'm also wondering if these need to be separate methods, or if they could be folded in the main functions.  E.g.:

``` javascript
 public reversed(): IterableIterator<DataClass> {
     return function * (): IterableIterator<DataClass> {
       let current = this.list;
       while (current.data !== END) {
         yield current.data as DataClass;
         current = current.prev;
       }
     }
   }
```

or even just

``` javascript
 public *reversed(): IterableIterator<DataClass> {
     let current = this.list;
     while (current.data !== END) {
       yield current.data as DataClass;
       current = current.prev;
     }
   }
```

I haven't tried these, but it seems unnecessary to have a method that just returns the result of another method.

dpvc (changes requested):
This looks good, but see my comment below.

artagnon:
Done; do check if it's alright.

ts/util/LinkedList.ts
```diff
@@ -224,48 +209,31 @@ export class LinkedList<DataClass> {
...
   /**
-   * Make the list iterable and return the data from the items in the list
+   * An iterator for the list in forward order
    *
-   * @return {{next: Function}}  The object containing the iterator's next() function
+   * @yield {DataClass} The next item in the iteration sequence
    */
-  public [Symbol.iterator](): Iterator<DataClass> {
+  public *[Symbol.iterator](): IterableIterator<DataClass> {
     let current = this.list;
```
dpvc, line 217: This needs to be `this.list.next` not just `this.list`.  Th function as it stands never returns any values, as `this.list.data` is always `END`.

dpvc (changes requested):
Neither iterator works as expected (though the fix is easy).  I'm concerned, however, that this suggests that you didn't actually test the code to see if it works, as both iterators always return empty lists.  Did you test the code?  And if so, how?

artagnon:
> Neither iterator works as expected (though the fix is easy). I'm concerned, however, that this suggests that you didn't actually test the code to see if it works, as both iterators always return empty lists. Did you test the code? And if so, how?

No, and I'm really clueless about how to test the change. Could you give me some pointers?

dpvc:
> I'm really clueless about how to test the change. Could you give me some pointers?

Here is a small program to test the iteration:

``` javascript
import {LinkedList} from './js/util/LinkedList';

const list = new LinkedList();

list.push('a', 'b', 'c');

console.log('The list is: ', Array.from(list));
console.log('Looping through the list:');

for (const item of list) {
  console.log(item);
}

console.log('The reversed list is: ', Array.from(list.reversed()));
console.log('Looping through the reversed list:');

for (const item of list.reversed()) {
  console.log(item);
}
```

run this using `node -r esm` and see that the output is as expected.  You should get:

```
The list is:  [ 'a', 'b', 'c' ]
Looping through the list:
a
b
c
The reversed list is:  [ 'c', 'b', 'a' ]
Looping through the reversed list:
c
b
a
```

dpvc:
Of course, you should also check an empty list, and a list with just one item as well.

artagnon:
Thanks; I can confirm that it works. Should we add automated tests?

dpvc:
> I can confirm that it works.

Excellent.

> Should we add automated tests?

We are working on a test framework, and there are already tests for much of the TeX input processing.  I had done some test for the `util` folder early on (in the `testsuite` branch), and I used those to test your code, but much more is needed, and those tests need to be converted to the framework used for the TeX input tests.

So for the moment, the answer is no.

dpvc (approved):
OK, the changes now look good.

dpvc:
I'm going to target this to 3.2.0, since the removal of the `toArray()` method is potentially a breaking change.

PR #627, opened 2021-03-12
[TextMacrosMethods] Fix deprecated var use

artagnon:


dpvc:
As I mentioned in a previous PR, the underscore is to prevent `tslint` from complaining about the fact that the `name` parameter is not used in the function itself.  This is the standard way to accomplish that (see the [tslint documentation](https://palantir.github.io/tslint/rules/no-unused-variable/), in particular, the `ignore-pattern`).  This technique is used through the codebase, and is not an error or deprecated use.

For that reason, I am closing this PR.

artagnon:
In this case, it's actually used in the first line of the function, no?

```js
const base = parser.ParseArg(name);
````

dpvc:
You are right.  My apologies.

PR #628, opened 2021-03-12
[loader] Get ready() and load() to look through CONFIG.load

artagnon:
As it currently stands, ready() and load() ask for a bunch of package
names, instead of looking through the readily-available list in
CONFIG.load. Change this, and deprecate the first argument.

EDIT: Some additional work is required to look through CONFIG.path and substitute it in CONFIG.load, but is this worth pursuing?

### DIFF
--- a/ts/components/loader.ts
+++ b/ts/components/loader.ts
@@ -80,12 +80,9 @@ export namespace Loader {
    * @param {string[]} names  The packages to wait for
    * @returns {Promise}       A promise that resolves when all the named packages are ready
    */
-  export function ready(...names: string[]): Promise<string[]> {
-    if (names.length === 0) {
-      names = Array.from(Package.packages.keys());
-    }
+  export function ready(..._names: string[]): Promise<string[]> {
     const promises = [];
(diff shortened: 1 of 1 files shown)


dpvc:
This will break MathJax.  The `ready()` function is a means of getting signaled when one more components are loaded, and it is something you can set up to use with any component, whether it was configured to be loaded or not.  For example, an extension may need to modify things in another extension.  E.g., the XyPic extension needs to modify the CHTML or SVG output jax, but may need to determine which is loaded, and the `ready()` function can be used to do that.  That way, you don't force both output jax to be loaded, but can modify the one that is used.  (XyPic doesn't currently use this, but it could and probably should.)  You can set up `ready()` promises for components that may never be loaded, but that is exactly what it is for.

Similarly, the `load()` function is used to load one or more components, whether they have been configured or not.  For example, the MathJax contextual menu uses this to load some of the assistive extensions when they are needed.  If you change this as you have done, you will prevent the MathJax from loading those extensions.

So these functions are correct as they stand.  For this reason, I am closing this PR.

PR #629, opened 2021-03-12
[loader] Clarify ready()'s docstring

artagnon:
ready()'s docstring is currently incorrect, and it's impossible to distinguish it from load().

### DIFF
--- a/ts/components/loader.ts
+++ b/ts/components/loader.ts
@@ -75,7 +75,7 @@ export interface MathJaxObject extends MJObject {
 export namespace Loader {
 
   /**
-   * Get a promise that is resolved when all the named packages have been loaded.
+   * Get a promise that is resolved when all the named packages are ready to be loaded (doesn't actually load them)
    *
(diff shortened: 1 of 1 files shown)


dpvc:
The comment is technically correct as is.  The `ready()` gives you a promise that is resolves when all the extensions are loaded.  It doesn't say anything about initiating that loading.  That is what `load()` does.  The `ready()` function doesn't load anything (as you note), but just signals you when something *is* loaded.  You can set up a `ready()` call in order to be notified when a component is loaded by some other action (e.g., you could set up a `ready()` call that would inform you if the explorer extension was loaded due to the user activating the explorer in the MathJax contextual menu).

It might be worth pointing out that `ready()` doesn't initiate the loading of the components, though its name certainly doesn't suggest that it does.  But the new wording you give is not correct: it doesn't resolve until the components actually *are* loaded, not just when they are ready to be loaded (aren't they always ready to be loaded?)

For this reason, I'm closing this PR.

PR #630, opened 2021-03-16
Fully clear tagging on equation start.  (mathjax/MathJax#2643)

dpvc:
This PR resolves an issue where the tagging information was not fully cleared when an equation begins.  This caused a problem if there were TagInfo objects on the stack when an extension is autoloaded (causing the equation to be rerendered).  The extra TagInfo would end up setting the initial `currentTag` item when the *next* equation is processed.  See the example at mathjax/MathJax#2643.


zorkow (approved):
lgtm.

PR #633, opened 2021-03-17
Tex tidy space constants

zorkow:
* Rewrites all spacing commands into numeric values wrt. `em`. Note, that the `Spacer` method now takes a numerical value and no longer a string. __I assume this is not considered a breaking change.__
* Uses the values from lengths utilities and removes the obsolete entries in `TexConstants`.


ts/input/tex/ams/AmsMappings.ts
```diff
@@ -112,7 +107,7 @@ new sm.CommandMap('AMSmath-macros', {
  */
 new sm.EnvironmentMap('AMSmath-environment', ParseMethods.environment, {
   'eqnarray*':   ['EqnArray', null, false, true, 'rcl',
-                  '0 ' + TexConstant.Length.THICKMATHSPACE, '.5em'],
+                  '0 ' + MATHSPACE.thickmathspace + 'em', '.5em'],
```
dpvc, line 110: There is an `em()` function in `util/lengths.ts` that handles converting the number to ems (with fewer decimals).  it would be good to use that here, and below rather than just adding `'em'`.  I see five other occurrences below.

dpvc (approved):
This looks good, except for the issue I mention about using the `em()` function.

I agree that it is not a breaking change.

dpvc:
Is this ready to go, then?  I will merge if so.

PR #634, opened 2021-03-17
Fix handling of italic correction for munderover and msubsup combinations.

dpvc:
This PR refactors some fo the previous accent code to better handle the italic correction differences in the bases of some constructs (`msub`, `msubsup`, and math-accent `munderover` nodes).  The previous code included some hacks that were competing with each other, and caused complications in handling the interactions between `munderover` and `msubsup` nodes.  This refactors some of the code to make that work more reliably, and moves more to the common wrappers so that the specific output routines are more uniform.

In the past, the italic correction was removed from the base character when it was output, and had to be added back in later, which made for complicated checking of whether the italic correction needs to be taken into account and when it needs to be added again.  The new approach is to always include the italic correction in the base and then compensate for it when subscripts are added, or when math accents are placed.  This makes for fewer cases, and allows the CHTML CSS to be reduced by removing the need for the `noIC` selector cases.

This PR also moves the `useIC` properties to the common wrappers (since they are inherited), and switches the default to `true` since that makes for fewer cases that need to be overridden (just `msub` and `msubsup`).

### DIFF
--- a/ts/output/common/Wrappers/scriptbase.ts
+++ b/ts/output/common/Wrappers/scriptbase.ts
@@ -61,6 +61,11 @@ export interface CommonScriptbase<W extends AnyWrapper> extends AnyWrapper {
   readonly baseIc: number;
 
   /**
+   * True if the base IC has already been added (when it was removed from the base)
+   */
+  readonly baseHasIc: boolean;
+
+  /**
    * True if the base is a single character
    */
   readonly baseIsChar: boolean;
@@ -122,6 +127,11 @@ export interface CommonScriptbase<W extends AnyWrapper> extends AnyWrapper {
   getBaseIc(): number;
 
   /**
+   * An adjusted italic correction (for slightly better results)
+   */
+  getAdjustedIc(): number;
+
+  /**
    * @return {boolean}  True if the base is an mi, mn, or mo (not a largeop) consisting of
    *                    a single unstretched character
    */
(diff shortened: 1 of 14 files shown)


ts/output/common/Wrappers/scriptbase.ts
```diff
@@ -455,16 +490,25 @@ export function CommonScriptbaseMixin<
       return (node.isToken && (node as MmlMo).getText() === '\u2015');
     }
 
+    /**
+     * @return {number}   The base child's width adjusted for the base italic correction
+     */
+    public getBaseWidth(): number {
+      const bbox = this.baseChild.getBBox();
+      return bbox.w * bbox.rscale - (this.baseRemoveIc ? this.baseIc : 0);
```
zorkow, line 498: Is that correct? If we remove the `ic` we take the base value otherwise we take none? Should that be the other way around?

zorkow (approved):
One question. Otherwise, the code is a lot cleaner now!

ts/output/common/Wrappers/scriptbase.ts
dpvc: No, this is right.  The characters normally include their italic correction (so the bases have italic correction already because that's normally added by TeX except in a few situations).  The `getBaseWidth()` function is supposed to return the width of the base when its italic correction is removed (rather than the actual bounding box size of the base, which includes it).  So that's why it is removed here.

dpvc: I will update the comment above.

dpvc:
> Otherwise, the code is a lot cleaner now!

Thanks!  I think moving things to the common wrappers and scriptbase and giving them names rather than just doing computation helped a lot in that respect.  Always helpful to go over the whole thing once it is in place, and the munderover/msubsup code was particularly intertwined and complicated, so that was a valuable exercise.

PR #635, opened 2021-03-17
Allow 'single-character' mi's to include trailing combining characters (mathjax/MathJax#2617)

dpvc:
This PR doesn't sound combining characters when counting the length of the content of an `mi` element in order to determine the `mathvariant` to use.  This allows things like `\not a` to render as an italic "a" rather than an upright one.

Resolves issue mathjax/MathJax#2617.

### DIFF
--- a/ts/core/MmlTree/MmlNodes/mi.ts
+++ b/ts/core/MmlTree/MmlNodes/mi.ts
@@ -45,7 +45,8 @@ export class MmlMi extends AbstractMmlTokenNode {
   /**
    * Pattern for single-character texts
    */
-  public static singleCharacter: RegExp = /^[\uD800-\uDBFF]?.$/;
+  public static singleCharacter: RegExp =
+    /^[\uD800-\uDBFF]?.[\u0300-\u036F\u1AB0-\u1ABE\u1DC0-\u1DFF\u20D0-\u20EF]*$/;
 
   /**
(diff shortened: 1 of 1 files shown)


zorkow (approved):
I note that you are making provisions for characters with combiners >1000.
I never got those to work. E.g., `'\u0041\u0300'.normalize()` works fine, however, `'\u0041\u1ab0'.normalize()` still gives me two characters. Am I doing something wrong?

dpvc:
> `'\u0041\u0300'.normalize()` works fine, however,  `'\u0041\u1ab0'.normalize()` still gives me two characters. Am I doing something wrong?

No, that is what should happen.  `'\u0041\u0300'.normalize()` gives you one character because there is a Unicode point U+00C0 for CAPITAL LETTER A WITH GRAVE, so the two can be combined into a single Unicode character.  But U+1AB0 is the COMBINING DOUBLE CIRCUMFLEX ABOVE, but there is no single-character glyph for A with two circumflex accents, so this has to stay two characters.  The point is that we want to count that as a single character, even though it is technically two, for the purposes of deciding whether it should be `mathvariant="italic"` or `mathvariant="normal"`.  The combining characters shouldn't make it count as two (causing it to become upright instead of italics), as it is semantically a single accented character, just like your first example.

PR #636, opened 2021-03-18
Add a small separation between base and scripts.  (mathjax/MathJax#2406)

dpvc:
This PR modifies the super- and subscript positions to leave a small amount of space between the base and the scripts.  In TeX, this is handled by italic correction values (that are different for each character), but we don't have that information in the font tables.  We may be able to add that during the font update this summer, so this is a bit of a hack to improve the layout until then.  It's not perfect, but it helps.  See, for example `\beta^p` before and after:

Before:  before

After: after

I also moved the comments that identified what `scriptbase` functions are for `msub` and `msup`, which end up making it look like functions were moved.

Addresses issue mathjax/MathJax#2406.


zorkow (approved):
lgtm.

PR #637, opened 2021-03-18
Ignore * in \\* and \\*[...] constructions.  (mathjax/MathJax#2433)

dpvc:
This PR causes the `\\` command to ignore a following asterisk, if there is one.  (The asterisk is for controlling page breaks, but it makes sense for MathJax to just ignore it.)

Resolves issue mathjax/MathJax#2433.

### DIFF
--- a/ts/input/tex/base/BaseMethods.ts
+++ b/ts/input/tex/base/BaseMethods.ts
@@ -1221,16 +1221,21 @@ BaseMethods.Cr = function(parser: TexParser, name: string) {
  */
 BaseMethods.CrLaTeX = function(parser: TexParser, name: string, nobrackets: boolean = false) {
   let n: string;
-  if (!nobrackets && parser.string.charAt(parser.i) === '[') {
-    let dim = parser.GetBrackets(name, '');
-    let [value, unit, ] = ParseUtil.matchDimen(dim);
-    // @test Custom Linebreak
-    if (dim && !value) {
-      // @test Dimension Error
-      throw new TexError('BracketMustBeDimension',
-                          'Bracket argument to %1 must be a dimension', parser.currentCS);
+  if (!nobrackets) {
+    if (parser.string.charAt(parser.i) === '*') {  // The * controls page breaking, so ignore it
+      parser.i++;
     }
-    n = value + unit;
+    if (parser.string.charAt(parser.i) === '[') {
+      let dim = parser.GetBrackets(name, '');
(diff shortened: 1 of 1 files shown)


dpvc:
PS, change the settings to ignore white-space changes for easier viewing.

ts/input/tex/base/BaseMethods.ts
```diff
@@ -1221,16 +1221,21 @@ BaseMethods.Cr = function(parser: TexParser, name: string) {
...
-    let [value, unit, ] = ParseUtil.matchDimen(dim);
-    // @test Custom Linebreak
-    if (dim && !value) {
-      // @test Dimension Error
-      throw new TexError('BracketMustBeDimension',
-                          'Bracket argument to %1 must be a dimension', parser.currentCS);
+  if (!nobrackets) {
+    if (parser.string.charAt(parser.i) === '*') {  // The * controls page breaking, so ignore it
+      parser.i++;
+    }
```
zorkow, line 1227: I assume `GetStar` would not work as it ignores spaces.

ts/input/tex/base/BaseMethods.ts
```diff
@@ -1221,16 +1221,21 @@ BaseMethods.Cr = function(parser: TexParser, name: string) {
...
+    if (parser.string.charAt(parser.i) === '[') {
```
zorkow, line 1228: This treatment has always been slightly problematic. Our newline is used in math array structures, so spaces after `\\` would actually have to be treated accordingly. That is, the current treatment is correct for AMS environments but should be ignored in `array` and `eqnarray`. 

In MJ2 base and AMS arrays were treated the same, so this could not really be resolved. However, now they are separated, so we could do that. 
But it's not for this PR to resolve this. Just pointing out that this is an issue waiting to happen.

zorkow (approved):
Can you please add a `TODO` note reflecting my comment?

ts/input/tex/base/BaseMethods.ts
dpvc: Yes, the direct check rather than `getStar()` was to not ignore spaces.

ts/input/tex/base/BaseMethods.ts
dpvc: OK, good to know.  I wasn't aware of the distinction.  We can keep it in mind, as you say.

dpvc:
> Can you please add a TODO note reflecting my comment?

Done.

PR #639, opened 2021-03-18
Fix script capital I (mathjax/MathJax-src#2555)

dpvc:
This PR fixes the `\mathsrc{I}` issue raised in mathjax/MathJax#2555.  It fixes the font data for both CHTML and SVG (and the common data between them).  I updated the code used to generate the data from the original v2 files, so the changes will not be lost if they need to be run in the future.  The issue was actually in the original v2 files, which had the wrong glyph mapped to U+2110.  (So this needs to be fixed in v2 as well.)


zorkow (approved):
lgtm.

PR #640, opened 2021-03-18
Better selectors for SVG path widths.  (mathjax/MathJax#2618)

dpvc:
This PR uses more specific CSS selectors for the SVG glyphs paths, and makes sure they apply to both `path` and `use` elements.

Resolves issue mathjax/MathJax#2618.


zorkow (approved):
lgtm.

PR #641, opened 2021-03-18
Remove role='presentation' attributes.  (mathjax/MathJax#2608)

dpvc:
This PR removes the unneeded `role="presentation"` attributes that are inserted by the assistive MathML extension.  Since these were being applied to custom element that had no role to begin with, there is no need for these.

Resolves issue mathjax/MathJax#2608.


zorkow (approved):
lgtm.

PR #642, opened 2021-03-18
Trim TeX and annotations before copying to clipboard.  (mathjax/MathJax#2574)

dpvc:
This PR trims TeX code and annotations before copying to the clipboard.

Resolves issue mathjax/MathJax#2574.


zorkow (approved):
lgtm.

PR #644, opened 2021-03-19
Make path resolving more flexible, and don't fail when loading entity files. (mathjax/MathJax#2650)

dpvc:
This PR adds a function list to use for resolving the path to a component file (i.e., package), and populates it with several default filters (the ones that were in the original `resolvePath()` method).  This allows extensions to add or remove the filters as needed, and is used by the MathML input component to handle loading entity files when needed.

In the past, the Entities object would try to load entity definitions when an entity wasn't found in its default table.  That works for node applications, but for in-browser use, which requires components, not just files, that caused MathJax to fail.  (Fortunately, the browser usually handles the entities, but it could cause problems with entities not known in the browser, and in some other edge cases).

This PR resolves that problem by adding a filter to remap attempts to load an entity file to instead load the "all entities" component.

Resolves issue mathjax/MathJax#2650.

### DIFF
--- a/ts/components/loader.ts
+++ b/ts/components/loader.ts
@@ -30,11 +30,19 @@ import {Package, PackageError, PackageReady, PackageFailed} from './package.js';
 export {Package, PackageError, PackageReady, PackageFailed} from './package.js';
 export {MathJaxLibrary} from './global.js';
 
+import {FunctionList} from '../util/FunctionList.js';
+
 /*
  * The current directory (for webpack), and the browser document (if any)
  */
 declare var __dirname: string;
 declare var document: Document;
+
+/**
+ * Function used to determine path to a given package.
+ */
+export type PathFilterFunction = (data: {name: string, original: string, addExtension: boolean}) => boolean;
+export type PathFilterList = {[name: string]: PathFilterFunction};
 
 /**
  * Update the configuration structure to include the loader configuration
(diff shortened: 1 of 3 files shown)


zorkow (commented):
I'd like to experiment with this first. But I have enough of code reviewing for now...

dpvc:
> I'd like to experiment with this first. But I have enough of code reviewing for now...

No problem.  Thanks for all the reviews.  You have done yeoman work today!

ts/components/loader.ts
```diff
@@ -65,10 +73,53 @@ export interface MathJaxObject extends MJObject {
     preLoad: (...names: string[]) => void;            // Indicate that packages are already loaded by hand
     defaultReady: () => void;                         // The function performed when all packages are loaded
     getRoot: () => string;                            // Find the root URL for the MathJax files
+    pathFilters: PathFilterList;                      // the filters to use for looking for package paths
   };
   startup?: any;
 }
 
+/**
+ * Functions to used to filter the path to a package
```
zorkow, line 82: typo.

ts/components/loader.ts
```diff
@@ -65,10 +73,53 @@ export interface MathJaxObject extends MJObject {
...
+      data.name = '[mathjax]/' + name.replace(/^\.\//, '');
+    }
+    if (data.addExtension && !name.match(/\.[^\/]+$/)) {
+      data.name += '.js';
+    }
+    return true;
+  },
+
+  /**
+   * Revursively replace path prefixes (e.g., [mathjax], [tex], etc.)
```
zorkow, line 110: typo

ts/components/loader.ts
```diff
@@ -65,10 +73,53 @@ export interface MathJaxObject extends MJObject {
...
+   */
+  prefix: (data) => {
+    let match;
+    while ((match = data.name.match(/^\[([^\]]*)\]/))) {
+      if (!CONFIG.paths.hasOwnProperty(match[1])) break;
```
zorkow, line 115: So what if the the path that is used actually has a `[path]` part and components need to be replace _after_ that?
We simply don't allow that?

zorkow (approved):
One question and a couple of typos.

ts/components/loader.ts
dpvc: I'm sorry, I'm not sure I understand the question.

Because this is a FunctionList, which is a PrioritizedList, you can add filters into the list at any location, so you can have a filter that adjusted the path after `[path]` has already been resolved, or you could add a filter that modifies the path before that.  It would depend on the need.  Prior to this PR we don't have any means of doing that (other than using the `source` configuration to directly map a request to its location).

I'm sure I'm not getting what you are driving at.

dpvc: For the record, we discussed this face-to-face, and I now understand that you were asking about an actual file name that contains `[...]`.  Since the substitutions are only made when (a) the `[...]` is at the start of the path name, and (b) only when there is an actual path for that prefix defined in the `paths` object, it is unlikely that this will cause a problem, and if it did, one could always use `./[prefix]...` to prevent this filter from modifying it.

PR #645, opened 2021-03-19
This PR corrects errors in the horizontal stretchy data for several characters.

dpvc:
This PR fixes the stretchy data for several characters where it was incorrectly translated from the v2 data.  In particular, this fixes the data for `\widehat` and `\widetilde`, as well as U+00AF (MACRON) and U+003D (EQUAL SIGN).

See also the comment below for additional changes that were added in a second commit.


dpvc:
I've added a commit that fixes additional data (adds the missing `variants` array for characters that need their `schar` array to be mapped to non-standard variants).  It also fixes the common mo wrapper to actually *use* the `schar` array (which was being ignored), which was the cause of mathjax/MathJax#2507, so that issue is fixed by this PR as well.

zorkow (approved):
lgtm.

PR #647, opened 2021-03-22
Update the MathDocument reset() command to allow resetting input and output jax.

dpvc:
As discussed in #570, we wanted a more general reset facility so that you can ask the MathDocument to reset, and it will reset its input and output jax automatically.  This PR extends the current `reset()` method to do that by allowing you to specify what items to be reset.  The default is still just the processed bits (since that's what the documented behavior is), but you can request the input or output jax, or all possible resets.  It adds the generic `reset()` methods to the input and output tax objects, and implements them in SVG and CHTML output to reset the font caches, and in TeX input to reset the tags.  It updates the Menu code to call the general input-jax reset function rather than including tex-specific code, and it updates the Startup object to create reset commands that use the new generic reset rather than making only a tex-specific one.


ts/core/InputJax.ts
```diff
@@ -1,3 +1,4 @@
+
```
zorkow, line 1: Added by accident?

zorkow (approved):
Just one whitespace line to remove. O/w lgtm.

ts/core/InputJax.ts
dpvc: Most likely.

PR #648, opened 2021-03-22
Add bolding data for the -tex-mathit variant (mathjax/MathJax#2502)

dpvc:
This PR adds data so that `\boldsymbol` will interact properly with `\mathit`.  So `\boldsymbol{\mathit{1}}` should produce `<mi mathvariant="bold-italic">1</mi>`.  Note, however, that the MathJax TeX fonts only include bold-italic alphabetic characters, so this will still not show a bold-italic 1.  That would require a font with more bold-italic coverage, such as STIX2.

Resolves issue mathjax/MathJax#2502


zorkow (approved):
lgtm.

PR #649, opened 2021-03-22
Allow stylesheet to be obtained even after it was created.  (mathjax/MathJax#2542)

dpvc:
This PR allows the `MathJax.svgStyleSheet()` or `MathJax.chtmlStyleSheet()` functions to return the stylesheet properly even after it has been created already, while still preventing the stylesheet from being updated unnecessarily.

Resolves issue mathjax/MathJax#2542.


zorkow (approved):
lgtm.

PR #650, opened 2021-03-22
Add ability to pre-define environments like for macros.  (mathjax/MathJax#2505)

dpvc:
This PR adds the ability to pre-define environments (just like macros) to the configmacros extension.  It adds a new configuration option `environments` like the `macros` one for the `tex` block.  

Resolves issue mathjax/MathJax#2505.


zorkow (approved):
lgtm.

PR #651, opened 2021-03-23
Improve heights of successively larger multi-character surds. (mathjax/MathJax#2658)

dpvc:
This PR adds the ability to make stretchy multi-character assemblies only use the full size of the extenders.  In TeX, this is used by stretchy surds, for example, to make them grow at a standard rate.  E.g., 

``` latex
\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{\sqrt{1-x^2}}}}}}}}}}
```

now produces

now

with this PR (which is consistent with LaTeX), but used to produce

then

without it (which is not consistent with LaTeX).

Resolves issue mathjax/MathJax#2658.

### DIFF
--- a/ts/output/common/Wrappers/mo.ts
+++ b/ts/output/common/Wrappers/mo.ts
@@ -106,6 +106,16 @@ export interface CommonMo extends AnyWrapper {
    * @return {number[]}        The height and depth for the vertically stretched delimiter
    */
   getBaseline(WHD: number[], HD: number, C: DelimiterData): number[];
+
+  /**
+   * Determine the size of the delimiter based on whether full extenders should be used or not.
+   *
+   * @param {number} D          The requested size of the delimiter
+   * @param {DelimiterData} C   The data for the delimiter
+   * @return {number}           The final size of the assembly
+   */
+  checkExtendedHeight(D: number, C: DelimiterData): number;
+
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ts/output/common/Wrappers/mo.ts
```diff
@@ -357,6 +367,22 @@ export function CommonMoMixin<T extends WrapperConstructor>(Base: T): MoConstruc
       return [h - d, d];
     }
 
+    /**
+     * Determine the size of the delimiter based on whether full extenders should be used or not.
+     *
+     * @param {number} D          The requested size of the delimiter
+     * @param {DelimiterData} C   The data for the delimiter
+     * @return {number}           The final size of the assembly
+     */
```
zorkow, line 376: This should just be an `@override` jsdoc.

zorkow (approved):
One minor issue. O/w lgtm.

ts/output/common/Wrappers/mo.ts
dpvc: I've never fully understood when to use `@override`.  You seem to be saying that if it is described in the interface, then the code that implements that interface should be marked `@override`.  In that case, most of the jsdoc in this class (and in lost of other places) should be turned into `@override`.  Perhaps another cleanup PR is in order?

PR #652, opened 2021-03-23
Update operator dictionary to the current state of Appendix C of the spec

dpvc:
This updates the operator dictionary to include some new characters that were not in Appendix C back when the data was created for v2.  There is also a correction to U+2032 (which looks like a deletion and addition since there are other characters added around it), and one character removed (not sure why they did that).

Not much to check, here, really.

### DIFF
--- a/ts/core/MmlTree/OperatorDictionary.ts
+++ b/ts/core/MmlTree/OperatorDictionary.ts
@@ -1288,5 +1288,10 @@ export const OPTABLE: {[form: string]: OperatorList} = {
 //
 //  These are not in the W3C table, but we need them for \widehat and \underline
 //
-OPTABLE['infix']['^'] = MO.WIDEREL;
-OPTABLE['infix']['_'] = MO.WIDEREL;
+OPTABLE.infix['^'] = MO.WIDEREL;
+OPTABLE.infix['_'] = MO.WIDEREL;
+
+//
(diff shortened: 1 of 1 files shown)


ts/core/MmlTree/OperatorDictionary.ts
```diff
@@ -1222,8 +1242,8 @@ export const OPTABLE: {[form: string]: OperatorList} = {
     '\u2AD9': MO.REL,        // element of opening downwards
     '\u2ADA': MO.REL,        // pitchfork with tee top
     '\u2ADB': MO.REL,        // transversal intersection
-    '\u2ADC': MO.REL,        // forking
     '\u2ADD': MO.REL,        // nonforking
+    '\u2ADD\u0338': MO.REL,  // nonforking with slash
```
zorkow, line 1246: So `2ADC` is supplanted by combining `2ADD` with a slash?
What if `2ADC` occurs in the input?

zorkow (approved):
One question.

ts/core/MmlTree/OperatorDictionary.ts
dpvc: It was remove from the table in Appendix C of the spec, so I removed it here.  I will add it back in at the bottom, with the extra additions.

PR #654, opened 2021-03-23
Allow MathML to mark mathaccents, so output is the same as TeX input.

dpvc:
This PR allows the `MmlMo` object to mark itself as having a math-accent so that MathML input will display the same as TeX input for these accents, without the need for the `data-mjx-accent="true"` attribute.  That attribute is no longer applied in serialized MathML output, but you can still use it on MathML input to control this internal setting (either to prevent an accent from getting it, or to force one that isn't usually marked that way).

This also fixes a problem with the `TeXAtom` and `mstyle` element's `notParent()` methods to have them not be parents only if they contain a single element (otherwise, they ARE the parent, effectively as an `mrow`).  Without this, something like `\lim_{n\to\infty}` would incorrectly mark the arrow as a mathaccent, as the group `n\to\infty` is in a `TeXAtom` and so the parent would incorrectly be the `munder`.

### DIFF
--- a/ts/core/MmlTree/MmlNodes/TeXAtom.ts
+++ b/ts/core/MmlTree/MmlNodes/TeXAtom.ts
@@ -65,7 +65,7 @@ export class TeXAtom extends AbstractMmlBaseNode {
    * @override
    */
   public get notParent() {
-    return true;
+    return this.childNodes[0].childNodes.length === 1;
   }
 
   /**
--- a/ts/core/MmlTree/MmlNodes/TeXAtom.ts
+++ b/ts/core/MmlTree/MmlNodes/TeXAtom.ts
@@ -65,7 +65,7 @@ export class TeXAtom extends AbstractMmlBaseNode {
    * @override
    */
   public get notParent() {
-    return this.childNodes[0].childNodes.length === 1;
+    return this.childNodes[0] && this.childNodes[0].childNodes.length === 1;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


ts/core/MmlTree/MmlNodes/TeXAtom.ts
```diff
@@ -65,7 +65,7 @@ export class TeXAtom extends AbstractMmlBaseNode {
    * @override
    */
   public get notParent() {
-    return true;
+    return this.childNodes[0].childNodes.length === 1;
```
zorkow, line 68: Can you add an additional guard `this.childNodes[0] && ...` ?

ts/core/MmlTree/MmlNodes/mo.ts
```diff
@@ -471,4 +490,17 @@ export class MmlMo extends AbstractMmlTokenNode {
     this.setProperty('primes', primes);
   }
 
+  /**
+   * Determine of the mo is a mathaccent character
```
zorkow, line 494: typo

ts/core/MmlTree/MmlNodes/mstyle.ts
```diff
@@ -58,7 +58,7 @@ export class MmlMstyle extends AbstractMmlLayoutNode {
    * @override
    */
   public get notParent() {
-    return true;
+    return this.childNodes[0].childNodes.length === 1;
```
zorkow, line 61: ditto.

zorkow (approved):
Couple of minor issues.

PR #655, opened 2021-03-23
Add the entities needed by TeX into the main entities table.

dpvc:
This PR adds the few entities needed by the TeX input jax into the main entity table so that we don't have to load the auxiliary tables for just one or two entities.  Also, replace the undefined `rquote` entity with `rsquo`.


zorkow (approved):
lgtm.

PR #656, opened 2021-03-24
Fix MathML spacing rules for mo elements.  (mathjax/MathJax#2392)

dpvc:
This PR modifies the MathML spacing rules for `<mo>` (and embellished `<mo>`) to correspond to Neil's explanation of the rules in mathjax/MathJax#2392.  This means that space is only added when the (embellished) operator is inside an mrow (or inferred mrow) that has more than one element.  This also makes the spacing between two successive `<mo>` elements be the maximum of the `rspace` of the first and the `lspace` of the second (as opposed to the sum of the two, as in the past). 

This only affects when MathML spacing has been selected for the output jax, not the TeX spacing.

Finally, this also fixes the `coreParent()` function, which used to skip over `<mstyle>` elements, and so would incorrectly report the parent for `<mstyle><mo>+</mo></mstyle>` as the `<mo>` element rather than the `<mstyle>`.  Similarly, the `isTopEmbellished()` function incorrectly reported the `<mo>` as the top embellished operator.

Resolve issue mathjax/MathJax#2392.

### DIFF
--- a/ts/core/MmlTree/MmlNodes/mo.ts
+++ b/ts/core/MmlTree/MmlNodes/mo.ts
@@ -150,7 +150,7 @@ export class MmlMo extends AbstractMmlTokenNode {
     let math = this.factory.getNodeClass('math');
     while (parent && parent.isEmbellished && parent.coreMO() === this && !(parent instanceof math)) {
       embellished = parent;
-      parent = (parent as MmlNode).Parent;
+      parent = (parent as MmlNode).parent;
     }
     return embellished;
   }
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -533,20 +533,11 @@ export class CommonWrapper<
                    Math.max(0, this.length2em(attributes.get('rspace'))) :
                    MathMLSpace(isScript, node.rspace));
     //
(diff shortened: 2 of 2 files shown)


zorkow (commented):
Two things:
* This only applies to `mrow` elements? Not any other "implied"  `mrow`?
* I have to admit I cannot really see the difference, but maybe I misunderstand what Neil's issue is.

Finally I think it is just wrong to ascribe meaning to a grouping element like `mrow`! But then again that is a major fault of MathML.

zorkow (approved):
f2f: lgtm.

dpvc:
For the historical record:

> This only applies to `mrow` elements? Not any other "implied" `mrow`?

The `isKind('mrow')` is actually an `instanceof` check under the hood, so it will be true for subclasses of `mrow` as well.  The `inferredMrow` is a subclass of `mrow`, it will also make that test true.  So the change applies to both explicit and implicit mrows.

> I have to admit I cannot really see the difference, but maybe I misunderstand what Neil's issue is.

It only affects the output when `mathmlSpacing: true` is set in the output jax options, as it is about the MathML spacing rules, not the (default) TeX spacing rules.

PR #657, opened 2021-03-24
Better handling of min height and width for SVG (so they don't disappear).

dpvc:
This PR attempts (again) to fix the problem with SVG element disappearing if they have 0 height or width.  This uses CSS `min-height` and `min-width` to make sure the SVG has some width and height, and also makes sure the `viewBox` has non-zero width and height (this time using the em-size to make it what should be one pixel minimum.


zorkow (approved):
lgtm.

PR #658, opened 2021-03-24
Have HTMLDomStrings skip node types that it can't process. (mathjax/MathJax#2662)

dpvc:
This PR arranges for the HTMLDomStrings object to skip node types that it doesn't recognize (it can handle text, comments, and Elements).  This fixes the problem reported in mathjax/MathJax#2662 where ProcessingInstructions would cause MathJax to crash.

Resolves issue mathjax/MathJax#2662.

### DIFF
--- a/ts/handlers/html/HTMLDomStrings.ts
+++ b/ts/handlers/html/HTMLDomStrings.ts
@@ -242,7 +242,7 @@ export class HTMLDomStrings<N, T, D> {
   }
 
   /**
-   * Handle an unknown node type (nodeType other than 1, 3, 7)
+   * Handle an unknown node type (nodeType other than 1, 3, 8)
    *
    * @param {N} node           The node to process
    * @param {boolean} ignore   Whether we are currently ignoring content
--- a/ts/handlers/html/HTMLDomStrings.ts
+++ b/ts/handlers/html/HTMLDomStrings.ts
@@ -242,6 +242,18 @@ export class HTMLDomStrings<N, T, D> {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ts/handlers/html/HTMLDomStrings.ts
```diff
@@ -241,6 +241,18 @@ export class HTMLDomStrings<N, T, D> {
     return [next, ignore];
   }
 
+  /**
+   * Handle an unknown node type (nodeType other than 1, 3, 7)
+   *
```
zorkow, line 246: Typo: `7 => 8`

ts/handlers/html/HTMLDomStrings.ts
```diff
@@ -241,6 +241,18 @@ export class HTMLDomStrings<N, T, D> {
...
+   * @param {N} node           The node to process
+   * @param {boolean} ignore   Whether we are currently ignoring content
+   * @return {N|T}             The next element to process
+   */
+  protected handleOther(node: N, _ignore: boolean): N | T {
```
zorkow, line 251: Why do we need `_ignore`?

zorkow (approved):
Minor things.

ts/handlers/html/HTMLDomStrings.ts
dpvc: Fixed.

ts/handlers/html/HTMLDomStrings.ts
dpvc: This is for in case someone needs to subclass this class in order to handle additional node types, so they have access to whether the node is being ignored or not.

PR #659, opened 2021-03-25
Add ability to specify additional textmacros packages.

dpvc:
This PR adds the ability to specify additional packages to be loaded in the textmacros text parser.  Use

``` javascript
MathJax = {
  tex: {
    textmacros: {
      packages: {'[+]', [...]}
    }
  }
}
```

to specify the packages to include, e.g., `textmacros: {packages: {'[+]': ['bbox']}` to allow `\bbox` inside '\text{}`.

This also makes the main `textBase` configuration a named one rather than a local one, so that it can be included in the package list (as `text-base`, just like `base` for the main TeX base package).   That means it will show up as a package in the Lab package list, so `v3-lab.js` should probably be modified to not create a checkbox for it, since it is not appropriate outside of `\text{}`.

### DIFF
--- a/ts/input/tex/textmacros/TextMacrosConfiguration.ts
+++ b/ts/input/tex/textmacros/TextMacrosConfiguration.ts
@@ -37,7 +37,7 @@ import './TextMacrosMappings.js';
 /**
  *  The base text macro configuration (used in the TextParser)
  */
-export const textBase = Configuration.local({
+export const textBase = Configuration.create('text-base', {
   handler: {
     character: ['command', 'text-special'],
     macro: ['text-macros']
@@ -103,8 +103,7 @@ export const TextMacrosConfiguration = Configuration.create('textmacros', {
     //  Create the configuration and parseOptions objects for the
     //    internal TextParser and add the textBase configuration.
     //
-    const textConf = new ParserConfiguration([]);
-    textConf.append(textBase);
+    const textConf = new ParserConfiguration(jax.parseOptions.options.textmacros.packages);
     textConf.init();
     const parseOptions = new ParseOptions(textConf, []);
     parseOptions.options = jax.parseOptions.options;      // share the TeX options
@@ -129,5 +128,10 @@ export const TextMacrosConfiguration = Configuration.create('textmacros', {
     //
     const config = data.data.packageData.get('textmacros');
     config.parseOptions.nodeFactory.setMmlFactory(config.jax.mmlFactory);
-  }]
+  }],
+  options: {
+    textmacros: {
+      packages: ['text-base']    // textmacro packages to load
+    }
+  }
 });
--- a/ts/input/tex/textmacros/TextMacrosConfiguration.ts
+++ b/ts/input/tex/textmacros/TextMacrosConfiguration.ts
@@ -37,7 +37,8 @@ import './TextMacrosMappings.js';
 /**
  *  The base text macro configuration (used in the TextParser)
  */
-export const textBase = Configuration.create('text-base', {
+export const TextBaseConfiguration = Configuration.create('text-base', {
+  parser: 'text',
   handler: {
     character: ['command', 'text-special'],
     macro: ['text-macros']
@@ -103,7 +104,7 @@ export const TextMacrosConfiguration = Configuration.create('textmacros', {
     //  Create the configuration and parseOptions objects for the
     //    internal TextParser and add the textBase configuration.
     //
-    const textConf = new ParserConfiguration(jax.parseOptions.options.textmacros.packages);
+    const textConf = new ParserConfiguration(jax.parseOptions.options.textmacros.packages, ['tex', 'text']);
     textConf.init();
     const parseOptions = new ParseOptions(textConf, []);
     parseOptions.options = jax.parseOptions.options;      // share the TeX options
--- a/ts/input/tex.ts
+++ b/ts/input/tex.ts
@@ -105,7 +105,7 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
    * @return {Configuration} The configuration object.
    */
   protected static configure(packages: (string | [string, number])[]): ParserConfiguration {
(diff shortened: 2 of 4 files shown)


ts/input/tex/textmacros/TextMacrosConfiguration.ts
```diff
@@ -37,7 +37,7 @@ import './TextMacrosMappings.js';
 /**
  *  The base text macro configuration (used in the TextParser)
  */
-export const textBase = Configuration.local({
+export const textBase = Configuration.create('text-base', {
```
zorkow, line 40: Is there still a need to bind and export this?

zorkow (commented):
One problem with this solution is that now all the `local` configurations are named and can be selected explicitly. That was originally what we wanted to avoid by allowing ephemeral configurations.

dpvc:
> One problem with this solution is that now all the `local` configurations are named and can be selected explicitly. 

It's not all of them, only the one from textmacros (and there is only one of it, just as there is only one `base` configuration).  That is necessary in order to make it selectable in the `tex.textmacros.packages` array, just as the `base` array is named and selectable in the main `tex.packages` array. This was to make things fully parallel between the two parsers.  

It would be possible to keep the text-base configuration local, and append it as before, but that would mean the two package lists operate differently.  We can do that, but I thought the consistency was useful.

Note that there are still local configurations in the `ams`, `newcommand`, and `configmacros` extensions, so the local configurations are still being used in other places where they are useful.

> That was originally what we wanted to avoid by allowing ephemeral configurations.

This one is not an ephemeral one, it is just like `base`, but for textmacros.  It was `local` before so it was private, meaning that you couldn't include it in `tex.packages` (not because it was ephemeral).  Unless you want to have a separate registry for textmacro packages, or a way or marking a package as being text macros rather than math macros, then this is a potential pitfall that people will have to avoid.

I'm happy to have the configurations be marked as text versus math, and I think that woudl make sense if we are going to make extensions to textmacros.

zorkow:
I agree that consistency is useful and that it is also very useful to be able to pass existing packages/configurations to the textparser. However, I still believe there is a need to pass macros to textparser without them becoming available globally.

I have continued a stream of consciousness via email.

zorkow:
I thought we had agreed on moving this to 3.2.

dpvc:
@zorkow, I just realized why I hadn't merged the `parser-packages-A` branch:  because it was targeted to this branch, and so now those changes are included here.  That means you may want to just look at the [first commit](https://github.com/mathjax/MathJax-src/pull/659/commits/9dfa6f18ab24517d22222314bb89c06e8c097d57) instead of the full branch (which now includes things you've already reviewed).  Sorry about that!

zorkow (approved):
Lgtm.

zorkow:
> @zorkow, I just realized why I hadn't merged the `parser-packages-A` branch: because it was targeted to this branch, and so now those changes are included here. That means you may want to just look at the [first commit](https://github.com/mathjax/MathJax-src/pull/659/commits/9dfa6f18ab24517d22222314bb89c06e8c097d57) instead of the full branch (which now includes things you've already reviewed). Sorry about that!

No problems. I've only looked at the  relevant commits.

pkra:
FYI this breaks existing extensions that are adding text macros using `TextBase.handler.macro.unshift('xhref-macros');` since TextBase was renamed TextBaseConfiguration.

dpvc:
That mechanism was a stop-gap approach until we had an official mechanism for adding to the text macros, which this PR provides.  Going forward, the proper way to add macros to the text processing is through the `textmacros.packages` configuration to the TeX input jax.  One defines a package as usual and adds that to the textmacros packages list just like you would for a math-mode package.

pkra:
> Going forward, the proper way to add macros to the text processing is through the textmacros.packages configuration to the TeX input jax. One defines a package as usual and adds that to the textmacros packages list just like you would for a math-mode package.

IIUC, this means the decision which macros are available in text mode can only be made by the person writing the configuration. If that's the case, then it seems unfortunate; as extension developer, I don't want to put that burden on my users.

dpvc:
> the decision which macros are available in text mode can only be made by the person writing the configuration

I'm sorry, I don't understand.  When you say "the configuration", which configuration are you talking about?  Do you mean the configuration being passed to the TeX input jax, or the configuration for the extension that adds text macros?  

If it is the TeX configuration, then this is exactly parallel to the current package mechanism for the math-mode macros.  The user must both load your extension and add it to the package list.  (Extensions don't modify the package list themselves.)  Similarly, if you want an extension that adds text-mode macros, you need to load that extension and configure the textmacros packages to include it.

It would be possible to have a tex extension that, when loaded in the tex packages, would add a package to the textmacros package list (you would probably need to set the package priority so that it modifies the textmacros packages before the textmacros package is configured).  I would not recommend that, however.

pkra:
>  When you say "the configuration", which configuration are you talking about? Do you mean the configuration being passed to the TeX input jax, or the configuration for the extension that adds text macros?

The former.

> If it is the TeX configuration, then this is exactly parallel to the current package mechanism for the math-mode macros. The user must both load your extension and add it to the package list. (Extensions don't modify the package list themselves.)

But an extension can include or load another extension easily enough.

> It would be possible to have a tex extension that, when loaded in the tex packages, would add a package to the textmacros package list [...]. I would not recommend that, however.

Fair enough and your call (obviously). Again, as an extension author, I (sometimes) don't want to put that burden on users of my extension; I'd rather have an \ensuremath-like behavior.

dpvc:
> But an extension can include or load another extension easily enough.

Can you give an example of what you mean by that?  

We may be thinking about different use cases.  I need to make sure that what is done works for web-based components, while your primary use is server-side.  The web-based components have to be a bit more careful about how they share code with other components.

Because there are two steps to using a TeX extension package (loading the code and adding it to the tex package list), I'm not clear on what you mean by "load another extension".  Do you mean accessing the code from another extension, or do you mean adding it to the tex packages list as well?

PR #661, opened 2021-03-26
Add support for restricting packages to particular named parsers (A)

dpvc:
This PR adds a parser name to each configuration (default is `tex`) and uses an array of names to indicate which packages are allowed for the given `ParserConfiguration` instance.  This allows the `text-base` configuration to be marked as for the text parser, so it will be rejected if included in the `tex.packages` list.

Pros:  This is easy to implement, and allows mixing and matching for packages with parsers, even if the package's parser is not a superclass of the given parser.

Cons:  You need to list all the superclasses explicitly if you want them to be allowed.  So to allow configuring the `TextParser` to accept TeX packages, you must indicate `['tex', 'text']` when creating the `ParserConfiguration` for the `TextParser`.

This is one proposal for how to handle this.

### DIFF
--- a/ts/input/tex/Configuration.ts
+++ b/ts/input/tex/Configuration.ts
@@ -74,6 +74,7 @@ export class Configuration {
                                   init?: ProtoProcessor<InitMethod>,
                                   config?: ProtoProcessor<ConfigMethod>,
                                   priority?: number,
+                                  parser?: string,
                                  } = {}): Configuration {
     let priority = config.priority || PrioritizedList.DEFAULTPRIORITY;
     let init = config.init ? this.makeProcessor(config.init, priority) : null;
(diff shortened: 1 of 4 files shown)


zorkow (approved):
I prefer this option.

ts/input/tex/Configuration.ts
```diff
@@ -284,13 +296,14 @@ export class ParserConfiguration {
    */
   public nodes: {[key: string]: any}  = {};
 
-
   /**
    * @constructor
    * @param {(string|[string,number])[]} packages A list of packages with
    *     optional priorities.
+   * @parm {string[]} parsers   The names of the parsers whose packages are supported
```
zorkow, line 303: Maybe rephrase a bit.

zorkow (approved):
I prefer this option.

ts/input/tex/Configuration.ts
dpvc: Done

PR #662, opened 2021-03-26
Add support for restricting packages to particular named parsers (B)

dpvc:
This PR adds a registry of parsers so that each new parser class registers itself under a given name.  The packages use that name to indicate which parser they are targeted at.  When the `ParserConfiguation` is created, it is given the name of the parser that it targets, and uses the registry to decide if the packages are appropriate.

Pros:  You don't have to list all the superclasses anywhere, as any subclass of parser A can load package targeted at parser A.

Cons:  You have less control over the packages in the sense that you can't exclude superclasses like you can in #661.  You also have to explicitly register the parser classes as you create them.

This is a second proposal for handling the situation.


zorkow (commented):
I prefer option #661 .

dpvc:
Closing in favor of #661

PR #663, opened 2021-03-26
Upgreek package

zorkow:
Implements [upgreek package](https://www.ctan.org/pkg/upgreek) using the `mathchar0mo` parsing method.


dpvc (approved):
One question about possibly reducing the size of the extension, but otherwise approved.

PR #665, opened 2021-03-26
Gensymb package

zorkow:
This implements the [gensymb package](https://ctan.org/pkg/gensymb) but only the good parts! In particular

* `\ohm` behaves as the normal `\Omega`, i.e., it can be affected by changes of fonts.
* All other symbols will remain in Roman font only.
* `\degree` and `\celsius` are mapped to the correct unicode points, instead of emulated with `^\circ`. This has the sideeffect, that unlike `gensymb` in LaTeX, `gensymb` in MathJax does __not__ throw a double superscript exception on `\degree\celsius`!


dpvc (approved):
This looks fine, but three questions:

1. Should `\ohm` generate U+2126 (OHM symbol) rather than U+03A9 (GREEK CAPITAL LETTER OMEGA)?
2. Should `\micro` produce an `<mo>`?  I'm not sure how units should be presented.
3. The MathJax fonts don't include characters for `\celsius`, `\perthousand`, and `\micro`, and I think there are no other extensions that add characters that we don't have in the fonts (though I may be wrong).  Do we want to change that precedent?  We can extend the character set during the font update this summer, so perhaps this isn't too important.

zorkow:
> This looks fine, but three questions:
> 
> 1. Should `\ohm` generate U+2126 (OHM symbol) rather than U+03A9 (GREEK CAPITAL LETTER OMEGA)?

Yes, it should. Changed.

> 2. Should `\micro` produce an `<mo>`?  I'm not sure how units should be presented.

True, they should probably all produce `mi` instead. E.g.,  `‰` should be treated similar to `%` which is an `mi`. We could even give them their unit class, although that would not work for `\micro` on its own.

> 3. The MathJax fonts don't include characters for `\celsius`, `\perthousand`, and `\micro`, and I think there are no other extensions that add characters that we don't have in the fonts (though I may be wrong).  Do we want to change that precedent?  We can extend the character set during the font update this summer, so perhaps this isn't too important.

I see what you mean. Both `\micro` and `\celsius` could be simulated, but I am a bit reluctant to do this as it takes away the semantic information inherent to the characters. So I would argue to change the precedent.

dpvc:
> So I would argue to change the precedent.

OK, I'm fine with that.  Remind me to add these into the next generation of the MathJax TeX fonts later this summer.

ts/input/tex/gensymb/GensymbConfiguration.ts
```diff
@@ -23,33 +23,42 @@
...
- * Ohm symbol as in gensymb. Usually upright, but can be affected by fonts.
+ * Handle characters that are known units.
+ * @param {TexParser} parser The current tex parser.
+ * @param {Symbol} mchar The parsed symbol.
  */
-new CharacterMap('gensymb-ohm', ParseMethods.mathchar7, {
-  ohm:            '\u03A9'
-});
+function mathcharUnit(parser: TexParser, mchar: Symbol) {
+  console.log(mchar.attributes);
```
dpvc, line 38: Don't forget to remove this log command.

zorkow: Good catch. Thanks.
Note that I've left `\micro` as a unit now.

dpvc: > Note that I've left `\micro` as a unit now.

Understood.

PR #666, opened 2021-03-28
support \mathsfit and \textsfit

tavin:
Implements https://github.com/mathjax/MathJax-src/issues/667.


dpvc:
Thanks for your interest in contributing to MathJax!

Was it turns out, this issue has been under discussion in mathjax/MathJax#2595, and there is a recent PR (#676) that handles this in a more general way, and that includes all the `\mathNN` (as well as `\symNN`) macros.  So while we appreciate the contribution, we won't be merging this particular request.

PR #670, opened 2021-03-29
Update SVG output so that following a link scrolls to the top of the tag.  (mathjax/MathJax#2588)

dpvc:
In Firefox, following a link in SVG output would scroll to the *bottom* of the linked content rather than the top.  In Safari, that would occur if the linked content was a `<text>` node, like from a `\tag` with `mtextInheritFont: true`.  This PR adjusts the output of elements with ID so that Firefox and Safari (and Chrome) all scroll to the top of the linked content.

That required shifting the element with the ID upward by its height, and then shifting its contents down by the sea amount (so that it is in the correct visual location, but the element's location for linking purposes is the top of the content.  A `<text>` element is also added so that Safari will scroll to the right location.

Since this required putting some unexpected elements as the children of the node with the `id` attribute, a new `firstChild()` method was added that skips over the unexpected nodes.  That also allows moving the `rect` that was used as the hotbox for an `href` to also be put inside the element with the `href` rather than just before it, as in the past.  (This corresponds to Peter's suggestion in mathjax/MathJax#2530.)  The `firstChild()` method skips the `href` element as well.

Resolves mathjax/MathJax#2588.

### DIFF
--- a/ts/output/svg/Wrapper.ts
+++ b/ts/output/svg/Wrapper.ts
@@ -141,16 +141,17 @@ CommonWrapper<
    * @returns {N}  The root of the HTML tree for the wrapped node's output
    */
   protected createSVGnode(parent: N): N {
+    this.element = this.svg('g', {'data-mml-node': this.node.kind});
     const href = this.node.attributes.get('href');
     if (href) {
       parent = this.adaptor.append(parent, this.svg('a', {href: href})) as N;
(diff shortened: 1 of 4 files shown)


zorkow (approved):
As my Mac is too decrepit, I could not check the issue/solution in Safari.
Firefox seems to work fine.

PR #671, opened 2021-03-29
Use a colon in link anchor ids so that links work in Safari in SVG output.  (mathjax/MathJax#2586)

dpvc:
This PR changes the tag id from `mjx-eqn-XXX` to `mjx-eqn:XXX`.  This is needed in order to get links to work in SVG in Safari.  I don't know exactly why that is needed, but it seems to need a character that requires escaping in URLs.  Go figure.

Resolves issue mathjax/MathJax#2586.


zorkow (approved):
Looks good. Again I did not manage to check in Safari.

PR #672, opened 2021-03-30
Add a 'nonce' attribute to inline stylesheets

christianp:
The output options now has a property 'nonce', which is a string to use as the value for a 'nonce' attribute on <style> tags created by the output jax. If it's not given, no 'nonce' attribute is added.

This fixes mathjax/MathJax#2665

Here's a tiny page to test:

```html
<!doctype html>
<html>
    <head>
        <script>
            window.MathJax = {
                chtml: {
                    nonce: 'hello'
                },
            };
        </script>
        <script src="es5/tex-chtml.js"></script>
    </head>
    <body>
        <p>This is \(x^2 + \int y\)</p>
    </body>
</html>
```

The `style.MJX-CHTML-styles` tag has an attribute `nonce="hello"`.

### DIFF
--- a/ts/output/common/OutputJax.ts
+++ b/ts/output/common/OutputJax.ts
@@ -98,7 +98,8 @@ export abstract class CommonOutputJax<
     displayIndent: '0',            // default for indentshift when set to 'auto'
     wrapperFactory: null,          // The wrapper factory to use
     font: null,                    // The FontData object to use
-    cssStyles: null                // The CssStyles object to use
+    cssStyles: null,               // The CssStyles object to use
+    nonce: null                    // The nonce value to apply to style tags created by MathJax
   };
 
   /**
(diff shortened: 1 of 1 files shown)


christianp:
The context menu also creates stylesheets, but it uses its own methods rather than the ones in OutputJax. I'm prepared to make the same changes to the mj-context-menu package, but where is the right place to pass it the nonce value? Is it possible to access the top-level options object from within the output jax and the context menu's constructors?

dpvc:
I don't think you will be able to get the menu to work, as there are lots of dynamically changing styles, from the position of the menu initially, to the placement of submenus, and other things that are not mediated by classes.  The creation of stylesheets is done here:

https://github.com/zorkow/context-menu/blob/8ddd26a41f834cd23b9bb20737dfae5fa9b05eb4/ts/css_util.ts#L274-L287

but this is inside a `namespace`, and is not exported, so it is not something you could get your hands on easily.

I think you may have to go without the menu.

zorkow:
The unorthodox use of namespaces are an artefact of the lack of  string `enum` types in Typescript `~v2.X` and my trying to achieve similar type checking behaviour as in `closure`. This will need an overhaul one of these days.

PR #673, opened 2021-03-30
Allow \middle to work when a style or color is open.  (mathjax/MathJax#2663)

dpvc:
The current implementation for `\middle` only works when there are no open style or color changes, and so it fails with

```
\left( \color{red} x \middle| y \right)
```

for example.  This PR introduces a `MiddleItem ` stack item that acts as a close item, so it will close any pending and style items.  The `LeftItem` looks for a `MiddleItem` and processes it (by adding the needed `mo`) without closing itself so that it will remain open until the following `RightItem` occurs.

In addition, the `\middle` and `\right` delimiters now save the color that they should be in, since `\left(\color{red}x\right)` colors the right brace red in actual LaTeX (and similarly for `\middle`), and apply those properly (they didn't before).

Finally, I made the `LeftRight` constructor take the actual delimiter (and color) as an argument so that you don't have to set it after it is created.

Resolves issue mathjax/MathJax#2663.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -122,7 +122,7 @@ namespace ParseUtil {
    * @param {string=} big Bigg command.
    */
   export function fenced(configuration: ParseOptions, open: string, mml: MmlNode,
-                         close: string, big: string = '') {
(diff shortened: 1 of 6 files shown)


zorkow (approved):
lgtm.

PR #674, opened 2021-03-30
Fix regression in tagging caused by #630

dpvc:
PR #630 was too aggressive about clearing the tag environment so that environments that allow multiple tags no longer worked.  This puts back the original `clearTag()` code, but resets the tag at the beginning of the equation so that it still solves the issue from mathjax/MathJax#2643.


zorkow (approved):
Tested and it works.

PR #675, opened 2021-03-31
Improve positioning of extenders for stretchy characters.  (mathjax/MathJax#2547)

dpvc:
This PR adjusts the CSS used for multi-glyph horizontal and vertical stretchy delimiters to make the pieces match up better.  It does this by making sure all the pieces are given the same height and depth (for horizontal) and width (for vertical).  The horizontal ones were the worst offenders, particularly for arrows, and this seems to resolve the problems, even when zoomed in or out.  I have tested in Firefox, Chrome, and Safari.

Resolves issue mathjax/MathJax#2547 and mathjax/MathJax#2642.


zorkow (approved):
Also tested on Brave and Edge.

PR #676, opened 2021-03-31
Add support for all \mathXYZ and \symXYZ macros using multi-letter <mi>.  (mathjax/MathJax#2595)

dpvc:
This PR adds support for multi-character `<mi>` elements when they are inside `\mathrm`, `\mathbf`, etc., and adds support for all the `\mathXYZ` and `\symXYZ` macros, so that all MathML variants can be produced.   This has been discussed extensively in mathjax/MathJax#2595, and this is the solution that we agreed on.

This also corrects an error in the variant names in the serialized output, and modifies the `autoOP` support to only turn normal variants into operators automatically (like `<mi>sin</mi>`), not the ones created via `\mathXYZ{}`.

Resolves issue mathjax/MathJax#2595.


ts/core/MmlTree/SerializedMmlVisitor.ts
```diff
@@ -203,6 +203,7 @@ export class SerializedMmlVisitor extends MmlVisitor {
     node.getProperty('variantForm') && this.setDataAttribute(data, 'alternate', '1');
     node.getProperty('pseudoscript') && this.setDataAttribute(data, 'pseudoscript', 'true');
     node.getProperty('mathaccent') && this.setDataAttribute(data, 'accent', 'true');
+    node.getProperty('autoOP') === false && this.setDataAttribute(data, 'auto-op', 'false');
```
pkra, line 206: This branch now has a merge conflict due to this line (since https://github.com/mathjax/MathJax-src/pull/654 was merged and removed the preceding line). 

Could you let me know whether this line should stay?

dpvc: Yes, this line should stay.

dpvc: But the preceding line should go.  Sorry if I misunderstood the question.

pkra: Thanks, Davide. All clear.

zorkow (approved):
This is great! SRE now gets `\mathrm{sin}` as `sine` rather than `normal s normal i normal n`.

dpvc:
Yes, that is one of the reasons I worked harder to get the multi-character `mi` elements, because you said it would help with SRE.  Glad that is one of the results!

PR #677, opened 2021-03-31
Add a11y enable options to menu handler so that they can be configured (mathjax/MathJax#2651)

dpvc:
This PR adds the a11y "enable" options to the menu handler so that a page author can include them in the MathJax configuration even if the extensions aren't loaded yet.  Without this, having `enableExporer`, for example, in the configuration would lead to an error about the option not being defined.  This defines them when the menu component is loaded, since it may load the a11y extensions later.

Resolves issue mathjax/MathJax#2651.

### DIFF
--- a/ts/ui/menu/MenuHandler.ts
+++ b/ts/ui/menu/MenuHandler.ts
@@ -167,7 +167,7 @@ export function MenuMathDocumentMixin<B extends A11yDocumentConstructor>(
       //    initially, and so would cause "undefined option" error messages
       //    if a user tries to configure them.  So we include them here.
       //    They are overridden by the options from the extensions when
-      //    those are laoded (via ...BaseDocument.OPTIONS).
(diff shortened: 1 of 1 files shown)


ts/ui/menu/MenuHandler.ts
```diff
@@ -162,6 +162,19 @@ export function MenuMathDocumentMixin<B extends A11yDocumentConstructor>(
      * @override
      */
     public static OPTIONS = {
+      //
+      //  These options are from the a11y extensions, which may not be loaded
+      //    initially, and so would cause "undefined option" error messages
+      //    if a user tries to configure them.  So we include them here.
+      //    They are overridden by the options from the extensions when
+      //    those are laoded (via ...BaseDocument.OPTIONS).
```
zorkow, line 170: typo

zorkow (approved):
Just a typo.

PR #678, opened 2021-03-31
Add ability to serialize annotation-xml content.

dpvc:
This PR adds a `serializeXML()` method for the DOMadaptor, and implements it in the HTML and Lite adaptors.  This is then used by the `XMLNode` class to get its serialized version for output in serialized MathML.  This allows `<annotation-xml>` element that contain HTML content with self-closing tags like `<input>` that don't usually produce valid XML to produce a serialized version that can be parsed by SRE as well as by MathJax.  This allows `<annotation-xml>` to be used with the assistive tools.

### DIFF
--- a/ts/adaptors/lite/Parser.ts
+++ b/ts/adaptors/lite/Parser.ts
@@ -355,9 +355,10 @@ export class LiteParser implements MinDOMParser<LiteDocument> {
     const attributes = adaptor.allAttributes(node).map(
       (x: AttributeData) => x.name + '="' + (CDATA[x.name] ? x.value : this.protectAttribute(x.value)) + '"'
     ).join(' ');
+    const content = this.serializeInner(adaptor, node, xml);
     const html =
       '<' + tag + (attributes ? ' ' + attributes : '')
-      + (SELF_CLOSING[tag] ? (xml ? ' />' : '>') : '>' + adaptor.innerHTML(node) + '</' + tag + '>');
+          + (content && !SELF_CLOSING[tag] ? `>${content}</${tag}>` : xml ? '/>' : '>');
     return html;
   }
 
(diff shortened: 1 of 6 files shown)


dpvc:
An example that used to fail is something like

``` html
  <semantics>
    <annotation-xml encoding="application/xhtml+xml">
      <input style="width: 1em;" xmlns="http://www.w3.org/1999/xhtml">
    </annotation-xml>
  </semantics>
</math>
```

This construction is used by the `\forminput` extension to insert input areas into MathJax expressions.

ts/adaptors/lite/Parser.ts
```diff
@@ -345,18 +345,19 @@ export class LiteParser implements MinDOMParser<LiteDocument> {
...
     const CDATA = (this.constructor as typeof LiteParser).CDATA_ATTR;
     const tag = adaptor.kind(node);
     const attributes = adaptor.allAttributes(node).map(
       (x: AttributeData) => x.name + '="' + (CDATA[x.name] ? x.value : this.protectAttribute(x.value)) + '"'
     ).join(' ');
     const html =
-      '<' + tag + (attributes ? ' ' + attributes : '') + '>'
-      + (SELF_CLOSING[tag] ? '' : adaptor.innerHTML(node) + '</' + tag + '>');
+      '<' + tag + (attributes ? ' ' + attributes : '')
+      + (SELF_CLOSING[tag] ? (xml ? ' />' : '>') : '>' + adaptor.innerHTML(node) + '</' + tag + '>');
```
zorkow, line 360: In XML any self-closing tag will close, regardless of whether there is `innerHTML` or not. Is that really intended?
In other words: `self_closing` is not the same as an empty tag?

zorkow (commented):
I have my doubts about whether self-closing tags are handled correctly.

ts/adaptors/lite/Parser.ts
dpvc: The purpose, here, is to get HTML to serialize as XML.  I'm not trying to handle all empty tags as self closing, as `<tag></tag>` is valid in that case.  This is to handle HTML where something like `<input...>` doesn't have a close tag or a `/>`, but needs to in order to be used in `<annotation-xml>` (unclosed tags cause parsing errors otherwise).  The `SELF_CLOSING` array lists those, and `innerHTML` uses that to generates HTML that doesn't have a close tag when it is not needed in HTML (like for `<input>`).  We don't want to do that for XML output.  Everything else gets a close tag already, so that is why `SELF_CLOSING` is the trigger for `/>` in this case.  Again, it is not about empty content (though it could be extended to do that, but that wasn't the intent).  

I do see, however, that `innerHTML` isn't the right thing to use, here, as the content needs to be serialized as XML as well.   So I will have to fix that.

dpvc:
> I have my doubts about whether self-closing tags are handled correctly.

OK, I adjusted the algorithm.  See if you like that better.

zorkow:
> > I have my doubts about whether self-closing tags are handled correctly.
> 
> OK, I adjusted the algorithm. See if you like that better.

I do.

zorkow (approved):

PR #681, opened 2021-04-07
svg mglyph: set valid alttext

pkra:
Replaces invalid alt and title with aria-label attribute.

Re-resolves mathjax/MathJax#2065 (which was fixed the same way in 2.x)


pkra:
Not sure if that's ideal; feel free to close.

dpvc:
Yes, this should be fixed.  Our style (tslint) requires the use of single quotes in place of the double-quotes, but I'll fix that afterward.

pkra:
Thanks, Davide.

PR #682, opened 2021-04-12
Fix mathaccent regression reported by Peter

dpvc:
This PR makes sure the parent element is not null before querying it.

This also fixes a second problem, where the base of an `mover` could be marked as a math accent.


pkra:
Thanks, Davide.

PR #683, opened 2021-04-14
A11y virtual attached keyexplorers

zorkow:
This takes care of item 1 in https://github.com/mathjax/MathJax-src/pull/561#issuecomment-787121895

Original problem was that key explorers where added to a node when they were switched on and removed when switched off. 
Since only the last explorer was supposed to stop propagation of the key events, if the order was messed up, propagation might be stopped too early. (E.g., switch on speech, braille, keyboard magnificatio. Then switch off braille and speech, and switch on speech: magnification works, speech does not).

Now key explorers are attached only once and retain an `attached` flag. This ensure that the order of the explorers remains the same.


ts/a11y/explorer.ts
```diff
@@ -145,17 +145,28 @@ export function ExplorerMathItemMixin<B extends Constructor<HTMLMATHITEM>>(
      * @param {ExplorerMathDocument} document The current document.
      */
     public attachExplorers(document: ExplorerMathDocument) {
-      this.attached = [];
```
dpvc, line 148: It looks like `this.attached` is no longer created or populated, but it is still used elsewhere in the code (in `rerender()` and `updateDocument()`).  Is that still correct?

dpvc (commented):
It looks like the `attached` property is no longer being populates, but is still used.  Can you check that that is as intended?

ts/a11y/explorer.ts
zorkow: Good catch. 
I've now also made the more restarting more robust on collapsed elements. This only works with the latest fixes I made to SRE today.
There's still a minor kink to work out: When a collapse is triggered with the mouse while the key explorers are deeper in the tree they effectively refocus on a node that is no longer there. But that is for 3.2 to work out.
the refocus does not y

zorkow: Solved that problem now. I will make a small PR once this PR is merged.

dpvc (approved):
OK, looks good.  I'll merge.

PR #686, opened 2021-04-19
Prevent border-spacing and border-collapse from being inherited from an outer table

dpvc:
This PR fixes a problem reported in WeBWorK (which is moving to v3).  The math was inside a larger table that had `border-spacing: 10px; border-collapse: separate`, and the arrays generated by MathJax within that outer table inherited those properties, making the spacing wrong for the math.   So I've added some CSS to the `mjx-mtable` element to reset these values.


zorkow (approved):
lgtm.

PR #687, opened 2021-04-21
Triggering links from the explorer.

zorkow:
Hitting return on the node that announces itself to be a link will trigger the link.
Unfortunately, this only works for CHTML, not in SVG.

I've tried by setting focus on the anchor element and continue propagating the event. 
Unfortunately, that only works sporadically. 

Also dispatching a new event like below did not do the trick.
``` javascript
      let event = new KeyboardEvent('keydown', {code: '13', bubbles: true});
      node.parentElement.dispatchEvent(event);
```

### DIFF
--- a/ts/a11y/explorer/KeyExplorer.ts
+++ b/ts/a11y/explorer/KeyExplorer.ts
@@ -292,6 +292,7 @@ export class SpeechExplorer extends AbstractKeyExplorer<string> {
     }
     if (this.active) {
       this.Move(code);
+      if (this.triggerLink(code)) return;
       this.stopEvent(event);
       return;
     }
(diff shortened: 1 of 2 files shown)


ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -301,6 +302,24 @@ export class SpeechExplorer extends AbstractKeyExplorer<string> {
...
+    if (code !== 13) {
+      return false;
+    }
+    let node = this.walker.getFocus().getNodes()?.[0];
+    let focus = node?.
+      getAttribute('data-semantic-postfix')?.
+      match(/(^| )link($| )/);
+    if (focus) {
+      // Works for CHTML only.
+      node.parentElement.click();
```
pkra, line 318: FWIW in the light walker, I basically set window.location.href to the href I get from the link.

dpvc:
> Also dispatching a new event like below did not do the trick.

It looks like

``` javascript
      node.parentNode.dispatchEvent(new MouseEvent('click'));
```

works for me, at least in Firefox (provided `node.parentNode` is the `<a>` node).

zorkow:
Thanks. I was clearly in the keyboard zone, as I had also experimented with `new Event('click')`, which did not work. D'oh!

zorkow:
This is now ready for review.

zorkow:
PS: This is the first time I am using optional chaining in MathJax code, as @pkra suggested it the other day.
It seems to be supported by all major browsers by now and the code transpiled by Typescript seems to be at least as fast as traditional boolean tests.

dpvc (approved):
Looks good.  I'll merge.

PR #689, opened 2021-04-23
Move mhchem files to ts/input/tex/mhchem and down-compile to ES5 (mathjax/MathJax#2673)

dpvc:
This PR resolves the problem with using the` mhchemparser` npm package (which is compiled to ES6) within node modules that call MathJax modules directly.  It includes the `mhchemparser` source files in the MathJax-sc/ts tree (as Martin originally asked), and compiles them along with the rest of MathJax.

The real solution is to have MathJax down-compile to ES5 rather than ES6.

Resolves issue mathjax/MathJax#2673.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,15 @@
 {
   "name": "mathjax-full",
-  "version": "3.1.2",
+  "version": "3.1.3",
   "lockfileVersion": 2,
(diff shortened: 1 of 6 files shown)


zorkow (approved):
I tested in a couple of use cases and it works fine.

PR #690, opened 2021-04-23
Fix problem with not closing empty tags in HTML serialization (mathjax/MathJax#2674)

dpvc:
This PR fixes a problem with HTML serialization (as opposed to XML serialization) where empty non-self-closing tags would not be properly closed.

Resolves issue mathjax/MathJax#2674.


zorkow (approved):
live

PR #691, opened 2021-04-27
Fix problem with detecting whether ex-height can be computed, and work around jsdom problems

dpvc:
This PR fixes a regression in 3.1.3 that causes liteDOM and jsdom adapters to get the ex-height badly wrong.  This was due to changes made to process CJK characters that also added some default height, which confused the test to see if the `exFactor` needed to be used.

There are also some problems with jsdom's implementation of `getComputesStyle()`, which doesn't actually inherit styles, it only lists the ones explicitly on the given element.  The CSSStyleDeclaration implementation is broken in a number of ways, and I've had to patch it before, but this issue would take some work.  This PR adds options to the jsdom to set the font size and family to use (just as we do with the LiteDOM), since `getComputedStyle()` can't be trusted.

Resolves issue mathjax/MathJax#2694.

### DIFF
--- a/ts/adaptors/jsdomAdaptor.ts
+++ b/ts/adaptors/jsdomAdaptor.ts
@@ -30,6 +30,8 @@ export class JsdomAdaptor extends HTMLAdaptor<HTMLElement, Text, Document> {
    * The default options
    */
   public static OPTIONS: OptionList = {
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #692, opened 2021-04-27
Only return a DOCTYPE if there is one

dpvc:
This PR fixes a problem where `adaptor.doctype()` could fail if the original document doesn't have a `<!DOCTYPE>` comment (in HTML and jsdom adaptors).  It also prevents the LiteDOM from returning a doctype if the original didn't have one.


zorkow (approved):
lgtm.

PR #693, opened 2021-04-27
Clear the stylesheet when used in a new MathDocument.  (mathjax/MathJax#2678)

dpvc:
This PR makes sure the output jax clear its internal stylesheet pointer when it is used in a new document.

Resolves issue mathjax/MathJax#2678.

### DIFF
--- a/ts/output/chtml.ts
+++ b/ts/output/chtml.ts
@@ -143,6 +143,13 @@ CommonOutputJax<N, T, D, CHTMLWrapper<N, T, D>, CHTMLWrapperFactory<N, T, D>, CH
   /**
    * @override
    */
+  public initialize() {
+    this.chtmlStyles = null;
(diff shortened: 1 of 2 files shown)


zorkow (approved):
I note there is a another `this.fontCache.clearCache();` in line 300. Would it make sense to let this use the `clearFontCache` method?

dpvc:
> Would it make sense to let this use the `clearFontCache` method?

Sure.  The reason I didn't was that that function was doing a lot of direct calls to `this.fontCache` already, so I considered the `clearCache()` call to be just one of those that manipulates the internal cache structure.  The `clearFontCache()` method is more for objects outside the SVG jax to call (so they don't have to handle the internal `fontCache` value).

I can change it if you want, but I don't really see the need.

PR #696, opened 2021-05-06
Update node-main for webpack 5

dpvc:
This PR updates the `webpack.config.js` file for the node-main component (which is what you get when you `require('mathjax')`) to work with the latest version of webpack to which we updated in 3.1.4.  This fixes the error that occurs when you do `require('mathjax')` with the `mathjax` npm package.  (The `mathjax-full` package works fine because it uses the source version rather than the webpacked version).

Resolves issue mathjax/MathJax#2683.


zorkow (approved):
lgtm.

PR #697, opened 2021-05-11
Mathtools update

dpvc:
This PR incorporates more of the mathtools package into the current `mathtools_package` branch, but since I merged `develop` into this branch, it doesn't make a good PR to `mathtools_package`, so I'm making it to `develop` instead. (Even if it had gone to `mathtools_package`, it wouldn't really have helped, since I ended up breaking the single `MathtoolsConfiguration.ts` into separate files, so the diff would not really show anything valuable anyway.)

This branch includes the files to make a MathJax Component for mathtools, and adds it into the dependencies and source listings, as well as the AllPackages file (so it will show up in the lab, in particular).  If we are going to separate out extra packages, this will need to be changed.

I ended up resetting the cramped style in tables and for explicit style changes (e.g., `\scriptstyle`), as in actual TeX, but did not do it for boxes, since that is more of a side-effect of how some macros are implemented using hboxes rather than a meaningful difference (as we discussed F2F).  I also added a `data-cramped` attribute to control it on `mstyle` and `table` elements so that the `\cramped` macros could be implemented without much overhead (and I made the mathtools `\math*lap` macros force non-cramped styles, as they do in actual TeX, since mathtools expects that distinction.  But it won't affect other boxes, just these mathtools macros.  I think that is a good compromise.

I also moved some code from BaseMethods to ParseUtil that I woudl have had to duplicate for some mathools macros (i.e., handling of under-over constructs, and checking the maximum macro counts).  Similarly, I moved the code for adjusting the row spacing in an alignment out of the implementation of `\\` (via `CrLatex`) and into the `ArrayItem` StackItem object, so that it could be used in mathtools' `\shortvdotswithin` macro.

I fixed an error with vertical centering (the 'c' option should center on the math axis), and normalized the spacing for some of the xArrows for consistency (and so that all packages define them the same way, as they are defined in more than one package).  

I added `\boxed` and `\framboxed` macros (which are defined in AMSmath, but were not in MathJax), as they are needed for mathtools' `\Aboxed` macro.

I added a `lookup()` utility function that gets a value from an object literal while checking that it is not an inherited value, and using a default value if not found.  E.g., `{l: 'left', r:'right'}[align] || 'center'` would not produce the expected result if `align` is `toString`.  Using `lookup(align, {l: 'left', r: 'right'}, 'center') resolves that problem.  (There are a number of places elsewhere in the code that are susceptible to this problem, and should be recoded using `lookup()`, but that can wait until our next code cleanup).

The `BaseMethods.Entry` function has been modified to handle text mode in the `cases*` environment (as it does for the `\cases` macro).  It looks like more changes here than there actually are, since reduced the indenting by exiting the function earlier for one conditional).  Viewing the difference with spacing changes ignored might help.

I refactored some code from the `newcommand` extension to move the parsing of a control sequence name and of optional argument counts into `NewcommandUtil`, as they are needed in mathtools as well.

Finally, I fixed a problem with CHTML tables where extra cells added to add out the rows wasn't getting the proper CSS if the column spacing was specified.  I also simplified some of the CSS for the cells.

And, of course, there are the mathtools files themselves.

I'm sorry that this is such a big set of changes.  If you want me to break out some of the changes to the non-mathtools files into separate PR's, I can do that.


pkra:
This branch does not compile for me. Here's what I get:

```bash
$ npm run compile

> mathjax-full@3.1.4 precompile
> npm run --silent clean:js


> mathjax-full@3.1.4 compile
> npx tsc

ts/input/tex/mathtools/MathtoolsMethods.ts:219:35 - error TS2345: Argument of type 'MmlNode' is not assignable to parameter of type 'MmlMstyle'.
  Type 'MmlNode' is missing the following properties from type 'MmlMstyle': setChildInheritedAttributes, texclass, updateTeXclass, getPrevClass, and 4 more.

219     MathtoolsUtil.setDisplayLevel(mml, style);
                                      ~~~

ts/input/tex/mathtools/MathtoolsMethods.ts:233:35 - error TS2345: Argument of type 'MmlNode' is not assignable to parameter of type 'MmlMstyle'.

233     MathtoolsUtil.setDisplayLevel(mml, style);
                                      ~~~

ts/input/tex/mathtools/MathtoolsMethods.ts:732:38 - error TS2550: Property 'entries' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the `lib` compiler option to 'es2017' or later.

732     for (const [id, value] of Object.entries(keys)) {
                                         ~~~~~~~


Found 3 errors.

```

dpvc:
It compiles for me.  What version of tsconfig are you using?  I'm at 4.2.4, as listed in the package.json (we updated all dependencies during the 3.1.4 release).

pkra:
> It compiles for me. What version of tsc are you using? I'm at 4.2.4, as listed in the package.json (we updated all dependencies during the 3.1.4 release).

I had cloned a fresh copy, checked out this branch, run `npm i` and `npm run compile` to get the above error.

pkra:
> I had cloned a fresh copy, checked out this branch, run npm i and npm run compile to get the above error.

I get same result after deleting the lockfile beforehand.

dpvc:
OK, I'm seeing it now.  I had made a local change to tsconfig.json to test something and forgot to revert it.  Sorry about that.  I have pushed a new commit to take care of the issues.

pkra:
Thanks, Davide. I can now compile this branch. Unfortunately, the develop branch no longer compiles for me which seems due to #653. I'll leave some details there since I can't think of a better place to put it.

ts/core/MmlTree/MmlNodes/mstyle.ts
```diff
@@ -75,10 +75,16 @@ export class MmlMstyle extends AbstractMmlLayoutNode {
...
+      prime = false;  // style change resets tex prime style
     }
     let displaystyle = this.attributes.getExplicit('displaystyle') as boolean;
     if (displaystyle != null) {
       display = (displaystyle === true);
+      prime = false;  // style change resets tex prime style
+    }
+    const cramped = this.attributes.getExplicit('data-cramped') as boolean;  // manual control of tex prime style
+    if (cramped != null) {
+      prime = cramped;
```
zorkow, line 87: You want to set `prime` only if `cramped` is not `undefined` or `null`? Can't you just set it to false in all of these cases? (i.e., `prime = !!cramped`)?

ts/core/MmlTree/MmlNodes/mtable.ts
```diff
@@ -122,10 +123,11 @@ export class MmlMtable extends AbstractMmlNode {
       columnalign: this.attributes.get('columnalign'),
       rowalign: 'center'
     });
+    const cramped = this.attributes.getExplicit('data-cramped') as boolean;
     const ralign = split(this.attributes.get('rowalign') as string);
     for (const child of this.childNodes) {
       attributes.rowalign[1] = ralign.shift() || attributes.rowalign[1];
-      child.setInheritedAttributes(attributes, display, level, prime);
+      child.setInheritedAttributes(attributes, display, level, cramped == null ? false : cramped);
```
zorkow, line 130: ditto.

ts/input/tex/ParseUtil.ts
```diff
@@ -450,6 +485,26 @@ namespace ParseUtil {
...
+   */
+  export function checkMaxMacros(parser: TexParser, isMacro: boolean = true) {
+    if (++parser.macroCount <= parser.configuration.options['maxMacros']) {
+      return;
+    }
+    if (isMacro) {
+      throw new TexError('MaxMacroSub1',
+                         'MathJax maximum macro substitution count exceeded; ' +
+                         'is here a recursive macro call?');
+    } else {
```
zorkow, line 501: No need for the else.

ts/input/tex/mathtools/MathtoolsConfiguration.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Create the paired-delimiters command map, and link it into the configuration.
+ * @param {ParserConfiguration} config   The current configuration.
+ */
+function initMathtools(config: ParserConfiguration) {
+  new CommandMap(PAIREDDELIMS, {}, {});
+  config.append(Configuration.local({handler: {macro: [PAIREDDELIMS]}, priority: -5}));
+}
+
+/**
+ * Add any pre-defined paried delimiters, and subclass the configured tag format.
```
zorkow, line 52: typo

ts/input/tex/mathtools/MathtoolsMethods.ts
```diff
@@ -0,0 +1,747 @@
...
+import {length2em, em} from '../../../util/lengths.js';
+import {lookup} from '../../../util/Options.js';
+import NewcommandUtil from '../newcommand/NewcommandUtil.js';
+import NewcommandMethods from '../newcommand/NewcommandMethods.js';
+
+import {MathtoolsTags} from './MathtoolsTags.js';
+import {MathtoolsUtil} from './MathtoolsUtil.js';
+
+/**
+ * The implementations for the macros and environemtns for the mathtools package.
```
zorkow, line 44: typo

ts/input/tex/mathtools/MathtoolsMethods.ts
```diff
@@ -0,0 +1,747 @@
...
+      displaystyle: true,
+      rowspacing: '.5em',
+      width: width || 'auto',
+      columnwidth: '100%',
+    };
+    return ParseUtil.setArrayAlign(item as ArrayItem, pos || 'c');
+  },
+
+  /**
+   * Replacement for the AMS HandleShove that incldues optional spacing values
```
zorkow, line 107: typo

ts/input/tex/mathtools/MathtoolsMethods.ts
```diff
@@ -0,0 +1,747 @@
...
+    const top = parser.stack.Top() as EqnArrayItem;
+    const isFlush = (top.getProperty('flushspaceabove') === top.table.length);
+    const arg = '\\mmlToken{mi}{}' + parser.GetArgument(name) + '\\mmlToken{mi}{}';
+    const base = new TexParser(arg, parser.stack.env, parser.configuration).mml();
+    let mml = parser.create('node', 'mpadded', [
+      parser.create('node', 'mpadded', [
+        parser.create('node', 'mo', [parser.create('text', '\u22EE')])
+      ], {width: 0, lspace: '-.5width', ...(isFlush ? {height: '-.6em', voffset: '-.18em'} : {})}),
+      parser.create('node', 'mphantom', [base])
+    ], {lspace: '.5width'});
```
zorkow, line 406: Can you spread this out a bit.

ts/input/tex/mathtools/MathtoolsMethods.ts
```diff
@@ -0,0 +1,747 @@
...
+      parser.string = ParseUtil.addArgs(parser, tex, parser.string.substr(parser.i));
+      parser.i = 0;
+    }
+  },
+
+  /**
+   * Implements \ndownarrow and \nuparrow via a terrible hack (visual only, no chance of this working with SRE).
+   *
+   * @param {TexParser} parser   The calling parser.
+   * @param {string} name        The macro name.
```
zorkow, line 575: `c` and `dy` missing.

ts/input/tex/mathtools/MathtoolsMethods.ts
```diff
@@ -0,0 +1,747 @@
...
+   */
+  NArrow(parser: TexParser, _name: string, c: string, dy: string) {
+    parser.Push(
+      parser.create('node', 'TeXAtom', [
+        parser.create('token', 'mtext', {}, c),
```
zorkow, line 580: What is `c`?

ts/input/tex/mathtools/MathtoolsMethods.ts
```diff
@@ -0,0 +1,747 @@
...
+        parser.create('node', 'mpadded', [
+          parser.create('node', 'mpadded', [
+            parser.create('node', 'menclose', [
+              parser.create('node', 'mspace', [], {height: '.2em', depth: 0, width: '.4em'})
+            ], {notation: 'updiagonalstrike', 'data-thickness': '.05em', 'data-padding': 0})
+          ], {width: 0, lspace: '-.5width', voffset: dy}),
+          parser.create('node', 'mphantom', [
+            parser.create('token', 'mtext', {}, c)
+          ])
```
zorkow, line 589: So you are overlaying the same arrow twice? That is a bit like poor man's bold (or whatever it is called). 
Is there some future way we could do that "properly" that is also SRE friendly?

zorkow (approved):
Some minor issues and typos. Otherwise OK.

ts/core/MmlTree/MmlNodes/mstyle.ts
dpvc: Not quite.  Here we are only adjusting the prime style if cramped is explicitly given, otherwise we retain the current prime style.  We don't reset it to false if the `data-cramped` attribute is not present.  So the check for not null is important.

ts/core/MmlTree/MmlNodes/mtable.ts
dpvc: Right, this one can be done via `!!cramped`, since tables do reset prime style.

ts/input/tex/ParseUtil.ts
dpvc: I know that you prefer to reduce indentation when possible, but the indentation sometimes has meaning.  In this case, the function will produce one of two error messages, neither of which is more important than the other.  The indentation of both shows that they are comparable alternatives.  Leaving out the "else" would make the second be the "main" message and the first be a subordinate one.  It makes the `isMacro` variable a break0out condition rather than a toggle that determines which of two equal things is being selected.  This structure preserves the logic of the function better than having the two messages at different indentations, which gives the second a prominence it doesn't deserve.

dpvc: Unless you strongly object, I'd like to keep it as is.

ts/input/tex/mathtools/MathtoolsMethods.ts
dpvc: It's the unicode character for either an up arrow or a down arrow (as provided by the `MathtoolsMappings.ts` file).

ts/input/tex/mathtools/MathtoolsMethods.ts
dpvc: No, it is not overplayed twice.  Notice that the second is inside an `mphantom`, so it is not visual.  It is only used for positioning purposes (the `space: '-.5width'` below means we are shifting the preceding (mpadded) `menclose` to the left by half the width of the arrow given in `c`.  That is how the "not" line is centered on the arrow (so that it will work in any font, regardless of the width of the arrow).

dpvc: As for making it SRE friendly, there don't seem to be unicode positions for up or down arrows with a diagonal slash (otherwise I'd just have called on those).  The closest is U+2908 and U+2909, which are explicit about having a horizontal slash.

ts/input/tex/mathtools/MathtoolsMethods.ts
dpvc: Done

dpvc:
@pkra, yes, the two branches do produce a conflict.  That will have to be resolves after one of them in merged.

PR #699, opened 2021-05-12
Force width of -explicitFont text (work around Safari bug)

dpvc:
This PR worked around a bug with WebKit that mis-measures regular text (like error messages) and so produces the wrong bounding boxes.  For example, a TeX error message will get the yellow background size wrong.  The problem is that the MJXZERO font seems to confuse WebKit (that is also the source of the problem with parts of characters disappearing).  I'd like to fix this during the font update this summer, but for now, this is a work-around.

The fix is to remove the MJXZERO font during the measuring process, and then specify the width explicitly on the final text element.  Not great, but it works for now.

You can test in Safari by causing any TeX error (e.g., `\end` with no name, or `\begin{xyz}`).

### DIFF
--- a/ts/output/chtml.ts
+++ b/ts/output/chtml.ts
@@ -208,7 +208,7 @@ CommonOutputJax<N, T, D, CHTMLWrapper<N, T, D>, CHTMLWrapperFactory<N, T, D>, CH
   /**
    * @override
    */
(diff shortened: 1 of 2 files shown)


zorkow (approved):
I don't have safari so have to trust that is works.

PR #700, opened 2021-05-13
Fix error where a second use of \| delim causes the wrong size to be used.

dpvc:
A change in 3.1.4 to make stretchy delimiters that end up using a different untied point produce the correct character had a side effect of making them unstretchable after the first use.  The `\Vert` double bar is one such, and so the second (and subsequent) uses of `\Vert` will not stretch.  E.g. `\left\Vert \frac 1 2 \right\Vert` will have the right-hand delimiter un-stretched.  This PR resolves the problem.

Resolves issue mathjax/MathJax#2697.


zorkow (approved):
I can't reproduce the problem. Neither locally nor on the demo site. Both `develop` and `Vert-delim` produce the same output.

dpvc:
@zorkow, here is an image I just made from the demo site:

Vert

Typesetting a second time will show both at the (incorrect) small size.

PR #701, opened 2021-05-13
Lazy typeset

dpvc:
This PR implements a lazy-typesetting extension that only processes an expression when it first comes into view.  This works best with SVG output, which doesn't require CSS changes when expressions are typeset, but can also be used with CHTML (which does).  I will make a second PR that improves the CHTML performance by changing how the CSS is updated.

This works by subclassing the MathItem and its `compile()` and `typeset()` methods so that before the item appears on screen, it is just replaced by a marker, and an IntersectionObserver is used to determine when it comes into view.  When it does, it is reprocessed and the original `compile()` and `typeset()` functions are used.

Because TeX expressions can depend on earlier ones (e.g., for function definitions, or equation numbers when automatic numbering is in effect), when a TeX expression comes into view, we compile all the earlier ones to make sure that everything is at it should be for this expression.  The only thing that won't work this way is forward references using `\ref` or `\eqref`.  (It would be possible to an an option to compile() the whole page and only do lazy typesetting, which is still an improvement.)


ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+
+  /**
+   * True if this item is a TeX MathItem
+   */
+  lazyTex: boolean;
+
+}
+
+/**
+ * The mixin for adding laxy typesetting to MathItems
```
zorkow, line 142: typo

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+     *
+     * @override
+     */
+    public compile(document: LazyMathDocument<N, T, D>) {
+      if (this.lazyCompile) {
+        if (this.state() < STATE.COMPILED) {
+          this.lazyTex = (this.inputJax.name === 'TeX');
+          this.root = document.mmlFactory.create('math');
+          this.state(STATE.COMPILED);
+        }
```
zorkow, line 178: Don't you need to set `lazyState` to `TYPESET` here as well?
Also `lazyState` is never really used only set. Could it be removed?

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+    /**
+     * Initially don't compile math, just use an empty math item,
+     *   then when the math comes into view (or is before something
+     *   that comes into view), compile it properly and mark the item
+     *   as only needing to be typeset.
+     *
+     * @override
+     */
+    public compile(document: LazyMathDocument<N, T, D>) {
+      if (this.lazyCompile) {
```
zorkow, line 173: The `if-if-else` is confusing. Could you turn it into a single guard:
```suggestion
      if (!this.lazyCompile || this.state() >= STATE.COMPILED) {
        super.compile(document);
￼        this.lazyState = STATE.TYPESET;
        return;
    }
```

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+    public lazyList: LazyList<N, T, D>;
+
+    /**
+     * A promise to make sure our compiling/typesetting is sequential
+     */
+    protected lazyPromise: Promise<void> = Promise.resolve();
+
+    /**
+     * The function used to typeset a set of lazy MathItems.
+     *   (uses requestIdleCallback if available, or setTimeout otherwise)
```
zorkow, line 293: Cool, `requestIdleCallback` is new to me!

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+        if (entry.isIntersecting) {
+          this.lazySet.add(id);
+          if (!this.lazyIdle) {
+            this.lazyIdle = true;
+            this.lazyProcessSet();
+          }
+        } else {
+          this.lazySet.delete(id);
+        }
+      }
```
zorkow, line 349: Please simplify this `if-if-else`.

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+     * @return {boolean}      True if there are TeX items to be typeset
+     */
+    protected compileEarlierItems(set: LazySet): boolean {
+      let math = this.earliestTex(set);
+      if (!math) return false;
+      let compile = false;
+      for (const item of this.math) {
+        if (item === math) break;
+        const earlier = item as LazyMathItem<N, T, D>;
+        if (earlier.lazyTex && earlier.lazyCompile) {
```
zorkow, line 409: Could you cast the list to `LazyMathItem<>[]`, run the loop on `item` only and  have a single guard we you break?
Would typescript be unhappy with that?

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+  };
+
+}
+
+/*==========================================================================*/
+
+/**
+ * Add lazy typesetting support a Handler instance
```
zorkow, line 463: typo

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+ * @param {Handler} handler   The Handler instance to enhance
+ * @return {Handler}          The handler that was modified (for purposes of chaining extensions)
+ *
+ * @template N  The HTMLElement node class
+ * @template T  The Text node class
+ * @template D  The Document class
+ */
+export function LazyHandler<N, T, D>(handler: HTMLHandler<N, T, D>): HTMLHandler<N, T, D> {
+  //
+  // Only update the document class if we can handle IntersectionObservers and idle callbacks
```
zorkow, line 474: Idle callbacks don't have to be available. You are using timeouts above.

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+      const callback = this.lazyHandleSet.bind(this);
+      this.lazyProcessSet = (typeof window !== 'undefined' && window.requestIdleCallback ?
+                          () => window.requestIdleCallback(callback) :
+                          () => setTimeout(callback, 10));
+    }
+
+    /**
+     * The function used by the IntersectionObserver to monitor the markers coming into view.
+     * When one (or more) does, add it to or remove it from the set to be processed, and
+     *   if added to the set, queue an idle task, if one isn't already pending.
```
zorkow, line 331: For clarification:
If a marker is removed, what happens to the idle task? Is that removed as well?

ts/ui/lazy/LazyHandler.ts
```diff
@@ -0,0 +1,483 @@
...
+        const math = this.lazyList.get(id);
+        if (!math) return;
+        if (entry.isIntersecting) {
+          this.lazySet.add(id);
+          if (!this.lazyIdle) {
+            this.lazyIdle = true;
+            this.lazyProcessSet();
+          }
+        } else {
+          this.lazySet.delete(id);
```
zorkow, line 347: So if there is already a processor set that keeps running?

zorkow (changes requested):
I have not had a chance to test it on any of our larger sample articles from the Simons set. 
I would still like to do this. In the mean time, however, here are already some of my comments.

zorkow:
PS: And from what I've seen, this is really cool. Great stuff!

ts/ui/lazy/LazyHandler.ts
dpvc: Looks like it can be removed.  It must have been left over from an earlier iteration.  Sorry about that.

ts/ui/lazy/LazyHandler.ts
dpvc: It's not the same thing.  In your setup, if `lazyCompile` is true but `this.state() >= STATE.COMPILED`, then `super.compile(document)` is called, but in mine it is not.  But I can do

```
      if (!this.lazyCompile) {
        super.compile(document);
        return;
      }
      if (this.state() < STATE.COMPILED) {
        this.lazyTex = (this.inputJax.name === 'TeX');
        this.root = document.mmlFactory.create('math');
        this.state(STATE.COMPILED);
      }
```

instead.

ts/ui/lazy/LazyHandler.ts
dpvc: It was new to me when I was looking up how to detect elements that come into view.

ts/ui/lazy/LazyHandler.ts
dpvc: I'm not sure what you have in mind (or what the objection is).  I have restructured using `break`, so perhaps that will do.

ts/ui/lazy/LazyHandler.ts
dpvc: No, because `this.math` is not an array, though it is iterable.  One could use `Array.from()` to make it an array, but since there could be hundreds (ben thousands) of equations you will be creating large arrays every time an expression is rendered.  The whole point of the iterable linked list for `this.math` was to be able to iterate it without needing to make it into an array.

It would be nice to be able to cast `this.math` to a list of `LazyMathItem<N, T, D>`, but the types are set up right for that because `this.math` is a `MathList<N, T, D>`, and that is based on `MathItem<N, T, D>`.  If you want, I could refactor that so that `MathList` has one generic (the item type), and we use `MathList<MathItem<N, T, D>>` instead of `MathList<N, T, D>`, so that this look could be

```
for (const item of this.math as `MathList<LazyMathItem<N, T, D>>) {...}
```

I haven't tried it, but I think that would work.

Alternatively, I could at least do 

```
        const earlier = item as LazyMathItem<N, T, D>;
        if (earlier === math || !earlier?.lazyCompile) {
         break;
        }
```

Which would you prefer?

ts/ui/lazy/LazyHandler.ts
dpvc: Fixed.

ts/ui/lazy/LazyHandler.ts
dpvc: Right.  Removed the comment (the test already was only for IntersectionObservers).

ts/ui/lazy/LazyHandler.ts
dpvc: If there is an idle task running, it will be working on a different set of equations (the previous set).  This new set will be fore the *next* idea task to process.  The point of removing it is if we happen to scroll quickly across some section of the text without stopping there, we should not typeset those expressions.  We don't try to stop a set from being typeset once we have sent it to an idle task.

Note that the marker DOM elements are created by the (lazy) `typeset()` method (and inserted into the document by the `updateDocument()` method).  They are removed during a `rerender()` call, which only occurs during the idle task.  Usually, the idle task will typeset the complete set of expressions in one go (so other javascript won't run during that process).  But if an expression needs to load an extension, then that does mean other code can run, so it is possible that one of the later expressions could be added to the next set of expressions to typeset.  While that is not desirable, I don' think it has any adverse consequences other than having the expression compiled and typeset a second time.  

I could try to do something about that, if you think not is important.

ts/ui/lazy/LazyHandler.ts
dpvc: Again, this is about removing an expression that you have scrolled past without stopping.  That is, if you start scrolling and reveal an expression marker, we add it into the set and queue an idle task.  But if you keep scrolling and the idle task hasn't run yet, the mark may move out of view, and we don't want to waste time typesetting it, so we remove it from the set.  When you stop scrolling and the idle task finally runs, it takes the expressions still in the set (those should be the ones still in view) and starts typesetting them after having created a new set to add expression to.  If one fo the expressions loads an extension and the idle task has to pause, you can add new expressions into the new list (or remove them from there), but can't affect the current idle-task list.  Those will be typeset even if they scroll out of view at this point.

It is possible that an expression is added to the set and then removed before the idle task runs, and it could be that the set is empty when it does.  Then the idle task does nothing, and new expressions are added to the new set to be processed by the next idle task.

Does that make sense?

dpvc:
> And from what I've seen, this is really cool. Great stuff!

Thanks!  I was pretty excited by it as well.  I think it will be an important addition to MathJax, and will make people with big pages happier.

On the other hand, I've just discovered that it doesn't play well with the assistive tools, so I need to look into what is happening there.

dpvc:
> I've just discovered that it doesn't play well with the assistive tools

I've pushed a commit that resolves the problem with the semantic enrichment, so it should be good to go now.  I also pushed changed that you requested, so can you have another look?

ts/ui/lazy/LazyHandler.ts
zorkow: Version 1 is seems overkill to avoid a single recurring cast. So the alternative is fine.

ts/ui/lazy/LazyHandler.ts
zorkow: Thanks.

ts/ui/lazy/LazyHandler.ts
zorkow: It does, and that's pretty much how I thought it worked. (And now have also practically experimented with.)
And I feel the gist of your last two comments should be part of the documentation for the `lazy` extension, to be kept for posterity.

zorkow (approved):
I've ran it now on the largest of our Elsevier articles and it really works very nicely!

ts/ui/lazy/LazyHandler.ts
dpvc: OK, made the change.

ts/ui/lazy/LazyHandler.ts
dpvc: I have added the two paragraphs to the comments for the method.

dpvc:
> I've ran it now on the largest of our Elsevier articles and it really works very nicely!

Thanks for the confirmation.  I'm excited to see how this is received by the community.

PR #702, opened 2021-05-13
Update the CHTML adaptive CSS process to insert new rules rather than replacing the stylesheet entirely.

dpvc:
The lazy-typeset extension causes equations to be updated frequently, and for CHTML, that means updating the CHTML stylesheet.  It turns out that that can be expensive on large pages.  This PR improves the updating of the CSS by inserting individual rules into the stylesheet rather than replacing it outright.  That means the browser only has to update a few rules rather than think they *all* have changed.  This brings the performance of the lazy-typesetting for CHTML to nearly the same as for SVG.

This also updates the way that usage information for the CSS rules is stored.  It introduces a new Usage class for holding the data and uses that for keeping track of the wrappers and characters that need to be included in the stylesheet.  This moves that information out of the wrappers and character instances (saving storage space in those that aren't used).

You might want to look at `ts/output/chtml/Usage.ts` for the implementation of the Usage object before trying to understand the changes to the other files.

### DIFF
--- a/ts/output/chtml.ts
+++ b/ts/output/chtml.ts
@@ -24,7 +24,7 @@
 import {CommonOutputJax} from './common/OutputJax.js';
 import {CommonWrapper} from './common/Wrapper.js';
 import {StyleList} from '../util/Styles.js';
-import {StyleList as CssStyleList} from '../util/StyleList.js';
+import {StyleList as CssStyleList, CssStyles} from '../util/StyleList.js';
 import {OptionList} from '../util/Options.js';
 import {MathDocument} from '../core/MathDocument.js';
 import {MathItem} from '../core/MathItem.js';
@@ -32,6 +32,7 @@ import {MmlNode} from '../core/MmlTree/MmlNode.js';
 import {CHTMLWrapper} from './chtml/Wrapper.js';
 import {CHTMLWrapperFactory} from './chtml/WrapperFactory.js';
 import {CHTMLFontData} from './chtml/FontData.js';
+import {Usage} from './chtml/Usage.js';
 import {TeXFont} from './chtml/fonts/tex.js';
 import * as LENGTHS from '../util/lengths.js';
 import {unicodeChars} from '../util/string.js';
@@ -121,10 +122,14 @@ CommonOutputJax<N, T, D, CHTMLWrapper<N, T, D>, CHTMLWrapperFactory<N, T, D>, CH
   public static STYLESHEETID = 'MJX-CHTML-styles';
 
   /**
-   *  Used to store the CHTMLWrapper factory,
-   *  the FontData object, and the CssStyles object.
+   *  Used to store the CHTMLWrapper factory.
    */
   public factory: CHTMLWrapperFactory<N, T, D>;
+
+  /**
+   * The usage information for the wrapper classes
+   */
+  public wrapperUsage: Usage<string>;
 
   /**
    * The CHTML stylesheet, once it is constructed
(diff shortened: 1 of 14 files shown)


ts/output/chtml.ts
```diff
@@ -152,31 +158,62 @@ CommonOutputJax<N, T, D, CHTMLWrapper<N, T, D>, CHTMLWrapperFactory<N, T, D>, CH
...
+        // Update the style sheet rules
+        //
+        const styles = new CssStyles();
+        this.addWrapperStyles(styles);
+        this.updateFontStyles(styles);
+        this.adaptor.insertRules(this.chtmlStyles, styles.getStyleRules());
+      }
       return this.chtmlStyles;  // stylesheet is already added to the document
     }
     const sheet = this.chtmlStyles = super.styleSheet(html);
```
zorkow, line 173: Is there a reason why this can't be done on `this.chtmlStyles` directly?

ts/output/chtml.ts
```diff
@@ -152,31 +158,62 @@ CommonOutputJax<N, T, D, CHTMLWrapper<N, T, D>, CHTMLWrapperFactory<N, T, D>, CH
...
-        });
+  protected addWrapperStyles(styles: CssStyles) {
+    if (this.options.adaptiveCSS) {
+      for (const kind of this.wrapperUsage.update()) {
+        const wrapper = this.factory.getNodeClass(kind) as any as typeof CommonWrapper;
+        wrapper && this.addClassStyles(wrapper, styles);
       }
-      super.addClassStyles(CLASS);
+    } else {
+      super.addWrapperStyles(styles);
```
zorkow, line 196: My usual: `else` to `if (!...) {... return;}`.

ts/output/chtml/Usage.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+/**
+ * @fileoverview  Keeps track of usage of font characters and wrappers
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+/**
+ * Class used for traking usage of font characters or wrppers
```
zorkow, line 25: typo

ts/output/common/OutputJax.ts
```diff
@@ -472,10 +467,27 @@ export abstract class CommonOutputJax<
   }
 
   /**
-   * @param {any} CLASS  The Wrapper class whose styles are to be added
+   * @param {CssStyles} styles   The style obejct to add to
```
zorkow, line 470: typo

ts/output/common/OutputJax.ts
```diff
@@ -472,10 +467,27 @@ export abstract class CommonOutputJax<
...
+   */
+  protected addFontStyles(styles: CssStyles) {
+    styles.addStyles(this.font.styles);
+  }
+
+  /**
+   * @param {CssStyles} styles   The style obejct to add to
```
zorkow, line 477: typo

zorkow (approved):
Some typos and one of my `if-else` complaints.
O/w lgtm.

ts/output/chtml.ts
dpvc: Sorry, I'm not sure I understand the question.  If you are asking about why the intermediate `styles` object in lines 166 to 169, the reason is that the information that comes from the Wrapper classes and font are just JSON objects, not actual style definition strings, as needed for `insertRules()`.  The `CssStyles()` object is the intermediary that can convert the JSON to the appropriate CSS strings.

ts/output/chtml.ts
dpvc: Ok, ok, I'll do it.

PR #703, opened 2021-05-13
Empheq and numcases packages

dpvc:
This is the preliminary implementation of the empheq and numcases packages.


ts/input/tex/empheq/EmpheqConfiguration.ts
```diff
@@ -0,0 +1,262 @@
+import {Configuration} from '../Configuration.js';
```
zorkow, line 1: Copyright is missing.

ts/input/tex/numcases/NumcasesConfiguration.ts
```diff
@@ -0,0 +1,170 @@
+import {Configuration} from '../Configuration.js';
```
zorkow, line 1: Copyright?

zorkow (commented):
A couple of comments before further review:
* As `EmpheqUtil` is alsoused in `Numcases`, while that uses nothing else from `Empheq`, I think it should be in a separate file so it can be packed separately. (E.g., if I want to pack `numcases` alone).
* From what I see in LaTeX  `numcases` and `subnumcases` are part of the `cases` pacakge. Should we not name it similar?

dpvc:
I have made the corrections you requested.  I also moved the `copyMml()` function to be a `copy()` method on the `MmlNode` object, since that is really where it belongs, and it would have been useful for some of the other extensions.  When this lands, I will go back and change those extensions to use it.

Can you check again?

ts/input/tex/cases/CasesConfiguration.ts
```diff
@@ -139,32 +175,38 @@ export const NumcasesMethods = {
...
+ */
+new MacroMap('cases-macros', {
   '&': 'Entry'
 }, NumcasesMethods);
 
 //
 //  Define the package for our new environment
 //
-export const NumcasesConfiguration = Configuration.create('numcases', {
+export const NumcasesConfiguration = Configuration.create('cases', {
```
pkra, line 203: The export is still called NumcasesConfiguration which seems odd.

dpvc: Thanks.  I missed that one, and will make a new commit to fix it.

ts/input/tex/cases/CasesConfiguration.ts
```diff
@@ -8,14 +8,23 @@ import {BeginItem, EqnArrayItem} from '../base/BaseItems.js';
 import {AmsTags} from '../ams/AmsConfiguration.js';
 import {StackItem, CheckType} from '../StackItem.js';
 import {MmlMtable} from '../../../core/MmlTree/MmlNodes/mtable.js';
-import {EmpheqUtil} from '../empheq/EmpheqConfiguration.js';
+import {EmpheqUtil} from '../empheq/EmpheqUtil.js';
 
+/**
+ * The StakItem for the numcases environment.
```
zorkow, line 14: typo

ts/core/MmlTree/MmlNode.ts
```diff
@@ -330,6 +347,36 @@ export abstract class AbstractMmlNode extends AbstractNode implements MmlNode {
     this.attributes.setList(attributes);
   }
 
+  /**
+   * @override
+   *
+   * @param {boolean} keepIds   True to copy id attributes, false to skip them.
+   */
+  public copy(keepIds: boolean = false): AbstractMmlNode {
```
zorkow, line 355: I am surprised typescript allows this, as it does not faithfully implement the interface. That could be the source of a future compile error.
Maybe add a comment?

ts/core/Tree/Node.ts
```diff
@@ -264,6 +269,18 @@ export abstract class AbstractNode implements Node {
...
 
 
+  /**
+   * @override
+   */
+  public copy() {
+    const node = (this as AbstractNode).factory.create(this.kind) as AbstractNode;
+    node.properties = {...this.properties};
+    for (const child of this.childNodes || []) {
+      if (child) node.appendChild(child.copy());
```
zorkow, line 279: missing braces

zorkow, line 279: Children can be empty in case of incomplete `subsup`, `underover`, etc. 
Maybe add a comment.

ts/input/tex/empheq/EmpheqUtil.ts
```diff
@@ -0,0 +1,228 @@
...
+import {AbstractTags} from '../Tags.js';
+import {MmlNode} from '../../../core/MmlTree/MmlNode.js';
+import {MmlMtable} from '../../../core/MmlTree/MmlNodes/mtable.js';
+import {MmlMtd} from '../../../core/MmlTree/MmlNodes/mtd.js';
+import {EmpheqBeginItem} from './EmpheqConfiguration.js';
+
+export const EmpheqUtil = {
+
+  /**
+   * Create the needed envonronment and process it by the give function.
```
zorkow, line 38: typo.

zorkow (approved):
Thanks for the changes. Just a few minor things left.

ts/core/Tree/Node.ts
dpvc: I have added a separate commit to address this.  The copied nodes were not being added to the various node lists in the `ParseOptions` object, and so the `msubsup` was not being converted to `msub` or `msup`.  So I've made a `ParseUtil` method `copyNode()` that calls `copy()` and then adds the new nodes to the proper lists.  It also modified `addNode()` to record if the node is added to a list that is other than the one for its node type (e.g., the `nonscript` list) so that its copies can be added to the same lists.

dpvc:
@zorkow, I have made a significant change by adding `copyNode()` to `ParseUtil`.  Can you re-review commit 3c40dd9?  Thanks.

ts/input/tex/ParseOptions.ts
```diff
@@ -173,6 +174,14 @@ export default class ParseOptions {
       list = this.nodeLists[property] = [];
     }
     list.push(node);
+    if (node.kind !== property) {
+      //
+      // If the list is not just for its kind, record that it is in this list
+      //   so that if it is copied, the copy can also be added to the list.
+      //
+      let lists = (NodeUtil.getProperty(node, 'in-lists') as string || '').split(',').concat(property).join(',');
+      NodeUtil.setProperty(node, 'in-lists', lists);
```
pkra, line 183: This attribute makes it to the output. If it should, could it get a data-* prefix?

dpvc: I'm not able to reproduce that.  Can you give an example that does it for you?

pkra: `\begin{equation} F(z,N) := e^{f(z,N)} := e^{z/N}N^{l-1}(1-e^{z/N})^{l-1}\prod _{j=1}^N \frac{1}{1-e^{zj/N}}, \tag{4}\end{equation}` gets me `in-lists=",fixStretchy"` on the `g` around the `:=`

pkra: Just the := is enough, actually.

dpvc: OK, thanks.  It turns out that this is a very special case, where two `mo` elements are combined into one.  The code that does this uses a function to copy all the internal properties from one of them to the combined `mo`, but unfortunately that function sets the attributes rather than the internal properties.  I will make a PR to fix it.  I've check the other uses of the function, and this appears to be the only place that occurs.

zorkow (approved):
Looks fine. 
I find the `in-lists` property a bit awkward, but we can work on that some other day.

PR #705, opened 2021-05-15
Remove munderover elements that have been cleaned.  (mathjax/MathJax#2691)

dpvc:
The `ParseOptions.nodeLists` object contains lists of elements generated during TeX input processing, and these are used by the post filters in order to do things like fix up `munderover` elements and combine `mo` relations.  Sometimes this means an element is removed and another put in its place.  When this happens, the element isn't currently being removed from its list, so later filters may try top process the removed element again.  This happens with the `cleanSubSup` and `moveLimits` filters.  If an `munderover` has been cleaned, it is still processed by `moveLimits`, sometimes leading to an error when the `munderover` has an empty child.

This PR adds a `removeFromList()` method to `ParseOptions` class that allows elements to be removed from its list.  Because the `for-of` loop will miss an element if you modify the array from within the loop, we keep a list of nodes to remove and remove them all at the end.

Resolves isse mathjax/MathJax#2691.

### DIFF
--- a/ts/input/tex/FilterUtil.ts
+++ b/ts/input/tex/FilterUtil.ts
@@ -243,7 +243,7 @@ namespace FilterUtil {
   let _moveLimits = function (options: ParseOptions, underover: string, subsup: string) {
     const remove: MmlNode[] = [];
     for (const mml of options.getList(underover)) {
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #706, opened 2021-05-15
Implement mml3 mathml extension and component

dpvc:
This PR ports the XSLT-based MML3 extension from v2 to v3, and adds the needed component file to make it a loadable MathML input extension.  This works both in the browser (through the native `XSLTProcessor` object) and in node (via the `saxon-js` npm package).  I have not added saxon-js to the dependencies, since I don't think it is highly used, and don't need the extra dependency under usual circumstances.  I expect this extension to be moved to the non-core extensions when we make that happen, and it can be a dependency at that point.  For now, it tells you to install saxon-js if you haven't done so, and we will document the need for it for this extension.  Saxon-js needs to preprocess the XSLT, so there are new npm scripts to create the needed JSON file, and to move it into place.  I've included the json file, here, but we could remove it since it can be generated from the source.

Of course, all this should be converted to actual support in MathJax, but for now, this is a stop-gap measure.

See mathjax/MathJax#2515.


ts/input/mathml/mml3/mml3-node.ts
```diff
@@ -0,0 +1,43 @@
+/**
```
zorkow, line 1: Copyright?

zorkow (approved):
On Copyright missing. I did not look at the xslt!

ts/input/mathml/mml3/mml3-node.ts
dpvc: Added.

PR #707, opened 2021-05-18
Colortbl package

dpvc:
This PR implements `\columncolor`, `\rowcolor` and `\cellcolor` from the `colortbl` TeX package.  This extension uses the color handling from the `color` package, so it checks if it is included and sets up the color model if it is not.  The `ArrayItem` StackItem is slightly refactored to move the generation of the MathML for the table so that it can be overridden to do modifications to the MathML.  The mtable output wrappers are modified so that the SVG and CHTML will handle an empty frame specification consistently (and this is used to make the table include the frame spacing but not outline, which makes colored rows and columns look better at the edges).


ts/input/tex/colortbl/ColortblConfiguration.ts
```diff
@@ -0,0 +1,150 @@
+import {ArrayItem} from '../base/BaseItems.js';
```
zorkow, line 1: Copyright? Fileoverview?

ts/input/tex/colortbl/ColortblConfiguration.ts
```diff
@@ -0,0 +1,150 @@
...
+import TexError from '../TexError.js';
+import {MmlNode} from '../../../core/MmlTree/MmlNode.js';
+
+import {TeX} from '../../tex.js';
+
+export type ColorData = {
+  cell: string;
+  row: string;
+  col: string[];
+};
```
zorkow, line 14: Would that not be better as an interface, in case we need to extend it at some point?

zorkow (approved):
On minor issue, otherwise OK.

PR #710, opened 2021-05-22
Add \noscript command

dpvc:
This PR implements the `\nonscript` command, which removes the following spacing command (like `\hskip`, `\kern`, `\hspace`, etc.).  This is used int macro definitions to handle not use spacing in scripts, when given explicitly.

This is handled by using a new StackItem for the `\nonscript` that checks what is pushed after it, and if an `mspace` (or an `mspace` inside an `mstyle`, as produced by `\,` and similar macros), then we add the space into a list to be post-filtered (once the `scriptlevel` is known).   The case of am `mstyle` containing an `mspace` is handled by wrapping it in an `mrow` so that we can test the `scriptlevel` (since the `mstyle` will have `scriptlevel="0"` set).  The `mrow` is removed in the post-filter if the space is being retained (otherwise the `mrow` together with its contents are removed).

### DIFF
--- a/ts/input/tex/base/BaseConfiguration.ts
+++ b/ts/input/tex/base/BaseConfiguration.ts
@@ -32,6 +32,8 @@ import * as bitem from './BaseItems.js';
 import {AbstractTags} from '../Tags.js';
 import './BaseMappings.js';
 import {getRange} from '../../../core/MmlTree/OperatorDictionary.js';
+import {MmlNode} from '../../../core/MmlTree/MmlNode.js';
+import ParseOptions from '../ParseOptions.js';
 
 /**
  * Remapping some ASCII characters to their Unicode operator equivalent.
@@ -89,6 +91,36 @@ function envUndefined(_parser: TexParser, env: string) {
   throw new TexError('UnknownEnv', 'Unknown environment \'%1\'', env);
 }
 
+/**
+ * Filter for removing spacing following \nonscript
+ * @param{ParseOptions} data The active tex parser.
+ */
+function filterNonscript({data}: {data: ParseOptions}) {
+  for (const mml of data.getList('nonscript')) {
+    //
+    //  If we are in script or script-script style
+    //    remove the space (either mspace or mrow containing it)
+    //    and remove it (and its contents) from the other lists.
+    //  Otherwise, if it is an mrow (which we added),
+    //    replace the mrow with its contents
+    //    and remove it from its list
+    //
+    if (mml.attributes.get('scriptlevel') > 0) {
+      const parent = mml.parent;
+      parent.childNodes.splice(parent.childIndex(mml), 1);
+      data.removeFromList(mml.kind, [mml]);
+      if (mml.isKind('mrow')) {
+        const mstyle = mml.childNodes[0] as MmlNode;
+        data.removeFromList('mstyle', [mstyle]);
+        data.removeFromList('mspace', mstyle.childNodes[0].childNodes as MmlNode[]);
+      }
+    } else if (mml.isKind('mrow')) {
+      mml.parent.replaceChild(mml.childNodes[0], mml);
+      data.removeFromList('mrow', [mml]);
+    }
+  }
+}
+
 
 /**
  * @constructor
@@ -135,19 +167,21 @@ export const BaseConfiguration: Configuration = Configuration.create(
       [bitem.MmlItem.prototype.kind]: bitem.MmlItem,
       [bitem.FnItem.prototype.kind]: bitem.FnItem,
       [bitem.NotItem.prototype.kind]: bitem.NotItem,
+      [bitem.NonscriptItem.prototype.kind]: bitem.NonscriptItem,
       [bitem.DotsItem.prototype.kind]: bitem.DotsItem,
       [bitem.ArrayItem.prototype.kind]: bitem.ArrayItem,
       [bitem.EqnArrayItem.prototype.kind]: bitem.EqnArrayItem,
       [bitem.EquationItem.prototype.kind]: bitem.EquationItem
-     },
-     options: {
-       maxMacros: 1000,
-       baseURL: (typeof(document) === 'undefined' ||
-                 document.getElementsByTagName('base').length === 0) ?
-             '' : String(document.location).replace(/#.*$/, '')
-     },
-     tags: {
-       base: BaseTags
-     }
+    },
+    options: {
+      maxMacros: 1000,
+      baseURL: (typeof(document) === 'undefined' ||
+                document.getElementsByTagName('base').length === 0) ?
+                '' : String(document.location).replace(/#.*$/, '')
+    },
+    tags: {
+      base: BaseTags
+    },
+    postprocessors: [[filterNonscript, -4]]
   }
 );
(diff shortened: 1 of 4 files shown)


ts/input/tex/base/BaseItems.ts
```diff
@@ -774,6 +774,45 @@ export class NotItem extends BaseItem {
...
+  }
+
+  /**
+   * @override
+   */
+  public checkItem(item: StackItem): CheckType {
+    if (item.isKind('mml') && item.Size() === 1) {
+      let mml = item.First;
+      //
+      //  Space macros like \, wrap with an mstyle to set scriptlevel=0 (so size is independent of level)
```
zorkow, line 793: typo _wrapped_?

ts/input/tex/base/BaseItems.ts
```diff
@@ -774,6 +774,45 @@ export class NotItem extends BaseItem {
...
+        if (mml !== item.First) {
+          mml = this.create('node', 'mrow', [item.Pop()]);
+          item.Push(mml);
+        }
+        //
+        //  Save the item for alter post-processing
+        //
+        this.factory.configuration.addNode('nonscript', item.First);
+      }
+    }
```
zorkow, line 812: As you can see from my comments above, I find the structure very confusing. 
So what do we want to do?
* Any `space` should be marked as `nonscript`
* If the `space` is wrapped in an `mstyle` we want to give it an extra `mrow` and mark that.
(There is no possibility that it is wrapped in multiple `mstyles`?)

Can you restructure the code somewhat to reflect this, even if it introduces some duplications? And/or update the comments?

ts/input/tex/base/BaseConfiguration.ts
```diff
@@ -89,6 +91,36 @@ function envUndefined(_parser: TexParser, env: string) {
...
 
+/**
+ * Filter for removing spacing following \nonscript
+ * @param{ParseOptions} data The active tex parser.
+ */
+function filterNonscript({data}: {data: ParseOptions}) {
+  for (const mml of data.getList('nonscript')) {
+    //
+    //  If we are in script or script-script style
+    //    remove the space (either mspace or mrow containing it)
```
zorkow, line 102: If I understand the item correctly, if the item is an `mrow` then we have added it.

zorkow (commented):
I am confused about the item (and the subsequent filter method). Please see my comments.

ts/input/tex/base/BaseItems.ts
```diff
@@ -774,6 +774,45 @@ export class NotItem extends BaseItem {
...
+   * @override
+   */
+  public checkItem(item: StackItem): CheckType {
+    if (item.isKind('mml') && item.Size() === 1) {
+      let mml = item.First;
+      //
+      //  Space macros like \, wrap with an mstyle to set scriptlevel=0 (so size is independent of level)
+      //
+      if (mml.isKind('mstyle') && mml.notParent) {
+        mml = NodeUtil.getChildren(NodeUtil.getChildren(mml)[0])[0];
```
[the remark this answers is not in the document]
dpvc: The preceding comment should read "Space macros like \` are wrapped with ...", which might make it clearer why we are looking inside for the `mspace`.  

Yes, you have it correct in the final comment above.

ts/input/tex/base/BaseItems.ts
```diff
@@ -774,6 +774,45 @@ export class NotItem extends BaseItem {
...
+      }
+      if (mml.isKind('mspace')) {
+        //
+        //  If the space is in an mstyle, wrap it in an mrow so we can test is scriptlevel.
```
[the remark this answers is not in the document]
dpvc: No, it can also fire if there is an explicit `mspace`, e.g., from `\hskip` or `\kern`.

ts/input/tex/base/BaseItems.ts
dpvc: Actually, "are wrapped" is probably better.

ts/input/tex/base/BaseConfiguration.ts
dpvc: Right.  Since only actual `mspace` elements or the `mrow` elements that we added above ever get added to the `nonscript` list, if we have an `mrow`, we know it is one that we added.

ts/input/tex/base/BaseItems.ts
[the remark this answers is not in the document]
zorkow: Sorry, I think I meant the `if` below.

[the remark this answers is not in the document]
dpvc: Yes, this `if` only happens with the first `if`, true.

ts/input/tex/base/BaseItems.ts
dpvc: I'll see what I can do.

zorkow (commented):
Thanks for the explanations. I think good commenting/structuring for `items` is important to preserve understanding what they do.  Having refactored many `item` objects and often spent some time working through all the cases, I know what I am talking about (Of course this is only a sample set of one. Still...)
I make a couple of suggestions in code.

ts/input/tex/base/BaseItems.ts
zorkow: I think what confused me most here is that they are all called `mml`.

ts/input/tex/base/BaseConfiguration.ts
```diff
@@ -89,6 +91,36 @@ function envUndefined(_parser: TexParser, env: string) {
...
+    //    and remove it (and its contents) from the other lists.
+    //  Otherwise, if it is an mrow (which we added),
+    //    replace the mrow with its contents
+    //    and remove it from its list
+    //
+    if (mml.attributes.get('scriptlevel') > 0) {
```
zorkow, line 108: Suggested comment:
```
// This space element has to be removed.
```

ts/input/tex/base/BaseConfiguration.ts
```diff
@@ -89,6 +91,36 @@ function envUndefined(_parser: TexParser, env: string) {
...
+      const parent = mml.parent;
+      parent.childNodes.splice(parent.childIndex(mml), 1);
+      data.removeFromList(mml.kind, [mml]);
+      if (mml.isKind('mrow')) {
```
zorkow, line 112: Suggested comment:
```
// The element was originally a mstyle > mspace element, so both have to be removed from the post-filter lists.
```

ts/input/tex/base/BaseConfiguration.ts
```diff
@@ -89,6 +91,36 @@ function envUndefined(_parser: TexParser, env: string) {
...
+        const mstyle = mml.childNodes[0] as MmlNode;
+        data.removeFromList('mstyle', [mstyle]);
+        data.removeFromList('mspace', mstyle.childNodes[0].childNodes as MmlNode[]);
+      }
+    } else if (mml.isKind('mrow')) {
```
zorkow, line 117: Suggested comment:
```
// The element was originally a mstyle > mspace element. As we retain the space we only have to remove the introduced `mrow`.
```

zorkow (commented):
Here's a few suggested comments for the post filter method.

dpvc:
@zorkow, OK, I've added more comments that I hope will help the situation.  See if that fits the bill.

zorkow (approved):
Thanks, that makes it a lot clearer.

PR #713, opened 2021-05-25
Add centernot package

dpvc:
This PR adds the `centernot` package, which defines `\centernot` and also a non-standard macro, `\centerOver{base}{symbol}` that places *symbol* centered over *base*, which may be useful in other situations.


zorkow (approved):
lgtm.

kno10:
It would be good if the `\centerOver` command were also available in latex - preferably added to the centernot package?

dpvc:
It's already there.  :-)

See the list of supported [macros starting with c](https://docs.mathjax.org/en/latest/input/tex/macros/index.html#c).

dpvc:
OOPS, sorry, I misunderstood.  You want it in LaTeX.  Got it.  I don't have any control over that.

kno10:
Since you added this command here, maybe you can make a pull request to the centernot latex package, to add it there, too? They should behave the same way, and you added this command. Right now, I made the following copy&paste modification of `\centernot`, and I guess this is what you had this in mind to be?
```
\makeatletter
\providecommand*{\centerOver}{\mathpalette\@centerOver}
\def\@centerOver#1#2#3{%
  \mathrel{%
    \rlap{%
      \settowidth\dimen@{$\m@th#1{#2}$}%
      \kern.5\dimen@
      \settowidth\dimen@{$\m@th#1=$}%
      \kern-.5\dimen@
      $\m@th#1{#3}$%
    }%
    {#2}%
  }%
}
\makeatother
```
(this is almost 1:1 the centernot, except that I replaced `\not` with `#3`.)

@oberdiek you seem to be responsible for the latex `centernot`  package, what do you think of `\centerOver`?

dpvc:
I'm not interested in making the pull request, but you may do so if you wish.  They probably won't like the upper-case "O", so may prefer `\centerover`.  I try to use upper-case in macro names for non-standard macros in MathJax in order to make it more clear that they are non-standard (and to avoid conflicts with macro names from packages I don't know).

PR #714, opened 2021-05-25
Go back to mhchem package, now that it is es5

dpvc:
This PR returns MathJax to using mhchem as an npm module rather than including the code directly.  The original package targeted ES6, and so caused problems in node usage.  We corrected that in 3.1.4 by supplying our own copy of it, but Martin has changed to ES5 at my request, so we can go back to the npm package.  This effectively undoes #689.


zorkow (approved):
lgtm.

PR #715, opened 2021-05-26
Implement \setOptions to change tex and package options

dpvc:
This PR implements a `\setOptions[package]{options}` macro that allows you to change TeX options and package options on the fly during the processing of the page.  For example,

```
$$\setOptions{tagSide=left}
E = mc^2 \tag{1}
$$

$$\setOptions{tagSide=right}
e^{\pi 1} + 1 = 0 \tag{2}
$$
```

will put the first tag on the left and the second on the right.  You can specify an optional package name to set the options for a package, as in 

```
\(\setOptions[amscd]{harrowsize=5em, colspace=0}\)
```

provided the package is already loaded.  (In not, then `\require` it first.)

If the `require` package is present, the `set options` package redefines `\require` so that it has an optional argument that gives options to set for the loaded package.  That is, `\require[harrowsize=5em]{amscd}` would load the `amscd` package and then set the `harrowsize`.

The `setoptions` package has the ability to allow or deny the setting of any option for any package, and can disable option setting for an entire package (for example, it disabled setting its own options so that users can override the settings of the page author).  There are defaults for the package and option values so that you don't have to list all everything, and could make it work either as a white list or a black list.  There are functions that can be overridden that filter the allowed packages, options, and values, so if the defaults are not specific enough, you can replace them with whatever testing you want to do.

The package is **not** included in the AllPackages list by default, so you have to include is explicitly if you want it.  That means you will have to at least think about what settings you want to allow to be changed.


ts/input/tex/setoptions/SetOptionsConfiguration.ts
```diff
@@ -0,0 +1,176 @@
...
+  'setoptions', {
+    handler: {macro: ['setoptions']},
+    config: setoptionsConfig,
+    priority: 3,  // must be less than the priority of the require package (which is 5).
+    options: {
+      setoptions: {
+        filterPackage: SetOptionsUtil.filterPackage,  // filter for whether a package can be configured
+        filterOption: SetOptionsUtil.filterOption,    // filter for whether an option can be set
+        filterValue: SetOptionsUtil.filterValue,      // filter for the value to assign to an option
+        allowPackageDefault: true,       // default for allowing packages if not erxplicitly set in allowOptions
```
zorkow, line 146: typo

ts/input/tex/setoptions/SetOptionsConfiguration.ts
```diff
@@ -0,0 +1,176 @@
...
+        allowOptionsDefault: true,       // default for allowing option that isn't explicitly set in allowOptions
+        allowOptions: expandable({       // list of packages to allow/disablow, and their options to allow/disallow
```
zorkow, line 148: typo

zorkow (approved):
Some typos. O/w lgtm..

PR #716, opened 2021-06-03
Properly produce initial CSS when stylesheet is loaded. 

dpvc:
This PR fixes the problem with CHTML output where the initial typeset fails to show the formulas (the CSS was missing the needed character data).


zorkow (approved):
I've tested in various scenarios and the problems we had observed have disappeared.
I will merge straight away as this is an important fix and will help me speed up my testing with then latest SRE.

PR #717, opened 2021-06-04
Provide switch for warn/error for invalid options

dpvc:
This PR changes invalid-option messages from fatal to warning (so MathJax will still run when a bad option is specified).  New configuration options are added to the `startup` block to control whether the message is a warning or a fatal one, and to provide a function that is called when there is an invalid option (so that you can trap these easier, and handle them in whatever way is desired).


zorkow (approved):
Nice. Thanks for the switch.

PR #720, opened 2021-06-07
Fix combining of mo elements so that properties don't become attributes

dpvc:
This PR resolves an issue raised by Peter in #703 where a property of one of the nodes becomes an attribute.  It turns out this is due to the `setProperties()` call, which actually sets attributes by default (and properties under certain circumstances).  The function is unfortunately named, and perhaps should be adjusted.  The fix is just to copy the properties by hand.


zorkow (approved):
lgtm.

zorkow:
In general the `NodeUtil` module needs refactoring. It is a remnant of the reimplementation of the `TexParser` to have a clean interface for nodes the while the data structures in `core` where still subject to change.

PR #721, opened 2021-06-08
Breaking changes for v3.2

dpvc:
This PR makes the breaking changes for v3.2:

* Removes the renaming of tex packages from their old mixed-case versions to all lower case.
* Removes the updating of SRE configuration options from the old A11y block.
* Moves `matchFontHeight` from the generic output jax to the CHTML output, since it doesn't apply to the SVG output.
* Removes the old `BBox.ts` and `CssStyles.ts` files from the output jax directories, since these were for backward compatibility from the we moved these files to `ts/util`.
* Moves the handling of `speechRules` and `locale` options to the menu code (as per comment in `explorer.ts`).


ts/ui/menu/Menu.ts
```diff
@@ -418,18 +418,22 @@ export class Menu {
         this.a11yVar<boolean>('subtitles'),
         this.a11yVar<boolean>('braille'),
         this.a11yVar<boolean>('viewBraille'),
-        this.a11yVar<string>('locale'),
-        this.a11yVar<string> ('speechRules'),
+        this.a11yVar<string>('locale', (value) => SRE.setupEngine({locale: value as string})),
```
zorkow, line 421: No need for parentheses.

ts/ui/menu/Menu.ts
```diff
@@ -418,18 +418,22 @@ export class Menu {
...
         this.a11yVar<boolean>('infoType'),
         this.a11yVar<boolean>('infoRole'),
         this.a11yVar<boolean>('infoPrefix'),
         this.variable<boolean>('autocollapse'),
-        this.variable<boolean>('collapsible', (collapse: boolean) => this.setCollapsible(collapse)),
-        this.variable<boolean>('inTabOrder', (tab: boolean) => this.setTabOrder(tab)),
-        this.variable<boolean>('assistiveMml', (mml: boolean) => this.setAssistiveMml(mml))
+        this.variable<boolean>('collapsible', (collapse) => this.setCollapsible(collapse)),
+        this.variable<boolean>('inTabOrder', (tab) => this.setTabOrder(tab)),
+        this.variable<boolean>('assistiveMml', (mml) => this.setAssistiveMml(mml))
```
zorkow, line 436: There should be no need for explicit lambda expressions. 
But at least we can get rid of parantheses.

ts/ui/menu/Menu.ts
```diff
@@ -402,13 +402,13 @@ export class Menu {
         this.variable<boolean>('semantics'),
         this.variable<string> ('zoom'),
         this.variable<string> ('zscale'),
-        this.variable<string> ('renderer', (jax: string) => this.setRenderer(jax)),
+        this.variable<string> ('renderer', (jax) => this.setRenderer(jax)),
```
zorkow, line 405: There should be no need for explicit lambda expressions.
But at least we can get rid of paranthese

ts/ui/menu/Menu.ts
```diff
@@ -402,13 +402,13 @@ export class Menu {
...
         this.variable<boolean>('alt'),
         this.variable<boolean>('cmd'),
         this.variable<boolean>('ctrl'),
         this.variable<boolean>('shift'),
-        this.variable<string> ('scale', (scale: string) => this.setScale(scale)),
-        this.variable<boolean>('explorer', (explore: boolean) => this.setExplorer(explore)),
+        this.variable<string> ('scale', (scale) => this.setScale(scale)),
+        this.variable<boolean>('explorer', (explore) => this.setExplorer(explore)),
```
zorkow, line 411: There should be no need for explicit lambda expressions.
But at least we can get rid of paranthese

zorkow (commented):
Just some cosmetics on the lambda expressions.
But I want to run some more tests before I approve. But after dinner.

dpvc:
I'm not sure I understand the comments about not needing the lambda expressions.  Do you mean use `this.setScale.bind(this)` instead?  I'm not sure that is either clearer or any more efficient (as the `bind` will end up making a new function anyway).

zorkow:
> I'm not sure I understand the comments about not needing the lambda expressions. Do you mean use `this.setScale.bind(this)` instead? I'm not sure that is either clearer or any more efficient (as the `bind` will end up making a new function anyway).

I was under the impression that Typescript would do the bind to `this` automatically. It worked for me the other day in my diagram explorer. But if in doubt, let's leave it like that. Still, I would drop the parentheses.

dpvc:
I'll check to see if it does or not, but yes, I can eliminate the parentheses if not.

zorkow:
Odd, I wonder why it works for my diagram-explorer. Maybe it's a fluke and might go wrong at some point. Or maybe because I also use the ES2017 library.

Anyway, I have not also found a bug in the smart preference menu for clearspeak. But that is actually on the SRE side. So one more thing to fix before I can make that release.

zorkow (approved):

PR #726, opened 2021-06-11
Use 'overflow: clip visible' for stretchy characters to avoid unwanted clipping. (mathjax/MathJax#2701)

dpvc:
This PR changes the overflow property used for the extenders in stretchy characters so that they clip only in the direction of the stretching.  This avoid accidentally clipping thin horizontal and vertical lines, for example.

It also adds the index to the `insertRule()` call in the HTMLadaptor so that it works in IE (which seems to require it), and wraps the call in `try/catch` so that rules that are only valid in modern browsers won't cause problems in IE (or other older browsers).

Resolves issue mathjax/MathJax#2701.

### DIFF
--- a/ts/adaptors/HTMLAdaptor.ts
+++ b/ts/adaptors/HTMLAdaptor.ts
@@ -69,7 +69,7 @@ export interface MinHTMLElement<N, T> {
   className: string;
   classList: DOMTokenList;
   style: OptionList;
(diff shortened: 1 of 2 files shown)


zorkow (approved):
Tested with a handful of browsers and worked fine everywhere except overline vanishes on IE 11 at 85% zoom. Otherwise fine on IE as well 

I don't think that is a deal breaker, so I approve.

PR #727, opened 2021-06-14
Fixed a typo

the-lazy-learner:
Line number 48 had a typo, of {OptionLis} instead of {OptionList}


dpvc:
Thanks.  Our repository follows the git-flow model, where changes are made on the develop branch.  Since yours is from and to the master branch, that doesn't fit with the policy.  I will make the corresponding change in the develop branch.

Thanks for spotting the error!

PR #729, opened 2021-06-15
Fix setting of movablelimits to false for \overset and similar macros.  (mathjax/MathJax#2709)

dpvc:
This PR adds a `ParseUtil` method for determining if a base needs to have `movablelimits` set to false, and then uses it in `\oerset`, `\underset`, `\overundetset`, and `ParseUtil.underOver()` (used by `\overbrace` and similar macros).  This makes sure all these situations work the same (currently there are three different criteria being used).

Resolves issue mathjax/MathJax#2709.


zorkow (approved):
Lgtm. Very nice and clean!

PR #734, opened 2021-08-18
Make CHTML use actual characters rather than CSS content specifications

dpvc:
This PR updates the CHTML output jax to use actual character data rather than CSS `content` rules to insert the characters used for display.  This allows the content of CHTML output to be selected and copied, as in v2 (note, however, that because the height/depth of the MathJax TeX fonts are near zero, the selection will not show very well -- this will be fixed as part of the update to the TeX fonts).  

This also changes the CHTML `TextNode` implementation to combine unknown characters into a single `utext` element rather than having a separate one for each character.  This allows combining characters to combine (e.g., multiple-character emoji or flags) when they are entered directly as text, even without `\text{}`).

### DIFF
--- a/ts/output/chtml/FontData.ts
+++ b/ts/output/chtml/FontData.ts
@@ -263,15 +263,12 @@ export class ChtmlFontData extends FontData<ChtmlCharOptions, ChtmlVariantData, 
    * @param {ChtmlDelimiterData} data  The data for the delimiter whose CSS is to be added
    */
   protected addDelimiterStyles(styles: StyleList, n: number, data: ChtmlDelimiterData) {
-    let c = this.charSelector(n);
-    if (data.c && data.c !== n) {
-      c = this.charSelector(data.c);
-    }
     if (!data.stretch) return;
+    const c = (data.c && data.c !== n ? this.charSelector(data.c) : this.charSelector(n));
     if (data.dir === DIRECTION.Vertical) {
-      this.addDelimiterVStyles(styles, c, data);
+      this.addDelimiterVStyles(styles, n, c, data);
     } else {
-      this.addDelimiterHStyles(styles, c, data);
+      this.addDelimiterHStyles(styles, n, c, data);
     }
   }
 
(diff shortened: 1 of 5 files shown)


ts/output/chtml/FontData.ts
```diff
@@ -436,7 +411,11 @@ export type CssMap = {[name: number]: number};
 export function AddCSS(font: CHTMLCharMap, options: CharOptionsMap): CHTMLCharMap {
   for (const c of Object.keys(options)) {
     const n = parseInt(c);
-    Object.assign(FontData.charOptions(font, n), options[n]);
+    const data = options[n];
+    if (data.c) {
+      data.c = data.c.replace(/\\[0-9A-F]+/ig, (x) => String.fromCodePoint(parseInt(x.substr(1), 16)));
+    }
```
dpvc, line 417: These three lines convert from the current `content` string like `\338` for U+0338 to an actual unicode string.  This is to avoid having to change all the font data files.  The font update for CHTML will resolve that problem and will remove these lines.

ts/output/chtml/FontData.ts
```diff
@@ -474,7 +451,11 @@ export type CssMap = {[name: number]: number};
 export function AddCSS(font: ChtmlCharMap, options: CharOptionsMap): ChtmlCharMap {
   for (const c of Object.keys(options)) {
     const n = parseInt(c);
-    Object.assign(FontData.charOptions(font, n), options[n]);
+    const data = options[n];
+    if (data.c) {
+      data.c = data.c.replace(/\\[0-9A-F]+/ig, (x) => String.fromCodePoint(parseInt(x.substr(1), 16)));
```
zorkow, line 456: long line.

zorkow (approved):
lgtm.

ts/output/chtml/FontData.ts
dpvc: Fixed.

KorigamiK:
How do I use these changes on my site?

PR #735, opened 2021-08-18
Provide support for dynamically loading font data

dpvc:
This PR adds the ability to specify font extensions that will be loaded dynamically when less-used characters are needed. You can specify any number of extensions, and these can contain characters from any number of variants, or additional delimiters.  The specification for a dynamic extension includes the file to load for the extension, along with the ranges of characters that it provides in each variant that it modifies.  When an output jax needs one of these characters or delimiters, the extension is loaded and set up, and the expression is reprocessed.

### DIFF
--- a/ts/components/startup.ts
+++ b/ts/components/startup.ts
@@ -56,6 +56,7 @@ export interface MathJaxConfig extends MJConfig {
     pageReady?: () => void;  // Function to perform when page is ready
     invalidOption?: 'fatal' | 'warn'; // Do invalid options produce a warning, or throw an error?
     optionError?: (message: string, key: string) => void,  // Function to report invalid options
+    loadAllFontFiles: false; // true means force all dynamic font files to load initially
     [name: string]: any;     // Other configuration blocks
   };
 }
@@ -67,7 +68,7 @@ export type MATHDOCUMENT = MathDocument<any, any, any>;
 export type HANDLER = Handler<any, any, any>;
 export type DOMADAPTOR = DOMAdaptor<any, any, any>;
 export type INPUTJAX = InputJax<any, any, any>;
-export type OUTPUTJAX = OutputJax<any, any, any>;
+export type OUTPUTJAX = OutputJax<any, any, any> & {font: any};
 export type COMMONJAX = CommonOutputJax<any, any, any, any, any, any, any, any, any, any, any>;
 export type TEX = TeX<any, any, any>;
 
@@ -297,7 +298,8 @@ export namespace Startup {
    * Setting Mathjax.startup.pageReady in the configuration will override this.
    */
   export function defaultPageReady() {
-    return (CONFIG.typeset && MathJax.typesetPromise ?
+    return (CONFIG.loadAllFontFiles && output.font ? output.font.loadDynamicFiles() : Promise.resolve())
+      .then(CONFIG.typeset && MathJax.typesetPromise ?
             MathJax.typesetPromise(CONFIG.elements) as Promise<void> :
             Promise.resolve());
   }
--- a/ts/components/startup.ts
+++ b/ts/components/startup.ts
@@ -68,7 +68,7 @@ export type MATHDOCUMENT = MathDocument<any, any, any>;
 export type HANDLER = Handler<any, any, any>;
 export type DOMADAPTOR = DOMAdaptor<any, any, any>;
 export type INPUTJAX = InputJax<any, any, any>;
-export type OUTPUTJAX = OutputJax<any, any, any> & {font: any};
+export type OUTPUTJAX = OutputJax<any, any, any>;
 export type COMMONJAX = CommonOutputJax<any, any, any, any, any, any, any, any, any, any, any>;
 export type TEX = TeX<any, any, any>;
 
@@ -298,7 +298,8 @@ export namespace Startup {
    * Setting Mathjax.startup.pageReady in the configuration will override this.
    */
   export function defaultPageReady() {
-    return (CONFIG.loadAllFontFiles && output.font ? output.font.loadDynamicFiles() : Promise.resolve())
+    return (CONFIG.loadAllFontFiles && (output as COMMONJAX).font ?
+            (output as COMMONJAX).font.loadDynamicFiles() : Promise.resolve())
       .then(CONFIG.typeset && MathJax.typesetPromise ?
             MathJax.typesetPromise(CONFIG.elements) as Promise<void> :
             Promise.resolve());
--- a/ts/output/common/FontData.ts
+++ b/ts/output/common/FontData.ts
@@ -155,7 +155,7 @@ export type DelimiterData = {
  * @template D  The DelimiterData type
  */
 export type DelimiterMap<D extends DelimiterData> = {
-  [n: number]: D;
+  [n: number]: D | DynamicFile;
 };
 
 /**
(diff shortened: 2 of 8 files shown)


ts/components/startup.ts
```diff
@@ -67,7 +68,7 @@ export type MATHDOCUMENT = MathDocument<any, any, any>;
 export type HANDLER = Handler<any, any, any>;
 export type DOMADAPTOR = DOMAdaptor<any, any, any>;
 export type INPUTJAX = InputJax<any, any, any>;
-export type OUTPUTJAX = OutputJax<any, any, any>;
+export type OUTPUTJAX = OutputJax<any, any, any> & {font: any};
```
zorkow, line 71: Is that a legal type?

ts/output/common/FontData.ts
```diff
@@ -225,6 +233,51 @@ export type FontParameters = {
...
+
+/**
+ * Object listing dynamic file data indexed by file name
+ */
+export type DynamicFileList = {[name: string]: DynamicFile};
+
+/**
+ * Map of characters that load a dynamic file
+ */
+export type DynamicCharMap = {[name: number]: DynamicFile};
```
zorkow, line 279: This should really be a JavaScript Map.

ts/output/common/FontData.ts
```diff
@@ -225,6 +233,51 @@ export type FontParameters = {
...
+  delimiters: DynamicRanges;      // delimiters in the file
+  setup: DynamicSetup;            // setup function to call to add the characters to the font
+  promise: Promise<void>;         // promise for when the file is loaded
+  failed: boolean;                // true when loading has failed
+};
+
+/**
+ * Object listing dynamic file data indexed by file name
+ */
+export type DynamicFileList = {[name: string]: DynamicFile};
```
zorkow, line 274: This should really be a JavaScript Map.

ts/output/common/FontData.ts
```diff
@@ -499,6 +554,11 @@ export class FontData<C extends CharOptions, V extends VariantData<C>, D extends
    */
   protected static defaultStretchVariants: string[] = [];
 
+  /**
+   * The dynamic file data
+   */
+  protected static dynamicFiles: DynamicFileList = {};
```
zorkow, line 560: ditto

ts/output/common/FontData.ts
```diff
@@ -561,12 +621,50 @@ export class FontData<C extends CharOptions, V extends VariantData<C>, D extends
...
+   * @param {CharMapMap} variants   The character data to be added
+   *
+   * @template C  The CharOptions type
+   * @template D  The DelimiterData type
+   */
+  public static dynamicSetup<C extends CharOptions, D extends DelimiterData>(
+    file: string, variants: CharMapMap<C>, delimiters: DelimiterMap<D> = []
+  ) {
+    this.dynamicFiles[file].setup = (font) => {
+      Object.keys(variants).forEach(name => font.defineChars(name, variants[name]));
```
zorkow, line 663: Can we start using `Object.entries` in situations like this? Or we should just have Maps everywhere.

zorkow (commented):
I believe we should start using Maps in places where we used JS objects as association lists, now that the implementation of the former is actually faster.

ts/output/common/FontData.ts
dpvc: We need to change the tsconfig to include additional libraries in the `libs` array.  I used `es2019` in the font-tools repository in order to allow for this, and would love to have it in MathJax-src as well.  Shall we do that?

ts/components/startup.ts
dpvc: Yes.  It is a join of the two types, and is needed in order to process the `loadAllFontFiles` option at

https://github.com/mathjax/MathJax-src/blob/aed4385da7ee81ea9d62c3304c75567f56437e21/ts/components/startup.ts#L301

But I can cast to `COMMONJAX` on that line instead, if you prefer.

ts/output/common/FontData.ts
dpvc: The reason this and the other map that you indicate (and a number of other ones like `CharMap` and `VariantMap`) are objects rather than `Map` instances is that these are used in object literals when the fonts are bing defined.  E.g., the `delimiter.ts` file for mathjax-modern begins

``` javascript
import {DelimiterMap, DelimiterData} from 'mathjax-full/js/output/chtml/FontData.js';
import {V, H} from 'mathjax-full/js/output/common/Direction.js';

export const delimiters: DelimiterMap<DelimiterData> = {
  0x28: {
    dir: V,
    sizes: [.997, 1.095, 1.195, 1.445, 1.793, 2.093, 2.393, 2.991],
    stretch: [0x239B, 0x239C, 0x239D],
    HDW: [.748, .248, .875]
  },
  0x29: {
    dir: V,
    sizes: [.997, 1.095, 1.195, 1.445, 1.793, 2.093, 2.393, 2.991],
    stretch: [0x239E, 0x239F, 0x23A0],
    HDW: [.748, .248, .875]
  },
  0x2D: {
    c: 0x2212,
    dir: H,
    stretch: [0, 0x2212],
    HDW: [0.583, 0.083, .778]
  },
  ...
```

which is a `DelimiterMap`.  While it would be possible to use a `Map` constructor here, via something like

``` javascript
export const delimiters: DelimiterMap<DelimiterData> = new Map([
  [0x28, {
    dir: V,
    sizes: [.997, 1.095, 1.195, 1.445, 1.793, 2.093, 2.393, 2.991],
    stretch: [0x239B, 0x239C, 0x239D],
    HDW: [.748, .248, .875]
  }],
  [0x29, {
    dir: V,
    sizes: [.997, 1.095, 1.195, 1.445, 1.793, 2.093, 2.393, 2.991],
    stretch: [0x239E, 0x239F, 0x23A0],
    HDW: [.748, .248, .875]
  }],
  [0x2D, {
    c: 0x2212,
    dir: H,
    stretch: [0, 0x2212],
    HDW: [0.583, 0.083, .778]
  }],
  ...
]);
```

this adds extra characters (two for each delimiter), making the file longer, and to see is less clear.  Also, there was a feature request some time ago about turning the font data into stringified JSON data, as apparently it is faster to JSON parse the string than to have the javascript compiler have to compile the large object.  I guess that could still be done for the array being passed to `new Map()`.  But this means I need to rewrite the font-generation code and rebuild all the font packages.  I'd rather put that off until later in case there are other changes that need to be made to the fonts.

Also, I mentioned `CharMap` and `VariantMap`, and there are a number of other such objects used in the font definitions.  It would be good to handle those all at once, so I'd like to put this off for now.  I'm also not sure if any of the PRs that are built on top of this one will need to be modified if we make the changes here, so I thin kit would be best to get all the font PRs merged, and then do the change over to maps, if that seems to be the way to go.

Is that acceptable?

ts/output/common/FontData.ts
dpvc: Here, the `Object.keys()` are for the `variants` object, which is a `CharMapMap`, and so that would need to be changes to a `Map()` as well.  That will affect all the font definitions, and again, I'd like to put this off until all the font PRs are merged, do a pass over the code for these kinds of changes, and then rebuild the font packages only once in response.

ts/output/common/FontData.ts
dpvc: I can do this one, but it will cause conflicts in later PRs, so I'd like to put it off until the font PRs are merged, and make the changes then, along with the other Map fixes, if that's OK.

dpvc:
I've made comments about keeping the maps as objects for now, and making a change after the font PRs are all merged, if that is OK.  So I'd like to merge this one now and do those changes later.  I've made a note to myself to do that.

ts/output/common/FontData.ts
zorkow: That does make sense. 
Ultimately, it might be a good idea to hide the maps in a wrapping class, with a corresponding `fromJson` and `toJson` method.

ts/output/common/FontData.ts
zorkow: Ditto. Let's put it onto the *todo* stack.

ts/output/common/FontData.ts
zorkow: Yes, let's add `es2019`.

zorkow (approved):
Thanks for cleaning up the type. Looks much better this way!

PR #737, opened 2021-08-18
Update build tools to work with extensions better

dpvc:
This PR update the build tools (and the web pack configuration in particular) to handle the situation for building an extension (like the fonts) more easily, and to handle imports that use node modules rather than relative paths.  This allows extensions to import from `mathjax-full` (or the future `@mathjax/src`) directly, and still have it work properly with the packing tools.

It also adds a `--no-subdirs` option to `makeAll` so that it won't build extensions from subdirectories.


zorkow (approved):
lgtm.

PR #738, opened 2021-08-18
Add versions to component files and check them when loaded

dpvc:
This PR adds support for version checking so that components know the version that was used to build them, and the component loader can check the component version to make sure it is the same as the one loading it.  If there is a mismatch, a (non-fatal) warning is issued.  (In the future, this could be extended to make better choices about whether the version mismatch should be fatal or not.)  The warnings can be suppressed via the `versionWarnings: false` option to the `loader` configuration.  

The idea is that, now that we are going to have separate repositories for fonts and TeX extensions, and people are going to be making their own extensions, we need a means of checking whether the extensions can work with the vision of MathJax that is loading them.  This is a first step in that direction, which at least indicates when a version mismatch occurs.

Note that most of the files in this PR are only changed to include an ID to use with the versions, so can be gone over quickly.

### DIFF
--- a/ts/components/version.ts
+++ b/ts/components/version.ts
@@ -1,0 +1,25 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2018-2021 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  The version of MathJax (iused to tell what version a component
+ *                was compiled against).
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+export const VERSION = '3.2.0';
--- a/ts/components/version.ts
+++ b/ts/components/version.ts
@@ -16,7 +16,7 @@
  */
 
 /**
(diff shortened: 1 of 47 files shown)


ts/components/version.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @fileoverview  The version of MathJax (iused to tell what version a component
```
zorkow, line 19: typo

ts/components/version.ts
```diff
@@ -0,0 +1,25 @@
...
+ *                was compiled against).
+ *
+ * @author dpvc@mathjax.org (Davide Cervone)
+ */
+
+export const VERSION = '3.2.0';
```
zorkow, line 25: This could be computed from the package.json.

zorkow (approved):
An alternative could be to import the `package.json`  and read the version from there. This way there is no need to keep both in sync.

ts/components/version.ts
dpvc: fixed.

dpvc:
OK, I've updated `webpack.common.js` to look up the version from the `package.json` file, and `version.ts` to either use that value (when in a webpacked version), or look up the version from `package.json` if run in node directly.  The lookup in version.ts is removed by webpack's tree shaking, so it doesn't enlarge the resulting file (yay!).  This works everywhere expect in the lab, where you need to define the global `PACKAGE_VERSION = "3.2.0"` or maybe `v3-lab` instead of `3.2.0`.

So I think this does what you want without adversely affecting the webpacked versions.  So good suggestion.

zorkow:
Yes, that's exactly what I had in mind: Version should be looked up during webpacking and is then set in stone.

zorkow (approved):
lgtm.

Another advantage is that we can now simply use `npm version` for version bumps.

PR #741, opened 2021-09-06
fix: Remove circular dependency

edeustace:
When bundling mathjax - you are warned that there is a circular
dependency:

```shell
code: CIRCULAR_DEPENDENCY, message: Circular dependency: node_modules/mathjax-full/js/input/tex/TexParser.js -> node_modules/mathjax-full/js/input/tex/ParseUtil.js -> node_modules/mathjax-full/js/input/tex/TexParser.js
```

This can lead to runtime errors as the bundler has to guess which module
to define first.

This commit removes this circular dependency by moving the functions in
ParseUtil that TexParser used into 2 new namespaces.

+ DimensionUtil -> matchDimen
+ StringUtil -> trimSpaces

I'm trying to be consistent with how the modules are laid out.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -31,86 +31,8 @@ import TexParser from './TexParser.js';
 import TexError from './TexError.js';
 import {entities} from '../../util/Entities.js';
 import {MmlMunderover} from '../../core/MmlTree/MmlNodes/munderover.js';
-
-
+import {Em} from '../../util/lengths.js';
 namespace ParseUtil {
-
-  // TODO (VS): Combine some of this with lengths in util.
-  const emPerInch = 7.2;
-  const pxPerInch = 72;
-  // Note, the following are TeX CM font values.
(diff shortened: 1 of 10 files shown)


zorkow:
Thanks for the PR. The issue's been on the todo list for some time, so I appreciate you taking it on.
There are three changes I would like you to make before we can merge:

* Rebase the PR onto the `mathjax:develop` branch (`master` branch always contains the latest stable version) 
* Move the content of `ts/input/tex/StringUtil.ts` into  `ts/util/string.ts`.
* Move the content of `ts/input/tex/DimensionUtil.ts` into `ts/util/lengths.ts`. 

That should similarly take care of the circular dependency and will put all the utilities into the module where they actually should go. There might even be some scope for future code rationalisation in `lengths.ts`.

edeustace:
Hi @zorkow - I've applied the changes you requested - however - I've noted some other circular deps that should also be fixed. I'm a bit time pressed to do all the fixes right now. So we can merge and re-visit, or if you want to keep it all on the branch?

PR #742, opened 2021-09-15
Fix problem with msubsup when subscript is blank (#2765)

dpvc:
This PR fixes an issue in CommonHTML output where `msubsup` output when the subscript is empty shows the superscript in the subscript position in some instances (see the initial issue mathjax/MathJax#2765).  This was caused by CSS that was overriding the `display` property for the super- and subscript elements, when it really only needed to be for the spacer.  The order in which the CSS for the msubsup and the scripts are output also affected the results.  The fix is to limit the CSS to just the spacer.

Resolves issue mathjax/MathJax#2765.


EricDunsworth:
@dpvc Did you mean to say "shows the **superscript** in the subscript position" in the OP?

dpvc:
@EricDunsworth, yes, thanks.  I've corrected the statement.

zorkow (approved):
lgtm

PR #743, opened 2021-09-15
Fix placement of surd when root extends above the horizontal line (mathjax/MathJax#2764)

dpvc:
This PR fixes a problem in SVG output for `mroot` when the height of the root extends above that of the surd itself, as in

```
\sqrt[\frac{1}{2}]{1}
```

In that case, the root symbol was too high.  This PR positions it correctly.

Resolves issue mathjax/MathJax#2764.


zorkow (approved):
Works (at least in the lab and Brave). lgtm.

PR #751, opened 2021-12-01
Add locale for Traditional Chinese

toto6038:


dpvc:
Thanks for your interest in supporting MathJax.  The section of code you are modifying here is not actually used in MathJax version 3.  The legacy code you are editing is used to patch the version 2 AsciiMath input jax into version 3 by incorporating much of the version 2 framework in order to allow it to operate, but the menu code that is part of the legacy MathJax.js file is not used.  So your edits will have no effect.  For that reason, I'm going to close this PR.

Note that the actual version 2 code is in the mathjax/MathJax repository in the legacy-v2 branch, and that [both Chinese versions](https://github.com/mathjax/MathJax/blob/3b461438246adfcf67690795fcc0ae6dc4e335fe/unpacked/MathJax.js#L1215-L1216
) are already included there (the legacy code here was branched off before the traditional Chinese locale was added to v2).  If you want to request that the menu tables be changed, that is the repository where it should be done.

PR #752, opened 2022-01-09
There's a small typo in the asciimath build.json

artagnon:


dpvc:
Thanks for catching the typo.  (The error currently has no effect, since only the base name of the component it used, which is probably why we didn't spot it ourselves.)

The problem with the PR is that it needs to be from the `develop` branch and targeted to the `develop` branch.  Can you rebase the branch from `develop` and updated the PR (or submit a new one if easier)?  Thanks!

PR #753, opened 2022-01-09
yarn: switch package manager from npm to yarn

artagnon:
yarn is a much better package manager than npm overall, and its linker
is far more sophisticated than the one from npm. Unfortunately, the
linker in question is dubbed as the "Plug n' Play feature" of yarn, and
MathJaX isn't quite ready to move to it. In this commit, we switch to
yarn berry, preserving the node_modules directory, to see if it piques
interest of the community to switch to the yarn linker.


dpvc:
Can you be clearer on what you mean by "much better package manager" and what advantages this change would provide for MathJax?

It looks like you are trying to use yarn's "zero install" approach, but I don't think that is appropriate for this repository.  For example, most of the packages that you have included are the developer packages needed to compile MathJax and build its components.  That is not something most users of MathJax will need to do.  There are only four packages needed to run MathJax, and those only install three others, so I would expect only 7 cached packages.

In any case, I'm not keen on including even zipped versions of these other tools as part of our own repository.  Is the only advantage to not have to do `yarn add mathjax-full` (or `npm install mathjax-full`)?  I think we need more information and a compelling reason to make this change, especially since `yarn add mathjax-full` does work now just fine.

artagnon:
The changes shown with this PR are somewhat misleading; sorry about that. The zero-install approach is just a side-effect; not a feature. The yarn dependency resolver and yarn linker are the compelling propositions. What this concretely means is that:

1. The dependency tree in the js world is somewhat complex: a package X might be providing package Y "implicitly", but adding a dependency on package X might pull in package Y into the dependency tree, breaking the linking process. The npm dependency resolver isn't sophisticated enough to specify these "implicit" dependencies, but the yarn dependency resolver is. The linking process in npm doesn't exactly fail, because it's handled in some ad-hoc way I don't fully understand, but in yarn, the linking process fails unless (a) all the upstream packages we depend are well-written with respect to specifying their dependencies, or (b) we specify the "peer dependencies" by hand, correcting for the deficiencies in the upstream packages.
2. The yarn linker is much faster than the npm linker, and does a better job of pruning out irrelevant things overall.

I've seen this npm-versus-yarn difference playing out in other projects with lots of dependencies, and it is perhaps not too relevant to MathJaX. Perhaps switching to the yarn linker won't yield much benefit for MathJaX either; I can't know unless I try it out. I was hoping that other people in the community with more yarn experience than me would comment on this issue.

artagnon:
Perhaps worth noting: I can't use the yarn linker in projects that include mathjax-full as a dependency. This would be the major user-facing change.

artagnon:
Closed due to inactivity.

PR #756, opened 2022-01-18
Fix typo in \DeclarePairedDelimiter*, and substitute arguments in pre and post sections. (mathjax/MathJax#2816, mathjax/MathJax#2758)

dpvc:
This PR fixes the names of the `\DecalrePairedDelimiter` macros (no `s`), but retains the old spellings for compatibility, at least for now.  It also does argument substitution on the pre and post material, as it does for the body.

Resolves issues mathjax/MathJax#2816 and mathjax/MathJax#2758.


zorkow (approved):
lgtm.

PR #757, opened 2022-01-18
Make sure explicit attributes added by \mmlToken are not removed. (mathjax/MathJax#2806)

dpvc:
This PR modifies `\mmlToken` to make sure the attributes specified by it are not removed but the `cleanAttributes` filter, even if their values are the default values.  This was causing the spacing for `\bmod` to be lost (and so it failed to have the proper space in super and subscripts).

Resolves mathjax/MathJax#2806.

### DIFF
--- a/ts/input/tex/FilterUtil.ts
+++ b/ts/input/tex/FilterUtil.ts
@@ -71,8 +71,10 @@ namespace FilterUtil {
       if (!attribs) {
         return;
       }
+      const keep = new Set((attribs.get('mjx-keep-attrs') || '').split(/ /));
+      delete (attribs.getAllAttributes())['mjx-keep-attrs'];
       for (const key of attribs.getExplicitNames()) {
-        if (attribs.attributes[key] === mml.attributes.getInherited(key)) {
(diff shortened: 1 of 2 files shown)


dpvc:
OK, thanks, I'll check into it.

pkra:
Sorry for not being more conclusive -- I'm testing issue2828, issue2806, issue2766, issue2764 and SRE4 final all at once. Lots of tiny visual changes but nothing too bad so far.

dpvc:
@pkra, no problem.  We appreciate your reporting the issues.

dpvc:
OK, I found it.  It is in #763, where I used `underaccent` instead of `accentunder`.  OOPS!  Because `underaccent` is not a MathML attribute, it is treated as one that you have added yourself, and so retains it in the output.  I'll make a fix for the attributes.  Thanks for the report, which will save me some more public embarrassment.

pkra:
Ah, sorry for misreporting -- I should've started by just updating from develop.

dpvc:
No problem.  I was pretty sure where the problem came from.

PR #758, opened 2022-01-18
Handle document created by parsing in XHTML.  (mathjax/MathJax#2788)

dpvc:
MathJax uses HTML parsing by default for parsing MathML strings (so that when we support HTML-in-MathML, it should work).  But in an XHTML document, the parsing structure is not the same as in HTML, so this PR alters the HTML adaptor to work with that structure as well.

### DIFF
--- a/ts/adaptors/HTMLAdaptor.ts
+++ b/ts/adaptors/HTMLAdaptor.ts
@@ -227,21 +227,21 @@ AbstractDOMAdaptor<N, T, D> implements MinHTMLAdaptor<N, T, D> {
    * @override
    */
   public head(doc: D) {
-    return doc.head;
+    return doc.head || (doc as any as N);
   }
 
   /**
    * @override
    */
   public body(doc: D) {
-    return doc.body;
+    return doc.body || (doc as any as N);
   }
 
   /**
    * @override
    */
   public root(doc: D) {
-    return doc.documentElement;
+    return doc.documentElement || (doc as any as N);
   }
 
   /**


zorkow (approved):
I see that the solution works, although I find it rather broad.

So for future considerations, here are a couple of points:
* When using the `HTMLAdaptor` I would assume it's behaviour to correspond to HTML. If there is no head or body something is wrong with the document and I (developer or client) would have to do something with that information.
* Instead of conflating xhtml and html we could have an `XHTMLAdaptor`.
* We do have a `chooseAdaptor` and a `browserAdaptor` which could ensure that the right one is chosen depending on the document.

dpvc:
The issue occurs where a document fragment is parsed.  In this case, it was the MathML that was being parsed as a string, like `<math><mi>x</mi></math>`, which is not a complete document.  I suppose we could put `<!DOCTYPE html><html><head><title>None</title></head><body>' and '</body></html>' around the math string instead, but it's not usually necessary.  (I include the `<title>` tags because it is a required tag for an HTML document.)  That could be done here:

https://github.com/mathjax/MathJax-src/blob/2dd53ce6c8af3c9cceba0baf014ea9b065130774/ts/input/mathml.ts#L139-L141

Would that be better?  There are a couple of other places that might be necessary (`adaptor.parse()` is used in several places).

dpvc:
The `chooseAdaptor` was really just for testing purposes so that the same code could be run in node and in the browser. 

If you wanted to use it for choosing between HTMLadaptor, the liteAdaptor (as it does now), and a proposed XHTMLadaptor, then any component that includes it would have to include all three adaptors, enlarging the component file for very little reason (there is no real reason to have the liteAdaptor in the browser, even if it is only 20K).

PR #759, opened 2022-01-18
Trim MathML string before parsing it.  (mathjax/MathJax#2805)

dpvc:
This PR trims whitespace from the ends of the serialized MathML sting before parsing it (so that we don't get unwanted text nodes on either side that cause MathJax to indicate multiple nodes were produced).

Resolves issue mathjax/MathJax#2805.


zorkow (approved):
Great PR! I used to put the `trim()` into our labs because I always found that error so annoying!

PR #762, opened 2022-01-24
Allow SRE to use resolverPath during startup

dpvc:
The addition of the paths, dependencies, etc, was being added in the `components/src/startup.js` file, which was loaded last by the combined components like `tex-chtml`.  This means that that data wasn't available to other components that are imported into the combined components (like SRE4).  This PR separates adding of paths, etc, into a startup init file, so that they can be added before other imports, making the default paths, etc, available to imported files, as they should be.  It also makes it easier to make combined configuration files by loading the init file rather than the lib/startup.js file, which wasn't very intuitive.

It also, fixed the sre webpack `libs` definition so that it shares the startup code.

Finally, I removed a closure function in `sre.js` that wasn't needed, since there are no local variables.

### DIFF
--- a/components/src/a11y/sre/webpack.config.js
+++ b/components/src/a11y/sre/webpack.config.js
@@ -5,7 +5,8 @@ module.exports = PACKAGE(
   '../../../../js',                   // location of the MathJax js library
   [                                   // packages to link to
     'components/src/input/mml/lib',
-    'components/src/core/lib'
+    'components/src/core/lib',
+    'components/src/startup/lib'
   ],
   __dirname                           // our directory
 );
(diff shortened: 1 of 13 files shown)


zorkow (approved):
Sorry, this is a very egoistic code review... 
I'll merge later. In the mean time I have the `delay` flag running in SRE, as well as preloading of the base locales for webpack distribution. The latter does not yet work in the lab, but will, once I have added `.js` extension to all imports in SRE, which is apparently good style anyway. This means we need a SRE v4.0.1 to include.

dpvc:
That all sounds good.  Looking forward to it.

PR #765, opened 2022-01-30
Move mml3 filter to an mmlFilter so that forceReparse isn't needed. (mathjax/MathJax#2718)

dpvc:
This PR move the MML3 filter from being a `preFilter` to being an `mmlFilter`.  That way it will always run and you don' have to specify `forceReparse: true`.  It also means that (in the browser) you don't have to serialize the math and then re-parse it, as is currently done, since the browser's XSLT parser can work on any DOM fragment.  The node version must serialize and re-parse, however.

There is a new `enableMml3` configuration option that controls whether the filter is added or not (so that if the extension is packaged with a combined configuration, it can be disabled).

This also make the `mmlFilters` public, as it should have been originally.

Resolves issue mathjax/MathJax#2718.

### DIFF
--- a/ts/input/mathml.ts
+++ b/ts/input/mathml.ts
@@ -76,7 +76,7 @@ export class MathML<N, T, D> extends AbstractInputJax<N, T, D> {
   /**
    * A list of functions to call on the parsed MathML DOM before conversion to internal structure
    */
-  protected mmlFilters: FunctionList;
+  public mmlFilters: FunctionList;
 
   /**
    * @override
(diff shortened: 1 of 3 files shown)


zorkow (approved):
lgtm.
Btw. I could not get mml3 extension to work in the lab yesterday. It could not load the `.json` file.

dpvc:
> I could not get mml3 extension to work in the lab yesterday. It could not load the .json file.

It should not be trying to load the `.json` file, as that should only occur when used in `node`.  It works in the lab for me (using the `update-lab` branch).

PR #766, opened 2022-01-31
Make sure all character data is included when adaptiveCSS is false.  (mathjax/MathJax#2724)

dpvc:
This PR fixes a problem introduced in v3.2.0 where setting `adaptiveCSS` to `false` would not produce the full CSS; it included all the wrapper CSS, but left out the unused character CSS.  This fixes that problem.

Resolves issue mathjax/MathJax#2724.

### DIFF
--- a/ts/output/chtml/FontData.ts
+++ b/ts/output/chtml/FontData.ts
@@ -185,7 +185,11 @@ export class CHTMLFontData extends FontData<CHTMLCharOptions, CHTMLVariantData, 
     //
     //  Add the styles for delimiters and characters
     //
-    this.updateStyles(styles);
+    if (this.options.adaptiveCSS) {
+      this.updateStyles(styles);
+    } else {
+      this.allStyles(styles);
+    }
     //
     //  Return the final style sheet
     //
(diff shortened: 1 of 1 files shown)


ts/output/chtml/FontData.ts
```diff
@@ -209,6 +213,35 @@ export class CHTMLFontData extends FontData<CHTMLCharOptions, CHTMLVariantData,
...
+      this.addDelimiterStyles(styles, N, this.delimiters[N]);
+    }
+    //
+    //  Add style for all character data
+    //
+    for (const name of Object.keys(this.variant)) {
+      const variant = this.variant[name];
+      const vletter = variant.letter;
+      for (const n of Object.keys(variant.chars)) {
+        const N = parseInt(n);
```
zorkow, line 234: Why do you need to parse that here? `variant.chars[n]` should return the value and in case the `continue` fires first, you don't need the actual `number`.

zorkow (approved):
One minor issue. O/w lgtm.

ts/output/chtml/FontData.ts
dpvc: This is just to make Typescript happy.  Since `Object.keys()` returns `string[]`, while `variant.char` is indexed by numbers; it is a `CharMap`:

https://github.com/mathjax/MathJax-src/blob/2dd53ce6c8af3c9cceba0baf014ea9b065130774/ts/output/common/FontData.ts#L58-L60

I could change that to being string indexed, if you prefer.  That doesn't seem to cause any type errors.

I've had to do this a couple of other times, as there are several other namer-indexed objects.  I could try converting them all to strings and see if anything blows up.

zorkow: I see. Strictly speaking Typescript is of course correct, as Object keys are always strings, so I suppose it is safer this way.

dpvc: OK, we can see about perhaps changing over to string-indexed objects rather than number-indexed in a future code clean-up

PR #767, opened 2022-01-31
Add a minimum height for accented characters.  (mathjax/MathJax#2766)

dpvc:
This makes the output for `mover` and `munderover` consistent with TeX when used for accents, and brings v3 in line with v2 output.

Resolves issue mathjax/MathJax#2766.


zorkow (approved):
lgtm.

PR #768, opened 2022-01-31
Make safe extension properly handle scriptlevel of 0. (mathjax/MathJax#2745)

dpvc:
The Safe extension was returning a number rather than a string when it filters the `scriptlevel` (and `scriptSizeMultipler`) attributes, and so a script level of 0 would end up reading as false rather than a string consisting of 0, and so the script level was filtered out.  This PR returns strings for these attributes.

Resolves issue mathjax/MathJax#2745.


zorkow (approved):
lgtm.

PR #769, opened 2022-02-02
Fix problem where some menu settings weren't sticky  (mathjax/MathJax#2786)

dpvc:
This PR adds some initialization to the menu code that makes sure that the initial settings of the menu take effect on a page reload.  This was not the case for the in-tab-order, assistive MML, renderer, and scale settings (argh!).

Resolves issue mathjax/MathJax#2786, and similar problems with the other settings.

@zorkow, activating the explorer seems to override the in-tab-order setting.  Is that intentional?

### DIFF
--- a/ts/ui/menu/Menu.ts
+++ b/ts/ui/menu/Menu.ts
@@ -371,6 +371,7 @@ export class Menu {
     this.initSettings();
     this.mergeUserSettings();
     this.initMenu();
+    this.applySettings();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


zorkow:
> @zorkow, activating the explorer seems to override the in-tab-order setting. Is that intentional?

Yes, it is. Once the explorer is on the elements need to be focusable using the tab as they now have aria role application. The in-tab-order only refers to the property of the context menu.
But I note that we set `tabIndex=1`  in the explorer. That should be `0`.

zorkow (approved):
lgtm.

PR #770, opened 2022-02-02
Make variables local in legacy code.  (mathjax/MathJax#2748)

dpvc:
This PR makes some variables local that were accidentally global variables in the AsciiMath code.

Resolves issue mathjax/MathJax#2748.


zorkow (approved):
lgtm.

PR #771, opened 2022-02-02
Make sure math-in-text forms an ORD atom with textmacros. (mathjax/MathJax#2828)

dpvc:
This PR forces the math that is typeset within text to be of TeX class ORD when the texmacros extension is in use (making it consistent with when textmacros is *not* in use).

Resolves issue mathjax/MathJax#2828.


pkra:
Thanks, Davide. I'll test and pull it into the AMS's main branch.

zorkow (approved):
lgtm.

PR #773, opened 2022-02-07
Have physics package match nested parentheses, fix spacing issues (mathjax/MathJax#2760, mathjax/MathJax#2831)

dpvc:
This PR causes nested delimiters to be handled properly (not closing the delimiters until the inner ones are also closed).  It also fixes the spacing issue around the delimiters (the physics package prevents the stretchy delimiters from getting TeX class INNER, so I've removed the properties that cause `setTeXclass()` to mark it as INNER so that it will be handled as a normal `mrow` with an OPEN at the front and a CLOSE at the end).  Also, some operators were being enclosed in `mi` rather than `mo`, producing the wrong spacing (e.g., in `\div` and `\curl`), so I've changed that as well to match the LaTeX output.  Finally, an expression that is missing its final delimiter would produce a math input error, so I've added the missing error message so that the missing close delimiter will be reported.

This resolves issues mathjax/MathJax#2760 and mathjax/MathJax#2831.

### DIFF
--- a/ts/input/tex/physics/PhysicsItems.ts
+++ b/ts/input/tex/physics/PhysicsItems.ts
@@ -86,8 +86,8 @@ export class AutoOpen extends BaseItem {
       this.getProperty('big') as string
     );
     //
-    //  Remove fence markers so it is treated as a regular mrow when
-    //  setting the tex class, so it is not class INNER (#2760)
+    //  Remove fence markers that would cause it to be TeX class INNER,
+    //  so it is treated as a regular mrow when setting the tex class (#2760)
     //
     NodeUtil.removeProperties(mml, 'open', 'close', 'texClass');
     return mml;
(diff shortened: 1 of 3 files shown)


ts/input/tex/physics/PhysicsItems.ts
```diff
@@ -64,20 +78,36 @@ export class AutoOpen extends BaseItem {
...
+    let mml = ParseUtil.fenced(
+      this.factory.configuration,
+      this.getProperty('open') as string,
+      super.toMml(),
+      this.getProperty('close') as string,
+      this.getProperty('big') as string
+    );
+    //
+    //  Remove fence markers so it is treated as a regular mrow when
+    //  setting the tex class, so it is not class INNER (#2760)
```
zorkow, line 90: Rephrase?

zorkow (approved):
Works fine. Only one minor request on a comment.

ts/input/tex/physics/PhysicsItems.ts
dpvc: Done

PR #775, opened 2022-02-08
Refactor node-based DOM adaptors to use a mixing for common code

dpvc:
This PR introduces a `NodeMixin` function for adding common methods and properties to a base DOM adaptor.  These methods are the ones that implement features that aren't available in node DOMs, like `getComputedStyles()` and measuring of DOM elements.

The jsdom and linkedom adaptors are refactored to use this mix-in, as is the liteDOM, though that is a bit more complicated, as the mix-in can't be used with an abstract class, and so we need to make a base lite class to use with the mix-in to get the full adaptor.  This means that the last four methods in `LiteBase` will be overridden by the mix-in, but I don't see any way around that.

### DIFF
--- a/ts/adaptors/NodeMixin.ts
+++ b/ts/adaptors/NodeMixin.ts
@@ -1,0 +1,160 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2022 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


zorkow (commented):
Are these the changes I've just requested in the previous PR?

dpvc:
> Are these the changes I've just requested in the previous PR?

Yes.  I also was thinking a refactoring was necessary.  :-)

zorkow (approved):
Nice! I like more red lines than green lines.

dpvc:
> I like more red lines than green lines.

Yes, red lines are good!

Issue #177, opened 2018-12-05
TeX input, amsmath: change `\sideset` implementation to generate multiscripts


pkra:
Cf. https://github.com/mathjax/MathJax/issues/1217.

AFAICT,  the current beta.3 still creates subsups.

Is this planned?

dpvc:
So far, the TeX jax produces (essentially) the same results as the v2 TeX input jax.  We have not adjusted individual macros at this point, except where necessary for changes in the infrastructure.  Improvements to the TeX output will come in a future release.

pkra:
Thanks, I understand that. Could you tell me if you are planning to include such a change in v3?

dpvc:
I would like it to use `mmultiscripts`, but (as pointed out in the discussion of #1217), it may not be a trivial change.

pkra:
Thanks.

PR #776, opened 2022-02-09
Re-implement \sideset using mmultiscripts.  (mathjax/MathJax#1217)

dpvc:
This re-implements the `\sideset` macro to use `mmultiscripts` rather than as a replacement macro that has a phantom duplicate of the base with super- and subscripts.  In order to produce the same output as LaTeX, a new property was added to allow `mmultiscripts` to control the alignment of the pre- and post-scripts; `\sideset` has the pre-script left-aligned, but `mmultiscripts` has them right-aligned, so this property is needed to allow left-alignment to be specified.  That required changes to the CHTML and SVG output to support the new property, and to the MathML serializer to output the property to MathML, and the MathML input jax to read the property on input.  For the latter, since the number of such properties is getting larger, the old if-then-else-if construction was replaced by a switch.

Resolves (closed) issue mathjax/Mathjax#1217 and #177.

### DIFF
--- a/ts/input/tex/ams/AmsMethods.ts
+++ b/ts/input/tex/ams/AmsMethods.ts
@@ -330,7 +330,7 @@ function splitSideSet(mml: MmlNode): [MmlNode, MmlNode] {
 }
 
 /**
- * Utility for checking if a \setset argument has scripts with an empty base.
+ * Utility for checking if a \sideset argument has scripts with an empty base.
  * @param {MmlNode} mml The node to check.
  * @return {boolean} True if the base is not and empty mi element.
  */
--- a/ts/input/tex/ams/AmsMethods.ts
+++ b/ts/input/tex/ams/AmsMethods.ts
@@ -240,42 +240,103 @@ AmsMethods.HandleOperatorName = function(parser: TexParser, name: string) {
  * @param {string} name The macro name.
  */
 AmsMethods.SideSet = function (parser: TexParser, name: string) {
-  const pre = parser.ParseArg(name);
-  const post = parser.ParseArg(name);
+  //
+  //  Get the pre- and post-scripts, and any extra material from the arguments
+  //
+  const [preScripts, preRest] = splitSideSet(parser.ParseArg(name));
+  const [postScripts, postRest] = splitSideSet(parser.ParseArg(name));
   const base = parser.ParseArg(name);
-  if (!checkSideSet(pre) || !checkSideSet(post)) {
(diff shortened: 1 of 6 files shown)


zorkow (commented):
I am all for moving `\sideset` to `mmultiscripts`, not only because it will improve voicing.

But there are a couple of points we might want to discuss, as the old implementation was 

In LaTeX only large operators are legal in `\sideset` (that's why I found it always useless!), so we never really implemented the exact macro behaviour. In my opinion this is an improvement.

On the other hand the LaTeX macro (and our old implementation) allows for non-empty base elements in the arguments, making them effectively sub/superscript elements, similar to our old implementation. As far as I understand the re-implementation, this is no longer possible.

Finally, this is potentially a breaking change, i.e., not for a patch release.

ts/input/tex/ams/AmsMethods.ts
```diff
@@ -233,6 +234,111 @@ AmsMethods.HandleOperatorName = function(parser: TexParser, name: string) {
...
+  //  If there are pre-scripts...
+  //
+  if (preScripts) {
+    //
+    //  If there is other material...
+    //
+    if (preRest) {
+      //
+      //  Replace the empty base of the prescripts with a phantom element of the
+      //    original base, with width 0 (so of the correct height and depth).
```
zorkow, line 260: Missing word?

ts/input/tex/ams/AmsMethods.ts
```diff
@@ -233,6 +234,111 @@ AmsMethods.HandleOperatorName = function(parser: TexParser, name: string) {
...
+    if (!mml || (mml.isInferred && mml.childNodes.length === 0)) return [null, null];
+    if (mml.isKind('msubsup') && checkSideSetBase(mml)) return [mml, null];
+    const child = NodeUtil.getChildAt(mml, 0);
+    if (!(mml.isInferred && child && checkSideSetBase(child))) return [null, mml];
+    mml.childNodes.splice(0, 1); // remove first child
+    return [child, mml];
+}
+
+/**
+ * Utility for checking if a \setset argument has scripts with an empty base.
```
zorkow, line 333:
```suggestion
 * Utility for checking if a \sideset argument has scripts with an empty base.
```

zorkow (approved):
Only some typos, o/w lgtm.

ts/input/tex/ams/AmsMethods.ts
dpvc: Not really, but perhaps could be better worded.  The phantom element is "of the correct height and depth".  I'll reword.

PR #777, opened 2022-02-10
Lazy extension update

dpvc:
This PR updates the lazy-typesetting extension to add several new features.

*  The ability to typeset all untypeset math on the page before printing it (so that it will print properly, even when parts of the page haven't been viewed).  (mathjax/MathJax#2711)
*  A new option to control the margin to use for the observer trigger (so that expressions slightly outside the viewport will be typeset in anticipation of scrolling.
*  The ability to mark some containers to force the math they contain to always be typeset, regardless of whether they are in the viewport.  This allows math that is used in JSXgraph, for example, to be placed properly.  (mathjax/MathJax#2720)

There are also a typo in a comment that is fixed, and a redundant definition of `rerender()` in the MathDocument interface declaration was removed.

The three bullet points are each done in a separate commit, so it might be easier to look at them separately, rather than mixed together, as in this PR.  The third is the most complicated, as there was considerably infrastructure needed to manage that.  The new `lazyAlwaysTypeset` option is either an array of container nodes whose contents should always be typeset (non-lazily), or a single string or array of strings that represent CSS selectors for the containers to use.  E.g., 

``` js
    lazyAlwaysTypeset: '.always-typeset'
```

would make all containers with `class="always-typeset"` to have their math typeset regardless of whether they appear in the viewport or not.

### DIFF
--- a/ts/ui/lazy/LazyHandler.ts
+++ b/ts/ui/lazy/LazyHandler.ts
@@ -22,7 +22,7 @@
  */
 
 import {MathDocumentConstructor, ContainerList} from '../../core/MathDocument.js';
-import {MathItem, STATE} from '../../core/MathItem.js';
+import {MathItem, STATE, newState} from '../../core/MathItem.js';
 import {HTMLMathItem} from '../../handlers/html/HTMLMathItem.js';
 import {HTMLDocument} from '../../handlers/html/HTMLDocument.js';
 import {HTMLHandler} from '../../handlers/html/HTMLHandler.js';
@@ -101,6 +101,8 @@ export class LazyList<N, T, D> {
 }
 
 /*==========================================================================*/
+
+newState('LAZYALWAYS', STATE.FINDMATH + 3);
 
 /**
  * The attribute to use for the ID on the marker node
@@ -249,16 +251,6 @@ export function LazyMathItemMixin<N, T, D, B extends Constructor<HTMLMathItem<N,
       }
     }
 
-    /**
-     * @override
-     */
-    public state(state: number = undefined, restore: boolean = false) {
-      //
-      // don't set the state if we are lazy processing
-      //
-      return (restore === null ? this._state : super.state(state, restore));
-    }
-
   };
 
 }
@@ -285,9 +277,19 @@ export interface LazyMathDocument<N, T, D> extends HTMLDocument<N, T, D> {
   lazyList: LazyList<N, T, D>;
 
   /**
+   * The containers whose contents should always be typeset
+   */
+  lazyAlwaysContainers: N[];
+
+  /**
    * A function that will typeset all the remaining expressions (e.g., for printing)
    */
   lazyTypesetAll(): Promise<void>;
+
+  /**
+   * Mark the math items that are to be always typeset
+   */
+  lazyAlways(): void;
 
 }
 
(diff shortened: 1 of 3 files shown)


ts/ui/lazy/LazyHandler.ts
```diff
@@ -334,21 +369,155 @@ B extends MathDocumentConstructor<HTMLDocument<N, T, D>>>(
      * Augment the MathItem class used for this MathDocument,
      *   then create the intersection observer and lazy list,
      *   and bind the lazyProcessSet function to this instance
-     *   so it can be used as a callback more easily.
+     *   so it can be used as a callback more easily. Add the
+     *   event listeners to typeset everytihng before printing.
```
zorkow, line 373:
```suggestion
     *   event listeners to typeset everything before printing.
```

ts/ui/lazy/LazyHandler.ts
```diff
@@ -334,21 +369,155 @@ B extends MathDocumentConstructor<HTMLDocument<N, T, D>>>(
...
+      // The state we need to go back to (COMPILED or TYPESET).
+      //
+      let state = STATE.LAST;
+      //
+      // Loop through all the math...
+      //
+      for (const item of this.math) {
+        const math = item as LazyMathItem<N, T, D>;
+        //
+        // If it is not laxy compile or typeset, skip it.
```
zorkow, line 480:
```suggestion
        // If it is not lazy compile or typeset, skip it.
```

ts/ui/lazy/LazyHandler.ts
```diff
@@ -346,9 +355,80 @@ B extends MathDocumentConstructor<HTMLDocument<N, T, D>>>(
...
+      //
+      //  Install print listeners to typeset the rest of the document before printing
+      //
+      if (window) {
+        let done = false;
+        const handler = () => {
+          !done && this.lazyTypesetAll();
+          done = true;
+        };
+        window.matchMedia('print').addListener(handler);  // for Safari
```
zorkow, line 408: That's curious. MDN says Safari knows `beforeprint`. But I am sure you've tested it.

ts/ui/lazy/LazyHandler.ts
```diff
@@ -584,6 +664,21 @@ B extends MathDocumentConstructor<HTMLDocument<N, T, D>>>(
...
 
+    /**
+     * @override
+     */
+    public render() {
+      //
+      // Get the containers whose content should always be typeset
+      //
+      const always = this.options.lazyAlwaysTypeset;
+      this.lazyAlwaysContainers = !always ? null :
```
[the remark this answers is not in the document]
dpvc: The `lazyAlwaysTypeset` option is not a boolean, but either a DOM element, a string used to select a DOM element, or  an array of those.  So it can't be the other way around (if I understand your comment).  Note that in the next line is

```
this.adaptor.getElements(Array.isArray(always) ? always : [always], this.document);
```

so `always` is being used to identify the elements to force to be always typeset (not lazy typeset).

ts/ui/lazy/LazyHandler.ts
dpvc: It was added in Safari v13, which is not all that long ago.  I wrote the code last year, and may have been running 12 at that point.  I can remove the support if you want, but it may help with other browsers that I haven't tested.  Removing it would mean that the previous lines could be simplified (since we wouldn't have to check for duplicate calls), so perhaps that is worth it.

dpvc: I tested that using only `before print` does work in the current Safari.  I suppose the real question then is how much backward-compatibility do we need?  There may be WebKit-based browsers (or tools lie WebView) that are not as up-to-date as Safari itself.

zorkow: Let's keep it then.

zorkow (approved):
lgtm.

PR #778, opened 2022-02-11
Add support for mglyph use of fontfamily/index.  (mathjax/MathJax#2298)

dpvc:
This PR adds support for using the `fontfamily` and `index` attributes on `mglyph`, to bring it more in line with v2.  This does not do the check that the requested character is actually in the specified font, however, which was done in the HTML-CSS output jax of v2, so the `alt` attribute will not be used in this case.

The common glyph output wrapper now creates a TextNode for the character indicated by `index`, and uses that to produce the SVG or CHTML output (and to compute its bounding box).

It now reports an error if the `mglyph` doesn't have a `src` attribute and is also missing either the `fontfamily` or `index` attributes.

Resolves issue mathjax/MathJax#2298.

### DIFF
--- a/ts/output/common/Wrappers/mglyph.ts
+++ b/ts/output/common/Wrappers/mglyph.ts
@@ -23,7 +23,7 @@
 
 import {AnyWrapper, WrapperConstructor, Constructor} from '../Wrapper.js';
 import {CommonTextNode} from './TextNode.js';
-import {TextNode} from '../../../core/MmlTree/MmlNodes/TextNode.js';
+import {TextNode} from '../../../core/MmlTree/MmlNode.js';
 import {BBox} from '../../../util/BBox.js';
 
 /*****************************************************************/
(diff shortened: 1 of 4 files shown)


ts/output/common/Wrappers/mglyph.ts
```diff
@@ -70,6 +77,11 @@ export function CommonMglyphMixin<T extends WrapperConstructor>(Base: T): Mglyph
      */
     public valign: number;
 
+    /**
+     * TextNode used for deprecated fontfamily/index use case
```
zorkow, line 81: That should be an `@override`

zorkow (approved):
One minor thing. O/w lgtm.

ts/output/common/Wrappers/mglyph.ts
dpvc: Is that because of the interface?  If so, shouldn't the other ones above also be `@override`?

dpvc: And the `getParameters()` below?

zorkow: Yes. `@override` ensures that the documentation is inherited from the definition in the superclass/interface. 
I had not spotted the `getParameters`, as the interface method already existed (and I clearly missed it the first time around). Effectively the `@override` can replace the first line of the comment in the method implementation.

zorkow:
Lgtm.

ts/output/common/Wrappers/mglyph.ts
dpvc: OK.  I'm afraid there are lots of files that have this wrong, then.  It would be good to do a documentation sweep before the next major release.

PR #779, opened 2022-02-11
Fix problems with verification and repair of malformed mtables.

dpvc:
This PR produces better results for repairing and reporting issues with malformed mtables.  There are verification options available that are supposed to control the error messages produced for bad MathML, and one controls whether tables should be repaired silently or produce `merror` elements to report the problem.  The `fixMtables` option did not work well, as it did not produce proper messages in some cases, and didn't always produce a repaired table.

The fixes here include:

*  Correct interface for return value of error (it was listed as `void` but returned an `error` node).
*  Make sure that errors for missing `mtr` or `mtd` elements are in a valid `mtable`, and produce the right output (a colored `mtd` when used outside an `mtr`, or a colored `mtr` when used outside a table, and a colored entry when missing the `mtd`.
*  Missing `mtr` and `mtd` elements are inserted to make a valid table regardless of the verification options (those only control the error reporting).

### DIFF
--- a/ts/core/MmlTree/MmlNodes/mtable.ts
+++ b/ts/core/MmlTree/MmlNodes/mtable.ts
@@ -137,10 +137,18 @@ export class MmlMtable extends AbstractMmlNode {
    * @override
    */
   protected verifyChildren(options: PropertyList) {
-    if (!options['fixMtables']) {
-      for (const child of this.childNodes) {
-        if (!child.isKind('mtr')) {
-          this.mError('Children of ' + this.kind + ' must be mtr or mlabeledtr', options);
+    for (const child of this.childNodes) {
+      if (!child.isKind('mtr')) {
+        const isMtd = child.isKind('mtd');
+        const factory = this.factory;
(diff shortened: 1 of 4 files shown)


dpvc:
> I've always been thinking that an alternative to this approach could be to combine all "non-row" elements between rows into a single `mtr`.

That sounds good.  I've pushed commits that does that.  See if that is more to your liking.

ts/core/MmlTree/MmlNodes/mtable.ts
```diff
@@ -137,10 +137,31 @@ export class MmlMtable extends AbstractMmlNode {
...
+    const factory = this.factory;
+    for (let i = 0; i < this.childNodes.length; i++) {
+      const child = this.childNodes[i];
+      if (child.isKind('mtr')) {
+        mtr = null;               // start a new row if there are non-mtr children
+      } else {
+        const isMtd = child.isKind('mtd');
+        //
+        //  If there is already an mtr for previous children, just remove the child
+        //    otherwise repalce the child with a new mtr
```
zorkow, line 150: typo

PR #780, opened 2022-02-11
Add ability for TeX input to force normal variant for CJK input.  (mathjax/MathJax#2744)

dpvc:
This PR adds an optional variant specification for the `RANGES` array that specifies how to handle raw unicode input in the TeX input jax.  This allows the CJK letters to be upright van when placed into an `mi` with one character.

This resolves mathjax/MathJax#2744.


zorkow (approved):
lgtm.

PR #781, opened 2022-02-15
Implement HTML in MathML token nodes.  (mathjax/MathJax#1128)

dpvc:
This PR implements the ability to include HTML within token elements in MathML (as allowed in HTML5).  It does so by implementing an HtmlNode class that holds the HTML tree.  This is inserted into the MathML token node child array whenever a non-MathML node is found.  Thus `<mtext>abc<input type="button" value="Push Me">def</mtext>` is allowed, and would produce an `mtext` element containing a button surrounded by some plain text.

Because the HTML is not currently sanitized (something that could be added to the `safe` extension), allowing HTML in token elements would be a security issue for sites that allow user-entered MathML.  For this reason, the MathML input jax has a new option `allowHtmlInTokenNodes` to control whether to allow it, and it is false by default.

The `makeNode()` method in `MathMLCompile.ts` has been broken into three parts, to make them shorter (it was getting a bit long) by the addition of a `createMml()` method that is used to create the internal MML node when everything has been verified, and an `unknownNode()` method for handling unknown nodes (either by making an `HtmlNode` or throwing an error).

Because token nodes can now contain multiple children (in the past we assumed there was only one text node), the normalization of the spaces in the text must be handled differently, so the `trimSpaces()` method has been broken into two pieces:  one standardized the whitespace to actual spaces (not tabs or newlines) and makes multiple spaces into single ones, and one that trims the initial and final space once all the text nodes are in place.  There are also some changes to the handling of the TextNode in the output wrappers.  In particular, the SVG output had to be modified to add extra `g` elements when there are multiple child nodes for a token node, so that the characters can be positioned properly by the `place()` method.  (In the past, since there was only one text element and its offset was 0 within the token node, it didn't matter, but now it does.)

Resolves issues mathjax/MathJax#1128 and mathjax/MathJax#487.

### DIFF
--- a/ts/input/mathml/MathMLCompile.ts
+++ b/ts/input/mathml/MathMLCompile.ts
@@ -24,6 +24,7 @@
 import {MmlFactory} from '../../core/MmlTree/MmlFactory.js';
 import {MmlNode, TextNode, XMLNode, AbstractMmlNode, AbstractMmlTokenNode, TEXCLASS}
 from '../../core/MmlTree/MmlNode.js';
+import {HtmlNode} from '../../core/MmlTree/MmlNodes/HTMLNode.js';
 import {userOptions, defaultOptions, OptionList} from '../../util/Options.js';
 import * as Entities from '../../util/Entities.js';
 import {DOMAdaptor} from '../../core/DOMAdaptor.js';
@@ -44,6 +45,7 @@ export class MathMLCompile<N, T, D> {
    */
   public static OPTIONS: OptionList = {
     MmlFactory: null,                   // The MmlFactory to use (defaults to a new MmlFactory)
+    allowHtmlInTokenNodes: false,       // True if HTML is allowed in token nodes
     fixMisplacedChildren: true,         // True if we want to use heuristics to try to fix
                                         //   problems with the tree based on HTML not handling
                                         //   self-closing tags properly
@@ -61,11 +63,11 @@ export class MathMLCompile<N, T, D> {
   /**
    *  The instance of the MmlFactory object and
    */
-  protected factory: MmlFactory;
+  public factory: MmlFactory;
   /**
    *  The options (the defaults with the user options merged in)
    */
-  protected options: OptionList;
+  public options: OptionList;
 
   /**
    *  Merge the user options into the defaults, and save them
@@ -125,7 +127,22 @@ export class MathMLCompile<N, T, D> {
         limits = true;
       }
     }
-    this.factory.getNodeClass(type) || this.error('Unknown node type "' + type + '"');
+    if (!this.factory.getNodeClass(type)) {
+      return this.unknownNode(type, node);
+    }
+    return this.createMml(type, node, texClass, limits);
+  }
+
+  /**
+   * Create an actual MmlNode tree from a given DOM node.
+   *
+   * @param {string} type      The type of MmlNode to create
+   * @param {N} node           The original DOM node that is being transcribed
+   * @param {string} texClass  The texClass specified on the node, if any
+   * @param {boolean} limits   True if fixed limits are to be used
+   * @return {MmlNode}         The final MmlNode tree
+   */
+  protected createMml(type: string, node: N, texClass: string, limits: boolean): MmlNode {
     let mml = this.factory.create(type);
     if (type === 'TeXAtom' && texClass === 'OP' && !limits) {
       mml.setProperty('movesupsub', true);
@@ -139,6 +156,21 @@ export class MathMLCompile<N, T, D> {
     this.checkClass(mml, node);
     this.addChildren(mml, node);
     return mml;
+  }
+
+  /**
+   * Handle unknown node by either creating an HTML node, or throwing an error.
+   *
+   * @param {string} type   The type of node being requested
+   * @param {N} node        The HTML node used to create it.
+   * @return {MmlNode}      The HtmlNode holding the node (or null)
+   */
+  protected unknownNode(type: string, node: N): MmlNode {
+    if (this.factory.getNodeClass('html') && this.options.allowHtmlInTokenNodes) {
+      return (this.factory.create('html') as HtmlNode<N>).setHTML(node, this.adaptor);
+    }
+    this.error('Unknown node type "' + type + '"');
+    return null;
   }
 
   /**
@@ -230,7 +262,7 @@ export class MathMLCompile<N, T, D> {
         mml.appendChild((this.factory.create('XML') as XMLNode).setXML(child, adaptor));
       } else {
         let childMml = mml.appendChild(this.makeNode(child));
-        if (childMml.arity === 0 && adaptor.childNodes(child).length) {
+        if (childMml.arity === 0 && adaptor.childNodes(child).length && !childMml.isKind('html')) {
           if (this.options['fixMisplacedChildren']) {
             this.addChildren(mml, child);
           } else {
@@ -240,6 +272,7 @@ export class MathMLCompile<N, T, D> {
         }
       }
     }
+    mml.isToken && this.trimSpace(mml);
   }
 
   /**
@@ -253,7 +286,7 @@ export class MathMLCompile<N, T, D> {
     if ((mml.isToken || mml.getProperty('isChars')) && mml.arity) {
       if (mml.isToken) {
         text = Entities.translate(text);
-        text = this.trimSpace(text);
+        text = this.normalizeSpace(text);
       }
       mml.appendChild((this.factory.create('text') as TextNode).setText(text));
     } else if (text.match(/\S/)) {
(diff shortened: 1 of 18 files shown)


pkra:
How does this work server-side? More generally, can I specify dimensions on the outermost HTML tag when something goes wrong (both client or server)?

pkra:
Thanks for taking the time to respond. FWIW I suspect it will be necessary to bring your-own-dimensions in the browser, too.

dpvc:
> I suspect it will be necessary to bring your-own-dimensions in the browser, too.

Maybe in some cases, but for now it seems to work fine in the browser with the examples that I've tested.

dpvc:
Based on Peter's comments, I implemented a different approach to allowing node applications to specify the dimensions of the embedded HTML.  It allows the dimensions to be used in the browser (optionally) as well as node.

To specify the dimensions, add `data-mjx-hwd="H D W"` where `H`, `D`, and `W` are the height, depth, and width of the HTML.  They can be in any units, but `em` units will work best.

Whether this attribute is used is handled via a new option to the output jax, `htmlHDW`, which can be set to `'auto'` (the default), `'ignore'`, `'use'`, or `'force`'.  When set to `ignore`, the `data-mjx-hdw` attribute is ignored, and the MathJax will try to measure the size of the HTML directly.  This works well in the browser, but not in the liteDOM, jsdom, linkedom, or other non-browser adaptors.  The `force` option means that MathJax will use the `data-mjx-hdw` values and will surround the HTML will additional nodes that force the HTML to have the given dimensions.  This would make the browser and node both have the same representation, not relying on the browser measurements.  Finally, `use` means that MathJax will assume the `data-mjx-hdw` values are correct and will use them without forcing the HTML to have the given dimensions.  Finally, `auto` means has MathJax determine which to use; this will be `ignore` when in the browser and `force` when in node.

Having accurate values for the `data-mjx-hdw` attribute is crucial to the quality of the output.  To that end, I have put together an HTML file that computes the needed values (see below).  The values depend on the surrounding font, so there is a place to enter that, as well.  You can also specify the font in the MathML token element that holds the HTML, as in `<mtext fontfamily="Arial"><div>...</div></mtext>` or `<mo style="font-family: arial"><div>...</div></mo>`.   Otherwise, MathJax will use the surrounding font.  The page below gives you a place to enter the HTML you want to measure and the font to use.  Press the "Compute HDW" and the HTML is shown below together with modified HTML that includes the needed `data-mjx-hdw` attribute.  You can copy that and replace the original HTML will it.

-----

``` html
<!DOCTYPE html>
<html>
<head>
<title>Compute HDW values for HTML in Token nodes</title>
<script>
function GetHDW() {
  const html = document.querySelector('mjx-html');
  const content = html.getBoundingClientRect();
  const baseline = document.querySelector('mjx-baseline').getBoundingClientRect();
  const em = parseFloat(window.getComputedStyle(html).fontSize);
  const h = baseline.top - content.top;
  const d = content.bottom - baseline.top;
  const w = content.right - content.left;
  return [h, d, w].map(x => (x / em).toFixed(3).replace(/\.?0+$/, '') + 'em').join(' ');
}
function ShowHDW() {
  const html = document.querySelector('#html').value;
  const content = document.querySelector('mjx-html');
  content.style.fontFamily = document.querySelector('#family').value;
  content.innerHTML = html
  const output = document.querySelector('#output');
  content.firstChild.setAttribute('data-mjx-hdw', GetHDW());
  output.innerHTML = content.innerHTML.replace(/</g, '&lt;');
}
</script>
<style>
mjx-measure {
  display: inline-block;
  border-left: 2px solid red;
  border-right: 2px solid red;
}
mjx-baseline {
  display: inline-block;
  height: 0;
  width: 0;
}
mjx-html {
  display: inline-block;
}
mjx-line {
  display: inline-block;
  height: 0;
  width: 1em;
  border-top: 1px solid blue;
}
#input {
  display: inline-block;
}
#input textarea {
  margin-bottom: 3px;
}
#input input[type="button"] {
  float: right;
}
</style>
</head>
<body>

<h1>Compute HDW values for HTML in Token nodes</h1>

<p id="input">
<textarea id="html" cols="80" rows="10">
<div>HTML</div>
</textarea><br>
Font family: <input type="text" id="family">
<input type="button" value="Compute HDW" onclick="ShowHDW()">
</p>
<h2>The HTML:</h2>
<p>
<mjx-line></mjx-line><mjx-measure><mjx-baseline></mjx-baseline><mjx-html>
&#xA0;
</mjx-html></mjx-measure><mjx-line></mjx-line>
</p>
<h2>The HTML with the HDW attribute:</h2>
<p id="output">
</p>

</body>
</html>
```

-----

Of course, you can use this as a basis for automating the process using something like puppeteer, if you wish.

This will probably cause merge conflicts with the branch that refactors the DOM adaptors, but that can be dealt with.

pkra:
I forget to share a real world use case: I recently built a custom extension to re-implement a TeX symbol font. For this, we created SVGs for the glyphs which we then push as a data-URI SVG string to end up in the img src / image href. This means currentColor in these SVGs won't work as intended - being able to add the SVG directly should help (and also avoid URI encoding bloat).

ts/input/mathml/MathMLCompile.ts
```diff
@@ -317,16 +350,24 @@ export class MathMLCompile<N, T, D> {
...
-  protected trimSpace(text: string): string {
+  protected normalizeSpace(text: string): string {
     return text.replace(/[\t\n\r]/g, ' ')    // whitespace to spaces
-               .replace(/^ +/, '')           // initial whitespace
-               .replace(/ +$/, '')           // trailing whitespace
                .replace(/  +/g, ' ');        // internal multiple whitespace
   }
 
+  /**
+   * @param {MmlNode} text  The token node whose leacing/trailing spaces should be removed
```
zorkow, line 362: typo

ts/input/mathml/MathMLCompile.ts
```diff
@@ -317,16 +350,24 @@ export class MathMLCompile<N, T, D> {
...
+   */
+  protected trimSpace(mml: MmlNode) {
+    let child = mml.childNodes[0] as TextNode;
```
zorkow, line 365: I thought there can now be multiple child nodes in a token. So why does that still only work on the first `textNode`?

zorkow (approved):
Just one question for clarification. O/w OK.

zorkow (changes requested):
I originally thought that this was a problem with the next PR `html-in-tex` but it already occurs here:
I get compilation errors when running typescript. More precisely, it claims that it can't find certain imports, although they are there (errors below). Could this be a circularity problem? Can you try running `npm run compile` in a clean client?

```
ts/core/MmlTree/MmlNodes/HtmlNode.ts:25:26 - error TS2307: Cannot find module '../../../core/DOMadaptor.js' or its corresponding type declarations.

25 import {DOMAdaptor} from '../../../core/DOMadaptor.js';
                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

ts/core/MmlTree/SerializedMmlVisitor.ts:28:24 - error TS2307: Cannot find module './MmlNodes/HTMLNode.js' or its corresponding type declarations.

28 import {HtmlNode} from './MmlNodes/HTMLNode.js';
                          ~~~~~~~~~~~~~~~~~~~~~~~~

ts/input/mathml/MathMLCompile.ts:27:24 - error TS2307: Cannot find module '../../core/MmlTree/MmlNodes/HTMLNode.js' or its corresponding type declarations.

27 import {HtmlNode} from '../../core/MmlTree/MmlNodes/HTMLNode.js';
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

ts/output/svg/Wrappers/mfenced.ts:28:69 - error TS2307: Cannot find module '../../common/Wrappers/Mfenced.js' or its corresponding type declarations.

28 import {CommonMfenced, CommonMfencedClass, CommonMfencedMixin} from '../../common/Wrappers/Mfenced.js';
```

ts/input/mathml/MathMLCompile.ts
dpvc: The trimming is of the initial spaces (which only occur on the first child) and the trailing spaces (which only occur on the last child, which is handled in lines 368 and 369 below.  So this is exactly what needs to be done to handle multiple child nodes.

dpvc:
> Can you try running `npm run compile` in a clean client?

I'll give it a try.

dpvc:
ARGH!  Capitalization problems (I wish Mac OS was case-sensitive).  I will fix.

zorkow:
> ARGH! Capitalization problems (I wish Mac OS was case-sensitive). I will fix.

Indeed `HTMLNode` vs `HtmlNode`... I did not even see that, I was just happy with it completing on `H...`. Sorry about that.

dpvc:
OK, taken care of.

zorkow (approved):

ts/input/mathml/MathMLCompile.ts
zorkow: Sorry had not seen the answer on the front page: Sounds good. Thanks.

PR #783, opened 2022-02-16
Add support for HTML within TeX

dpvc:
This PR adds a `texhtml` extension for the TeX input jax that allows HTML within TeX input.  This is done by enclosing the HTML in a special tag, `<tex-html>...</tex-html>` that is processed specially by the TeX input jax:  it inserts `\texHTML{...}` around a serialized version of the HTML so that it can be part of a TeX string.  The `\texHTML` macro then turns the HTML into the internal `mtext` node containing an `HtmlNode` that handles the HTML in the internal MathML representation.  So for example:

```
$$3 + <tex-html><input type="text" id="answer" size="10"></tex-html> = 10$$
```

When used in an HTML page, the HTML will already be in the page as actual DOM nodes, but when used with `MathJax.tex2chtml()` or `MathJax.tex2svg`, the DOM elements will not be parsed as HTML, so you should use the `\texHTML` macro directly, as in

``` js
MathJax.tex2svg('3 + \texHTML{<input type="text" id="answer" size="10">} = 10', {display: true});
```

If you used `'3 + <tex-html><input type="text" id="answer" size="10"></tex-html> = 10'` as the string, then the tags would be interpreted as less-than signs followed by their names as variables, and so on.

If you want an expression that will work in both a conversion string and in the document DOM itself, you can use `\texHTML{<tex-html>...</tex-html>}` around the HTML.  In the DOM, the `<tex-html>...</tex-html>` will insert another `\texHTML{...}`, but the `\texHTML` macro will remove that redundant macro, if there is one in its argument string.  Similarly, when used in a string setting, `\texHTML` will remove the outer `<tex-html>` node, if there is one.

So this can be used in both the browser and in node applications, and the same TeX expression can be made to work in both.  As with the HTML-in-token-nodes, you can use `data-mjx-hdw` attributes to specify the size of the HTML in node (or in the browser, when the `htmlHDW` document option is set to `'use'` or `'force'`).

In its usual use case in the browser, the HTML will come from the DOM already, and so MathJax doesn't include HTML sanitization.  Because of this, however, the `texHTML` extension does represent a security risk on sites that allow user content, if they don't sanitize the user input themselves.  For this reason, there is an `allowTexHTML` option for the TeX input jax that must be enabled in order for the `\texHTML` macro to be used.  That way, even if someone used `\require{texhtml}`, they won't be able to use it unless the web page has specifically authorized its use.  (This should also be able to be controlled by the Safe extension, but not every web site is using it that should be, and since this is such a large risk, it seemed best to make the default be disabled, even if the website loads the extension explicitly.)

To make `<tex-html>` work, the `HTMLDomStrings` object was enhanced so that its `includeHtmlTags` option is more general.  It is a little hard to hook that into the TeX extension, since the extension doesn't have access to the output jax (where the dom-string parser is stored) or to the user options before the document is created (where the `includeHtmlTags` option could be set).  So this extension patches the `HTMLDomStrings` class's `OPTIONS` value directly.  That's a bit of a hack, but there doesn't seem to be a better way.

### DIFF
--- a/ts/input/tex/texhtml/TexHtmlConfiguration.ts
+++ b/ts/input/tex/texhtml/TexHtmlConfiguration.ts
@@ -24,7 +24,7 @@
 
 import {Configuration} from '../Configuration.js';
 import TexParser from '../TexParser.js';
-import {CommandMap} from '../SymbolMap.js';
+import {MacroMap} from '../SymbolMap.js';
 import {ParseMethod} from '../Types.js';
 import ParseOptions from '../ParseOptions.js';
 import TexError from '../TexError.js';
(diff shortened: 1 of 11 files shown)


dpvc:
I've set the `texhtml` extension so that `\require` will not load it.  I said that was in the Safe extension above, but it is really in the `require` extension.  That makes the `allowTexHTML` option less critical, since now it requires explicitly adding `texhtml` to the package list (and loading it if not using a `-full` component).  That's probably enough, so maybe `allowTexHTML` can be removed?

pkra:
Will it be possible to include TeX input inside the tex-html element?

dpvc:
> Will it be possible to include TeX input inside the tex-html element?

I knew someone was going to ask about that, but I didn't think would happen before it was even released.  :-)

The current answer is no.  That is a complication I was planning to address later.  There are some technical details that make this somewhat complicated.

Do you have a particular use-case in mind?  It might help to know what situations will be needed to be handled.

pkra:
> Do you have a particular use-case in mind?

Immediate: better control over anchors by inserting them directly (cf. https://github.com/mathjax/MathJax/issues/2620#issuecomment-769084617). 

Possibly others (but nothing short term).

dpvc:
> Immediate: better control over anchors by inserting them directly

Isn't that going to make the result pretty non-semantic?  The contents of the embedded HTML would now be the SVG or CHTML output of MathJax, not the internal MathML that would have produced it, and so what gets passed to SRE would be tag soup, not something it could integrate very well into its own analysis of the outer expression.  I expect SRE will simply punt on the content of HTML embedded in MathML.

Note that any TeX inside the HTML would be its own top-level expression, which would have its own menu, its own independent SRE data, and so on.  Using the current `href` handling seems much less complicated.  In the issue tracker you cite, I gave you code for modifying the anchor after it is created, and although you never responded to my final comment, I thought that I addressed the concerns you voiced there.  If you need more control, I think that approach could be modified to handle that, or there are other approaches that would modify the code that creates the anchor in the first place.  These seem much better integrated than embedding TeX-in-HTML-in-TeX simply for adjusting the anchors.  What you are asking for is really about how the output is produced, not about the input, so changing from a simple MatheML tag with an `href` attribute to an `mtext` containing an `a` containing MathJax output for another expression seems pretty heavy-handed.

pkra:
Apologies - I misunderstood what this PR is doing.

> Isn't that going to make the result pretty non-semantic? [...] I expect SRE will simply punt on the content of HTML embedded in MathML.

I agree that would be pretty bad. I wouldn't have expected it.

> Note that any TeX inside the HTML would be its own top-level expression, which would have its own menu, its own independent SRE data, and so on.

Again I wouldn't have expected that and I see how that would be a problem client-side. It would open an interesting avenue - thinking beyond single expressions - but I think this would be the wrong way to go for that.

Anyway, It turns out I don't have use cases for this PR so please ignore my comments.

dpvc:
> Apologies - I misunderstood what this PR is doing.

No problem.  Can I asked what you had in mind?

pkra:
> No problem. Can I asked what you had in mind?

I suppose my gut reaction would be: a "more generic `\text{}`". 

(I realize that the example above is an input; for interactive elements I would also expect nested math mode and accessibility issues there. But that's just speculation.)

dpvc:
> a "more generic \text{}".

I'm sorry, I know I'm being dense, but I'm really not sure what that means.  What additional features are you thinking of?

> for interactive elements I would also expect nested math mode and accessibility issues there.

The situation for accessibility is certainly something that needs to be more fully explored.  I think SRE will need to ignore the HTML content for now.  The explorer may be able to hand it over to the screen reader for regular processing (he says hopefully).  I do hope to address nested math in the future.

ts/input/tex/texhtml/TexHtmlConfiguration.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import {Configuration} from '../Configuration.js';
+import TexParser from '../TexParser.js';
+import {MacroMap} from '../SymbolMap.js';
+import {ParseMethod} from '../Types.js';
+import ParseOptions from '../ParseOptions.js';
+import TexError from '../TexError.js';
+import {HTMLDocument} from '../../../handlers/html/HTMLDocument.js';
+import {HtmlNode} from '../../../core/MmlTree/MmlNodes/HtmlNode.js';
+import {HTMLDomStrings} from '../../../handlers/HTML/HTMLDomStrings.js';
```
zorkow, line 33:
```suggestion
import {HTMLDomStrings} from '../../../handlers/html/HTMLDomStrings.js';
```

zorkow (changes requested):
* One problem with path names.
* Needs to merge `better-visitors`

But even after merging `better-visitors` I am getting a number of Typescript errors: `'Node' is not generic`. Even with the latest Typescript 4.7.

dpvc:
OK, I've merged in `better-visitors` and fixed up the typing issues that that caused.  I think it is ready, now.

zorkow (approved):
Two points:
* semantic enrichment does currently not work on nodes containing HTML.
* For the documentation: you might want to present the part about `allowTexHtml` at the beginning.

dpvc:
> semantic enrichment does currently not work on nodes containing HTML.

We could potentially mark these in some way to allow SRE to ignore them for now, if that helps.  perhaps put them into a semantics element as an annotation or something.

PR #784, opened 2022-02-16
Refactor usage of all packages to reduce redundant code

dpvc:
This PR refactors some of the component files to reduce redundancy, and to make it clearer which packages are not included in AllPackages.

Here, we make `registerTeX()` more flexible so that it can be used in `all-packages`, and have `tex-full` load `all-packages` since it already does all the work.


zorkow (approved):
lgtm.

PR #785, opened 2022-02-18
Update XSLT to produce better results in mml3 extension

dpvc:
This PR adjusts the XSLT slightly to improve the results of the MML3 extension.  In particular, it makes the lines look better, improves the spacing, does a better vertical alignment (in particular, for `.` and `,`), and makes the long-division sign look better.


zorkow (approved):
I can't really comment on that.

PR #786, opened 2022-02-19
Fix problem where errors during mhchem argument collection are not properly handled.  (mathjax/MathJax#2835)

dpvc:
This PR isolates the call to `mhchemParser` and in order to handle the different error format used by the TeX parser and the mhchem parser, as reported in mathjax/MathJax#2835.

Resolves issue mathjax/MathJax#2835.


zorkow (approved):
lgtm.

PR #787, opened 2022-02-19
Place super- and subscripts properly around \vcenter elements.

dpvc:
This PR fixes a problem where super- and subscript were placed in the position for the UN-centered content, rather than the centered content.


zorkow (approved):
lgtm.

PR #797, opened 2022-04-19
Include CSS to reset border-collapse in CHTML output.  (mathjax/MathJax#2861)

dpvc:
This PR resolves the issue raised in mathjax/MathJax#2861 where attribute settings on a surrounding `<table>` can bleed into MathJax CHTML output.  It adds CSS to prevent that.


zorkow (approved):
lgtm.

PR #798, opened 2022-04-20
Update svg output to properly handle token elements with multiple child nodes. (mathjax/MathJax#2836)

dpvc:
This PR resolves an issue where SVG output of token nodes with more than one child would get the horizontal position of the children incorrect.  This was due to the incorrect handling of the `element` property.  The fix is to wrap the children in a `<g>` tag when there is more than one, and position that, rather than assuming the last child is the only child, as was previously done.

There is also a comment typo fix, and a set in the CHTML output to avoid empty nodes when there are no children in a token node with an explicit font.

Resolves issue mathjax/MathJax#2836.

### DIFF
--- a/ts/output/chtml/Wrappers/TextNode.ts
+++ b/ts/output/chtml/Wrappers/TextNode.ts
@@ -69,6 +69,7 @@ CommonTextNodeMixin<CHTMLConstructor<any, any, any>>(CHTMLWrapper) {
     const adaptor = this.adaptor;
     const variant = this.parent.variant;
     const text = (this.node as TextNode).getText();
+    if (text.length === 0) return;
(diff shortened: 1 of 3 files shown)


dpvc:
PS, this PR was extracted from the `html-in-mml` branch so that it can be included in the 3.2.1 release.  That required rebasing the `html-in-mml` and `html-in-tex` branches, which is good to have done anyway, since `develop` had lots of updates since they were branched from it, so this makes for cleaner PRs for those branches.   (There were some PRs onto those branches that were merged that I've rebased onto `develop` and merged, since they were already approved but didn't actually depend on those branches.)

zorkow (approved):
lgtm.

PR #799, opened 2022-04-21
Handle border and padding CSS in CHTML and SVG output

dpvc:
This PR allows CHTML and SVG output to handle border and padding CSS values properly.  In the past, CHTML would show the borders and padding, but didn't take it into account in its bounding boxes, so things like square roots and stretchy delimiters would not produce the correct sizes, and sub- and super-scripts would be misplaced.  SVG did not handle these properties at all.

We fix the problem by introducing a new `getOuterBBox()` function that gets the bounding box taking borders and padding into account (the original `getBBox()` gets the size of the content without borders or padding, and is needed in that form to handle the internal layout properly).  This new method is called when a layout needs the size of a child node in elements that don't have inferred rows, which never have borders or padding, so most of the files changed here are just to substitute `getOuterBBox()` for `getBBox()` in those cases.

The other main change is in the SVG wrapper class, which adds `handleBorders()` in order to implement the borders, and adds a `dx` property that is used to adjust the horizontal position of the wrapped element when there are borders or padding on the left.  Note that SVG only implements solid, dashed, and dotted border styles, and no rounded corners, just rectangular ones, but you can specify individual sides and have them colored and sized independently.

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -328,7 +328,7 @@ export class CommonWrapper<
     for (const [name, side] of BBox.StyleAdjust) {
       const x = this.styles.get(name);
       if (x) {
-        (obox as any)[side] += this.length2em(this.styles.get(name), 1, obox.rscale);
+        (obox as any)[side] += this.length2em(x, 1, obox.rscale);
       }
     }
     return obox;
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -315,13 +315,33 @@ export class CommonWrapper<
   }
 
   /**
+   * Return the wrapped node's bounding box that includes borders and padding
+   *
+   * @param {boolean} save  Whether to cache the bbox or not (used for stretchy elements)
+   * @return {BBox}  The computed bounding box
+   */
+  public getOuterBBox(save: boolean = true): BBox {
+    const bbox = this.getBBox(save);
+    if (!this.styles) return bbox;
+    const obox = new BBox();
+    Object.assign(obox, bbox);
+    for (const [name, side] of BBox.StyleAdjust) {
+      const x = this.styles.get(name);
+      if (x) {
+        (obox as any)[side] += this.length2em(this.styles.get(name), 1, obox.rscale);
+      }
+    }
+    return obox;
+  }
+
+  /**
    * @param {BBox} bbox           The bounding box to modify (either this.bbox, or an empty one)
    * @param {boolean} recompute   True if we are recomputing due to changes in children that have percentage widths
    */
   protected computeBBox(bbox: BBox, recompute: boolean = false) {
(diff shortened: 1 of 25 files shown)


ts/output/svg/Wrapper.ts
```diff
@@ -210,6 +231,98 @@ CommonWrapper<
...
+    }
+    const f = SVGWrapper.borderFuzz;
+    const bbox = this.getOuterBBox();
+    const [h, d, w] = [bbox.h + f, bbox.d + f, bbox.w + f];
+    const paths: [number, number][][] = [
+      [[-f, h], [w, h], [w - width[1], h - width[0]], [-f + width[3], h - width[0]]],
+      [[w, -d], [w, h], [w - width[1], h - width[0]], [w - width[1], -d + width[2]]],
+      [[-f, -d], [w, -d], [w - width[1], -d + width[2]], [-f + width[3], -d + width[2]]],
+      [[-f, -d], [-f, h], [-f + width[3], h - width[0]], [-f + width[3], -d + width[2]]]
+    ];
```
zorkow, line 258: I find this very confusing. Could you pre-compute the four values (i.e., `w - width[1]`...) and reuse them? That way they would also get a meaningful name.

ts/output/common/Wrapper.ts
```diff
@@ -314,14 +314,34 @@ export class CommonWrapper<
...
+   * @return {BBox}  The computed bounding box
+   */
+  public getOuterBBox(save: boolean = true): BBox {
+    const bbox = this.getBBox(save);
+    if (!this.styles) return bbox;
+    const obox = new BBox();
+    Object.assign(obox, bbox);
+    for (const [name, side] of BBox.StyleAdjust) {
+      const x = this.styles.get(name);
+      if (x) {
```
zorkow, line 330: `x` is never used. Collapse to
```suggestion
      if (this.styles.get(name)) {
```

zorkow (approved):
Just a few minor things. O/w lgtm.

ts/output/svg/Wrapper.ts
```diff
@@ -164,6 +177,13 @@ CommonWrapper<
     if (styles) {
       this.adaptor.setAttribute(this.element, 'style', styles);
     }
+    BBox.StyleAdjust.forEach(([name, , lr]) => {
+      if (lr !== 0) return;
```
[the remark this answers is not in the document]
dpvc: Not in this case, since the data are frequently `null`, and the result of this test would change in that case.

ts/output/common/Wrapper.ts
dpvc: OOPS!, meant to replace the duplicate call in the next line.

ts/output/svg/Wrapper.ts
dpvc: Done.

PR #800, opened 2022-04-26
Sre v4 Integration

zorkow:
Integration of new SRE v4. Depends on SRE v4.0.4.

## Changes to SRE integration

SRE can now be imported directly via the file given in its distribution. To keep import clean `sre.ts` now imports all API methods and exports them for use elsewhere. Consequently modules depending on SRE only have to import `sre.js`. Files with old type handling (`d.ts` files) are removed.

## Changes to the explorer

Most changes are wrt. the new handling of API methods and updating to new locales. The most significant change in logic is that `setupEngine` now returns a promise itself, indicating when locale loading has been complete. Consequently previously awkward constructions with `sreReady` are rewritten in `explorer.ts` but also in `semantic-enrich.ts`.

## Changes to the components

The component integrating SRE is now build similar to other components:

* New `sre` component directory takes care the webpacking. Main file `sre.js` and configuration in `sre_config.js` 
* No more need to distinguish between `sre-node` and `sre-browser`. 
* Affects components depending on sre: `semantic-enrich` and `a11y`
* The PR also contains the new `init.js` component subsuming the `startup.js` calls implemented by @dpvc 

## New component

The PR also includes a new exemplary component `tex-chtml-full-speech` that demonstrates how SRE locales can be included via `json` imports in a MathJax distribution. This ensures that MathJax's internal `MathMaps` map is usually empty, keeping builds small and also allowing our lab still to work as usual.


ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -219,29 +221,34 @@ export class SpeechExplorer extends AbstractKeyExplorer<string> {
...
+        SpeechExplorer.updatePromise = new Promise((res) => {
+          Sre.sreReady()
+            .then(() => Sre.setupEngine({locale: options.locale}))
+            .then(() => {
+              // Important that both are in the same block so speech explorers
+              // are restarted sequentially.
+              this.Speech(this.walker);
+              this.Start();
+            })
+            .then(() => res());
```
dpvc, line 236: I'm not sure this handling of promises does what you intend it to (but I may be misjudging the intent).  If you want the `this.Speech(); this.Start()` calls to be performed in the other that the original `Start()` (and `Update()`)  methods are called, then that is not guaranteed.

To see that, consider the code:

``` js
let promise = Promise.resolve();

const delay = n => new Promise((resolve, reject) => setTimeout(resolve, n*1000));

const test = (n, message) => {
  console.log('adding promise for "' + message + '"');
  promise.then(() => {
    console.log('setting promise for "' + message + '"');
    promise = new Promise((resolve, reject) => {
      console.log('waiting for "' + message + '"');
      delay(n).then(() => {console.log(message)});
    });
  }).catch(err => console.error(err));
}

test(1, 'first call');
test(.5, 'second call');
```

Here, `promise` is your `updatePromise`, `delay()` acts as `sreReady()` (and simply does a delay to imitate the potential delay in that call), and `console.log(message)` represents `this.Speach(...); this.Start();`.  Running this produces:

```
adding promise for "first call"
adding promise for "second call"
setting promise for "first call"
waiting for "first call"
setting promise for "second call"
waiting for "second call"
second call
first call
```

Note that the second call is made before the first call.  To understand this, first note that the two `test()` calls are both made before any promises execute (the two `adding promise...` messages at the top).  That means that the same `updatePromise` is used for both, and so there are two `then()` calls made on that same promise.  When there is an idle time, the first `then()` is run, which changes `updatePromise` (the `setting promiser "first call"` message) and returns and reenters the idle state.  That allows the newly created promise to run (`waiting for "first call"`), which sets up the delay for the function that logs the first call.  Then we enter idle state again, and that allows the second promise that was attached to the original `updatePromise` from `test(.5)` to run (`setting promise for "second call"`), which then replaces the `updatePromise` with another new promise (wiping out the one created by the first `test(1)` call without ever using it, so that promise has no purpose), and again enters idle state, allowing the newly created promise to run (`waiting for "second call"`), setting up the second delay.  These two delays are independent of each other, and so when the shorter one (the second call) is satisfied after .5 seconds, its `console.log()` is performed (`second call`), and finally the other delay is reached and its message reported (`first call`).  So the order of execution of the central code of the promises is reversed. 

If you want to actions of `test(1)` and `test(.5)` to be performed sequentially, regardless of the delays), you need to update `promise` at the time `test()` is called.  Since `promise.then()` itself returns a promise, you can use that to update `promise`, as follows:

``` js
let promise = Promise.resolve();

const delay = n => new Promise((resolve, reject) => setTimeout(resolve, n*1000));

const test = (n, message) => {
  console.log('adding promise for "' + message + '"');
  promise = promise.then(() => {
    console.log('waiting for "' + message + '"');
    return delay(n).then(() => {console.log(message)});
  }).catch(err => console.error(err));
}

test(1, 'first call');
test(.5, 'second call');
```

This simplifies the process and avoids an explicit `new Promise()` that isn't really needed, while guaranteeing that the actions caused by `test()` will be performed sequentially.  Here the output is

```
adding promise for "first call"
adding promise for "second call"
waiting for "first call"
first call
waiting for "second call"
second call
```

Here, the first `test(1)` call updates `promise` (unlike in the original) using `promise.then()` so that `promise` now won't resolve until the `delay()` and `console.log()` are complete (note that the promise from `delay().then()` is returned, and so that is the promise that is used for the new value of `promise`).  Then when `test(.5)` is called, its promise is chained off the one from `test(1)` (unlike the first code above, where the promise from test(1)` is overwritten and never used).  So the second `delay()` is not performed until after the first one is completed.  That is, we don't even start waiting for the second call until after the first call is made.

So if you are trying to use `updatePromise()` to serialize the functions within the promise, then you aren't currently doing that.  I would suggest the following:

``` js
      SpeechExplorer.updatePromise = SpeechExplorer.updatePromise.then(async () => {
        return Sre.sreReady()
          .then(() => Sre.setupEngine({locale: options.locale}))
          .then(() => {
            // Important that both are in the same block so speech explorers
            // are restarted sequentially.
            this.Speech(this.walker);
            this.Start();
          });
      })
        .catch((error: Error) => console.log(error.message));
```

(I can't make a commit suggestion because of the subtracted lin in the middle, and I didn't want to give a partial code block here.)

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -253,31 +260,39 @@ export class SpeechExplorer extends AbstractKeyExplorer<string> {
...
-        options.domain + '-' + options.style;
-    }
+    SpeechExplorer.updatePromise.then(() => {
+      SpeechExplorer.updatePromise = new Promise((res) => {
+        Sre.sreReady()
+          .then(() => Sre.setupEngine({modality: options.modality,
+                                   locale: options.locale}))
+          .then(() => this.region.Update(this.walker.speech()))
+          .then(() => res());
+      });
```
dpvc, line 270: Similar to the discussion about the previous `updatePromise`, I think this needs to become

``` js
    SpeechExplorer.updatePromise = SpeechExplorer.updatePromise.then(async () => {
      return Sre.sreReady()
        .then(() => Sre.setupEngine({modality: options.modality,
                                     locale: options.locale}))
        .then(() => this.region.Update(this.walker.speech()));
    });
```

Note also that the lines below (271 to 279) will be done before the content of this promise is run, so it doesn't matter if those lines are before or after the setting of `updatePromise()`.  I think that is OK, but wanted to check with you to be sure.

components/src/source.js
```diff
@@ -66,8 +66,8 @@ export const source = {
   'a11y/semantic-enrich': `${src}/a11y/semantic-enrich/semantic-enrich.js`,
   'a11y/complexity': `${src}/a11y/complexity/complexity.js`,
   'a11y/explorer': `${src}/a11y/explorer/explorer.js`,
-  '[sre]': (typeof window === 'undefined' ? `${src}/../../js/a11y/sre-node.js` :
-            `${src}/../../node_modules/speech-rule-engine/lib/sre_browser.js`),
+  'a11y/sre': `${src}/a11y/sre/sre.js`,
+  '[sre]': `${src}/../../es5/sre`,
```
dpvc, line 70: This is actually the original location without `source.js`, so there is no reason to set it here.  This line can be removed.

components/src/dependencies.js
```diff
@@ -58,7 +58,7 @@ export const dependencies = {
 export const paths = {
   tex: '[mathjax]/input/tex/extensions',
   mml: '[mathjax]/input/mml/extensions',
-  sre: '[mathjax]/sre/' + (typeof window === 'undefined' ? 'sre-node' : 'sre_browser')
+  sre: '[mathjax]/sre'
```
dpvc, line 61: Since this is only used to get the math maps, I'm wondering if `sre` is the best name to use.  Maybe `mathmaps` or `sre-maps` or something might be better, and it could be set to `[mathjax]/sre/mathmaps`.  This would require corresponding changes to `sre_config.js` above (lines 8 and 9).

dpvc (changes requested):
This looks good, but I think there is an issue with the promise handling in KeyExplorer.  See the discussion in the comments to the code.

ts/a11y/sre.ts
```diff
@@ -16,51 +16,71 @@
...
- */
-const SRE_DELAY = 100;         // in milliseconds
-const SRE_TIMEOUT = 20 * 1000; // 20 seconds
+  export type speechGenerator = SpeechGenerator;
 
-/**
- * A promise that resolves when SRE is loaded and ready, and rejects if
- * SRE can't be loaded, or does not become ready within the timout period.
- */
-export const sreReady = function() {
```
dpvc, line 48: Since v3.2.1 is not supposed to be an update that includes breaking changes, I'm wondering if we shouldn't continue to export an `sreReady()` function that interfaces with the new promise structure of SRE, at least for the time being.  What do you think?

components/src/dependencies.js
zorkow: This is what I effectively have been talking about yesterday. Had not seen your comment then. But would renaming `sre` not be considered a breaking change?

zorkow: The `sre` path now maps to `mathmaps` directory. `sre_config.js` is adapted accordingly.

dpvc: > But would renaming sre not be considered a breaking change?

Technically, so is removing the `sre_browser.js` and `sre-node.js` files that this was used to access.  This value wasn't really used publicly, so it should be OK.  The `[sre]` path was used to obtain the proper version of SRE for browser or node (not the directory itself, but the actual file), and that is no longer necessary, so the change should be a non-issue for users.

components/src/a11y/sre/sre_config.js
```diff
@@ -5,8 +5,8 @@ import {Package} from '../../../../js/components/package.js';
 if (MathJax.startup) {
 
   // Combine path with user defined path and resolve it wrt. MathJax path.
-  combineDefaults(MathJax.config.loader, 'paths', {sre: '[mathjax]/sre'});
-  let path = Package.resolvePath('[sre]', false) + '/mathmaps';
+  combineDefaults(MathJax.config.loader, 'paths', {});
```
dpvc, line 8: This line is not needed.

ts/a11y/sre.ts
```diff
@@ -78,6 +78,18 @@ export namespace Sre {
...
+ * A promise that resolves when SRE is loaded and ready, and rejects if
+ * SRE can't be loaded, or does not become ready within the timout period.
+ *
+ * @deprecated
+ */
+export const sreReady = function() {
+  return new Promise<void>((resolve, reject) =>
+    Sre.sreReady().then(() => resolve())
+      .catch((error: Error) => reject(error.message || error)));
+};
```
dpvc, line 91: Here again, I think this can be simplified.  Since Sre.sreReady() is already a promise, I think all you need is 

```sugestion
export const sreReady = ()  => Sre.sreReady();
```

dpvc (approved):
Two small changes requested, but otherwise looks good.

ts/a11y/sre.ts
zorkow: I had that at first, but then, we don't have a catch anymore. I am happy to do this, but it does change the behaviour a bit.

dpvc: Because the old `sreReady()` used polling to determine when SRE was ready, together with a timeout so that if SRE wasn't ready after a certain time, it would time out and report an error, the `catch()` was to handle that case (I didn't know much about promises then and should have simplified the usage).  The `catch()` in the current `sreReady` ends up essentially throwing the error message, so you can use

```suggestion
export const sreReady = ()  => Sre.sreReady().catch((error: Error) => {throw (error.message || error)});
```

to get the same result.

dpvc: Thought I think it would be fin to leave off the `catch()` and simply let the full error object go through rather than just the message string.

zorkow: OK, I'll leave off the catch.

zorkow: > Here again, I think this can be simplified. Since Sre.sreReady() is already a promise, I think all you need is
> 
> ```
> export const sreReady = ()  => Sre.sreReady();
> ```

Put back my original export line  `export const sreReady = Sre.sreReady;` which is even shorter...

dpvc: Just making sure that `Sre.sreReady` doesn't use `this` internally, as that would no longer work with the definition as is.  I guess `sreReady = Sre.sreReady.bind(Sre)` would do it in that case.

PR #802, opened 2022-05-01
Add \data to html extension

ysulyma:
This PR is a continuation of #791 (which incorrectly committed to the master branch). It adds the `\data` command to the HTML extension to allow adding `data-*` attributes to MathJax content. Documentation is in https://github.com/mathjax/MathJax-docs/pull/299 (`MathJax-docs` doesn't seem to have a `develop` branch).


zorkow (approved):
Thanks for making those changes. This looks good now. 
As this is a new feature, it will need to be in the next feature release, which should be pretty soon.
We will merge as soon as 3.2.1 is out of the way.

PR #804, opened 2022-05-10
Locale cleanup

zorkow:
These are the final adaptations to SRE's new locale handling. In particular,

* We ignore bad locale options provided by local settings. While non-existent locales are handled by SRE falling back to an existing default (e.g., English), currently each explorer retains the incorrect locale. Now the first explorer will correct the options to contain the default locale. 
* This also ensures that there will always be a valid locale selected in the `Accessibility > Speech > Language` menu.
* SRE now has its own mapping for locales from iso to language name. We make use of this now in MJ.

Note, that this currently works only with the next patch release of SRE. I'll do that tonight, and will push one more commit with the correct version number in `package.json`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "esm": "^3.2.25",
         "mhchemparser": "^4.1.0",
         "mj-context-menu": "^0.6.1",
-        "speech-rule-engine": "^4.0.4"
+        "speech-rule-engine": "^4.0.5"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


zorkow:
> Note, that this currently works only with the next patch release of SRE. I'll do that tonight, and will push one more commit 
> with the correct version number in package.json.

That is now done. Please have a look.

dpvc (approved):
looks good.

PR #805, opened 2022-05-11
Allow return value of false to be used in SymbolMap parsers to continue the lookup process.

dpvc:
This PR fixes a problem introduced by #788 that was originally to the html-in-tex branch, and relied on changes in that branch that affect the processing of SymbolMaps.  This PR adds those needed changes so that a parser method that returns false will cause further symbol lookups to occur.  This is needed by the changes for `\operatorname` to work properly.

### DIFF
--- a/ts/input/tex/SymbolMap.ts
+++ b/ts/input/tex/SymbolMap.ts
@@ -70,6 +70,13 @@ export interface SymbolMap {
 
 }
 
+/**
+ * @param {ParseResult} result    The result to check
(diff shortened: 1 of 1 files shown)


pkra:
Thank you for the quick response! I'll test this today.

zorkow (approved):
lgtm.
I leave it up to you if you want to merge now or wait for more of @pkra's feedback.

pkra:
I think I found a slight issue with the latest SRE and this patch (introducing horizontal spacing sometimes).

I'll follow up by email.

PR #806, opened 2022-05-12
Make U+2061 through U+2064 have TeX class NONE so they don't affect spacing.

dpvc:
This PR makes the TeX class for U+2061 through U+2064 be class NONE rather than ORD so that when SRE inserts these, the spacing doesn't change.


pkra:
Thanks!

PR #808, opened 2022-05-13
Remove movablelimits property if the accent base has it set

dpvc:
If the base of an accent comes from `\operatorname`, for example, it will have the `movablelimits` property set to `true`, and so will convert to `msup` in in-line mode, causing incorrect rendering.  This PR makes sure `movable limits` is removed from an accent base to prevent that.


pkra:
Thanks!

PR #809, opened 2022-05-16
Wrap sideset in an TeXAtom of class OP

dpvc:
This PR resolves the issue raised by Peter with `\sideset{}{^*}{\sum}_{a\mid A_y}` where the final underscore should place its argument under the sum rather than in subscript position.  This is done by placing everything into a TeXAtom of class OP so that the limits are placed above and below in displaystyle.

The positioning is not quite the same as in LaTeX, however.  You could get the proper placement using

```
\sideset{}{^*}{\sum}_{a\mid A_y\hphantom{*}}
```

but I am hesitant to add such phantoms automatically as they would have to be removed in textstyle, and would also require special processing of subscript and superscripts that follow the `\sideset`.  That could be done, but would require additional StackItems to handle it.

### DIFF
--- a/ts/input/tex/ams/AmsMethods.ts
+++ b/ts/input/tex/ams/AmsMethods.ts
@@ -324,17 +324,19 @@ AmsMethods.SideSet = function (parser: TexParser, name: string) {
     mml = postScripts;
   }
   //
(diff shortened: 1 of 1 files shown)


zorkow (approved):
I tested and it works.
l

PR #810, opened 2022-05-16
Fix relative width in mpadded, and correct indentation

dpvc:
This PR corrects the width of an `mpadded` element used in the `amscd` extension so that it is an addition to the width rather than the width itself.  This was lost reign the v2 to v3 translation, apparently.  It also fixes some indentation problems in the code.


zorkow (approved):
I actually think that the `+` got lost when we moved from `mu` to `em`. So there could be other little issues of that nature lurking somewhere.

zorkow:
Had a browse through the M2 code. The other (ex)-occurrences of `mu` are fine.

PR #814, opened 2022-05-17
Changing "node-based application" to "node-based applications"

thendotshikota:
Changed "node-based application" to "node-based applications" in line 104


dpvc:
Thanks for the fix!

dpvc:
Thanks for the fix!

PR #815, opened 2022-05-18
Properly set parent for newly created mtd

dpvc:
This PR fixes a problem in `EmpheqUtils` where an `mtd` element is created and inserted into an `mtr` child list without properly setting the parent node.  There was a line that was supposed to do that, but apparently wan't working.  This PR replaces that line with one that sets the parent explicitly (not sure why I didn't do that originally).

It also fixes an issue where nodes could be added to the ParseOptions' `nodeList` for a list with an empty name (when the `in-lists` string is empty, the `split(/,/)` produces an array with an empty entry).

This resolves the issue Peter was having with his `subnumcases` environment.


zorkow (approved):
lgtm.

PR #817, opened 2022-05-20
Add support for color models to \colorbox

pkra:
Part of MathJax/MathJax#2871

### DIFF
--- a/ts/input/tex/color/ColorMethods.ts
+++ b/ts/input/tex/color/ColorMethods.ts
@@ -121,14 +121,13 @@ ColorMethods.DefineColor = function (parser: TexParser, name: string) {
  * @param {string} name The name of the control sequence.
  */
 ColorMethods.ColorBox = function (parser: TexParser, name: string) {
-  const cname = parser.GetArgument(name);
+  const model = parser.GetBrackets(name, "");
+  const cdef = parser.GetArgument(name);
   const math = ParseUtil.internalMath(parser, parser.GetArgument(name));
-  const colorModel: ColorModel = parser.configuration.packageData.get('color').model;
-
-  const node = parser.create('node', 'mpadded', math, {
-    mathbackground: colorModel.getColor('named', cname)
+  const colorModel = parser.configuration.packageData.get("color").model;
+  const node = parser.create("node", "mpadded", math, {
+    mathbackground: colorModel.getColor(model, cdef),
   });
-
   NodeUtil.setProperties(node, padding(parser.options.color.padding));
   parser.Push(node);
 };
--- a/ts/input/tex/color/ColorMethods.ts
+++ b/ts/input/tex/color/ColorMethods.ts
@@ -139,16 +139,17 @@ ColorMethods.ColorBox = function (parser: TexParser, name: string) {
  * @param {string} name The name of the control sequence.
  */
 ColorMethods.FColorBox = function (parser: TexParser, name: string) {
-  const model = parser.GetBrackets(name, "");
+  const fmodel = parser.GetBrackets(name, 'named');
   const fname = parser.GetArgument(name);
+  const cmodel = parser.GetBrackets(name, fmodel);
   const cname = parser.GetArgument(name);
   const math = ParseUtil.internalMath(parser, parser.GetArgument(name));
   const options = parser.options.color;
-  const colorModel = parser.configuration.packageData.get("color").model;
+  const colorModel = parser.configuration.packageData.get('color').model;
 
   const node = parser.create('node', 'mpadded', math, {
-    mathbackground: colorModel.getColor(model, cname),
-    style: `border: ${options.borderWidth} solid ${colorModel.getColor(model, fname)}`
+    mathbackground: colorModel.getColor(cmodel, cname),
+    style: `border: ${options.borderWidth} solid ${colorModel.getColor(fmodel, fname)}`
   });
 
   NodeUtil.setProperties(node, padding(options.padding));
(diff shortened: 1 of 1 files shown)


pkra:
Feel free to close & replace with another.

dpvc:
Would you consider adding support for that in `\fcolorbox` as well?

pkra:
> Would you consider adding support for that in \fcolorbox as well?

Happy to.

I wasn't sure what you had in mind with 

> Also, the `padded()` function should be exported.

on the issue thread.

dpvc:
> I wasn't sure what you had in mind with
> > Also, the padded() function should be exported.
> 
> on the issue thread.

It means add "export" to this line:

https://github.com/mathjax/MathJax-src/blob/9f588aaf7ed8e7f38bfb636157bf29a94ee4804d/ts/input/tex/color/ColorMethods.ts#L40

pkra:
> It means add "export" to this line:

Oh, padding, not padded :see_no_evil:. That makes sense then.

dpvc:
OOPS!, my fault.

pkra:
> OOPS!, my fault.

I feel like I should've been able to figure that one out :rofl: but thank you.

pkra:
`\fcolorbox` supports four patterns:

```
{⟨frame color ⟩}{⟨background color ⟩}{⟨text⟩}
[⟨model-list⟩]{⟨frame spec-list⟩}{⟨background spec-list⟩}{⟨text⟩}
[⟨fr. model-list⟩]{⟨fr. spec-list⟩}[⟨backgr. model-list⟩]{⟨backgr. spec-list⟩}{⟨text⟩}
{⟨frame color ⟩}[⟨background model-list⟩]{⟨background spec-list⟩}{⟨text⟩}
```

Currently, MathJax supports pattern 1.
My commit tries to add support for pattern 2.

Let me know what you think.

ts/input/tex/color/ColorMethods.ts
```diff
@@ -140,15 +139,16 @@ ColorMethods.ColorBox = function (parser: TexParser, name: string) {
  * @param {string} name The name of the control sequence.
  */
 ColorMethods.FColorBox = function (parser: TexParser, name: string) {
+  const model = parser.GetBrackets(name, "");
   const fname = parser.GetArgument(name);
   const cname = parser.GetArgument(name);
```
dpvc, line 144: I think the way to do this is
```suggestion
  const fmodel = parser.GetBrackets(name, 'named');
  const fname = parser.GetArgument(name);
  const cmodel = parser.GetBrackets(name, fmodel);
  const cname = parser.GetArgument(name);
```

and then use `fmodel` to get the color model for the frame and `cmodel` to get the model for the background color.  So you will need `frameModel` in addition to `colorModel` below.  (I haven't tested it, but this should be the right idea.)

ts/input/tex/color/ColorMethods.ts
```diff
@@ -121,14 +121,13 @@ ColorMethods.DefineColor = function (parser: TexParser, name: string) {
  * @param {string} name The name of the control sequence.
  */
 ColorMethods.ColorBox = function (parser: TexParser, name: string) {
-  const cname = parser.GetArgument(name);
+  const model = parser.GetBrackets(name, "");
```
dpvc, line 124: Note that our coding style uses single quotation marks, not double, for string constants (tslint should flag this).  This occurs several times below as well.

dpvc (changes requested):
Thanks for the update.  I've made a few comments about how to handle the `\fcolorbox` situation.

ts/input/tex/color/ColorMethods.ts
pkra: Thanks! I'll try to find time this week.

pkra:
Unexpected slot showed up so I tried to wrap this up.

I noticed that something like `\fcolorbox{127, 127, 127}[RGB]{155, 233, 168}{x}` would create ` style="border: 2px solid 127, 127, 127"` (whereas I'd expect an error of some kind).

I've been wondering what to do about tests.  I tried to spin up the v3 lab in mathjax-dev but failed (first, that eval bug, then systemjs stuff). I could get the scripts to work (without esm) but it's not clear to me what to do with them.

Let me know what you need from me here.

pkra:
Thanks for your comments.

> Yeah, the named color model doesn't do any checking

Ok, I'll ignore that then (as an issue with getColor / getColorByName)

> @zorkow may be able to tell you about how to handle the tests.

Thanks!

pkra:
@dpvc @zorkow was there anything else you need from me on this?

ts/input/tex/color/ColorMethods.ts
```diff
@@ -121,14 +121,13 @@ ColorMethods.DefineColor = function (parser: TexParser, name: string) {
  * @param {string} name The name of the control sequence.
  */
 ColorMethods.ColorBox = function (parser: TexParser, name: string) {
-  const cname = parser.GetArgument(name);
+  const model = parser.GetBrackets(name, '');
```
dpvc, line 124: Does `''` need to be `'named'`, like you do in `FColorBox` below?

dpvc:
I'm good with it, pending the one small change I just noted.  Can you change it from draft to a regular PR?

@zorkow, are you OK with the tests?

ts/input/tex/color/ColorMethods.ts
dpvc: I'm OK either way (probably `''` is better as it is shorter), but would like to see the same treatment both here and at 142, whichever it is.

pkra: Done.

dpvc:
Thanks.  Looks good.

PR #822, opened 2022-06-02
Make sure we get the <math> element from mml3 conversion.  (mathjax/MathJax#2879)

dpvc:
It turns out that the XSLTProcessor produces different results in Firefox and Chrome, and the code that obtained the `<math>` element didn't work in both, leading to an error message about `div` elements.  This PR uses a more robust approach to finding the `<math>` element.  We know here will be one (and only one) because this runs after the initial parsing of the math string, which checks for a single `<math>` element.

Resolves issue mathjax/MathJax#2879.


zorkow (commented):
lgtm.

zorkow (approved):

PR #823, opened 2022-06-02
Fix issue where lazy typesetting would re-typeset expressions unnecessarily.  (mathjax/MathJax#2873)

dpvc:
The changes to the lazy typesetting in v3.2.1 introduced an error where previously typeset expressions were re-typeset when they shouldn't be.  This lead to errors about labels already existing, in addition to the overhead caused by re-typesetting expression that didn't need it (more and more as you scroll).

This was caused by the fact that setting the document's state forced the math elements to be set to that state as well, whereas we only want to change the sate of the elements that have appeared on screen.  This is fixed here by restoring the `state()` method that was removed in 3.2.1.  It uses `restore === null` to determine whether the state is being changed by the lazy loader (and so prevents state changes on MathItems that should not be updated).

We also ignore non-TeX expressions when looking for the earlier TeX expressions that might need to be compiled.

Resolve issue mathjax/MathJax#2873.

### DIFF
--- a/ts/ui/lazy/LazyHandler.ts
+++ b/ts/ui/lazy/LazyHandler.ts
@@ -216,6 +216,15 @@ export function LazyMathItemMixin<N, T, D, B extends Constructor<HTMLMathItem<N,
     }
 
     /**
+     * Only update the state if restore is true or false (null is used for setting lazy states)
+     * @override
+     */
+    public state(state: number = null, restore: boolean = false) {
+      if (restore !== null) super.state(state, restore);
+      return super.state();
+    }
+
+    /**
      * Initially, just insert a marker for where the math will go, and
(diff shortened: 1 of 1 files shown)


zorkow (approved):
The test I've put together, which gave errors on the previous version now works fine. Hence, lgtm.

Would be good if we could automate that test. Not sure how, though.

dpvc:
> Would be good if we could automate that test. Not sure how, though.

Probably need to run it in puppeteer in order to get the IntersectionObserver and associated events.

dpvc:
> I am not sure if this will allow to simulate a scroll event.

I haven't tried it, but it should.  I would expect that `window.scrollTo()` would cause the IntersectionObserver to fire on any newly-revealed nodes.  I believe there is still a virtual viewport even through the window is not being displayed.

Issue #818, opened 2022-05-20
For whatever reason eval being executed in production mode


solomax:
TL;DR;

for whatever reason this https://github.com/mathjax/MathJax-src/blob/master/ts/components/version.ts#L37 code being executed at my app :(

Long version:
We are using MathJax to draw beautiful formulas at Apache OpenMeetings: https://github.com/apache/openmeetings/blob/master/openmeetings-web/src/main/front/wb/package.json#L22
And `browserify` as build tool

We are using some sort node-like configuration https://github.com/apache/openmeetings/blob/master/openmeetings-web/src/main/front/wb/src/wb-tool-stat-math.js#L4

Everything was working as expected (and still working in DEV mode)
BUT for whatever reason in PRODUCTION `eval` is being called to determine the MathJax version :(

I've hacked this by adding `window.PACKAGE_VERSION = '3.2.1';` to my source code
But this a bit dirty :(

Maybe it would be possible to introduce some eval-free mode?
Or maybe I'm doing something wrong?
Or maybe it would be possible to store version in some pre/post build step? :)))

Thanks in advance for the help :)

dpvc:
MathJax uses the webpack DefinePlugin to replace `PACKAGE_VERSION` by the actual version:

https://github.com/mathjax/MathJax-src/blob/9f588aaf7ed8e7f38bfb636157bf29a94ee4804d/components/webpack.common.js#L49-L58

(so web components will have the value) and tries to look it up at run time otherwise (for node applications).  So if you are using some other package manage, you will need to do something similar.  It looks like maybe the [browserify-replace](https://www.npmjs.com/package/browserify-replace) package might be able to do the job for you.

Sorry that this update has had this negative impact for you.

solomax:
Thanks!
was able to properly work-around this :)

ryangr-texthelp:
Just to chip in here, we have also run into this issue.

Unfortunately, for us, we are using `mathjax-full` as a `dependency` for our library so we are not bundling it directly. Consumers of our library may use a bundler, but it would be difficult for us to try and enforce this as part of a consumer build process.

I understand the need for `PACKAGE_VERSION` to be defined, and it's great that it happens as part of the build process, but equally it can be hardcoded in source and overwritten at runtime if required.

If we go down this route, we're going to have to have separate instructions for Rollup, Webpack, Browserify, and so on.... simply for reading the version number.

ShaMan123:
[...]

PR #824, opened 2022-06-02
Make version.ts use a constant and create the file during the build process.  (#818)

dpvc:
This PR resolves the build issue that was reported with v3.2.1 due to the dynamic lookup of the version number by creating the `version.ts` file via `components/bin/version`, which is run during the travis-ci publish process (and can be run by hand at any point).

Resolve issue #818

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ node_js:
 - 17
 sudo: false
 script:
+- components/bin/version
 - npm install
 - npm run compile
 - npm run make-components
--- a/components/bin/version
+++ b/components/bin/version
@@ -1,0 +1,60 @@
+#! /usr/bin/env node
+
+/*************************************************************
+ *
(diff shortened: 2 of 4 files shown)


zorkow (approved):
lgtm. 

Since I've just seen the line: Do we know if travis works with stable again?

dpvc:
> Do we know if travis works with stable again?

Last time I found out by the fact that the publish run failed, and I tracked down the error message.  Could always try that again.  I will see if I can find something indicating the current state of affairs.

dpvc:
The default OS used by Travis-CI is ubuntu 16.04, unless you request otherwise.  It looks like node.js v18 changed the library version they use, and that ubuntu 16.04 doesn't have it, nor does 18.04; you need 18.10 or 20.04 to allow it to work.  Travis-CI does have a 20.04 option, so I will switch to that and see if that allows the use of the stable version of node.js again.

PR #825, opened 2022-06-09
Better types

dpvc:
This PR improves the type checking of the MmlNodes and Wrappers by adding more generics to reduce the use of `any` types.  This allows the removal of some `as MmlNode` and other casting of types, but means there are more generics to specify for many classes.

To make all this work, the mix-ins needed to be updated.  This involves adding interfaces for both the mix-in instance and classes that weren't used before (or were only used in a rudimentary way).  The complicated part was that the mix-in base class is not allowed to take generics from the class that is extending it, so the mechanism for handling the mix-in had to be adjusted to work around that.  This is done by having the SVG and CHTML classes us a function that take generics and that function creates a class that extends the mix-in using he generics from the function (not from the extended class).  

For example, the CHTML wrapper for the MmlMi is created via:

``` js
/**
 * The ChtmlMi wrapper class for the MmlMi class
 */
export const ChtmlMi = (function <N, T, D>(): ChtmlMiClass<N, T, D> {

  const Base = CommonMiMixin<
      N, T, D,
      CHTML<N, T, D>, ChtmlWrapper<N, T, D>, ChtmlWrapperFactory<N, T, D>, ChtmlWrapperClass<N, T, D>,
      ChtmlCharOptions, ChtmlVariantData, ChtmlDelimiterData, ChtmlFontData, ChtmlFontDataClass,
      ChtmlMiClass<N, T, D>
    >(ChtmlWrapper);

  return class ChtmlMi extends Base implements ChtmlMiNTD<N, T, D> {

    /**
     * @override
     */
    public static kind = MmlMi.prototype.kind;

  };

})<any, any, any>();
```

Here, we use `function <N, T, D>()` to make a function that depends on three generics (the DOM Node, Text, and Document classes), and then get the base mix-in class using those generics.  Then we create the `ChtmlMi` class from that base and return it.  We use the generics `<any, any, any>` for `<N, T, D>`, but within the definition of `ChtmlMi`, the generic types are honored; that is, the checking that something that is supposed to be `N` is really `N` and not `T`, and that these are as expected in the use of `N`, `T`, and `D` within the mix-in base class.  This typechecking was not being done before.

In addition, we change the `ts/output/common/OutputJax.js` to just `ts/output/common.js` for consistency with `ts/output/chtml.js` and `ts/output/svg.js`, and use a common `dom` property rather than the jax-specific `chtml` and `element` properties, again for consistency among the output jax.

Finally, the names of the classes and interfaces are put into camel-case for better consistency of naming.

This looks like a big update, and it is, but many of the changes follow a common pattern (adding the interfaces and associated generics, and updating the mix-in usage), are to remove casting that is no longer needed, or just change the names of things (camel-case or move to `dom` name).  This is mostly just about getting the types right, and the code itself is changed very little other than that.


ts/core/Tree/Visitor.ts
```diff
@@ -27,50 +27,51 @@ import {NodeFactory} from './NodeFactory.js';
 /**
  * The type for the functions associated with each node class
  */
-export type VisitorFunction = (visitor: NodeFactory<Node, NodeClass>, node: Node, ...args: any[]) => any;
+export type VisitorFunction<N extends Node<N, C>, C extends NodeClass<N, C>> =
```
zorkow, line 30: Some extra comments.

ts/core/Tree/Visitor.ts
```diff
@@ -27,50 +27,51 @@ import {NodeFactory} from './NodeFactory.js';
...
+  (visitor: NodeFactory<N, C>, node: N, ...args: any[]) => any;
 
 /*****************************************************************/
 /**
  *  Implements the Visitor interface
  */
 
-export interface Visitor {
+export interface Visitor<N extends Node<N, C>, C extends NodeClass<N, C>> {
```
zorkow, line 38: Ditto.

ts/output/chtml.ts
```diff
@@ -47,7 +48,11 @@ import {unicodeChars} from '../util/string.js';
  * @template D  The Document class
  */
 export class CHTML<N, T, D> extends
-CommonOutputJax<N, T, D, CHTMLWrapper<N, T, D>, CHTMLWrapperFactory<N, T, D>, CHTMLFontData, typeof CHTMLFontData> {
+CommonOutputJax<
+  N, T, D,
+  ChtmlWrapper<N, T, D>, ChtmlWrapperFactory<N, T, D>, ChtmlWrapperClass<N, T, D>,
+  ChtmlCharOptions, ChtmlVariantData, ChtmlDelimiterData, ChtmlFontData, ChtmlFontDataClass
+> {
```
zorkow, line 55: I think it is worth explaining those in more detail, i.e., not just the template types.

ts/output/chtml/Wrapper.ts
```diff
@@ -59,20 +60,22 @@ export const SPACE: StringMap = {
...
+ *  The type of the ChtmlWrapper class (used when creating the wrapper factory for this class)
  */
-export interface CHTMLWrapperClass extends AnyWrapperClass {
-
-  kind: string;
+export interface ChtmlWrapperClass<N, T, D> extends CommonWrapperClass<
+  N, T, D,
+  CHTML<N, T, D>, ChtmlWrapper<N, T, D>, ChtmlWrapperFactory<N, T, D>, ChtmlWrapperClass<N, T, D>,
+  ChtmlCharOptions, ChtmlVariantData, ChtmlDelimiterData, ChtmlFontData, ChtmlFontDataClass
+> {
```
zorkow, line 78: Add comments for those.

ts/output/chtml/WrapperFactory.ts
```diff
@@ -16,43 +16,36 @@
...
 CommonWrapperFactory<
-  CHTML<N, T, D>,
-  CHTMLWrapper<N, T, D>,
-  CHTMLWrapperClass,
-  CHTMLCharOptions,
-  CHTMLDelimiterData,
-  CHTMLFontData
+  N, T, D,
+  CHTML<N, T, D>, ChtmlWrapper<N, T, D>, ChtmlWrapperFactory<N, T, D>, ChtmlWrapperClass<N, T, D>,
+  ChtmlCharOptions, ChtmlVariantData, ChtmlDelimiterData, ChtmlFontData, ChtmlFontDataClass
```
zorkow, line 43: Same as for `WrapperClass`. Might be worth copying the comment?

zorkow (approved):
Disclaimer: I did not go through all files in fine detail. I got as far as `chtml` and its `Wrappers` and skim read the rest.
I only have one question about a casting. 
Otherwise, I think it is very good cleanup. I like that we can get rid of a lot of casting. 
However, I am a bit worried about the "mountain" of new generics and extensions in some of the interfaces. I think they need some additional commenting. I made some comments where I believe some extra comments should go. They could probably be centralised. I.e., if the `Wrapper` and `WrapperFactory` interfaces are well documented then there is no need to do the same in each of the actual wrapper node classes.
Please have a look.

zorkow:
PS: I also think it is worthwhile keeping your explanations for the PR around somewhere. Either in the code or better yet in our documentation somewhere.

ts/core/MmlTree/MmlNode.ts
```diff
@@ -808,9 +813,9 @@ export abstract class AbstractMmlNode extends AbstractNode implements MmlNode {
     merror.attributes.set('data-mjx-message', message);
     if (options['fullErrors'] || short) {
       let mtext = this.factory.create('mtext');
-      let text = this.factory.create('text') as TextNode;
+      let text = this.factory.create('text') as any as TextNode;
       text.setText(options['fullErrors'] ? message : this.kind);
-      mtext.appendChild(text);
+      mtext.appendChild(text as any);
```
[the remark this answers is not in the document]
dpvc: I think this may have been left over from an earlier stage before the cast to TextNode two lines earlier.  I will remove it.

dpvc:
> I think they need some additional commenting. I made some comments where I believe some extra comments should go.

OK, I've made the changes your requested.  See if the new comments are what you were looking for.

> I also think it is worthwhile keeping your explanations for the PR around somewhere

OK, I think it can be in the documentation somewhere.  We need a lot more of that, I'm afraid.

PR #826, opened 2022-06-09
Better visitors

dpvc:
This PR abstracts the visitor classes to allow visitors on arbitrary node classes, and adds a `walkTree()` method to the abstract wrapper class, so it is always available.

(The original impetus for this PR was to make it possible to use the visitor class to define a visitor on the wrapped nodes rather than only on the internal MathML nodes.  This is used in the line breaking code to make visitor to determine automatic line break positions.)


zorkow (approved):
lgtm.

PR #830, opened 2022-06-18
Use handleRetries() to handle issue with rerender after explorer is enabled

dpvc:
This PR resolves an issue where enabling the explorer for the first time causes an error (due to the dynamic loading of maps?).  The retry caused by SRE needs to be trapped by `handleRetriesFor()`, which is what this PR does.


zorkow (approved):
lgtm.

PR #831, opened 2022-06-18
Make sure a11y.speechRules in initialized

dpvc:
This PR fixes a problem where the explorer domain and style menus are not initialized when first activated.  The problem is that the `a11y.speechRules` option, which is a compound of the `sre.domain` and `sre.style` options, is not initialized.  The fix is to initialize it during the creation of the explorer document based on the `sre` options.


zorkow (approved):
lgtm.

PR #834, opened 2022-06-27
Handle labels nested in aligned environments (mathjax/MathJax#2889)

dpvc:
This PR allows labels nested within `aligned`, `split`, and `gather` environments to apply to the outer equation.  It does so by bubbling the `label` property up to the outer `TagInfo` object in the `end()` method, if there isn't already a label in that one (and carrying a label down into the new `TagInfo` in the `start()` method), and clearing the label when a tag is assigned to it.

This resolves issue mathjax/MathJax#2889.


zorkow (approved):
lgtm.

PR #835, opened 2022-07-05
Don't typeset mathml inserted by the AssistiveMML extension. (mathjax/MathJax#2892)

dpvc:
This PR prevents MathML that is the result of the AssistiveMML extension from being processed by the MathML input jax (in the event that the page is retypeset).  It also switches some `let` commands to `const`, and uses `set.values()` rather than `Array.from(set)` for efficiency.

Resolves issue mathjax/MathJax#2892.


zorkow (approved):
lgtm.
Hopefully we can get rid of that soon.

PR #837, opened 2022-07-13
Make sure the scriptspace CSS is included for any script class

dpvc:
With CHTML's dynamic CSS handling, the CSS needed for properly handling the `scriptspace` font parameter was only being output when `msubsup` or `mmultiscripts` was used.  This PR make sure the CSS is present for any script-based class (`msub`, `msup`, `munderover`, etc.).


zorkow (commented):
lgtm.

PR #838, opened 2022-07-19
Fix handling of template macros with no actual arguments. (mathjax/MathJax#2901)

dpvc:
If a macro definition has a template, but requires no arguments (e.g., `\def\a\b{...}`) then the template is not checked and not properly removed.  This PR makes sure the template matches its usage, and removes it.

Resolves issue mathjax/MathJax#2901.


zorkow (approved):
lgtm.

PR #841, opened 2022-08-08
Properly locate speech string when not in first child element. (mathjax/MathJax#2911)

dpvc:
Fix problem with semantic-enrich's `getSpeech()` method.  It currently only checks the first child node (recursively to the bottom of the tree), as the termination condition will end the loop when the child node returns an empty string, so the second and subsequent children are never checked.

Resolves issue mathjax/MathJax#2911.


zorkow (approved):
lgtm.

PR #832, opened 2022-06-27
Update handling of brackets in algined environments to current LaTeX behavior (mathjax/MathJax#2888)

dpvc:
This PR updates the aligned and alignedat environments to handle the optional bracket argument consistent with current versions of LaTeX, where no spaces are allowed before the bracket argument.

Resolves issue mathjax/MathJax#2888.

PR #842, opened 2022-08-09
Handle bracket arguments for align environments more like LaTeX.

dpvc:
This PR corrects #832 for mathjax/MathJax#2888 so that the behavior is more consistent with LaTeX.  It's not that the bracket is only scanned if it is immediately after the environment name (as in #832), but rather than it is only removed and used for alignment when it is exactly one of `[t]`, `[b]`, or `[c]`.  Otherwise, it is retained and processed normally.

The fix is handled in `setArrayAlign()` by adding `parser` and `i` arguments that, if present, will cause `parser.i` to be set to `i` if the bracket is not one of the allowed values.

There is also a typo (in two places) in an error message that is corrected here.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -426,17 +426,23 @@ namespace ParseUtil {
    * Sets alignment in array definitions.
    * @param {ArrayItem} array The array item.
    * @param {string} align The alignment string.
+   * @param {TexParser?} parser The current tex parser.
+   * @param {number?} i The position to return to if the alignment isn't t, b, or c.
    * @return {ArrayItem} The altered array item.
    */
-  export function setArrayAlign(array: ArrayItem, align: string): ArrayItem {
+  export function setArrayAlign(array: ArrayItem, align: string, parser?: TexParser, i?: number): ArrayItem {
     // @test Array1, Array2, Array Test
-    align = ParseUtil.trimSpaces(align || '');
+    if (!parser) {
+      align = ParseUtil.trimSpaces(align || '');
+    }
     if (align === 't') {
       array.arraydef.align = 'baseline 1';
     } else if (align === 'b') {
       array.arraydef.align = 'baseline -1';
     } else if (align === 'c') {
       array.arraydef.align = 'axis';
+    } else if (parser) {
+      parser.i = i;
     } else if (align) {
       array.arraydef.align = align;
     } // FIXME: should be an error?
--- a/ts/input/tex/ams/AmsMethods.ts
+++ b/ts/input/tex/ams/AmsMethods.ts
@@ -61,9 +61,10 @@ AmsMethods.AmsEqnArray = function(parser: TexParser, begin: StackItem,
                                       align: string, spacing: string,
                                       style: string) {
   // @test Aligned, Gathered
-  const args = (parser.getCodePoint() === '[' ? parser.GetBrackets('\\begin{' + begin.getName() + '}') : '');
+  const i = parser.i;
+  const args = parser.GetBrackets('\\begin{' + begin.getName() + '}');
   const array = BaseMethods.EqnArray(parser, begin, numbered, taggable, align, spacing, style);
-  return ParseUtil.setArrayAlign(array as ArrayItem, args);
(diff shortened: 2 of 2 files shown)


dpvc:
There are a bunch of edge-case strangenesses with the LaTeX implementation, and this is one.  The `[l]` is not a valid alignment, and so it, and everything after it, should become part of the actual cell, which means (to me) that the required argument for the `alignedat` environment is missing.  That means I would expect an error about that argument being missing.

Apparently, however, LaTeX reads the bracket value and the argument before processing the brackets, and then later if the bracket is not `t`, `b`, or `c`, it reinserts it at the beginning of the cell.  That means `\begin{alignedat}[x]{2}...` will end up being treated as `\begin{alignedat}{2} [x]...`, which is counter-intuitive to me.  Is that the output you would expect?

While it would be possible to duplicate this edge case, is that really the right thing to do?  We have not always implemented the behavior for incorrect input.

pkra:
Ah, this whole thing is annoying. The AMS naturally uses fixed versions of everything in production and it's sometimes hard to reproduce new LaTeX / XeLaTeX behavior. 

To skip ahead:

> While it would be possible to duplicate this edge case, is that really the right thing to do? We have not always implemented the behavior for incorrect input.

I think it's fine that we have to fix any issues due to authoring mistakes. So it's fine as is and we'll fix this one in the source.

Just for completeness:

> Apparently, however, LaTeX reads the bracket value and the argument before processing the brackets, and then later if the bracket is not `t`, `b`, or `c`, it reinserts it at the beginning of the cell. That means `\begin{alignedat}[x]{2}...` will end up being treated as `\begin{alignedat}{2} [x]...`, which is counter-intuitive to me. Is that the output you would expect?

I don't quite follow this one. For me, LaTeX `\begin{alignedat}[x]{2}...` won't have the `[x]` show up in the PDF output while `\begin{alignedat}{2} [x]...` would have it show up. 

But again, the AMS is fine with things as is.

pkra:
In any case, thank you for taking another look!

dpvc:
> Ah, this whole thing is annoying.

Sorry, didn't mean to annoy you with it.  Just trying to make the result be more consistent with current versions of LaTeX.

> I don't quite follow this one. For me, LaTeX `\begin{alignedat}[x]{2}...` won't have the `[x]` show up in the PDF output 

I think this is the new behavior after the recent changes in LaTeX (around 2020).  For me,

``` latex
\begin{equation}
  \begin{alignedat}
    [x]{2} a &=b & c &= d
  \end{alignedat}
\end{equation}
```

produces

brakets

The same for

``` latex
\begin{equation}
  \begin{alignedat}[x]{2} 
    a &=b & c &= d
  \end{alignedat}
\end{equation}
```

> the AMS is fine with things as is.

OK, great.  Thanks again for the thorough testing.

pkra:
Thanks for the example. I don't see this on, e.g., overleaf (and not on my machine where tlmgr claims I'm up to date). 

But that just prove the point, I'd say :grin: 

Thanks again for taking the time for the additional comment!

dpvc:
> I don't see this on, e.g., overleaf

It may depend on the document class or packages loaded (macros and environments can change with these, especially the processing of edge cases).

Here is an example that works in Overleaf as I indicated above, using `amsart` environment.

``` latex
\documentclass{amsart}

\title{test}
\author{Davide Cervone}
\date{August 2022}

\begin{document}

\begin{equation}
  \begin{alignedat}
    [x]{2} a &= b & c &= d
  \end{alignedat}
\end{equation}

\end{document}
```

I have been using `amsart` as the source of "default" behavior for MathJax.

zorkow (approved):
lgtm.

pkra:
> Here is an example that works in Overleaf as I indicated above, using amsart environment.

Wow. So with my "usual" test file, I still don't see the behavior. 

But if I create a fresh project, I do see the behavior you describe.

That's just lovely...

Thanks again for the follow up.

dpvc:
> That's just lovely...

Pretty peculiar.  I wonder if you created the original test project before 2020 and it is retaining the version of LaTeX and packages involved at that time.  That's the only thing that I can think of that would account for that.

dpvc:
I suppose the LaTeX log file would show the versions used...

PR #844, opened 2022-08-16
Handle missing and invalid brackets in alignedat like LaTeX does.

dpvc:
This PR fixes a problem reported by Peter where `alignedat` would typeset its number argument when there is no bracket argument.  It also handles invalid bracket arguments like LaTeX does (the number argument is parsed after the brackets, then the brackets are typeset).  These issues were introduced by #842, which would reset the `parser.i` value to before the brackets when they were empty to invalid.  Now we re-insert the bracket argument if it is non-empty and invalid (like LaTeX does).

After #842,

``` latex
\begin{alignedat}{2} a \end{alignedat}
```

would typeset as `2a` when it should just be `a`.  This PR should resolve that.

Also,

``` latex
\begin{alignedat}[x]{2} a \end{alignedat}
```

would produce `[x]2a`, but now should produce `[x]a` as LaTeX does.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -427,10 +427,9 @@ namespace ParseUtil {
    * @param {ArrayItem} array The array item.
    * @param {string} align The alignment string.
    * @param {TexParser?} parser The current tex parser.
-   * @param {number?} i The position to return to if the alignment isn't t, b, or c.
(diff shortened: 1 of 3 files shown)


dpvc:
I've added a commit that fixes similar problems with the `multlined` environment in the `mathtools` package.

zorkow (approved):
lgtm.

PR #845, opened 2022-08-17
fix: update STIX2 type imports to new names

pkra:


pkra:
Not sure if you're interested but since I did it in our production branch with your help, I thought I should at least open it.

dpvc:
I've merged it, but the stix2 branch is not going to be merged into develop or master.  It was there to help as we developed the alternate font support, but the other fonts are going to be stored in a separate repository and as separate npm packages.  Some will be merged into the `mathjax` npm package so they will be part of the CDN distribution.  Not all the details are worked out, but in any case, this brach was a convenience one for you as we were developing STIX2 support.

pkra:
Thanks for the clarification.

PR #846, opened 2022-08-22
Make sure stretchy embellished operators use the proper scaling when determineing their widths

dpvc:
This PR fixes a problem reported by Peter where some horizontally stretched characters are stretched to the wrong length.  For example:

``` latex
\bbox[red]{\begin{smallmatrix}
 - \otimes _k tk \\
  \xleftarrow {\hspace{50pt}}\\
  \xleftarrow {\hspace{50pt}}\\
  \xleftarrow {\hspace{50pt}}\\
  \operatorname {Hom}_k(tk,-)
\end{smallmatrix}}
```

in SVG output.  An incorrect relative scaling value was being used (with an embellished operator, the relative scaling of the core has to include all the nesting parent elements).

### DIFF
--- a/ts/output/common/Wrappers/mrow.ts
+++ b/ts/output/common/Wrappers/mrow.ts
@@ -181,7 +181,8 @@ export function CommonMrowMixin<
         //  Stretch the stretchable children
         //
         for (const child of stretchy) {
-          child.coreMO().getStretchedVariant([H, D]);
+          const rscale = child.coreRScale();
+          child.coreMO().getStretchedVariant([H / rscale, D / rscale]);
         }
       }
     }
--- a/ts/output/common/Wrappers/mtable.ts
+++ b/ts/output/common/Wrappers/mtable.ts
@@ -681,7 +681,7 @@ export function CommonMtableMixin<
         //  Stretch the stretchable children
         //
         for (const child of stretchy) {
-          child.coreMO().getStretchedVariant([W]);
+          child.coreMO().getStretchedVariant([Math.max(W, child.getBBox().w) / child.coreRScale()]);
         }
       }
     }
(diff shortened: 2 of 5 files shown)


ts/output/common/Wrappers/mrow.ts
```diff
@@ -181,7 +181,8 @@ export function CommonMrowMixin<
         //  Stretch the stretchable children
         //
         for (const child of stretchy) {
-          child.coreMO().getStretchedVariant([H, D]);
+          const rscale = child.coreRScale();
+          child.coreMO().getStretchedVariant([H / rscale, D / rscale]);
```
zorkow, line 185: Compute a constant before the loop rather than performing the division every time.

ts/output/common/Wrappers/mtable.ts
```diff
@@ -681,7 +681,7 @@ export function CommonMtableMixin<
         //  Stretch the stretchable children
         //
         for (const child of stretchy) {
-          child.coreMO().getStretchedVariant([W]);
+          child.coreMO().getStretchedVariant([Math.max(W, child.getBBox().w) / child.coreRScale()]);
```
zorkow, line 684: ditto

zorkow (approved):
Minor complaint about too many divisions.
Otherwise, lgtm.

ts/output/common/Wrappers/mrow.ts
dpvc: The division is by a potentially different value of `rscale` for each child, so can't be computed ahead of time.

ts/output/common/Wrappers/mtable.ts
dpvc: Same here.  The `child.coreRScale()` can be different for each child.  This finds the relative scaling of the core `mo` in each child, and since some children might be embellished operators with several changes of relative scaling on the way, and others might not, the scaling value has to be computed for each child separately.

PR #847, opened 2022-08-22
Updates to support alternate fonts in CHTML

dpvc:
This PR finalizes the updates needed for additional font support by updating the CHTML output to handle the alternate fonts being prepared in the fonts repository.

* Provide for updating the CSS when a new font is loaded
* Creates an `addFontIds()` function that inserts the needed font ID data into a font definition (to save hading the same `f` option in each of the character's data).
* Updates the creation of CSS for the stretchy characters to include the font explicitly, when needed
* Assumes the ZERO font will have a space that is one em wide (making it easier to handle in-line breakpoints for linebreaking).  This affects both SVG and CHTML output, and it will make it not work with the current MathJax_Zero font, whose space is .25 em, so the spacing will be too small when used with that font.  (It will be replaced in the next release, as the while MathJax TeX font set will be replaced by a new one.)
* Adds support for CHTML font extensions (making sure the CSS is properly updated).
* Removes `defaultVariantClasses` as they are never actually used (no CSS is created for these classes)
* Fix an issue with the `build` command so that file names with no alphanumeric characters are handled properly (the new additional fonts have such names).

### DIFF
--- a/ts/output/chtml/DynamicFonts.ts
+++ b/ts/output/chtml/DynamicFonts.ts
@@ -26,7 +26,7 @@ import {ChtmlCharMap, ChtmlCharData} from './FontData.js';
 /**
  * Used by dynamic font extensions to add font identifiers to ranges of characters in given variants
  */
-export function AddFontIds(ranges: {[variant: string]: {[id: string]: ChtmlCharMap}}) {
+export function AddFontIds(ranges: {[variant: string]: {[id: string]: ChtmlCharMap}}, prefix?: string) {
   const data: {[variant: string]: ChtmlCharMap} = {};
   for (const id of Object.keys(ranges)) {
     const map = ranges[id];
@@ -41,7 +41,11 @@ export function AddFontIds(ranges: {[variant: string]: {[id: string]: ChtmlCharM
           if (!data[3]) {
             data[3] = {};
           }
-          data[3].f = id;
+          if (prefix) {
+            data[3].F = prefix + '-' + id;
+          } else {
+            data[3].f = id;
+          }
         }
       }
       Object.assign(data[variant], chars);
--- a/ts/output/chtml/DynamicFonts.ts
+++ b/ts/output/chtml/DynamicFonts.ts
@@ -25,14 +25,17 @@ import {ChtmlCharMap, ChtmlCharData} from './FontData.js';
 
 /**
  * Used by dynamic font extensions to add font identifiers to ranges of characters in given variants
+ *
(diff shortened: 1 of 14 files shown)


ts/output/chtml/FontData.ts
```diff
@@ -38,6 +38,7 @@ export * from '../common/FontData.js';
 export interface ChtmlCharOptions extends CharOptions {
   c?: string;                   // the content value (for css)
   f?: string;                   // the font postfix (for css)
+  F?: string;                   // the full font css class (for extensions)
```
zorkow, line 41: Maybe use a different small letter? I find this confusing.

ts/output/chtml/DynamicFonts.ts
```diff
@@ -0,0 +1,55 @@
...
+      if (id) {
+        for (const c of Object.keys(chars)) {
+          const data = chars[parseInt(c)] as ChtmlCharData;
+          if (!data[3]) {
+            data[3] = {};
+          }
+          if (prefix) {
+            data[3].F = prefix + '-' + id;
+          } else {
+            data[3].f = id;
```
zorkow, line 47: See comment below on `F` vs `f`.

ts/output/common.ts
```diff
@@ -208,6 +208,7 @@ export abstract class CommonOutputJax<
    */
   protected unknownCache: UnknownVariantMap;
 
+
```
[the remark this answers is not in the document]
dpvc: Probably I had inserted something else above it and then deleted it but forgot to delete the blank line.

ts/output/common/FontData.ts
```diff
@@ -911,9 +924,9 @@ export class FontData<C extends CharOptions, V extends VariantData<C>, D extends
   public createVariant(name: string, inherit: string = null, link: string = null) {
     let variant = {
       linked: [] as CharMap<C>[],
-      chars: (inherit ? Object.create(this.variant[inherit].chars) : {}) as CharMap<C>
-    } as V;
-    if (link && this.variant[link]) {
+      chars: Object.create(inherit ? this.variant[inherit].chars : {}) as CharMap<C>
+    } as any as V;
```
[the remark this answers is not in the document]
dpvc: Either works, but I will change to `unknown`.  There are other places where `as any as` also occurs, so perhaps we need to add that to the cleanup.

ts/output/chtml/FontData.ts
dpvc: OK, i've changed to `ff` for `full font`.

PR #849, opened 2022-08-26
delete arguments.callee

masx200:
https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Reference/Functions/arguments/callee


dpvc:
Closing:  see #850.

PR #852, opened 2022-09-01
New font location

dpvc:
This PR changes the directory structure for the font data from `ts/output/<jax>/fonts/<name>` to `ts/output/fonts/<name>/<jax>` so that additional fonts are easier to handle (all their font data is in one directory rather than split into separate output jax directories).  This also makes it possible to use a single configuration path to locate an extra font, so that third-party fonts are easier to handle, and so that if the user switches the output jax, the same font path can be used (with the new jax appended).  We also add a new `output` option where the common options for both output jax can be placed, so that configuration will be the same for both.

Most of the changes in this PR are for renaming the various font files, so it looks like a lot, but mostly is just file name changes.


dpvc:
This PR is going to be replaced by a new one that uses the default font from an npm package rather than having it in the main repository, so I'm closing this one in favor of the new one to come.

PR #853, opened 2022-09-04
Update lite DOM

dpvc:
This PR removes the `CDATA_ATTR` array for the LiteDOM and always translated entities both during parsing of attributes and during serialization.  It also allows the `safe` extension to be used with the LiteDOM (not sure why it was excluded before, as it seems to work just fine).


zorkow (approved):
lgtm.

PR #856, opened 2022-09-13
Adjust how \end{...} is handled for user-defined environments to avoid a potential recusion loop

dpvc:
This PR changes how the `\end` is handled for a user-defined environment so that `parser.Parse()` isn't called, avoiding a potential infinite loop.

In this PR, the end material from the `\newenvironment` is inserted into the parser string (with `addArgs()` so buffer overflow is checked) along with a following `\end{...}`, but we record the fact that we have already added the end material by setting a stack environment value so that the second `\end{}` will actually finish the environment.  This avoids the recursive loop triggered by something like

``` latex
\newenvironment{a}{x}{y\end{a}}
\begin{a} ... \end{a}
```

and it now generates the proper error about `\end{a}` without a `\begin{a}`.


zorkow (approved):
Tested and lgtm.

PR #858, opened 2022-09-15
Change usage of thickmathspace to numeric value in \bmod.  (mathjax/MathJax#2902)

dpvc:
This PR removes the explicit use of `thickmathspace` in `\bmod` and uses the numeric equivalent instead.  It also makes a variable for `em(MATHSPACE.thickmathspace)` and uses that so that it doesn't have to be computed multiple times.

Resolves issue mathjax/MathJax#2902.


zorkow (commented):
lgtm.

zorkow (approved):

PR #860, opened 2022-09-15
Handle \operatorname{} like named functions. (mathjax/MathJax#2899)

dpvc:
This PR makes operators from `\operatorname` and `\DeclareMathOperator` act the same as named functions like `\sin`.  in particular, they get the invisible function apply when followed by an argument.

Resolves issue mathjax/MathJax#2899.


zorkow (approved):
lgtm.

PR #861, opened 2022-09-15
Refactor MathML visitors to make MathMLVisitor work like SerializedMml. (mathjax/MathJax#2920)

dpvc:
This PR makes the MathMLVisitor work more like the SerializedMmlVisitor in terms of handling of attributes and converting TeXAtom nodes to mrows.  Common code has been moved to the common MmlVisitor class that is the root for both.  This invalid moving the static data items from SerializedMmlVisitor to MmlVisitor, as well as the protected methods for managing the attribute lists, so the big chunk of additions to MmlVisitor is just moving the code from SerializedMmlVisitor.  The handling of TeXAtom nodes as a special case has been removed in favor of a translation table for the node kind.  Support for the new HTML nodes has been added to the MathMLVisitor.

Resolves issue mathjax/MathJax#2920.

### DIFF
--- a/ts/core/MmlTree/MmlVisitor.ts
+++ b/ts/core/MmlTree/MmlVisitor.ts
@@ -111,7 +111,7 @@ export class MmlVisitor extends AbstractVisitor<MmlNode> {
    */
 
   /**
-   * @param {MmlNode} node  The node whose kind is nedded
+   * @param {MmlNode} node  The node whose kind is needed
    * @return {string}       The MamlML node name for that kind
    */
   protected getKind(node: MmlNode): string {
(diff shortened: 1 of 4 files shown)


ts/core/MmlTree/MmlVisitor.ts
```diff
@@ -58,10 +92,95 @@ export class MmlVisitor extends AbstractVisitor<MmlNode> {
...
+   */
+  public visitHtmlNode(_node: HtmlNode<any>, ..._args: any[]): any {}
+
+  /***********************************************/
+  /**
+   * Utilities for handling attributes
+   */
+
+  /**
+   * @param {MmlNode} node  The node whose kind is nedded
```
zorkow, line 114: typo

zorkow (approved):
lgtm

PR #862, opened 2022-09-21
Add version check to entities component.

dpvc:
This PR adds the version check to the entities component (which is usually added in the `lib` file from the `build.json` file, but this component doesn't have one, so it must be added manually).  This will avoid a warning message about the component not having a version number.

This resolves part of mathjax/MathJax-demos-node#51.


zorkow (approved):
lgtm.

PR #872, opened 2022-11-10
Rewrite dynamic submenus

zorkow:
Refactors the MathJax context menu:

* Rewrites the clearspeak sub menu by moving some of the methods that compute the menu items in the dynamic submenu from SRE into MathJax, where they really belong. In addition, these computations are moved from `explorer.ts` to its own dedicated file `SpeechMenu.ts`. This way it can be included into `MenuHandler` directly and makes `explorer.ts` independent of any references to menu code, which should help to eventually remove circularities.
* Adds a `RadioCompareItem` class that allows to specify a customisable comparator method. The ordinary `RadioItem` only compares wrt. to the variable it is defined for. This is needed for the `ClearSpeak` submenu, where the radio buttons for smart preference setting and in the full selection box are difficult to control otherwise. In fact, they were previously only correct if only one of the preferences was not set to `Auto`. O/w they were incorrect  in the selection box.
* Refactors the `Annotation` submenu as a dynamic submenu and moves all corresponding code into `AnnotationMenu.ts`. This ensures that it only needs to be  included in the `Menu.ts` file, making `MJContextMenu` independent of MathJax specific code. This would also allow us to move MJContextMenu into the actual context menu package at some point.

**Also note, this will only work with SRE 4.1.0-beta.3!**


ts/a11y/SpeechMenu.ts
```diff
@@ -0,0 +1,209 @@
...
+ */
+
+/**
+ * @fileoverview Clearspeak preference menu.
+ *
+ * @author v.sorge@mathjax.org (Volker Sorge)
+ */
+
+import { SpeechExplorer } from './explorer/KeyExplorer.js';
+import { ExplorerMathItem } from './explorer.js';
```
dpvc, line 25: Traditionally, we have not included spaces in the braces.  Is this a new style that you want to start using?  Personally, I don't care for it, but I know a lot of people do it,.

ts/a11y/SpeechMenu.ts
```diff
@@ -0,0 +1,209 @@
...
+function csPrefsVariables(menu: MJContextMenu, prefs: string[]) {
+  let srVariable = menu.pool.lookup('speechRules');
+  let previous = Sre.clearspeakPreferences.currentPreference();
+  csPrefsSetting = Sre.clearspeakPreferences.fromPreference(previous);
+  for (let pref of prefs) {
+    menu.factory.get('variable')(menu.factory, {
+      name: 'csprf_' + pref,
+      setter: (value: string) => {
+        csPrefsSetting[pref] = value;
+          srVariable.setValue(
```
dpvc, line 49: Indenting is off, here.

ts/a11y/SpeechMenu.ts
```diff
@@ -0,0 +1,209 @@
...
+        id: 'clearspeak-default',
+        variable: 'speechRules'
+      },
+      {
+        type: 'radio',
+        content: 'Current Preferences',
+        id: 'clearspeak-' + previous,
+        variable: 'speechRules'
+      },
+      { type: 'rule' },
```
dpvc, line 116: Again, we haven't uses extra spaces in this setting.  Perhaps spread out the object over three lines?

ts/a11y/SpeechMenu.ts
```diff
@@ -0,0 +1,209 @@
...
+ */
+function smartPreferences(previous: string, smart: string, locale: string) {
+  const prefs = Sre.clearspeakPreferences.getLocalePreferences();
+  const loc = prefs[locale];
+  if (!loc) {
+    return [];
+  }
+  const items = [
+    { type: 'label', content: 'Preferences for ' + smart },
+    { type: 'rule' }
```
dpvc, line 137: Extra spaces, again.

ts/ui/menu/AnnotationMenu.ts
```diff
@@ -0,0 +1,188 @@
...
+import {MJContextMenu} from './MJContextMenu.js';
+import {MmlNode} from '../../core/MmlTree/MmlNode.js';
+import {SelectableInfo} from './SelectableInfo.js';
+import * as MenuUtil from './MenuUtil.js';
+
+/**
+ * The annotation types to look for in a MathItem. These are options set in the
+ * Menu object.
+ */
+let annotationTypes: {[type: string]: string[]} = {};
```
dpvc, line 35: The type `{[type: string]: string[]}` gets used several times.  Do you think a type declaration like

``` js
type ANNOTATIONTYPES = {[type: string]: string[]};
```

would be in order?

ts/ui/menu/AnnotationMenu.ts
```diff
@@ -0,0 +1,188 @@
...
+ */
+export function setAnnotationBox(box: SelectableInfo) {
+  annotationBox = box;
+}
+
+/**
+ * The info box for showing annotations (created by the Menu object that
+ * contains this MJContextMenu)
+ */
+let annotationBox: SelectableInfo = null;
```
dpvc, line 49: I'm a little leary about making these values global rather than tied to a context instance like the `MJContextMenu` instance.  It is possible that there are two separate MathDocuments (say one each for two frames in a window) and so two separate contexts where annotations could be used.  I've tried to avoid globals like this and make everything be tied to an instance so that MathJax can have multiple separate contexts at once.  That's one reason these were part of `MJContextMenu` originally.

ts/ui/menu/RadioCompare.ts
```diff
@@ -0,0 +1,77 @@
...
+  protected role = 'menuitemradiocompare';
+
+  /**
+   * @override
+   */
+  public static fromJson(
+    _factory: ParserFactory,
+    {content: content, variable: variable, id: id, comparator: comparator}:
+    {content: string, variable: string, id: string,
+     comparator: (variable: string, id: string) => boolean}, menu: Menu): Radio {
```
dpvc, line 45: These lines are hard to parse (what are the arguments and what the types.  Perhaps moving `, menu: Menu` to a separate line and `): Radio {` to the line after that would help.

dpvc (changes requested):

ts/a11y/SpeechMenu.ts
zorkow: I think it is `prettier` which adds those automatically. Those spaces are probably from an experimental implementation in SRE. I don't add them manually.
We can make a decision once we decided on linting/formatting for v4.

ts/a11y/SpeechMenu.ts
zorkow: Code moved over from SRE, so that is the automatic formatting by `prettier`.

zorkow: Also fixed the indentation.

ts/a11y/SpeechMenu.ts
zorkow: Removed.

ts/ui/menu/AnnotationMenu.ts
zorkow: Added the type.

ts/ui/menu/RadioCompare.ts
zorkow: I used `prettier` to make it more reader friendly. Have a look to see if this would be helpful to use in the future.

ts/ui/menu/AnnotationMenu.ts
zorkow: I've removed all globals now and even caching is handled by a handshake Array.

zorkow:
* Takes care of review suggestions
* Removes all globals from `annotationMenu`. Caching for a node is handled by a handshake Array. 

Question: Do we allow for multiple annotations of the same type? If not we could replace the Array with a Map.

PTAL.

ts/ui/menu/RadioCompare.ts
dpvc: Yes, that looks better.

ts/ui/menu/AnnotationMenu.ts
```diff
@@ -0,0 +1,155 @@
...
+ * Returns a method to create the dynamic submenu for copying annotations.
+ * Clears the annotation cache parameter.
+ *
+ * @param {[string, string][]} cache The annotation cache.
+ */
+export function copyAnnotations(cache: [string, string][]) {
+  return (menu: MJContextMenu, sub: Submenu) => {
+    const annotations = cache.slice();
+    cache.length = 0;
+    return createAnnotationMenu(menu,  sub, annotations,
```
dpvc, line 67: remove extra space in `menu,  sub`.

ts/a11y/SpeechMenu.ts
```diff
@@ -0,0 +1,211 @@
...
+  for (let lang of Sre.locales.keys()) {
+    if (lang === 'nemeth') continue;
+    radios.push({type: 'radio', id: lang,
+                 content: Sre.locales.get(lang) || lang, variable: 'locale'});
+  }
+  radios.sort((x, y) => x.content.localeCompare(y.content, 'en'));
+  return menu.factory.get('subMenu')(menu.factory, {
+    items: radios, id: 'Language'}, sub);
+}
+MJContextMenu.DynamicSubmenus.set('A11yLanguage', localeMenu);
```
dpvc, line 211: The `DynamicSubmenus` are run whenever the menu is posted.  Does the locale menu have to be regenerated every time?  Or can it just be created dynamically when the Menu object is created and then be used from there.

My understanding is that the dynamic menus are for things that change from equation to equation (like the annotations), but global preferences and such should be able to put in the global menu once and continued to be used for all expressions.  But I may be missing something.

ts/a11y/SpeechMenu.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+  if (box) {
+    items.splice(2, 0, box);
+  }
+  return menu.factory.get('subMenu')(menu.factory, {
+    items: items,
+    id: 'Clearspeak'
+  }, sub);
+}
+MJContextMenu.DynamicSubmenus.set('Clearspeak', clearspeakMenu);
```
dpvc, line 192: Same question about dynamic menus here as for the locale menu below.  Do these preferences have to be regenerated for each expression, or can they be reused once created the first time?

ts/ui/menu/AnnotationMenu.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * @param {MJContextMenu} menu The context menu for which the submenu is constructed.
+ * @param {Submenu} submenu The submenu to attach elements to.
+ * @param {[string, string][]} annotations The annotations to use for the submenu.
+ * @param {() => void} action The action to perform when the annotation is selected.
+ * @return {Submenu} The newly created submenu.
+ */
+function createAnnotationMenu(menu: MJContextMenu, submenu: Submenu,
+                              annotations: [string, string][],
+                              action: () => void): Submenu {
```
dpvc, line 140: I had to change `Submenu` to `SubMenu` as the return type (and import `SubMenu`) in order to get this branch to compile without errors.  (Your favorite: capitalization ambiguities.)

ts/ui/menu/Menu.ts
```diff
@@ -557,14 +557,18 @@ export class Menu {
...
     this.originalText.attachMenu(menu);
-    this.annotationText.attachMenu(menu);
     this.mathmlCode.attachMenu(menu);
     this.zoomBox.attachMenu(menu);
     this.checkLoadableItems();
     this.enableExplorerItems(this.settings.explorer);
-    menu.showAnnotation = this.annotationText;
-    menu.copyAnnotation = this.copyAnnotation.bind(this);
-    menu.annotationTypes = this.options.annotationTypes;
+    const cache = [];
```
dpvc, line 564: I had to make this

``` js
  const cache: [string, string][] = [];
```

to get it to compile without errors.

ts/a11y/SpeechMenu.ts
```diff
@@ -0,0 +1,211 @@
...
+let csPrefsSetting: {[pref: string]: string} = {};
+
+/**
+ * Generator of all variables for the Clearspeak Preference settings.
+ * @param {MJContextMenu} menu The current context menu.
+ * @param {string[]} prefs The preferences.
+ */
+function csPrefsVariables(menu: MJContextMenu, prefs: string[]) {
+  let srVariable = menu.pool.lookup('speechRules');
+  let previous = Sre.clearspeakPreferences.currentPreference();
```
dpvc, line 42: I had to update to SRE version 4.1.0-beta.3 to get the `currentPreference()` method to be available.  Maybe that need to be added to `package.json`?

dpvc (approved):
I had to add/change a couple of types in order to get this branch to compile, but other than that it looks good.  I'm confused about why one thing works, but will chat with you about that when we meet next.

ts/a11y/SpeechMenu.ts
zorkow: This works here. We now avoid recomputing the menu.

ts/a11y/SpeechMenu.ts
zorkow: Does not work here, because the Clearspeak menu is adapted according to the current position of the browser to present the "smart preferences".

ts/a11y/SpeechMenu.ts
zorkow: Updated `package.json` and `package-lock.json`

ts/ui/menu/AnnotationMenu.ts
zorkow: I also had compilation problems now. Not sure why that did not happen before. Odd that.

ts/ui/menu/Menu.ts
zorkow: Same here. Also solved.

PR #875, opened 2022-12-01
Fix infinite loop when <tex-html> is used more than once (mathjax/MathJax#2948)

dpvc:
This PR fixes a problem where an infinite loop occurs when `<tex-html>...</tex-html>` is used more than once in an equation.

Resolves issue mathjax/MathJax#2948.


zorkow (approved):
lgtm.

PR #876, opened 2022-12-01
Fix parsing of vertical bar delimiters in physics package (mathjax/MathJax#2973)

dpvc:
This PR fixes the parsing of `|` delimiters in the physics package.  The code that tests for nested delimiters was considering the closing `|` to be the opening delimiter of a nested set of `|` delimiters, leading to an error about missing close delimiters.  The solution is to put the check for nested delimiters *after* the check for closing delimiter to allow for open and close delimiters to be the same character.

Resolves issue mathjax/MathJax#2973.


zorkow (approved):
lgtm.

PR #877, opened 2022-12-02
Expose `StyleList` value contained in `Styles` object

rhysd:
Today I have created a library to render Mathjax LiteDOM into React element tree.

https://github.com/rhysd/react-mathjax-component

One problem I'm facing is that there is no way to access to the key-value pairs of style value stored in `Styles` object. Since React element requires to specify styles through an object like `{maxWidth: '100px'}`, CSS text does not work. But `Styles` object only provides a way to get CSS styles as string. Currently my package [re-parses](https://github.com/rhysd/react-mathjax-component/blob/721b674b236786ad0517f897797df3b81f5bc8c5/litedom.ts#L14) the CSS text returned from `Styles` object but this is redundant (`Styles` generates a CSS string and my package restores the original key-value pairs of style object).

This PR fixes the problem by exposing the internal `StyleList` object.

### DIFF
--- a/ts/util/Styles.ts
+++ b/ts/util/Styles.ts
@@ -405,6 +405,13 @@ export class Styles {
   }
 
   /**
+   * @return {StyleList} The object to map style names to the values
+   */
+  public get styleList(): StyleList {
+    return {...this.styles};
+  }
+
+  /**
    * @param {string} name   The name of the style to set
    * @param {string|number|boolean} value The value to set it to
    */


dpvc:
OK, I'm not opposed to this idea in principle, but there are a couple of issues.  First, we follow the git-flow model, so pull requests need to be made from a branch that originates from the `develop` branch and are targeted to that branch, so you would need to rebase your `expose-stylelist` branch to originating on the `develop` branch.

Second, I would not want to give access to the actual styles object, as that would allow the user to modify the internal styles structure, which is a bad idea, since that structure needs to be maintained properly to be consistent (e.g., when `border` is changed, `border-left`, `border-right`, etc. must be updated).  So I would recommend making a copy of the object, e.g.,

``` js
  public get styleList(): StyleList {
     return {...this.styles};
   }
```

so that the user gets a private copy of the data rather than the live copy.

rhysd:
Hi @dpvc,

thank you for the points. I changed the base branch from `master` to `develop`, and modified the patch to return a copied object.

dpvc:
Thanks for making the changes!

PR #878, opened 2022-12-02
Only activate explorer if in a browser

dpvc:
This PR sets the `enableExplorer` option to `false` if there is no `window` variable, so that it is only activated in interactive contexts, and puts off creating the RegionPool until it is needed (in case it is not enabled).  Now that the explorer is included as part of the combined components, it needs to play nice with node-based applications that use the combined configurations, so references to the `window` and to HTML element properties can cause problems.

I started to convert the explorer files to use generics rather than explicit `HTMLElement` references and use the DOMAdaptor to handle the HTML interaction.  But because some pieces of SRE itself use HTMLElements directly, that probably is not worth it, as even if the MathJax portions are properly handled, the SRE parts could still fail.

The changes here are enough to avoid errors when using the examples in the "simple" directory of the MathJax-demos-node repository.


zorkow (approved):
lgtm. I will merge that.

PR #881, opened 2022-12-09
Adds the latex units package.

zorkow:
Implementation of the old units package  (https://www.ctan.org/pkg/units). Altogether three commands and two options. 
But I was primarily after the `\nicefrac` command.


dpvc (approved):
Looks good.

PR #885, opened 2023-01-21
Work around combining characters being handled as 0-width automatically by some browsers

dpvc:
This PR fixes a problem where some accents are misplaced.  Combining accents like U+0302 are handled in some fonts by making zero-width characters that have negative left bearing (so they extend to the left over the character that precedes them).  Modern browser will position combining characters automatically, whether they are zero-width or not (and they do vertical positioning so that accents are placed over both lower-case and capital letters properly).

MathJax expects the browser to respect the widths of the characters in the font, so when the browser makes it zero-width automatically, MathJax misplaces the accent.  This PR makes sure that the accents are properly placed regardless of whether the character is zero-width or not, and regardless of whether the browser forces zero-width for ones that are non-zero in the font itself.  This is done by setting the width of the `<mjx-c>` to the expected width of the character (and removing the padding that used to make it that wide), and using `text-align: right` so that a zero-width character will move to the right and so be placed properly, while a full-width character will be placed in the right place to begin with.

### DIFF
--- a/ts/output/chtml/FontData.ts
+++ b/ts/output/chtml/FontData.ts
@@ -39,6 +39,7 @@ export interface ChtmlCharOptions extends CharOptions {
   c?: string;                   // the content value (for css)
   f?: string;                   // the font postfix (for css)
   F?: string;                   // the full font css class (for extensions)
+  cmb?: boolean;                // true if this is a combining character
 }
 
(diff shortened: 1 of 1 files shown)


zorkow (approved):
lgtm.

PR #886, opened 2023-01-23
Fix vertical separation for x-arrows (#mathjax/MathJax#2981).

dpvc:
This PR fixes the vertical separation between `\xrightarrow` and the item above it.  It was due to the depth of the strut used to make sure that all the over items are aligned vertically on their baselines, but the value of the depth was based on earlier font handling, where the stretchy characters had the height of their extender, not the assembly as a whole.  The new font settings have all elements in the assembly having the same height and depth in order to improve alignment among the parts, and that is the maximum of the heights and depths of all the parts.  That makes the arrows taller than they were, and so a smaller strut is needed.

Resolves issue mathjax/MathJax#2981.


zorkow (approved):
lgtm.

PR #891, opened 2023-01-30
Fix CHTML so that depth is correct for empty mpadded elements.

dpvc:
This PR fixes a bug in CHTML output for `mpadded` elements that have no content.  The height and/or depth was not being properly handled in the output unless there was content, so this PR adds an empty box so that the depth shows properly.

For example:

``` html
  <mpadded height="1em" depth="1em" width="1em" background="red" />
  <mi>x</mi>
</math>
```

would not have the proper depth before this PR.


zorkow (approved):
Tested with given expression. Lgtm.

PR #893, opened 2023-02-06
Use proper scaling factors for placement of top and bottom when scaled (mathjax/MathJax#2983)

dpvc:
The placement of top and bottom in `munderover` and related constructs were not correct if the `munderover` is scaled, as in `\huge\vec x`.  This was due to the use of the wrong scaling factors in determining the height of the base and the horizontal position of the super- or subscript.  This PR uses the correct scaling factors.

Resolve issue mathjax/MathJax#2983.


zorkow (approved):
lgtm.

PR #894, opened 2023-02-07
Allow (embellished) operators in \underset, \overset, and \underoverset to specify accent="true"

dpvc:
Currently, `\underset`, `\overset`, and `\underoverset` force `accent="false"` on `mo` elements that are set on top or bottom of the base.  This PR allows a (possibly embellished) operator that has an explicit `accent="true"` to set the `accent` or `accentunder` attributes of the `munder`, `mover`, or `munderover` that is produced.  That allows, for example

``` latex
\underset{\mmlToken{mo}[accent="true"]{⎵}}{abcabc}
```

to properly handle the underbrace as an accent (i.e., without reducing its scale).   (Without this PR, the accent always gets `accent="false"`, and so is reduced in size.  There is an additional but where that causes the stretchy character not to stretch sufficiently, but that is treated in the fix-stretchy-widths` branch.)

### DIFF
--- a/ts/input/tex/base/BaseMethods.ts
+++ b/ts/input/tex/base/BaseMethods.ts
@@ -626,11 +626,10 @@ BaseMethods.Overset = function(parser: TexParser, name: string) {
   // @test Overset
   const top = parser.ParseArg(name);
   const base = parser.ParseArg(name);
+  const topMo = top.coreMO();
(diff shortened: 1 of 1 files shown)


zorkow (commented):
lgtm.

zorkow (approved):

PR #895, opened 2023-02-07
Map \models to correct Unicode point, and don't use variant form for \vDash (mathjax/MathJax#2543)

dpvc:
This PR corrects the unicode value for `\models` and removes the variant form from `\vDash` (this will be correct once we go to the mathjax-modern font).

Note, however, that the TeX/LaTeX version of `\models` does not appear the same as the U+22A7 glyph in the mathjax-modern font.  TeX uses the equivalent of `\def\models{\mathrel{|\!=}}`; should we do that as well?

Resolves issue mathjax/MathJax#2543.


zorkow (approved):
lgtm.

PR #896, opened 2023-02-08
Allow borders to inherit surrounding color in SVG (mathjax/MathJax#2939)

dpvc:
This PR resolves a problem where borders that don't specify an explicit color were getting the wrong color in SVG output due to using 'currentColor', which caused them to get the color of the text surrounding the math as a whole.  Instead, they should just not set the fill color and let it be inherited.

Resolves issue mathjax/MathJax#2939.


zorkow (approved):
lgtm.

PR #897, opened 2023-02-11
Add \underline and several other macros to textmacros. (mathjax/MathJax#2969)

dpvc:
This PR adds `\underline`, `\llap`, `\rlap`, the phantom and `\smash` macros, and `\mmlToken` to the `textmacros` package so they can be used in text mode, as they are allowed in LaTeX text-mode.

Resolve issue mathjax/Mathjax#2969.


zorkow (approved):
lgtm.

PR #901, opened 2023-02-14
Better handling of scriptminsize (mathjax/MathJax#2975)

dpvc:
This PR moves the checking of `scriptminsize` to later in the calculation of the scaling factor so that `mathsize` and `font-size` CSS are included in the test, and also changes the default value to a better one (see mathjax/MathJax#2975).

Resolves issue mathjax/MathJax#2975.


zorkow (approved):
lgtm.

PR #902, opened 2023-02-14
Allow \columncolor in the array preamble.  (mathjax/MathJax#2955)

dpvc:
This PR allows `\columncolor` to be used in an array environment preamble (where it is supposed to be used).  We allowed it in the top row in previous versions because we were not processing all the preamble options until now.  This PR allows it to be used both in the preamble and the first row of the array.

Resolves issue mathjax/MathJax#2955.


zorkow (approved):
lgtm.

PR #903, opened 2023-02-15
Combine unknown characters into a common <text> element.  (mathjax/MathJax#2672)

dpvc:
This PR groups runs of unknown characters into a single `<text>` element in SVG output, just as was implemented in CHTML in #734 in commit 115c5423.  It also makes the computation of the bounding box consistent with that, since text with combining characters will have a different width when grouped together than when handled separately.  (The bounding box computations were wrong for CHTML, but since most of the CHTML doesn't depend on those computations, not wasn't caught.)  For example, with `\text{🇦🇩}`, the output was correct but the bounding box computation was wrong.

Resolves issue mathjax/MathJax#2672.

### DIFF
--- a/ts/output/common/Wrappers/TextNode.ts
+++ b/ts/output/common/Wrappers/TextNode.ts
@@ -159,6 +159,7 @@ export function CommonTextNodeMixin<
         bbox.w = w;
       } else {
         const chars = this.remappedText(text, variant);
+        let utext = '';
         bbox.empty();
         //
         // Loop through the characters and add them in one by one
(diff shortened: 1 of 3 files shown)


ts/output/common/Wrappers/TextNode.ts
```diff
@@ -159,38 +159,61 @@ export function CommonTextNodeMixin<
...
+
+    /**
+     * @param {BBox} bbox        The bounding box to update
+     * @param {number} h         The height to use
+     * @param {nunber} d         The depth to use
+     * @param {number} w         The width to add
+     */
+    protected updateBBox(bbox: BBox, h: number, d: number, w: number) {
+      bbox.w += w;
+      if (h > bbox.h) bbox.h = h;
```
zorkow, line 212: Braces.

ts/output/common/Wrappers/TextNode.ts
```diff
@@ -159,38 +159,61 @@ export function CommonTextNodeMixin<
...
+      if (d > bbox.d) bbox.d = d;
```
zorkow, line 213: dito

zorkow (approved):
lgtm with minor changes.

ts/output/common/Wrappers/TextNode.ts
dpvc: I find that harder to read, but I've made the changes.

PR #904, opened 2023-02-15
Provide an option for placement of equation numbers in TeX input.  (mathjax/MathJax#2977)

dpvc:
This PR adds an option to specify the vertical placement of tags in TeX expressions.  The main purpose is to center the tags for long expressions that have line breaks.

Resolves issue mathjax/MathJax#2977.


zorkow (approved):
lgtm.

Issue #899, opened 2023-02-13
ES6 class keyword used in "ES5" code due to bundled speech-rule-engine dependency 


hallvors:
Hi, I came across an issue that causes a syntax error if one tries to use MathJax with the PrinceXML JavaScript engine. I suppose it will also cause trouble in many other older browser that do not support ES6 `class`.

> prince: https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-svg.js:1: error: SyntaxError: unexpected token reserved("class")

It is a little confusing that a URL which contains the string "es5" uses the class keyword :) .

I checked out the repository and the TS compiler does not output class statement anywhere if I run `npm run compile`. However, the `make-components` command that uses Node.js scripts and Webpack  to compile all the sub-components MathJax consists of, unfortunately outputs stuff like
```
class Q extends o.AbstractAudioRenderer {
```
or
```
e.Span = class {
```
Now, there's no MathJax component defining an AbstractAudioRenderer. This seems to come from a dependency which gets bundled into the Mathjax scripts by Webpack (but not compiled / transpiled to ES5 in the process - Webpack configurations tend to assume stuff in dependencies does not need transpiling). The dependency that ships ES6 code which gets bundled is most likely  https://www.npmjs.com/package/speech-rule-engine

Is it possible to enable Babel loader or similar transpiling for this dependency? I know it works as-is with most browsers out there, but at least the URL saying es5 would be more truthful if it were possible to also transpile speech-rule-engine :)

Thanks for your attention and the amazing library.

dpvc:
Thanks for reporting this issue.  You are correct that the speech-rule-engine is the culprit, here, and should be transpired for the `es5` directory.

One work-around for now would be to edit `components/webpack.common.js` and change 

https://github.com/mathjax/MathJax-src/blob/8565f9da973238e4c9571a86a4bcb281b1d98d9b/components/webpack.common.js#L135-L136

to 

``` javascript
      test: new RegExp(dirRE + quoteRE(path.sep) + '.*\\.js$|speech-rule-engine'),
      exclude: new RegExp(quoteRE(path.join(path.dirname(__dirname), 'es5') + path.sep) + '|.json$'),
```

which will cause the speech-rule-engine files to be transpired, excluding its `.json` files.  Then repack the needed component:

``` shell
components/bin/makeAll components/tex-svg
```

That seems to do the trick in my (limited) testing.

dpvc:
[...]

PR #905, opened 2023-02-21
Update webpack configuration to downcompile SRE so es5 directory really is es5.  (#899)

dpvc:
The current combined components in the `es5` directory include SRE as is, which is ES6 (it includes `class` declarations and such), so the results are not actually ES5.  This PR changes the `webpack.common.js` to force SRE to be down compiled, and add the needed regenerator-runtime library to make it work.  This makes the resulting files larger, but does make them actual ES5 files.

Resolves issue #899.

### DIFF
--- a/components/webpack.common.js
+++ b/components/webpack.common.js
@@ -133,7 +133,6 @@ const MODULE = function (dir) {
     // NOTE: for babel transpilation
     rules: [{
       test: new RegExp('(?:' + dirRE + '|speech-rule-engine)' + quoteRE(path.sep) + '.*\\.js$'),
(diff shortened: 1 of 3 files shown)


zorkow (commented):
Would it be easier to offer an es5 compiled SRE?

dpvc:
Looks like the next version of SRE is not going to be able to be down compiled to ES5, so I'm marking this as a draft for now.

We discussed making two versions of MathJax, and ES5 and ES6, so we can either go with only ES6, or have the ES5 not be fully ES5 (not too good), or have the ES5 version NOT include SRE by default.

dpvc:
Closing this, as the new ES6/ES5 configuration makes it unnecessary.

PR #906, opened 2023-02-21
Properly handle \over within braket and physics packages. (mathjax/MathJax#3000)

dpvc:
This PR modifies the braket and physics packages so that they properly handle `\over` constructs as the actual LaTeX packages do.  Although `\frac` is recommended, `\over` can still be used, and the current code for these packages doesn't handle that properly.  (For bracket, `\Set{a\over b}` includes braces around the a as well as the whole fraction, and for physics, `\sin(a\over b)` generates an error message about missing delimiters.)  This PR also fixes some spacing issues with `\Set`.

Because braket and physics use `BraketItem` and `AutoOpen` items that have `isOpen` set to `true`, they act as delimiters for `\over`, and need to take that into account.  Because `\over` works by collecting the current node list from the preceding open `StackItem` using its `toMml()` method, and saving that in the `num` (numerator) property of the `StackItem`, the `BraketItem` and `AutoOpen` items need to take that behavior into account.  This PR uses the `inferred` argument to `toMml()` (which is set to false when the `OverItem` calls it) to tell whether to just do the super `toMml()`, or to do its full version (which adds the delimiters).

We also need to have the braket package have vertical bars mark the ends of `\over` constructs, and for physics, the closing delimiters must also end the `\over` construction.  So this PR modifies the `Bar` and `AutoClose` methods to handle an `OverItem` as the top item, and to use a fake `CloseItem` to end the fraction.  For braket items, the material up to the last bar is stored in a `barNodes` list (similar to the `num` property for `\over`) so that additional fractions can occur between bars, and after the last bar.  The `toMml()` method then re-inserts that list into the final MathML.

A similar fake close item is used in the physics case, but because parenthesis balancing is being done, that has to be taken into account, so that `\sin(a(b\over c))` will properly handle the faction (which should produce the equivalent of `\sin(\frac{a(b}{c)})` with the inner parentheses being part of the fraction, and only the outer ones being stretchy).

Finally, the spacing for `\Set` and its bar (if any) is corrected.

Resolves issue mathjax/MathJax#3000.

### DIFF
--- a/ts/input/tex/base/BaseItems.ts
+++ b/ts/input/tex/base/BaseItems.ts
@@ -316,12 +316,12 @@ export class OverItem extends BaseItem {
         NodeUtil.setAttribute(mml, 'linethickness',
                               this.getProperty('thickness') as string);
       }
-      if (this.getProperty('open') || this.getProperty('close')) {
+      if (this.getProperty('ldelim') || this.getProperty('rdelim')) {
         // @test Choose
         NodeUtil.setProperty(mml, 'withDelims', true);
         mml = ParseUtil.fixedFence(this.factory.configuration,
-                                   this.getProperty('open') as string, mml,
-                                   this.getProperty('close') as string);
+                                   this.getProperty('ldelim') as string, mml,
+                                   this.getProperty('rdelim') as string);
       }
       return [[this.factory.create('mml', mml), item], true];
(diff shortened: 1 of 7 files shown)


zorkow (approved):
lgtm.

PR #907, opened 2023-02-21
Make vertical bar have the correct TeX class.  (mathjax/MathJax#2938)

dpvc:
When used in prefix or postfix positions, the operator dictionary gives `|` TeX class of OPEN or CLOSE, but TeX creates it as ORD in these locations.  This PR forces the TeX input jax to make it ORD, while still allowing MathML to treat them as OPEN and CLOSE in a construction like `<mrow><mo>|</mo><mi>x</mi><mo>|</mo></mrow>` that corresponds to `\left|x\right|` (where they *do* have OPEN and CLOSE class).

Resolves issue mathjax/MathJax#2938 and part of mathjax/MathJax#2894.


zorkow (approved):
lgtm.

PR #908, opened 2023-02-22
Fix scaling of tooltips when text font is inherited. (mathjax/MathJax#2957)

dpvc:
This PR fixes a problem where the scaling for `mtext` elements in tooltips is wrong when `mtextInheritFont` is true (or `mtextFont` is set).  This was due to the CSS `font-size: 70%`, which wasn't being taken into account with the measuring of the font size of the inherited font.  The solution used here is to remove the 70% scaling and setting the `scriptlevel` to 1 (which is a 70.7% scaling).


zorkow (approved):
lgtm.

PR #909, opened 2023-02-22
Allow SVG character stroke-width to be specified. (mathjax/MathJax#2859)

dpvc:
This PR provides a configuration option that controls the stroke-width used for characters in SVG output.  This allows the author to adjust the thickness to better match he surrounding font, if needed.  The current default is 3, but I prefer values more like 10 or even 15 on MacOS.  But the AMS would like to use 0.

This is accomplished by providing a per-wrapper function that adds the wrapper-specific CSS so that it can use the output jax options to control the styles.

Resolves isses mathjax/MathJax#2859.

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -29,7 +29,7 @@ import {Property} from '../../core/Tree/Node.js';
 import {unicodeChars} from '../../util/string.js';
 import * as LENGTHS from '../../util/lengths.js';
 import {Styles} from '../../util/Styles.js';
-import {StyleList} from '../../util/StyleList.js';
+import {StyleList, CssStyles} from '../../util/StyleList.js';
 import {OptionList} from '../../util/Options.js';
 import {CommonOutputJax} from '../common.js';
 import {CommonWrapperFactory} from './WrapperFactory.js';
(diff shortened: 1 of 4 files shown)


pkra:
Please don't add this just because of my report. We won't use it this way anyway.

dpvc:
As I mentioned, I prefer a larger number myself, so may use this myself, and since the new default font has thinner lines than the older MathJax TeX font, it may be that some people may want to beef them up a bit with a higher setting.

For your usage, of course you can do `new SVG({blacker: 0})` to get the 0 value you originally requested.  But changing it via CSS that overrides the default is still an option.

pkra:
> But changing it via CSS that overrides the default is still an option.

Just fyi / developer feedback: I never use the default CSS. I even rewrite the output to have simpler styles to work with. The additional maintenance on our end is not much and thus worth it.

dpvc:
> I never use the default CSS.

Just for clarity, there are two reasons that the CSS is more complex:  (1) it helps to make sure that MathJax CSS doesn't affect other content on the page, and (2) it helps make MathJax CSS more specific so that other CSS won't override MathJax's CSS so readily.  Because MathJax doesn't know what other content or CSS is on the page, it can't tell if a simpler CSS is sufficient.  But since you control your page's content and CSS, you can, and so simplifying the CSS is reasonable.

ts/output/common/Wrapper.ts
```diff
@@ -271,6 +279,16 @@ export class CommonWrapper<
     }
   };
 
+  /**
+   * Add any styles for this wrapper class
+   *
+   * @param {CssStyles} styles   The styles object to extend
+   * @param {JX} _jax            The output jax whose style sheet is being modified (in case options are needed)
+   */
```
zorkow, line 287: That should just be an `@override`.

zorkow (approved):
lgtm with minor changes.

PR #910, opened 2023-02-22
Add option to not use xlink namespace. (mathjax/MathJax#2929)

dpvc:
This PR adds an option to the SVG output jax to control whether the `href` attributes in `use` elements will include the `xlink` namespace or not.  It once was necessary, but is not any longer in most browsers.

Resolves issue mathjax/MathJax#2929.


zorkow (approved):
lgtm.

PR #911, opened 2023-02-22
Don't typeset within <select> or <option> nodes. (mathjax/MathJax#3006)

dpvc:
This PR adds `<select>` and `<option>` tags to the `skipHtmlTag` list, since the contents of `<option>` is text-only, and shouldn't be modified by MathJax.

Resolves issue mathjax/MathJax#3006.


zorkow (approved):
lgtm.

PR #912, opened 2023-02-23
Make sure multiple elements are typset in order. (mathjax/MathJax#2999)

dpvc:
This PR refactors the `HTMLDocument.findMath()`method so that the order of the math is preserved properly.  This is done by changing the order of the nested loops.  In the past, the outer loop was on the containers, and then each Jax was called to find its math and the results were merged into the document list.  This failed to work properly because the sorting is based on the index of the string in which the math was found, but the list of strings is separate for each container, so they get interleaved when merged.  (See mathjax/MathJax#2999 for an example.)

Now the order is to loop over the jax first, and then over the containers, so that all the strings for all the containers are in one larger list, and there is no interleaving.  At least not for a single typeset call.

Resolves issue mathjax/MathJax#2999.


zorkow (approved):
lgtm.

PR #913, opened 2023-02-24
Allow separate formatting for \eqref{}.  (mathjax/MathJax#2736)

dpvc:
This adds a separate function for formatting the output for `\eqref`.  The default is to use the `formatTag()` function unless that is overridden with the `ref option in the `tagformat` configuration block.

Resolves issue mathjax/MathJax#2736.


zorkow (approved):
lgtm.

PR #914, opened 2023-02-24
Add ability to define active characters as macros. (math/MathJax#2967)

dpvc:
This PR adds the ability for the `configmacros` extension to create active characters tied to macros.  The new `active` option in the `configmacros` block of the `tex` configuration defines active characters in the same way as `macros` does for control sequences.

Resolves issue mathjax/MathJax#2967.


zorkow (approved):
lgtm.

PR #915, opened 2023-02-27
Better checking of improperly nested AMS environments. (mathjax/MathJax#1892)

dpvc:
This PR redoes the check for improperly nested AMS environments, which currently allows them to be used in situations where they are not allowed in LaTeX (e.g., there is no error if they are not a top-level environments, so `align` can be used inside an array, for example, where it shouldn't be allowed).  So we introduce a test for being at the top level (the task top must be a `start` item with no stored nodes).

On the other hand, `gather` should allow nested `align` and other environments (but not another `gather`), so we use a `nestable` property and ad an extra parameter to tell if the current environment is a second `gather`.

All of this makes the `eqnenv` global property no longer necessary.

Resolves issue mathjax/MathJax#1892.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -535,12 +535,18 @@ namespace ParseUtil {
   /**
    *  Check for bad nesting of equation environments
    */
-  export function checkEqnEnv(parser: TexParser) {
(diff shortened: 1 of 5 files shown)


zorkow (approved):
lgtm.

PR #916, opened 2023-02-28
Use left alignment and no indent for forced inline breaks when inline breaking is off. (mathjax/MathJax#3005)

dpvc:
When automatic inline breaking is turned off, a forced line break in an inline expression ends up using the `displayAlign` and `displayIndent` values rather than being left aligned and unindented.  This PR forces left alignment and 0 indent in that case.

Resolves one issue from mathjax/MathJax#3005.


zorkow (approved):
lgtm.

PR #917, opened 2023-02-28
Make \frak consistent with latex, and remove non-standard \scr and \bbFont. (mathjax/MathJax#1780)

dpvc:
This PR makes `\frak` consistent with LaTeX's version (that takes an argument rather than acts as a switch), and removes the non-standard `\src` and `\bbFont`, which are no longer needed.  It also makes a constant `VARIANT` to make references to `TexConstant.Variant` smaller (and should compress better).

Resolves issue mathjax/MathJax#1780.


zorkow (commented):
The `TexConstant` namespace is now outdated. It would be easier to simply get rid of the namespace and import elements like `Variant` directly. We can still do `import * as TexConstant...` 
Should we do that for 4.0? It should not be too much work as it is restricted to the `tex` module only.

zorkow (approved):
O/w fine, modulo the previous comment. However, if we agree to get rid of `TexConstant` it should be done in a separate PR.

PR #919, opened 2023-03-03
Fix several issues with inline breaks. (mathjax/MathJax#3005)

dpvc:
This PR fixes a number of problems with inline breaks in SVG and CHTML output.

* The old approach of using `font-size` to scale a 1em space to the proper size turns out not to work as well as hoped, since it runs into the minimum font size settings of most browsers.  Instead, we now use `letter-spacing` to adjust the size of the space without scaling.  The space is still 1em, so we are usually removing space from it.

* For CHTML output, the `mjx-linestack` and `mjx-linebox` nodes were not being set to `display: inline` as needed for inline breaks because the math node wasn't setting the `breakable` attribute on its children.  This was due to a change in the property name from `breakable` to `process-breaks` in an earlier PR that missed changing the name in the math wrapper, and also to the fact that some CSS was using `data-mjx-breakable` rather than just `breakable`.

* We now use CSS to process explicit breaks in CHTML using `::before` and `::after` selectors on the line boxes that are produced for explicit breaks rather than adding `mjx-break` elements for that.  (Note that the space character is 1em wide in the MJX-BRK font, so `letter-spacing: -.999em` makes it nearly 0 width, but still allowing a break to occur there.  It has to be non-zero width for that in some browsers.)

* The marking of breakpoints in `common.ts` has been modified to include `mspace` elements (moving common code to a helper function), to handle `linebreakstyle="after"` to be honored, and to allow breaks when `lspace` or `rspace` have been specified, none of which were handled before.  The `svg.ts` handling of breaks is modified in a similar way to accommodate the changes.

* Finally, the handling of the line spacing for inline breaks is improved.

Resolves issues from mathjax/MathJax#3005

### DIFF
--- a/ts/output/chtml/Wrappers/mrow.ts
+++ b/ts/output/chtml/Wrappers/mrow.ts
@@ -100,15 +100,27 @@ export const ChtmlMrow = (function <N, T, D>(): ChtmlMrowClass<N, T, D> {
         display: 'inline-table',
         width: '100%'
       },
-      'mjx-linestack[data-mjx-breakable]': {
-        display: 'inline',
-        width: 'initial',
-      },
-      'mjx-linestack[data-mjx-breakable] > mjx-linebox': {
+      'mjx-linestack[breakable]': {
         display: 'inline'
       },
-      'mjx-break[newline]::after': {
-        display: 'block'
+      'mjx-linestack[breakable] > mjx-linebox': {
+        display: 'inline'
+      },
+      'mjx-linestack[breakable] > mjx-linebox::before': {
+        'white-space': 'pre',
+        content: '"\\A"'
+      },
+      'mjx-linestack[breakable] > mjx-linebox::after': {
+        'white-space': 'normal',
+        content: '" "',
+        'letter-spacing': '-.999em',
+        'font-family': 'MJX-BRK'
+      },
+      'mjx-linestack[breakable] > mjx-linebox:first-of-type::before': {
+        display: 'none'
+      },
+      'mjx-linestack[breakable] > mjx-linebox:last-of-type::after': {
+        display: 'none'
       },
       'mjx-linebox': {
         display: 'block'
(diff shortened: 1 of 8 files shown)


ts/output/chtml/Wrappers/mrow.ts
```diff
@@ -100,15 +100,27 @@ export const ChtmlMrow = (function <N, T, D>(): ChtmlMrowClass<N, T, D> {
...
       },
-      'mjx-linestack[data-mjx-breakable] > mjx-linebox': {
+      'mjx-linestack[breakable] > mjx-linebox': {
         display: 'inline'
       },
-      'mjx-break[newline]::after': {
-        display: 'block'
+      'mjx-linestack[breakable] > mjx-linebox::before': {
+        'white-space': 'pre',
+        content: '"\\A"'
```
zorkow, line 111: Why do we need this `content`?

ts/output/svg.ts
```diff
@@ -379,19 +380,17 @@ CommonOutputJax<
       //
       const [mml, mo] = wrapper.childNodes[0].getBreakNode(line);
       const forced = !!(mo && mo.node.getProperty('forcebreak'));
-      if (i || forced) {
+      if (forced && mo.node.attributes.get('linebreakstyle') === 'after') {
+        const k = mml.parent.node.childIndex(mml.node) + 1;
+        const next = mml.parent.childNodes[k+1];
```
zorkow, line 385: spacing.

zorkow (approved):
Minor things. O/w lgtm.

ts/output/chtml/Wrappers/mrow.ts
dpvc: For inline math that contains explicit line breaks (via `\\` in TeX or `<mo linebreak="newline">` in MathML), the linebreaking algorithm will insert `mjx-linestack` and `mix-linebox` elements.  The latter are normally `display: block`, which make the line break, but for in-line math, the `mjx-linebox` nodes are `display: inline` and on't force a line brake.  This `content` and `white-space` is what forces the linebreak at the end of the `mjx-linebox` in in-line math.

PR #920, opened 2023-03-06
Update CSS for assistive-mml. (mathjax/MathJax#2936)

dpvc:
This PR adjusts the CSS for the assistive MathML output so that it fits the space of the math expression exactly, as the current arrangement sometimes allows it to be larger than the MathJax output.  It also adds a `clip-path` since `clip` is deprecated.  Leaving the `clip` CSS allows for support of IE11 (though I'm not sure how much of MathJax still works in IE).


zorkow (approved):
lgtm.

PR #921, opened 2023-03-07
Implement \char and add \U for raw unicode characters.

dpvc:
This PR implements a version of the low-level `\char` command for inserting characters by number (decimal, octal, or hex) or by character (e.g., `\char'\%` or `\char'A`).  In actual TeX, these are font positions in TeX fonts, and are limited to 0 to 127, but here, they are Unicode code points, and can be 1 to 6 hex digits.  So `\char"222B` is the integral sign, for example.  These are mapped to MathML nodes using the `Other()` function from the base methods (so uses the `RANGES` list from the operator dictionary to determine the node type to use).

We also add a `\U` macro for inserting a character into the TeX input string by Unicode hex code point.  So `\U{222B}` is the integral sign.  This is different from `\unicode{x222B}` as the latter always produces an `mtext` node, while the former will be treated as though the unicode character was part of the original TeX string (and `\char"222B` will place the integral sign in an `mo` element without further processing).

Another example of the difference is that `\unicode{x25}` produces `<mtext>%</mtext>`, `\char"25` will produce `<mo>%</mo>`, and `\U{25}` will produce a percent that comments out the rest of the line.

These new macros are available in text mode when `textmacros` is loaded; `\U` is available in text mode without `textmacros`, and in the contents of the second argument to `\mmlToken{}{...}`.  That is, `\mmlToken{mo}{\U{222B}}` would produce `<mo>&#x222B;</mo>`.

### DIFF
--- a/ts/input/tex/unicode/UnicodeConfiguration.ts
+++ b/ts/input/tex/unicode/UnicodeConfiguration.ts
@@ -31,6 +31,7 @@ import {ParseMethod} from '../Types.js';
 import ParseUtil from '../ParseUtil.js';
 import NodeUtil from '../NodeUtil.js';
 import {numeric} from '../../../util/Entities.js';
+import {Other} from '../base/BaseConfiguration.js';
 
 // Namespace
 export let UnicodeMethods: Record<string, ParseMethod> = {};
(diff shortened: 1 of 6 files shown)


ts/input/tex/unicode/UnicodeConfiguration.ts
```diff
@@ -89,8 +90,83 @@ UnicodeMethods.Unicode = function(parser: TexParser, name: string) {
...
+ * @param {string} name The name of the macro.
+ */
+UnicodeMethods.Char = function (parser: TexParser, _name: string) {
+  let match;
+  let next = parser.GetNext();
+  let c = '';
+  const text = parser.string.substr(parser.i);
+  if (next === '\'') {
+    match = text.match(/^'(?:([0-7]{1,7}) ?|(\\\S)|(.))/u);
+    if (match) {
```
zorkow, line 123: Can we `break` here on `!match` to reduce nesting depth.

zorkow (approved):
One minor thing. O/w lgtm.

ts/input/tex/unicode/UnicodeConfiguration.ts
dpvc: No, I don't think so.  There is no loop or `switch` statement to terminate, and we do want to fall through to the `if (!c)` that produces the error message at line 150.  So I don't think `break` works here.

PR #923, opened 2023-03-07
Add requested macros and environments now that we can support them.

dpvc:
This PR adds a number of macros and environments that have been requested in the past, now that we have fonts that contain the needed characters.  These include `\AA`, `\iddots`, `\oiint` and `\oiiint`, plus I added `\dddot` and `\ddddot` accents, and the missing `\ointop`.  The `TEXCLASS.OP` has been removed from the big operators, since those will be inherited from the operator dictionary, so no need to waste the space on redundant values.  The unicode value for `\backslash` has been changed to just `\`. 

The new environments are `darray` for an array in display mode, and `math` and `displaymath`.  Since `\begin{xyz}...\end{xyz}` are parsed as math-mode delimiters, and these default to display mode, the `math` environment requires the ability to change the MathItem's `display` property to `false` for this environment.  That required adding the active MathItem to the ParseOptions instance so that the `math` environment could change it.  In turn, that means the TeX input jax can't cache the `display` value, since it can change during the parsing process.

Resolves:
* mathjax/MathJax#566
* mathjax/MathJax#795
* mathjax/MathJax#996
* mathjax/MathJax#1269
* mathjax/MathJax#1937
* mathjax/MathJax#2838
* mathjax/MathJax#2878

### DIFF
--- a/ts/input/tex/base/BaseMethods.ts
+++ b/ts/input/tex/base/BaseMethods.ts
@@ -1649,7 +1649,6 @@ BaseMethods.Equation = function (
   numbered: boolean,
   display: boolean = true
 ) {
-  ParseUtil.checkEqnEnv(parser);
   parser.configuration.mathItem.display = display;
   parser.stack.env.display = display;
   parser.Push(begin);
@@ -1800,4 +1799,3 @@ BaseMethods.MathChoice = function(parser: TexParser, name: string) {
 
 
 export default BaseMethods;
-
--- a/ts/input/tex/base/BaseMethods.ts
+++ b/ts/input/tex/base/BaseMethods.ts
@@ -1512,8 +1512,7 @@ BaseMethods.BeginEnd = function(parser: TexParser, name: string) {
   }
   ParseUtil.checkMaxMacros(parser, false);
   parser.parse('environment', [parser, env]);
-};
-
+ };
 
 /**
  * Handle array environment.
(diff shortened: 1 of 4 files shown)


ts/input/tex/base/BaseMethods.ts
```diff
@@ -1791,3 +1800,4 @@ BaseMethods.MathChoice = function(parser: TexParser, name: string) {
 
 
 export default BaseMethods;
+
```
zorkow, line 1803: Extra whitespace.

ts/input/tex/base/BaseMethods.ts
```diff
@@ -1641,8 +1641,17 @@ BaseMethods.IndentAlign = function (parser: TexParser, begin: StackItem) {
...
+  parser: TexParser,
+  begin: StackItem,
+  numbered: boolean,
+  display: boolean = true
+) {
+  ParseUtil.checkEqnEnv(parser);
+  parser.configuration.mathItem.display = display;
+  parser.stack.env.display = display;
   parser.Push(begin);
   ParseUtil.checkEqnEnv(parser);
```
zorkow, line 1656: This is now already checked.

zorkow (approved):
Some minor things.

ts/input/tex/base/BaseMethods.ts
dpvc: Sorry about that.  There was a merge conflict when I cherry-picked this change into a separate branch, and I missed that in resolving the conflict.  I'll remove the second check.

dpvc: Actually, it's the first check that should be removed.

PR #926, opened 2023-03-09
Allow space for labels when deciding on a table's width for line breaking its columns.

dpvc:
This PR takes the label and padding properly into account for deciding the width for a table when determining column break widths.  In particular, it does the right thing for centered tables.


zorkow (approved):
lgtm.

PR #927, opened 2023-03-14
Better control over breaking in tables, and over vertical alignment of nodes containing breaks.

dpvc:
The motivation for this PR is to perform better line break alignments within table cells, but that lead to some other improvements as well.  In 4.0.0-alpha, when line breaking occurs, the baseline of the result is the baseline of the top line.  That is not always optimal, and leads to bad breaking in `align` and other equation alignment environments.  For example, 

``` latex
\begin{align}
a+b+c+d+e &= a+b+c
\end{align}
```

might break as

bad line breaks

This PR adds the ability to specify how cells that have linebreaks are aligned, and uses that to set up the `align` and similar environments so the first column aligns on the *last* line while the right-hand column aligns on the top.  So you get something like

good line breaks

instead.  This makes the connection from left to right work more naturally.

There is a new macro `\breakAlign{type}{align}` that lets you set the alignment for a cell, row, or the entire table (type = c, r or t) to the given alignment (t, b, m, c for top, bottom, middle, or center).  This may be extended in the future to allow alignment on any line (by number). The difference between middle and center is that center is always centered, even if one line, while middle takes effect only when there are line breaks.  For a row for the table, you can enter multiple characters, one for each column.  This macro can be used inside any array or alignment table.

The implementation of these alignments is done in the (generally inferred) `mrow` element, which looks for a `data-vertical-align` attribute on the parent element.  The `mtable` and `mtr` elements have `data-break-align` attributes that set the `data-vertical-align` on their `mtd` elements (through the inheritance mechanism).

This new functionality for `mrow` essentially duplicated that of the `TeXAtom` element when the `texClass` was `VCENTER`, `VBOX`, or `VTOP`, and so those classes have been removed, and are no longer handled by `TeXAtom`, but rather by the nested `mrow` elements themselves.  So those classes have been removed, and the `TeXAtom` classes have been simplified to only deal with changing the `texClass` of their contents.  The `center` value is used to handle the `VCENTER` satiation, which is centered regardless of whether there are line breaks, whereas `middle` is used for situations where a single line should be placed on the normal baseline position, but multiple lines should be centered.

The `data-vertical-align` attribute can be used on any linebreak container, like `msqrt` or `mpadded` or `mtd` in order to  control how the element is aligned when it breaks.  The `msqrt` and `mroot` elements have a default value for `data-vertical-align` of `bottom` so that they align on their bottom rather than top lines, as in

sqrt

though perhaps `middle` would be better.  Of course, one can always use `\vcenter`, `\vtop` or `\vbox` to align the `\sqrt` in any way desired.

The `EqnArray` method of `base/BaseMethods.ts` now has an extra parameter that specifies the `data-break-align` list for the `mtable` that it creates.  

The `ArrayItem` in `base/BaseItems.ts` is modified to handle the `\breakAlign` macro, and `EqnArrayItem` to make the `EndEntry()` and `EndRow()` methods be able to call the `super` versions.

The output wrappers for `mrow` and `TeXAtom` are modified to handle the functionality moved from `TeXAtom` to `mrow`.


ts/input/tex/base/BaseMethods.ts
```diff
@@ -55,6 +55,20 @@ const MmlTokenAllow: {[key: string]: number} = {
 };
 
 
+export function splitAlignArray(align: string, n: number = Infinity) {
+  const list = align.replace(/\s+/g, '').split('').map((c: string) => {
+    const name = {t: 'top', b: 'bottom', m: 'middle', c: 'center'}[c];
```
zorkow, line 60: Maybe not call variable `c`. That is confusing.

ts/output/common/Wrappers/mrow.ts
```diff
@@ -291,6 +302,26 @@ export function CommonMrowMixin<
       bbox.clean();
     }
 
+    /**
+     * Adjust bbox vertical alignment. (E.g., for \vbox, \vcenter.)
+     */
+    protected vboxAdjust(bbox: BBox) {
+      const n = this.breakCount;
+      const valign = this.parent.node.attributes.get('data-vertical-align');
+      if (valign === 'bottom' && n) {
```
zorkow, line 311: Swap to test `n` first.

ts/output/common/Wrappers/mrow.ts
```diff
@@ -291,6 +302,26 @@ export function CommonMrowMixin<
...
+        this.dh = (n ? bbox.d - this.lineBBox[n - 1].d : 0);
+      } else if (valign === 'center' || (valign === 'middle' && n)) {
```
zorkow, line 313: Why don't we need a breakcount on `center`?
Also swap `n`.

zorkow (approved):
Minor issue. O/w fine.

ts/output/common/Wrappers/mrow.ts
dpvc: As I mention in the initial PR discussion, `center` is centered regardless of linebreaks.  This is what `\vcenter` does, whereas `middle` only centers if there are line breaks.  Without this distinction, you would need to keep the `\vcenter` code in the `TeXAtom` node, which was always a bit out of place there but was there because I was trying to keep all the non-MathML functionality in one place.  We now have various extensions to MathML attributes (e.g., `data-padding` for `menclose`, and this `data-vertical-align`, and a number of table border extensions, etc.), so that isolation of non-MathML material no longer holds, and it seemed better to go back to `TeXAtom` just handling the change of `texClass`.

zorkow: Right. Sorry about not making the connection. I had read the explanation when I started the review, but by the time I looked at this file (after lunch...) the information had left my brian.

ts/input/tex/base/BaseMethods.ts
```diff
@@ -55,6 +55,20 @@ const MmlTokenAllow: {[key: string]: number} = {
 };
 
 
+export function splitAlignArray(align: string, n: number = Infinity) {
```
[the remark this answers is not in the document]
dpvc: Fixed

ts/output/common/Wrappers/mrow.ts
dpvc: Done

PR #929, opened 2023-03-16
Don't treat \overrightarrow as a mathaccent.  (mathjax/MathJax#3010)

dpvc:
This PR prevents `\overrightarrow` and similar macros from treating the stretchy character as a math accent (which doesn't contribute to the width even if it extends beyond the base).  The arrow should count in the width when the base is smaller than the arrow.

Resolves issue mathjax/MathJax#3010.


zorkow (approved):
lgtm.

PR #930, opened 2023-03-16
Handle mrow with class INNER as a unit.  (mathjax/MathJax#3018)

dpvc:
This PR fixes a problem where an `mrow` that comes from `\left...\right` isn't treated as TeX class `INNER` by the item that follows it.  The fix is to return the `mrow` itself (not the last child of the `mrow`) in the `setTeXclass()` method.

For example, the spacing is wrong for `a\left(b\right)c` without this PR.

Resolves issue mathjax/MathJax#3018


zorkow (commented):
lgtm.

zorkow (approved):

PR #931, opened 2023-03-16
Add nulldelimiterspace for null delimiters.  (mathjax/MathJax#985)

dpvc:
This PR causes null delimiters to get the font's `nulldelimiterspace` that TeX specifies should be used in place of the null delimiter.  We identify a null delimiter as one whose width is 0 (easy first check that will rule out most delimiters), whose text is blank, show TeX class is OPEN or CLOSE, and that has an explicit `fence` attribute (which will have come from the `\left..\right` processing).  We add the extra `nulldelimiterspace` to the box on the right.

Resolves issue mathjax/MathJax#985


zorkow (approved):
lgtm.

PR #932, opened 2023-03-17
Properly handle arrays with dashed frame borders. (mathjax/MathJax#2972)

dpvc:
This PR allows tables to have frames where some sides are dashed rather than solid.  In the past, if all four sides of the table frame were included, they all had the style of the left edge (solid or dashed).  If not all the sides where included, then they were all solid.  Now that border CSS styles are implemented in both output jax, we use dashed or solid borders when needed.  When all the borders are included and the same style, the `frame` attribute is used, as before.  If not all the sides are included, but they are all solid, we use the `menclose` element with the proper sides, as before.  Otherwise (mixed styles, or all are dashed but not all side are included), we use a new `data-frame-style` to record the styles.  These are converted to real styles in the common output jax.  The reason to do it in the output jax is that it knows the width of the lines that it is using elsewhere, and I plan to allow that to be configurable; the TeX input jax has no knowledge of the output jax or its options, so can't set he proper border width itself, so we mark that borders should be used, and le the output jax take care of it.  This does mean that native MathML renderers will not have the border in that case (they will if they are all solid, but not in the mixed case).  One could use a TeX input jax post-filter to modify the `mtable` elements to have actual `style` attributes if one needed to have native MathML output, or one could use CSS targeting the `data-frame-style` attribute to do it (it would take 15 styles).  But I'm not that concerned about native MathML, since mathML core doesn't handle most of the `mtable` attributes anyway.

This PR also refactors some of the ColumnParser code to break up the `process()` method into smaller pieces (the diff doesn't do such a good job of making that clear), and includes fixes to remove some blank attributes (`columnlines` and `rowlines`) when there is a frame but no internal lines.

Resolves issue mathjax/MathJax#2972.

### DIFF
--- a/ts/input/tex/ColumnParser.ts
+++ b/ts/input/tex/ColumnParser.ts
@@ -43,8 +43,8 @@ export type ColumnState = {
   cwidth: string[];                     // the explicit column widths
   cspace: string[];                     // the column spacing
   clines: string[];                     // the column lines
-  cstart: string[];                     // the '>' declarations (not currently used)
-  cend:   string[];                     // the '<' declarations (not currently used)
+  cstart: string[];                     // the '>' declarations
+  cend:   string[];                     // the '<' declarations
   cextra: boolean[];                    // the extra columns from '@' and '!' declarations
   ralign: [number, string, string][];   // the row alignment and column width/align when specified
 }
@@ -206,11 +206,10 @@ export class ColumnParser {
    */
   protected setPadding(state: ColumnState, array: ArrayItem) {
     if (!state.cextra[0] && !state.cextra[state.calign.length - 1]) return;
+    const i = state.calign.length - 1;
     const cspace = state.cspace;
-    array.arraydef['data-array-padding'] = [
-      cspace[0] || '.5em',
-      cspace[state.calign.length - 1] || '.5em'
-    ].join(' ');
+    const space = (!state.cextra[i] ? null : cspace[i]);
+    array.arraydef['data-array-padding'] = `${cspace[0] || '.5em'} ${space || '.5em'}`;
   }
 
   /**
--- a/ts/input/tex/ColumnParser.ts
+++ b/ts/input/tex/ColumnParser.ts
@@ -131,60 +131,86 @@ export class ColumnParser {
       this.columnHandler[c](state);
     }
     //
-    // Set the column alignments
+    // Set array definition values
     //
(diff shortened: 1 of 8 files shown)


ts/input/tex/ColumnParser.ts
```diff
@@ -131,62 +131,88 @@ export class ColumnParser {
...
+   * @param {ColumnState} state   The current state of the parser
+   * @param {ArrayItem} array     The array stack item to adjust
+   */
+  protected setPadding(state: ColumnState, array: ArrayItem) {
+    if (!state.cextra[0] && !state.cextra[state.calign.length - 1]) return;
+    const cspace = state.cspace;
+    array.arraydef['data-array-padding'] = [
+      cspace[0] || '.5em',
+      cspace[state.calign.length - 1] || '.5em'
+    ].join(' ');
```
zorkow, line 213: This could be done more efficiently in a template literal:
```suggestion
    array.arraydef['data-array-padding'] = 
       `${cspace[0] || '.5em'} ${cspace[state.calign.length - 1] || '.5em'}`
```

ts/input/tex/base/BaseItems.ts
```diff
@@ -1085,6 +1058,45 @@ export class ArrayItem extends BaseItem {
...
+    // Map sides to their styles
+    //
+    const sides = new Map(this.frame);
+    //
+    // If all style are the same,
+    //   If all four sides are present, use a frame of the correct type
+    //   Otherwise, if the given sides are solid, use menclosed (with no padding)
+    //
+    const fstyle = this.frame.reduce((fstyle, [, style]) => style === fstyle ? style : '', this.frame[0][1]);
+    if (fstyle) {
```
zorkow, line 1076: Spread out over two lines.

zorkow (approved):
Some minor issues, o/w lgtm. `\hdashline` works nicely, too.

General remark: If I recall correctly, dashed lines are not a base feature of LaTeX `array` environments, but part of an extension package `ardshln (sic?)`. Would it be worthwhile to fully implement that at some point?

ts/input/tex/ColumnParser.ts
dpvc: Fixed.

ts/input/tex/base/BaseItems.ts
dpvc: Done.

PR #933, opened 2023-03-17
Remove Safari hack that is not needed with v4.  (mathjax/MathJax#3023)

dpvc:
This CSS hack is no longer needed in v4.  See issue mathjax/MathJax#3023.


zorkow (approved):
Nice!

PR #935, opened 2023-03-18
Use mspace for \allowbreak, etc., and fix issues with mspace breaking.

dpvc:
This PR changes `\allowbreak`, `\goodbreak`, etc. to use `mspace` elements rather than `mo` elements, since `mo` can become embellished, which can cause strange behavior.  In particular since `mtext` elements are space-like (because they can sometimes be all spaces, but the spec doesn't make an exception for when they contain non-space characters), that means that `\text{abc}\break\text{def}` would produce `<mtext>abc</mtext><mo linebreak="newline"/><mtext>def</mtext>`, and that would be an embellished operator, and the break would occur *outside* the embellished operator, so *before* the first `mtext`.  That is not what most people would intend.  Using `mspace` avoids that problem.

This has been a perennial problem with how `mtext` is treated as space-like even it is not all space characters.  I'm wondering if we should ignore the specification on this and make `mtext` only be space-like if it only contains whitespace characters?  I'm already ignoring the fact that `space` is only supposed to be breakable when its width is 0, and allow breaks at `mspace` even with non-zero width.

In any case, this also fixes a number of problems with breaking at `mspace` that turned up when trying to make this change:

* Separate out the check for whether `mspace` can be broken, to avoid duplicate code.
* Allow breaks when the width is zero (this was only partially correct).
* Fix problems with SVG inline breaks at `mspace`.
* Remove use of `data-allowbreak` attribute, which is not needed (now that we handle zero-width `mspace`)

### DIFF
--- a/ts/core/MmlTree/MmlNodes/mspace.ts
+++ b/ts/core/MmlTree/MmlNodes/mspace.ts
@@ -89,12 +89,22 @@ export class MmlMspace extends AbstractMmlTokenNode {
    * @override
    */
   public get hasNewline() {
+    const linebreak = this.attributes.get('linebreak');
+    return (this.canBreak && (linebreak === 'newline' || linebreak === 'indentingnewline'));
+  }
+
+  /**
+   * @return {boolean}   True if mspace is allowed to break, i.e.,
+   *                     no height/depth, no styles, and no background color.
+   */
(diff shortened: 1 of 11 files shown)


zorkow (approved):
lgtm.

PR #937, opened 2023-03-18
Handle multiple spaces in \text{}.

dpvc:
This PR reduces multiple spaces to a single space in `\text{}` and other math-mode material, as would happen in actual TeX, and simplifies the definition of `\ ` (since TeX doesn't seem to break expressions at this space, it can be teated as a non-breaking space in math mode).


zorkow (approved):
Lgtm.

However, looking at the code, I note that `substr` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr). Another thing we might want to clean up before releasing 4.0.

dpvc:
> Another thing we might want to clean up before releasing 4.0.

I already have it on my list.

PR #939, opened 2023-03-18
Add a tooltip for MathML verification short errors to show the full error.

dpvc:
This PR adds a tooltip to the `merror` element that gives the full error message when short errors are being used.  This brings it in line with the "Math input error" and "Math output error" messages, which show the full message as a tooltip.


zorkow (approved):
lgtm.

PR #940, opened 2023-03-19
Add checks for valid mathvariant

dpvc:
Because the mathvariant is critical to the operation of the output jax, having an incorrect math variant will lead to "Math output error" messages.  This PR adds checks within the MathML input jax and the common output jax to make sure that the variant exists in the output font in use.  This allows the output to be produced, but still gives the user information that something is wrong.

### DIFF
--- a/ts/core/MmlTree/MmlNode.ts
+++ b/ts/core/MmlTree/MmlNode.ts
@@ -828,7 +828,7 @@ export abstract class AbstractMmlNode extends AbstractNode<MmlNode, MmlNodeClass
     }
     if (options.checkMathvariants) {
       const variant = this.attributes.getExplicit('mathvariant') as string;
-      if (variant && !MATHVARIANTS.has(variant)) {
+      if (variant && !MATHVARIANTS.has(variant) && !this.getProperty('ignore-variant')) {
         this.mError(`Invalid mathvariant: ${variant}`, options, true);
       }
     }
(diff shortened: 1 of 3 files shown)


ts/core/MmlTree/MmlNode.ts
```diff
@@ -303,6 +315,7 @@ export abstract class AbstractMmlNode extends AbstractNode<MmlNode, MmlNodeClass
   public static verifyDefaults: PropertyList = {
     checkArity: true,
     checkAttributes: false,
+    checkMathvariants: true,
```
zorkow, line 318: Why do we need an extra option here? Could that not just be subsumed by `checkAttributes`?

ts/output/common/Wrapper.ts
```diff
@@ -812,7 +812,12 @@ export class CommonWrapper<
     if (!this.node.isToken) return;
     const attributes = this.node.attributes;
     let variant = attributes.get('mathvariant') as string;
-    if (!attributes.getExplicit('mathvariant')) {
+    if (attributes.getExplicit('mathvariant')) {
+      if (!this.font.getVariant(variant)) {
+        console.warn(`Invalid variant: ${variant}`);
+        variant = 'normal';
```
zorkow, line 818: Is `normal` the variant we really want? Even for single, Latin characters?

zorkow (commented):
A few questions.

zorkow (approved):

PR #941, opened 2023-03-19
Make promise-based call use startup.promise to avoid concurency

dpvc:
The promise-based calls like `MathJax.typesetPromise()` and `MathJax.tex2chtmlPromise()` currently shouldn't be called in parallel; that is, you should wait for the first promise to resolve before calling the second one.  Users don't always take this into account, and that can cause problems if the typesetting from the first call requires an extension to be loaded, since the second will run while the first waits for the extension.  That can cause problems with things like automatic equation numbering and macro definitions that must be done in sequence.  With the fact that fonts now can cause extensions to be loaded (for less-used groups of characters), this means the problem will be more severe in version 4.

This PR uses the existing `MathJax.startup.promise` to chain the promise-based calls so that they will always occur in the proper sequence, and later calls will wait for earlier ones to finish before they execute.

For example, in

```
MathJax.startup.promise.then(() => console.log(1));
MathJax.tex2chtmlPromise('\color{red}{x}').then(() => console.log(2));
MathJax.startup.promise.then(() => console.log(3));
MathJax.tex2chtmlPromise('y').then(() => console.log(4));
MathJax.startup.promise.then(() => console.log(5));
```

since the first `tex2chtmlPromise()` call loads the color extension, it will wait for that to occur, and the second `tex2chtmlPromise()` call will run before the first one completes.  So the output produced might be something like

```
1
3
4
5
2
```

(this was actually produced by the above code in a node application).

With this PR, the output will be

```
1
2
3
4
5
```

as expected.

To do this, the `typesetPromise()` code has been refactored into a separate function, since it must run for the initial typeset od the page *before* the `startup.promise` is resolved (since it initially resolves only after the first typeset pass).

This will require documentation changes.

### DIFF
--- a/ts/components/startup.ts
+++ b/ts/components/startup.ts
@@ -300,8 +300,19 @@ export namespace Startup {
   export function defaultPageReady() {
     return (CONFIG.loadAllFontFiles && output.font ? output.font.loadDynamicFiles() : Promise.resolve())
       .then(CONFIG.typeset && MathJax.typesetPromise ?
-            MathJax.typesetPromise(CONFIG.elements) as Promise<void> :
+            typesetPromise(CONFIG.elements) :
             Promise.resolve());
+  }
+
+  /**
+   * Perform the typesetting with handling of retries
+   */
+  export function typesetPromise(elements: any[]) {
+    document.options.elements = elements;
+    document.reset();
+    return mathjax.handleRetriesFor(() => {
+      document.render();
+    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


dpvc:
PS, because the v3-lab overrides `startup.ready()` with a function that never calls the `MathJax.startup.defaultReady()` function, which is where the `MathJax.startup.promise` is resolved, we have to do that by hand.  Add

```
MathJax.startup.promiseResolve();
```

right after [this line](https://github.com/mathjax/MathJax-dev/blob/dd2b3f28b06f7a105cd2a56ce2c2e89466a704a9/lib/v3-lab.js#L901).

zorkow (approved):
lgtm.

PR #942, opened 2023-03-22
Add SVG Image and Speech Text as Show As and Copy As menu items

dpvc:
This PR adds three new items to the `Show Math As` and `Copy to Clipboard` menu items:  "SVG Image", "Speech Text", and "Error Message".  They are enabled when the required data is available.  There is also a new item that allows filtering out the SRE semantic attributes when showing or copying the MathML format.

To do this, we store a copy of the speech text (if any) in the `MathItem.outputData` object, so the menu code can pick it up later.  The handling of SVG images is more complicated, as even ion the current output is SVG, we may not be able to use the existing SVG if it is in-line and in-line line breaking is allowed (as that is not a single SVG image), or if the `fontCache` is set to `global` (since that won't include the needed glyph path definitions).  So we need to be able to regenerate the SVG under some circumstances.

For that, we store a copy of the `Menu` object's `jax` list in the `MJContextMenu` object so that we can tell if the SVG output jax is available for regenerating the SVG (or in case the output is set to CHTML, it can still produce an SVG image).

The `MJContentMenu` object's `post()` method is where these menus items are enabled/disabled, and that has been refactored so that each menu gets its own function to enable/disable it.

A new variable is added in the `Menu` class in order to handle the filtering of semantic attributes, new SelectableInfo boxes for the SVG, Speech, and Error message output, and new commands for posting and copying these values.  The SelectableInfo copy function is modified so that it can copy the output of the new data (which is not always in a `pre` element).

The SVG output is a bit complicated to produce.  First, because SVG output can cause dynamic font ranges to be loaded, it has to use promises when it calls the SVG jax to generate new SVG.  But the SelectInfo box must be posted within the handler for the menu command that is being processed.  That means we need to post the info box before the SVG is ready, and must fill it in afterward.  A bit awkward, but when no dynamic loading occurs, you don't notice the difference.

If the existing SVG is acceptable for use (no inline breaks, and local or no font cache), then we use that for the output, otherwise, we need to regenerate it from the internal MathML using the SVG jax directly.  We do that by copying the internal MathML and removing any in-line break markers, and temporarily setting the font cache to `local` if it was `global` originally, then calling the SVG jax's `toDOM()` function to generate the SVG output.

In either case, the SVG output is modified to include an XML declaration `<?xml ...>` needed for stand-alone files, and the `<defs>` object is either inserted or modified to include the minimum CSS needed for the SVG output.  There result is then posted to the information box or copied to the clipboard.


zorkow (approved):
I am afraid merging PR #872 now leads to conflicts. Sorry about that. 
Otherwise just some minor issues.

PR #943, opened 2023-03-22
Don't include all of SRE in the menu component

dpvc:
This PR prevents the complete SRE package from being included in the menu component.  The original `import` cases that to occur.  The Menu object is unusual in that it has to load a11y components like the explorer on the fly, so we don't want to include them directly, here, or there will be two independent copies, one from the explorer component and one from the menu component.  It would work OK in combined components that include both, but not in the case where components are loaded separately.


zorkow (approved):
lgtm.

PR #945, opened 2023-03-23
Move handling of scrolling to CSS and allow tagged expressions to scroll

dpvc:
Normally, when an expression has equation numbers or tags, the `mjx-container` gets a minimum width wide enough to hold the equation with its numbers (to prevent the numbers from overlapping the equation).  But when the `overflow` attribute is `scroll` or `truncate`, we don't want the minimum width.  This PR moves the settings for scroll and truncate into CSS and overrides the `min-width` in that case.


zorkow (approved):
lgtm.

PR #946, opened 2023-03-24
Properly include prefix in build paths

dpvc:
This PR fixes a problem with the `prefix` option added in commit d34a9bb by making sure the prefix is included in the directory path when the `lib` files are created.


zorkow (approved):
lgtm.

PR #947, opened 2023-03-27
Update BaseMethods.ts

1217965058:
fix \ref macro quote the equation does not work properly


1217965058:
I find I'm wrong,it is Unnecessary

PR #948, opened 2023-03-27
Honor the rule_thickness and surd_height font parameters

dpvc:
This PR improves the handling of the `rule_thickness` and `surd_height` font parameters.  In the past, the `rule_thickness` was used where `surd_height` should be (TeX uses `rule_thickness` for both, but we allow them to be different), and the rule in fractions was not using `rule_thickness` since the MathML default for the `linethickness` attribute is `medium`, which was fixed at .06em.

Here, we modify the `length2em()` method of the `common/Wrapper.ts` to use the font's `rule_thickness` as the default for `medium` (and use percentages to get `thin` and `thick` values).  This allows `mfrac` to get the correct line thickness.  We also change the `msqrt` wrappers to use `surd_height` for the thickness of the line over the radicand.

It turns out that the results in CHTML is thinner than in SVG output, so we introduce a `rule_factor` to multiply the CHTML thicknesses by in order to get a comparable visual thickness.  Since this is in the font parameters, it can be adjusted to suit the font properly.

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -1178,11 +1178,8 @@ export class CommonWrapper<
       scale = this.bbox.scale;
     }
     const t = this.font.params.rule_thickness;
-    return lookup(length as string, {
-        medium: () => t,
-        thin: () => (2 / 3) * t,
-        thick: () => (5 / 3) * t
-    }, () => LENGTHS.length2em(length as string, size, scale, this.jax.pxPerEm))();
+    const factor = lookup(length as string, {medium: 1, thin: 2 / 3, thick: 5 / 3}, 0);
+    return factor ? factor * t : LENGTHS.length2em(length as string, size, scale, this.jax.pxPerEm);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


ts/output/common/Wrapper.ts
```diff
@@ -1177,6 +1177,16 @@ export class CommonWrapper<
...
+    const t = this.font.params.rule_thickness;
+    if (length === 'medium') {
+      return t;
+    }
+    if (length === 'thin') {
+      return (2 / 3) * t;
+    }
+    if (length === 'thick') {
+      return (5 / 3) * t;
+    }
```
zorkow, line 1189: Can you get the coefficient from a mapping of length to number and then  
`return coeff ? coeff * t : LENGTHS....` ?
While that means we get an extra multiplication on `length === 'medium'`, we avoid the linear complexity comparison.

zorkow (changes requested):
One rewrite to reduce complexity.

ts/output/common/Wrapper.ts
dpvc: OK, done.  Please check that this is OK.

dpvc:
@zorkow, can you re-review this?  Thanks!

zorkow (approved):
Sorry for not approving that one. I had looked at it before.

PR #950, opened 2023-03-31
Use promises when changing the renderer so that dynamic files can be loaded properly.

dpvc:
This PR introduces promises into the process of setting the output jax so that when the page is re-rendered, if other files need to be loaded (like autoloaded extension or font dynamic ranges), the processing will proceed properly.

The issue is described in the [MathJax User's Forum](https://groups.google.com/g/mathjax-users/c/9bhqmYfNGws), and is triggered by having an autoloaded macro (like `\verb` or `\bbox`) on the page, chaining the renderer using the MathJax contextual menu, then reloading the page.


zorkow (approved):
lgtm.

PR #951, opened 2023-04-07
Removes merge conflicts from package files.

zorkow:
Some cleanup of package files:

* Removes merge conflict markers that must have been introduced when `post-linebreaks` was merged.
* Updates to version number to 4.0.0-alpha.1
* Updates `mathjax-modern-font` to new beta.2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,24 +1,16 @@
 {
   "name": "mathjax-full",
-<<<<<<< HEAD
-  "version": "3.2.2",
-=======
   "version": "4.0.0-alpha.1",
->>>>>>> origin/post-linebreaks
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "mathjax-full",
-<<<<<<< HEAD
-      "version": "3.2.2",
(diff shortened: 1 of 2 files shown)


zorkow:
I am not sure if the version bump to alpha.1 should be included here as it was done outside the usual timeline, but I also don't think it matters much.

dpvc:
Thanks for catching the package-lock problem.  The post-linebreaks merge was problematic in a number of ways, and I guess I messed it up.

I'd like to make this version v4.0.0-beta.1 rather than alpha.1, since there is a separate branch that is alpha.1, and we talked about making this a beta release followed closely by the official 4.0 release.

I am going to need to make a beta.3 version of the mathjax-modern-font based on changes I've made since we talked, and will make beta.3 versions of all the fonts for the v4.0.0-beta.1 version, so I'm going to wait with this for now, and make an PR to it that makes those changes before merging it.  Is that OK?

dpvc:
On second thought, I'll make the PR for the alpha->beta change and merge this.  We can deal with the font version later.

PR #952, opened 2023-04-08
More reliable handling of stretchy assemblies

dpvc:
This PR improves the handling of multi-character assemblies for stretchy characters, mostly in CHTML output, but a small change in SVG.  It makes the horizontal and vertical stretchies work more consistently (horizontals used to be fake tables).  Now both are handled using extenders that are 100% (or 50% for assemblies with a middle piece like braces), with transparent borders of the proper size where the end pieces go.  The ends are in 0-height or width boxes and are positioned to overlap the borders properly.  When there is a middle piece, the two extenders are 50% width, and the borders have to cover half the middle piece (this wasn't being properly handled before).

For CHTML, if a combining character is used in the assembly, we run into the issue with some browsers handling non-zero-width combining characters as zero width while others don't, so must use the trick of right-aligning the character in a `mjx-c` node of the proper width that we introduced earlier for single-character accents.  That code was refactored in order to be able to reuse it here.

These changes allow the CHTML css styles to be simplified.


zorkow (approved):
lgtm.

PR #956, opened 2023-04-22
Remove forced breaks that SVG inserted when switching to CHTML

dpvc:
This PR removes the `force break` property that SVG uses to manage in-line line breaks so that when switching from SVG to CHTML, we don't get unwanted line breaks.

To test, use an in-line expression with potential line breaks using SVG output and use the menu to switch to CHTML.  Without this PR, the CHTML in-line expression will have forced breaks at each potential break.  With the PR, it should render properly in CHTML.


zorkow (approved):
lgtm.

PR #957, opened 2023-04-22
Work around WebKit bug that causes baseline alignment problems.  (mathjax/MathJax#2866)

dpvc:
This PR works around a bug in WebKit that causes text to not be aligned properly on the baseline.  it does so by forcing the heights of all characters in a given MathML node to be the same height.  In particular, for `\text{}`, the baseline should now be the same throughout.  Its ugly, but it works, at least within a single node.

Resolves issue mathjax/MathJax#2866.


zorkow (commented):
lgtm.

zorkow (approved):

PR #958, opened 2023-04-26
Handle dash during font-change in `\operatorname`. (mathjax/MathJax#3038)

dpvc:
This PR fixes a conflict with the usage of `multiLetterIdentifiers` between `\operatorname` and font-changing macros like `\mathsf`.  The solution is to preserve the `multiLetterIdentifiers` value if it exists when the font changes, so that an outer `\operatorname` setting will be inherited by an inner font change.

I've also added a check for a failing match of the `multiLetterIdentifiers`, though that really shouldn't happen (but did in this case because of the wrong pattern being used).

Resolves issue mathjax/MathJax#3038.

### DIFF
--- a/ts/input/tex/base/BaseConfiguration.ts
+++ b/ts/input/tex/base/BaseConfiguration.ts
@@ -192,7 +192,7 @@ export const BaseConfiguration: Configuration = Configuration.create(
     },
     options: {
       maxMacros: 1000,
-      identifierPattern: /^[a-zA-Z]+/,  // pattern for multLetterIdentifiers in \mathrm, etc.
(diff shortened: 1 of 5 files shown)


ts/input/tex/base/BaseConfiguration.ts
```diff
@@ -192,6 +192,7 @@ export const BaseConfiguration: Configuration = Configuration.create(
     },
     options: {
       maxMacros: 1000,
+      identifierPattern: /^[a-zA-Z]+/,  // pattern for multLetterIdentifiers in \mathrm, etc.
```
zorkow, line 195: typo in comment

zorkow (approved):
lgtm.

PR #959, opened 2023-05-22
Make it easier to find a specific input jax by name

dpvc:
This PR alters the `inputJax` array of the `MathDocument` to include named keys (as well was indices) for accessing the various input jax.  That way you can use `doc.inputJax.TeX` or `doc.inputJax.MathML` to access the needed input jax, rather than assuming it is `doc.inputJax[0]` or `doc.inputJax[1]`.  I leave the array as is, since it is itterable and some places loop through the jax.  Also, the output jax name is filtered when used to construct the URL for loading the jax when menu initialization causes an alternative jax to be loaded, for security.


zorkow (approved):
lgtm.

PR #960, opened 2023-05-22
Vertical spacing for horizontal stretchy characters more like TeX

dpvc:
This PR changes the vertical spacing for horizontal stretchy assemblies.  In TeX, these have the height and depth of the extender character, but in MathML, they are the maximum height and depth of any character in the assembly.  MathJax implements the latter, but this makes things like `\overrightarrow` have more height and separation from the base than TeX does.  In the TeX fonts, some of the characters used for extenders have height/depth that is different from their inked area in order to adjust the spacing, but the fonts used in MathJax don't have that (though potentially it could be added by hand).

This PR adds a font parameter that controls how MathJax uses the extender height/depth versus the total assembly's.  Its value is a linear interpolation value between the two values, defaulting to 1/3 of the extender's value and 2/3 of the total, which seems to do pretty well for the mathjax-modern and mathjax-stix2 fonts.  This factor is only used when `mathmlSpacing` is false; for MathML spacing, the fully assembly height is always used.

To test this, you need a new copy of the fonts that has the needed extender data, but I haven't made the npmjs packages for that, yet.  Let me know if you want to try it out, and I'll publish one.

### DIFF
--- a/ts/output/chtml/FontData.ts
+++ b/ts/output/chtml/FontData.ts
@@ -435,9 +435,20 @@ export class ChtmlFontData extends FontData<ChtmlCharOptions, ChtmlVariantData, 
    * @param {ChtmlDelimiterData} data  The data for the delimiter whose CSS is to be added
    */
   protected addDelimiterHStyles(styles: StyleList, n: number, c: string, data: ChtmlDelimiterData) {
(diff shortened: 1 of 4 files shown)


zorkow (approved):
lgtm.

PR #961, opened 2023-05-22
Properly handle italic correction for -tex-mit, and for subscripts with multi-character bases

dpvc:
This PR fixes spacing issues with the `-tex-mit` pseudo variant used for `\mathit` and `\it` by adding italic correction information in a property `oc` (other correction) that is only used by the last character in a string of characters in that variant.  That is, `\mathit{office}` and `{\it office}` will have the italic correction only on the last letter rather than between every letter, as in `office`, where the spacing between the f's is too wide, for example.  This allows superscripts and subscripts to be placed properly after such characters (e.g., `\mathit{f}^0` used to be too close).

This PR also fixes an inconsistency with TeX's placement of super and subscripts when the base is multi-character.  In TeX, `{UU}_0^1` has the scripts vertically aligned above each other, whereas in `U_0^1` they are offset to match the italic correction.  Before this patch, MathJax would use italic correction offsets in both places; after the patch, it works as in TeX.

This PR also requires a new copy of the fonts in order to include the needed `oc` data.  If you want to run it, I will publish a new npm package with the needed data.

### DIFF
--- a/ts/output/chtml/FontData.ts
+++ b/ts/output/chtml/FontData.ts
@@ -497,6 +497,9 @@ export class ChtmlFontData extends FontData<ChtmlCharOptions, ChtmlVariantData, 
     const font = options.ff || (letter ? `${this.cssFontPrefix}-${letter}` : '');
     const selector = 'mjx-c' + this.charSelector(n) + (font ? '.' + font : '');
     styles[selector] = {padding: this.padding(data, options.ic || 0)} as StyleData;
+    if (options.oc) {
+      styles[selector + ':last-child'] = {'padding-right': this.em(data[2] + options.oc)};
+    }
     this.checkCombiningChar(options, styles[selector]);
   }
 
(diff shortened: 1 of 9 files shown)


ts/output/chtml/FontData.ts
```diff
@@ -496,7 +496,10 @@ export class ChtmlFontData extends FontData<ChtmlCharOptions, ChtmlVariantData,
     const letter = (options.f !== undefined ? options.f : vletter);
     const font = options.ff || (letter ? `${this.cssFontPrefix}-${letter}` : '');
     const selector = 'mjx-c' + this.charSelector(n) + (font ? '.' + font : '');
-    styles[selector] = {padding: this.padding(data, options.ic || 0)} as StyleData;
+    styles[selector] = {padding: this.padding(data, options.oc || options.ic || 0)} as StyleData;
```
zorkow, line 499: Pull this disjunction out into a `const` for readability.

zorkow (approved):
One small thing.

PR #962, opened 2023-05-22
Properly place super- and subscripts for stretched operators. (mathjax/MathJax#3049)

dpvc:
This PR fixes a problem where superscripts and subscripts aren't placed properly on an operator that has been stretched.  There is an example in the initial issue

Resolves issue mathjax/MathJax#3049.


zorkow (approved):
lgtm.

PR #964, opened 2023-06-02
Es6 modules sre fixes

zorkow:
* Fixes the es modules paths for sre
* Updates to sre v4.1.0-beta.5
* Updates the `package-lock` file.
    *  In particular there are references to the `font-tools`, which I assume stem from you local copy. This is now removed
    * Some of the other modules have updates to their minor versions.
In case there is a problem with the `package-lock`, I can make the PR against the first commit, which does not contain those changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,78 +30,34 @@
         "webpack-cli": "^4.9.2"
       }
     },
-    "../font-tools": {
-      "name": "@mathjax/font-tools",
-      "version": "0.1.0",
-      "extraneous": true,
-      "license": "Apache-2.0",
-      "dependencies": {
-        "mathjax-full": "^3.2.0"
-      },
-      "devDependencies": {
-        "@types/node": "^12.12.31",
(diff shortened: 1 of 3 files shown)


dpvc:
This looks good.  I see that the package-lock still included babel stuff, which I removed from package.json as it is no longer needed.  So I suppose it might be good to delete the package-lock file and rebuild it from npm install so that it will not have any old baggage.

dpvc (approved):

zorkow:
> This looks good. I see that the package-lock still included babel stuff, which I removed from package.json as it is no longer needed. So I suppose it might be good to delete the package-lock file and rebuild it from npm install so that it will not have any old baggage.

I did just that, but `babel` is still in there. According to ```npm list --depth=4``` the `tslint` package depends on that. We should be able to get rid of it, once we switch to `eslint`.

dpvc:
OK, thanks for looking into the babel thing.

PR #965, opened 2023-06-03
Update dev dependencies.  Fix new type warnings from tsc v5

dpvc:
This PR updates the dependencies to their latest versions, and in particular updates to Typescript v5 (in order to handle the `moduleResolution` of `nodenext` that we set in the ES6-modules branch).  This update causes a few warnings about comparing a number to a value of type `number | string`, so the other changes are to handle those situations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,17 +17,17 @@
       },
       "devDependencies": {
         "copyfiles": "^2.4.1",
-        "diff": "^5.0.0",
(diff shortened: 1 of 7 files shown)


dpvc:
You should probably view the file changes with shitespace changes hidden, since the `collapse.ts` file has had the indenting changed.

zorkow (approved):
lgtm.

PR #966, opened 2023-06-15
Update use of skew values from accents to allow better placement. (mathjax/MathJax#3051)

dpvc:
This PR adjusts how some accents are placed.  The skew values stored in the font are used to move an accent to its proper position over a base character (e.g., to move the accent over the point in *A* rather than over the center of the width of the *A*).  These values re being taken from the Opentype MATH table accent data, but that table specifies that both the base skew and the accent skew should be used (these two skewed positions are to be aligned).  But currently, MathJax only uses the base skew.  This PR adds the ability to use the accent skew as well.  It also fixes an error where the skewed position for the top accent was also used for the bottom character in `munderover` causing the bottom to the improperly shifted to the left.

his makes it possible to resolve issue mathjax/MathJax#3051.

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -741,13 +741,11 @@ export class CommonWrapper<
    */
   protected copySkewIC(bbox: BBox) {
     const first = this.childNodes[0];
-    if (first) {
-      if (first.bbox.sk) {
-        bbox.sk = first.bbox.sk;
-      }
-      if (first.bbox.dx) {
-        bbox.dx = first.bbox.dx;
-      }
+    if (first?.bbox?.sk) {
+      bbox.sk = first.bbox.sk;
+    }
+    if (first?.bbox?.dx) {
+      bbox.dx = first.bbox.dx;
     }
     const last = this.childNodes[this.childNodes.length - 1];
     if (last?.bbox?.ic) {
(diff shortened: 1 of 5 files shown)


ts/output/common/Wrapper.ts
```diff
@@ -741,14 +741,16 @@ export class CommonWrapper<
    */
   protected copySkewIC(bbox: BBox) {
     const first = this.childNodes[0];
-    if (first?.bbox.sk) {
-      bbox.sk = first.bbox.sk;
-    }
-    if (first?.bbox.dx) {
-      bbox.dx = first.bbox.dx;
+    if (first) {
+      if (first.bbox.sk) {
```
zorkow, line 745: Will `bbox` exist here, since you explicitly check for it later in line !753?

I don't think we gain much with the `if` in line 744. It would  look considerably cleaner otherwise.

zorkow (approved):
One concern and a style question. O/w good.

ts/output/common/Wrapper.ts
dpvc: I guess line 744 should be `if (first?.bbox) {` so that you don't need the question mark check repeated in the other two.  But I can change it to doing that twice, as `first?.bbox?.sk` and `first?.bbox?.dx` if you prefer.

PR #968, opened 2023-06-20
Fix memory leak in textmacros package.  (mathjax/MathJax#3061)

dpvc:
This PR fixes a memory leak that occurs when the `textmacros` extension is used.  It was caused by not clearing the node lists that are part of the `ParseOptions` object, and not removing the parsers from the `parsers` list when they are done.  That means these will continue to build up over time.  We fix the problem by clearing these lists when the `mml()` function is called, and clearing the configuration itself when `internalMath` first starts parsing text-mode material.

In addition, since we are using a separate `ParseOptions` object, and the node lists are stored there, the nodes created for text elements are not being added to the main `TexParser` node lists, so we add code to copy the lists over from our parser to the parent TeX parser.

Resolves issue mathjax/MathJax#3061.

### DIFF
--- a/ts/input/tex/textmacros/TextParser.ts
+++ b/ts/input/tex/textmacros/TextParser.ts
@@ -98,9 +98,8 @@ export class TextParser extends TexParser {
    */
   protected copyLists() {
     const parseOptions = this.texParser.configuration;
-    const lists = this.configuration.nodeLists;
-    for (const name of Object.keys(lists)) {
-      for (const node of lists[name]) {
+    for (const [name, list] of Object.keys(this.configuration.nodeLists)) {
+      for (const node of list) {
         parseOptions.addNode(name, node);
       }
     }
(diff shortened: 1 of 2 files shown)


ts/input/tex/textmacros/TextParser.ts
```diff
@@ -86,11 +86,27 @@ export class TextParser extends TexParser {
...
             this.nodes.length === 1 ? this.nodes[0] : this.create('node', 'mrow', this.nodes));
   }
 
+  /**
+   * Copy the node list from the text parser to the TeX parser
+   */
+  protected copyLists() {
+    const parseOptions = this.texParser.configuration;
+    const lists = this.configuration.nodeLists;
+    for (const name of Object.keys(lists)) {
```
zorkow, line 102: Now that we are using `ES2020` in `tsconfig` you could just use 

```JavaScript
for (const [name, list] of Object.entries(list)) {
```

zorkow (approved):
One suggestion. O/w lgtm.

zorkow (approved):
One suggestion. O/w lgtm.

PR #970, opened 2023-07-11
Upgrades to latest version of SRE.

zorkow:
Upgrades to `SRE v4.1.0-beta.7`. I had to deprecate `beta.6` as the `prepublish` scripts did not work as expected and it was published without the `mathmaps`.
* Reroutes `commonjs` sources to the correct directory.
* A little fix to avoid Euro Braille showing up as a speech locale.
* Otherwise we now have Korean!


dpvc:
I added a fix to `components/bin/package.json`, which needed the `cjs` directory change as well (it is copied to other directories when commonjs format is needed).  I will merge.

PR #971, opened 2023-07-14
Allow inline-breaking within top-level nodes added by SRE

dpvc:
This PR allows `mrow` nodes that were added by SRE to include in-line line breaks, so that TeX output will work as expected even after enrichment.  For example,

```
\(f(x+y+z+x+y+z+x+y+z+x+y+z+x+y+z+x+y+z)\)
```
should allow line breaks when it appears in a container with a small width.

We also fix some problems with SVG output when the top-level element is an `mstyle` or `semantics`, as in

```
\(\large x+y+z+x+y+z+x+y+z+x+y+z+x+y+z+x+y+z\)
```

When an `mstyle`, `semantics`, or `mrow` with `data-semantics-added` is encountered during in-line break processing, it is marked with the `inline-breaks` property, and that is used in CHTML output to add an attribute that is used to set `display: inline` in the CSS so that the breaks will be able to take effect (the default `inline-block` doesn't allow the breaks).  

The common in-line breaking code mishandled a potential breakpoint that has `linebreakstyle="after"`, and that has been fixed, here (by changes to the `markNext` variable).  The operator with `linebreakstyle="after"` doesn't need to be marked for a breakpoint (this lead to an empty `svg` node in the SVG output that was getting in the way of determining the spacing for the first svg in the next line).  That means we don't need the `newline` variable any longer in the SVG code that processes the breakpoints, as that was trying to compensate for the incorrect break for the `linebreakstyle="after"`. 

We also properly scale the SVG breakpoints, so the `\large` example above has the correct spacing.

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -626,9 +626,14 @@ export class CommonWrapper<
    */
   public getBreakNode(bbox: LineBBox): [WW, WW] {
     const [i, j] = bbox.start || [0, 0];
-    if (this.node.isEmbellished) return [this, this.coreMO()] as any as [WW, WW];
-    if (this.node.isToken || !this.childNodes[i]) return [this, null] as any as [WW, WW];
-    return this.childNodes[i].getBreakNode(this.childNodes[i].getLineBBox(j));
+    if (this.node.isEmbellished) {
+      return [this, this.coreMO()] as any as [WW, WW];
+    }
+    const childNodes = (this.childNodes?.[0]?.node?.isInferred ? this.childNodes[0].childNodes : this.childNodes);
+    if (this.node.isToken || !childNodes[i]) {
+      return [this, null] as any as [WW, WW];
+    }
+    return childNodes[i].getBreakNode(childNodes[i].getLineBBox(j));
   }
 
   /**
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -629,7 +629,7 @@ export class CommonWrapper<
     if (this.node.isEmbellished) {
       return [this, this.coreMO()] as any as [WW, WW];
     }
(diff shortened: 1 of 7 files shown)


ts/output/common/Wrapper.ts
```diff
@@ -626,9 +626,14 @@ export class CommonWrapper<
    */
   public getBreakNode(bbox: LineBBox): [WW, WW] {
     const [i, j] = bbox.start || [0, 0];
-    if (this.node.isEmbellished) return [this, this.coreMO()] as any as [WW, WW];
-    if (this.node.isToken || !this.childNodes[i]) return [this, null] as any as [WW, WW];
-    return this.childNodes[i].getBreakNode(this.childNodes[i].getLineBBox(j));
+    if (this.node.isEmbellished) {
+      return [this, this.coreMO()] as any as [WW, WW];
+    }
+    const childNodes = (this.childNodes?.[0]?.node?.isInferred ? this.childNodes[0].childNodes : this.childNodes);
```
zorkow, line 632: What if `this.childNodes` does not exist? According to the test, this is a possibility. This might throw an error in the next condition.

ts/output/svg.ts
```diff
@@ -264,6 +264,7 @@ CommonOutputJax<
    * @param {number} h   The height of the SVG to create
    * @param {number} d   The depth of the SVG to create
    * @param {number} w   The width of the SVG to create
+   * @return [N, N]      The svg element and its initial g child
```
zorkow, line 267:
```suggestion
   * @return {[N, N]}      The svg element and its initial g child
```

zorkow (approved):
Two small things. O/w lgtm.

ts/output/common/Wrapper.ts
dpvc: It actually always will exist; I got too aggressive with the `?`.

PR #972, opened 2023-07-15
Fix bussproofs

zorkow:
This is a fix for the bussproofs issue that occurs during abstract bounding box computation. The problem here is that some of the `mrow` elements introduced for measuring width will not have a parent. So I introduced two guards for methods that throw errors otherwise.

I assume setting `isStack` to `true` is reasonable, as for the time being, we don't really want line breaking in proof tree expressions.


dpvc (approved):
Looks good.  I will merge.

PR #974, opened 2023-07-19
Update version to beta.3

dpvc:
This PR includes several changes needed for version 4.0.0-beta.3.  These fixes include:

* Using the correct root path for webpacked node-main.js (for when you use `require("mathjax")` or `import "mathjax"` or the same for `mathjax-full`.
* Replacing `mjs/root.js` in the webpacked combined configuration files with a new `components/root-pack.mjs` that does the same thing as `__dirname` does in CommonJS webpacking (it replaces it with the directory `/`).  This is not ideal, but it prevents the inclusion of the directory name where the webpacked file was created (e.g., `/home/travis/mathjax/MathJax-src/...`).  The v3 versions had a broken default directory in node as well, so this is no worse than in the past.
* Fixing some comments (in particular some copyright dates that should not have been ranges).

### DIFF
--- a/components/mjs/node-main/node-main.js
+++ b/components/mjs/node-main/node-main.js
@@ -58,6 +58,8 @@ if (path.basename(dir) === 'node-main') {
     return REQUIRE(name.charAt(0) === '.' ? path.resolve(ROOT, name) :
                    name.charAt(0) === '[' ? Package.resolvePath(name) : name);
   };
+} else {
+  CONFIG.paths.mathjax = dir;
 }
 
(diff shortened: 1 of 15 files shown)


zorkow (approved):
Please run `npm install` before merging, so `package-lock.json` will also be updated to `beta.3`.

dpvc:
Yes, I'm in the process of building the beta.3 font packages, and was going to update the package-lock.json after they are in place, but you beat me to it.  Almost there.

PR #976, opened 2023-07-24
Prettier and lint

zorkow:
This is the first formatting change, just using prettier. There are a few things that you might find shocking, for instance, how it breaks up formatting of comments (e.g, in `output/common.ts`).

Anyway, have a look. I am currently linting, working through errors that are easy to resolve like comments.


dpvc (changes requested):
I didn't look at every file, but took a sampling and commented on a few things in hopes that there is some level of control that can be used to adjust them.  But if not, we can go with how this is.  I wonder if the Emacs typescript mode can be configured to match this format?

The only important change I had was in the package.json file.

zorkow:
Now superseded by PR #1123

PR #982, opened 2023-08-11
Change Symbol to Token to avoid conflict with native Symbol in webpacked files. (mathjax/MathJax#3072)

dpvc:
This PR changes the `Symbol` class name to `Token` so that it doesn't conflict with the native `Symbol` object in webpacked files.  Webpacking of ESM files handles the exports from the component `lib` files as global variables, and that means the `Symbol` object will overwrite the native one, and since the webpack initialization code refers to `Symbol` before that `const Symbol` is set up, that cause a compile error.  (The old CommonJS `modules.export` approach didn't use separate global variables, so didn't have that problem.)

Along with this, it renames `SymbolMap` to `TokenMap`, and changes variables called `symbol` to `token`, where appropriate.

Unfortunately, this touches a lot of files.

Resolves issue mathjax/MathJax#3072.


zorkow (approved):
I did not review all files in detail. I built, tested in the lab and with my own test page. It works fine.

PR #983, opened 2023-08-11
Fix bug with \text{} in textmacros package. (mathjax/MathJax#3082)

dpvc:
This PR fixes a problem introduced by a last minute change in respond to PR review where I didn't change `keys()` to `entries()` as I should have, and apparently failed to test the result.  Argh!

Resolves issue mathjax/MathJax#3082.


zorkow (approved):
I wonder why TS did not catch that?

dpvc:
> I wonder why TS did not catch that?

I don't know, but am glad Peter found it.

PR #985, opened 2023-08-14
Remove install script

### LINKED ISSUE #984: installation of v4 not working with yarn
Hi MathJax developers,

Thanks for your constant work and improvement on this awesome software.

I'm currently using version 3.2.2 and my first attempt to migrate to version 4 failed already during installation with yarn: `yarn add mathjax-full@4.0.0-beta.3` (I don't have npm installed and using pnpm might also not work).

A quick look into `package.json` shows that you added a custom `install` script (`npm run -s link:full`) that relies on `npm`.
Looking a bit further into the script, it adds the symlink `node_modules/mathjax-full` to the source code folder if there is the `node_modules` folder. This seems to be necessary only during development with source code from git but not if I want to load mathjax-full from a package repository because in this case the package will be install anyways into the `node_modules/mathjax-full` folder.

Would it be possible to remove the custom `install` script and then call `npm run -s link:full` in your dev environment? I see that `.travis.yml` is also using `npm install`. In that case a call to `npm run -s link:full` after the default install should work too.

Kind regards,
Marcel

MarcelBolten:
This PR will address and close #984.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ sudo: false
 script:
 - components/bin/version
 - npm install
+- npm run -s link:full
 - npm run -s build-all
 branches:
   only:
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,7 +7,6 @@
     "": {
       "name": "mathjax-full",
       "version": "4.0.0-beta.3",
-      "hasInstallScript": true,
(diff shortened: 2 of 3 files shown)


dpvc:
Thanks for reporting the issue and for the fix.

PR #986, opened 2023-08-14
Feature: Adding LaTeX subexpressions into the tree

zorkow:
The PR computes and adds corresponding LaTeX commands to subexpressions in a `latex` attribute. This allows SRE to generate Euro Braille.

* We mainly keep pointers to the start and end of a command in the main parser.
* For environments and commands we use two auxiliary attributes to collate LaTeX expressions: `latex` and `itemLatex` in `BaseItem`. The latter retains information on the current command or environment, the former can be overwritten when more of the actual LaTeX command is completed.
* Care has to be taken when computing sub- and superscripts.

**Please note the following:**

* I have left the `latex` attribute in place for the time being as this is what the current version of SRE expects. We can update that to `data-mjx-latex` with the next version of SRE.
* I have also changed default Braille setting from Nemeth to Euro for better testing. However, we can change that back before the next release.


zorkow:
I also note that the PR contains a number of odd commits from @dpvc . Maybe because I branched this off of the alpha release?

dpvc:
I've made a new branch `sre-latex-braille-ext` (using `-` rather than `_` so you can tell it is from me) that is based off of the current develop branch, and where I cherry-picked the individual entries from the original branch.  This takes the 4.0.0-alpha branch out of the picture, eliminating the unwanted commits from that branch.  If I merge develop into the original SRE branch, then the results show no differences, so I think I did it OK, but you might want to check it to make sure that it is correct.  If you are happy with is, make a PR based off that branch, and close this one.

I suspect we will need to do the same for the explorer rewrite branch, since it has also merges 4.0.0-alpha (and the original SRE LaTeX branch), in order to get a clean PR onto the new SRE branch.

zorkow:
Superseded by PR #1004

PR #988, opened 2023-08-15
Remove incorrect frame from multline environment. (mathjax/MathJax#3083)

dpvc:
This PR fixes a problem with the `multline` environment where it was incorrectly getting a frame.  That was due to changes with how the `frame` attribute was being handled, and the fact that I had used an empty `frame` attribute to add the `framespacing` without a frame.  That was a misuse of the `frame` attribute, which I knew would come back to bite me, and now it did.

The fix is to change `multline` to use `data-array-padding` instead of `framespacing` for the spacing.  This also requires a fix in the CHTML output to properly handle that change as well.

Resolves issue mathjax/MathJax#3083.


zorkow (approved):
lgtm.

PR #989, opened 2023-08-15
Make sure \operatorname stays an OP after enrichment. (mathjax/MathJax#3084)

dpvc:
This PR makes sure that the result of `\operatorname{}` keeps TeX class OP after a round-trip through enrichment by removing the `noAutoOP` from its definition.  I'm not sure why it was there originally, but isn't needed.

Resolves issue mathjax/MathJax#3084.


zorkow (approved):
lgtm.

PR #990, opened 2023-08-15
Fix spacing before \vcenter, \vtop, \vbox.  (mathjax/MathJax#3085)

dpvc:
This PR fixes the spacing before `\vcenter`, `\vtop`, and `\vbox` macros.  The problem was that the previous element's class wasn't being properly recorded, so nob spacing was being computed for it (the TeX spacing is based on the TeX classes of adjacent pairs of nodes).

Resolves issue mathjax/MathJax#3085


zorkow (approved):
lgtm.

PR #991, opened 2023-08-17
Remove errant mjx-linestrut from SVG output. (mathjax/MathJax#3087)

dpvc:
This PR removes the code that adds `mjx-linestrut` node to embellished operators in SVG output for in-line math with line-breaks allowed.  It seems to have been left over when copying the CHTML output `toEmbellishedCHTML()` to make the SVG version.

Resolves issue mathjax/MathJax#3087.


zorkow (approved):
lgtm.

PR #992, opened 2023-08-22
Include default font configuration so that it will be loaded when input/chtml or input/svg is loaded

dpvc:
This PR adds the default font to the default configuration for `node-main` so that when an output jax is loaded, it will load the default font automatically (rather than the "no-font" that is built into the `input/chtml` and `input/svg` components).  This allows you to do

``` js
require("mathjax-full").init({loader: {load: ['input/tex', 'output/svg']}}).then(...)
```

without having to include configuration for the font by hand.  You can still override the font by setting the `loader.paths` and `output.font` values in the argument to `init()` if you want to use a different font.


zorkow (approved):
lgtm.

PR #993, opened 2023-08-22
Prevent file system paths from being included in mml3 extension.

dpvc:
Webpack will replaced `import.meta.url` by a literal string for the path of the file on the system where it is being run.  That means files that include `import.meta.url` will have references to the host file system where webpack is run (in our case, the travis-ci servers).  The `beta.3` version worked around this issue in the `ts/components/loader.ts` file by replacing `#root/root.js` with one that uses `/` rather than `import.meta.url`; but `ts/input/mathml/mml3` also includes references to `import.meta.url`, and so the `mml3` extension was still incorrectly including the full path name.

This PR fixes that problem by having the `mml3` extension call on `#root/root.js` to get its root directory and making a path relative to that, rather than trying to look up the equivalent to `__dirname` using `import.meta.url`.  That means, in particular, that we don't need the `#mml3` pseudo-module only longer, so that is removed from the `package.json` and `tsconfig` files that included it.

In addition, we modify the `root-pack.js` file (that replaces `#root/root.js` when webpack is used) to look for a MathJax configuration `MathJax.config.loader.paths.mathjax` or `MathJax.config.__dirname` (the latter being set up by `node-main`) in order to try harder to find a good value for the root directory when used from webpacked files.  Also, the `esRoot()` function was renamed `mjxRoot()`, as that is more meaningful.

Finally, the `mml3.serf.json` file also included full file pathnames based on my own file hierarchy, so the package scripts used to create it have been modified to end up using `/tmp` rather than my full path to MathJax-src, so at least that is more generic.


zorkow (commented):
I did not get it to work in the lab. Is there anything I need to do?

dpvc:
> I did not get it to work in the lab. Is there anything I need to do?

Hmmm.  None of this should affect the usage in the lab (it is all for node-based use).  It works in the lab for me, at least for this simple test case:

``` xml
<math><mstack><mn>123</mn></mstack></math>
```

Can you say more about what is going wrong for you?

zorkow:
That was a local issue. I rebuilt everything and it now works in the lab.

zorkow (approved):
lgtm.

PR #998, opened 2023-08-27
Set TeX class OP for multi-letter mo elements, as in v2.  (mathjax/MathJax#3095)

dpvc:
In version 2, an `mo` with multiple letters, like `<mo>Hom</mo>`, would be given the TeX class of OP when it doesn't appear explicitly in the operator table.  This was lost in the v3 translation, and this PR reinstates that.  In addition, it refactors the code that does the checking, since it is used in two places (the `get texClass()` getter, and the `checkOperatorTable()` initialization routine.  It also adjusts the `getRange()` function in the `OperatorTable.ts` to always return a value.

The pattern used to identify a multi-letter operator is a property of the `MmlMo` object, so that it can be adjusted during initialization, if desired.

Resolves issue mathjax/MathJax#3095.


zorkow (approved):
lgtm.

PR #999, opened 2023-08-30
Use superscript for scriptChild in msubsup.  (mathjax/MathJax#3097)

dpvc:
This PR fixes a problem with the placement of `msubsup` scripts when when the depth of either script is large.  The problem was due to the `getU()` method using `this.scriptChild`, where `sciptChild` was always set to `childNodes[1]`; for `msubsup` nodes, it should be `childNodes[2]`, the superscript.  The solution is to override `scriptChild` in the `CommonMsubsupMixin` wrapper.  (I also cached the superscript depth and subscript height values that are used in several places).

The issue would show up if the superscript has large depth, as in 

```
$$\int_0^{\Rule{3pt}{0pt}{20pt}}$$
```

which produces

bad-int

or if the subscript has large depth, as in

```
$$x_{\Rule{3pt}{0pt}{20pt}}^2$$
```

which produces

bad-x

With this PR, both produce the proper results:

good

Resolves issue mathjax/MathJax#3097.

### DIFF
--- a/ts/output/common/Wrappers/msubsup.ts
+++ b/ts/output/common/Wrappers/msubsup.ts
@@ -415,6 +415,13 @@ export function CommonMsubsupMixin<
     /**
      * @override
      */
+    public get scriptChild(): WW {
+      return this.supChild;
+    }
+
+    /**
+     * @override
+     */
     public getUVQ(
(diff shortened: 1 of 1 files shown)


zorkow (approved):
lgtm.

PR #1000, opened 2023-09-11
Propery handle placement of scripts on \vcenter, \vbox, and \vtop (mathjax/MathJax#3091)

dpvc:
This PR fixes a problem where super- and subscripts were not placed properly on the results of `\vcenter`, `\vbox`, and `\vtop` elements.  These are implemented using `mpadded` nodes with `data-vertical-align` attributes and a `vbox` property to mark them as vboxes.  The problem is that the `getBaseCore()` method should stop at an `mpadded` element being used as a vbox, but was continuing down into it incorrectly if there was only one child element.  The fix is to stop when the `mpadded` has the `vbox` property.

Resolves issue mathjax/MathJax#3091.


zorkow (approved):
lgtm.

PR #1001, opened 2023-09-14
Prevent setting a propery that has a getter.  (mathjax/MathJax#3098)

dpvc:
This PR resolves a problem that occurs when `combineConfig()` or `combineWithMathJax()` are used to merge an object into the `MathJax._` object and the item to be set is already set and has a getter.  This occurs due to the way exports are handled in ESM modules, where the exports are in an object each with a getter (so that they can't be modified, as imports are supposed to be immutable).  That didn't seem to be the case for the CJS components we used to use.  As a consequence, changing the renderer now causes an error since both the CHTML and SVG components include the common output code, and both set the properties of `MathJax._.output.common`, which are all getters.  This produces errors.

Resolves the console error issue from mathjax/MathJax#3098.

### DIFF
--- a/ts/components/global.ts
+++ b/ts/components/global.ts
@@ -78,18 +78,21 @@ export function isObject(x: any): boolean {
  * Combine user-produced configuration with existing defaults.  Values
  * from src will replace those in dst.
  *
- * @param {any} dst      The destination config object (to be merged into)
- * @param {any} src      The source configuration object (to replace defaul values in dst}
- * @return {any}         The resulting (modified) config object
+ * @param {any} dst         The destination config object (to be merged into)
+ * @param {any} src         The source configuration object (to replace defaul values in dst}
+ * @param {boolean} check   True when combining into MathJax._ to avoid setting getter properties
+ * @return {any}            The resulting (modified) config object
  */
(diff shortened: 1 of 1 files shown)


ts/components/global.ts
```diff
@@ -78,18 +78,21 @@ export function isObject(x: any): boolean {
  * Combine user-produced configuration with existing defaults.  Values
  * from src will replace those in dst.
  *
- * @param {any} dst      The destination config object (to be merged into)
- * @param {any} src      The source configuration object (to replace defaul values in dst}
- * @return {any}         The resulting (modified) config object
+ * @param {any} dst         The destination config object (to be merged into)
+ * @param {any} src         The source configuration object (to replace defaul values in dst}
```
zorkow, line 82: typo

ts/components/global.ts
```diff
@@ -78,18 +78,21 @@ export function isObject(x: any): boolean {
...
+ * @param {boolean} check   True when combining into MathJax._ to avoid setting getter properties
```
zorkow, line 83: typo

zorkow (approved):
typos.
o/w lgtm.

PR #1002, opened 2023-09-14
Update check for spacelike mml nodes to be more sensible.  (mathjax/MathJax#3098)

dpvc:
This PR adjusts the way that `mspace` and `mtext` nodes are treated as spacelike.  The spec says they are always spacelike, but that causes all kinds of problems, like making `{\text{a}=\text{b}}` being treated as an embellished operator.  This interferes with line-breaking, for example, since a break at the equal sign will be moved to outside the text elements, which is not what is usually wanted.  Similarly, 

``` xml
<math>
  <mtext>a</mtext>
  <mspace linebreak="newline">
  <mo>=</mo>
  <mtext>b</mtex>
</math>
```

is also an embellished operator, and so treated as a single unit.

This PR makes `mtext` be spacelike only if it consists only of spaces, and doesn't have styles or background colors. Similarly, `mspace` will not be spacelike if it has an explicit `linebreak` attribute (or if it is not breakable due to other attributes).  This is not according to the spec, but makes more sense, particularly for how we use `mtext` and `mspace` in translating TeX to MathML.

### DIFF
--- a/ts/core/MmlTree/MmlNodes/mtext.ts
+++ b/ts/core/MmlTree/MmlNodes/mtext.ts
@@ -51,11 +51,18 @@ export class MmlMtext extends AbstractMmlTokenNode {
   }
 
   /**
-   * <mtext> is always space-like according to the spec
+   * <mtext> is always space-like according to the spec,
+   * but we make it so only if it contains only spaces and
+   * doesn't have certain attributes
+   *
    * @override
    */
   public get isSpacelike() {
-    return true;
+    const attributes = this.attributes;
+    const spaces = !!this.getText().match(/^\s*$/);
+    return spaces && (attributes.getExplicit('mathbackground') === undefined &&
+                      attributes.getExplicit('background') === undefined &&
+                      attributes.getExplicit('style') === undefined);
   }
 
 }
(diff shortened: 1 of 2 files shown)


ts/core/MmlTree/MmlNodes/mtext.ts
```diff
@@ -51,11 +51,18 @@ export class MmlMtext extends AbstractMmlTokenNode {
...
+   *
    * @override
    */
   public get isSpacelike() {
-    return true;
+    const attributes = this.attributes;
+    const spaces = !!this.getText().match(/^\s*$/);
+    return spaces && (attributes.getExplicit('mathbackground') === undefined &&
+                      attributes.getExplicit('background') === undefined &&
+                      attributes.getExplicit('style') === undefined);
```
zorkow, line 65: Should those attributes go into a dedicated constant, in case we need more in the future?

zorkow (approved):
One comment.
O/w fine.

ts/core/MmlTree/MmlNodes/mtext.ts
dpvc: There are a couple of places where this type of check is being made, and I was thinking about making a service function to check them.  I'll look into that.

dpvc: I have added `hasOneOf()` to the `Attributes` object to handle this, and made lists for the styles to use for `space` and `mtext` elements.  I also added a `hasExplicit()` method to the `Attributes` object to make checking for explicit attributes more natural, and `unset()` to make removing an attribute easier.  I edited a number of other files to use these new functions.  So there are a bunch of new edits here.  Do you want to review again?

PR #1003, opened 2023-09-14
Don't include extra linebreak positions after an explicit break.

dpvc:
This PR fixes a problem with CHTML output where in-line expressions with explicit line breaks could end up with extra vertical space at the explicit break if the part after it starts with an operator and is too wide for the container.  To do with, we mark the nodes with explicit linebreaks, and if we are just after an explicit line break, we don't mark an operator as breakable (preventing an extra line break from occurring right after an explicit one).  In the CHTML output, we don't insert breakpoints before explicit linebreaks (preventing an extra line break from occurring before that).

A test case is to use 

``` html
<div style="width:5em; border: 1px solid black">
\(a + \text{bbbbbbbbbbbbbbbbb} \\ = \text{cccccccccccccccc} + d\)
</div>
```

to see that there are no extra blank lines in CHTML output.

### DIFF
--- a/ts/output/chtml/Wrapper.ts
+++ b/ts/output/chtml/Wrapper.ts
@@ -260,7 +260,7 @@ CommonWrapper<
    */
   protected handleSpace() {
     const adaptor = this.adaptor;
-    const breakable = !!this.node.getProperty('breakable');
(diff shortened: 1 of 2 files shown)


zorkow (commented):
lgtm.

PR #1004, opened 2023-09-18
Sre latex braille ext

zorkow:
Properly rebased version of #986 .
Thanks to @dpvc !

### DIFF
--- a/ts/input/tex/TexParser.ts
+++ b/ts/input/tex/TexParser.ts
@@ -162,20 +162,54 @@ export default class TexParser {
     if (!str) {
       return;
     }
+    if (input === '\\') {
+      str = '\\' + str;
+    }
+    // TODO: Simplify. Add case for sub/sup sup/sub by checking number of children.
+    if (node.attributes.get('latex') === '^' && str !== '^') {
+      // console.log(53);
+      // console.log(node.childNodes.length);
+      node.childNodes.forEach(x => console.log(x));
+      if (str === '}') {
+        this.updateChild(node.childNodes[2], 0);
+        this.composeLatex(node, '^', 0, 2);
+        return;
+      }
+      node.childNodes[2].attributes.set('latex', str);
+      this.composeLatex(node, '^', 0, 2);
+      return;
+    }
+    if (node.attributes.get('latex') === '_' && str !== '_') {
+      // console.log(54);
+      // console.log(node.childNodes.length);
+      node.childNodes.forEach(x => console.log(x));
+      if (str === '}') {
+        this.updateChild(node.childNodes[1], 0);
+        this.composeLatex(node, '_', 0, 1);
+        return;
+      }
+      node.childNodes[1].attributes.set('latex', str);
+      this.composeLatex(node, '_', 0, 1);
+      return;
+    }
     if (str === '}') {
       this.updateChild(node, 0);
       return;
     }
-    if (input === '\\') {
-      str = '\\' + str;
-    }
     node.attributes.set('latex', str);
   }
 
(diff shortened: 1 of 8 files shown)


ts/input/tex/TexParser.ts
```diff
@@ -135,10 +141,12 @@ export default class TexParser {
    * @return {ParseResult} The output of the parsing function.
    */
   public parse(kind: HandlerType, input: ParseInput): ParseResult {
-    return this.configuration.handlers.get(kind).parse(input);
+    this.saveI.push(this.i);
+    let result = this.configuration.handlers.get(kind).parse(input);
+    this.updateResult(input[1], this.saveI.pop());
```
dpvc, line 146: I think you can do this without needing an array for `saveI`.  Since the value is pushed here and then popped below, with it only be used in one place later (and only to top value is used), you can use a local variable to hold the previous value, save the new value in `saveI` where it can be read later, and then put the local value it back afterward.  That is, you could do

``` js
  public parse(kind: HandlerType, input: ParseInput): ParseResult {
    const i = this.saveI;
    this.saveI = this.i;
    let result = this.configuration.handlers.get(kind).parse(input);
    this.updateResult(input[1], i);
    this.saveI = i;
    return result;
  }
```

The make the change to `Push()` listed below.

ts/input/tex/TexParser.ts
```diff
@@ -195,6 +203,12 @@ export default class TexParser {
    * @param {StackItem|MmlNode} arg The new item.
    */
   public Push(arg: StackItem | MmlNode) {
+    if (arg instanceof BaseItem) {
+      arg.startI = this.saveI.pop();
+      this.saveI.push(arg.startI);
+      arg.stopI = this.i;
+      arg.startStr = this.string;
+    }
```
dpvc, line 211: If you change the `saveI` as suggested above, then this would become

``` js
    if (arg instanceof BaseItem) {
      arg.startI = this.saveI;
      arg.stopI = this.i;
      arg.startStr = this.string;
    }
```

No need to pop and then push the value.

If you don't want to do that change, then at least use

``` js
    arg.startI = this.saveI[this.saveI.length - 1];
```

instead of popping and then re-pushing the same value.

ts/input/tex/TexParser.ts
```diff
@@ -66,6 +66,11 @@ export default class TexParser {
    */
   public currentCS: string = '';
 
+  /**
+   * A stack to save the string positions when we restart the parser.
+   */
+  private saveI: number[] = [0];
```
dpvc, line 72: If you make the changes I suggest to `saveI`, it should just be a number, here, not an array.

dpvc (changes requested):
I have checked the PR and it does seem to work as advertised.  I did mark the changes we talked about of leaving `nemeth` as the default rather than `euro`, and of adding `data-` before `latex` and `itemLatex` attributes.  

I also suggest a change for the `saveI` handling that doesn't use an array (it uses the call stack for that instead).  That is just a suggestion if you want to use it.

Otherwise, I'm good with the PR.

ts/input/tex/TexParser.ts
zorkow: Simplified `saveI`.

zorkow:
> One question I had: can you tell me the difference between the `itemLatex` and `latex` attributes?

`latex` contains the final latex commands. `itemLatex` can contain incomplete elements. It might be that we don't need the latter, but I would like to have a large enough test set before messing with it again.

zorkow:
I've made the required changes. PTAL.

dpvc (approved):
Looks good.  Thanks for the changes.

PR #1006, opened 2023-09-28
Fixes issue with wrapped script bases.

zorkow:
This is a fix for the issue that rendering of script elements with an base element that is wrapped in a `TexAtom` fails.
E.g., ${\bf a}^b$ fails, while $a^b$ works. But he latter only by accident.
 
The problem is in the `getBaseCore` method, where we loop into the base element. However, the loop works effectively on the node and not the wrapper, thus we have to reset both `core` and `node`, otherwise we end up with the innermost text node and an error in the `baseCharZero` method.


dpvc:
OOPS!  Good catch.  Thanks!

PR #1007, opened 2023-09-28
Fix/missing glob

zorkow:
Adds the missing `-g` option for `rimraf` in the `clean:lib` script. Otherwise the `lib` directories are not removed recursively.


dpvc:
Looks good.

PR #1010, opened 2023-11-01
Refactors all `substr` to `substring` calls

zorkow:
`substr` is deprecated and needs to be replaced by  `substring` (or `slice`). This PR replaces all occurences of `substr` in the `ts` code base with `substring` calls.  It uses the following approach:

* `substr(n)` $\to$ `substring(n)`
* `substr(0, n)` $\to$ `substring(0, n)`
* `substr(a, b)` $\to$ `substring(a, a + b)` if it is clear from context that $b >= 0$. Sometimes, I've added a comment for explanation.
*  `substr(a, b)` $\to$ `b < 0 ? '' : substring(a, a + b)` in all other cases.

### DIFF
--- a/ts/a11y/complexity/collapse.ts
+++ b/ts/a11y/complexity/collapse.ts
@@ -505,7 +505,7 @@ export class Collapse {
 
     const attributes = node.attributes.getAllAttributes();
     for (const name of Object.keys(attributes)) {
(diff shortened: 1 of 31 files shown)


zorkow:
Next PR will do the same for components files.

dpvc (approved):
Looks good.

PR #1011, opened 2023-11-01
Refactors substr to substring on component files.

zorkow:
All `substr` to `substring` rewrites are obvious here. Even the one on the `font` string.


dpvc (approved):
Looks good.

PR #1013, opened 2023-11-09
Fix make-cjs-components script to work without mjs files being built first

dpvc:
This PR resolves the problem where the webpacking failed when building the cjs bundle without having first build the mjs files.  It was due to the `webpack.common.cjs` file using `require.resolve()` command to get full path names; since `webpack.common.cjs` is in the `components` directory, it was using the main `package.json` file, and so the pseudo-packages like `#js` were resolving to the `mathjax-full/mjs` direcrtory rather than the `mathjax-full/cjs` directory.

This resolves the problem by moving the `fullPath()` function to a separate `fullpath.cjs` file in the `components/mjs` and `components/cjs` directories, so that the cjs file will use the `components/cjs/package.json` mappings instead.

It also adds a test to the `link:full` script to only make the link if it is not already there, so it is safe to run the script more than once.

### DIFF
--- a/components/bin/link-full
+++ b/components/bin/link-full
@@ -3,6 +3,7 @@ const path = require('path');
 
 const dir = path.join(__dirname, '..', '..');
 
-if (fs.existsSync(path.join(dir, 'node_modules'))) {
+if (fs.existsSync(path.join(dir, 'node_modules')) &&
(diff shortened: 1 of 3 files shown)


zorkow (approved):
lgtm.

PR #1014, opened 2023-11-09
Better measuring of text that is in a native font rather than MathJax fonts

dpvc:
This PR fixes a problem with determining the width of text that is in a system font rather than in the MathJax fonts.  These come from `\text{}` then `mtextInheritFont` is true, or from error messages (when `merrorInheritFont` is true, which it is by default), or from characters that aren't in the MathJax fonts, like Chinese characters, or when a `font-family` is specified in a `\style{}` command or MathML `style` attribute.

You can't test this in the lab, as the issue only occurs when the math is typeset outside the DOM via `MathJax.tex2chtml()` or one of the other conversion functions, and then inserted into a container with a `font-size` that is other than 100%.

### DIFF
--- a/ts/output/chtml.ts
+++ b/ts/output/chtml.ts
@@ -286,8 +286,7 @@ CommonOutputJax<
     //    and call to getBBox().w in TextNode.ts)
     //
     if (width !== null) {
-      const metrics = this.math.metrics;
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #1016, opened 2023-11-11
Fix default option list to {}.  (mathjax/MathJax#3128)

dpvc:
This PR fixes a problem where the default options list was `null` rather than `{}`.  So, for example, `new SVG()` would produce an error.

Resolves issue mathjax/MathJax#3128.


zorkow (approved):
lgtm.

PR #1017, opened 2023-11-11
Add ability to not show latex attributes, and reorganize settings menu

dpvc:
This PR adds a menu option to remove the `data-latex` and `data-latexItem` attributes from the "Show Math As MathML" menu item, as these can get quite long, making it hard to see the MathML structure.  We also reorganize the Math Settings submenu to push the MathML filters into an additional submenu level.  Finally, we move the filtering of the SRE attributes into the `MmlVisitor` so the attributes are never added, rather than using a string replacement after the fact.

### DIFF
--- a/ts/ui/menu/MmlVisitor.ts
+++ b/ts/ui/menu/MmlVisitor.ts
@@ -106,7 +106,7 @@ export class MmlVisitor<N, T, D> extends SerializedMmlVisitor {
   protected getAttributeList(node: MmlNode): PropertyList {
     const list = super.getAttributeList(node);
     if (this.options.filterTex) {
-      delete list['data-latex'], list['data-latexItem'];
+      delete list['data-latex'], list['data-latex-item'];
     }
     if (this.options.filterSRE) {
       const keys = Object.keys(list).filter(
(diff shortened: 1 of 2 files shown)


ts/ui/menu/MmlVisitor.ts
```diff
@@ -97,4 +100,22 @@ export class MmlVisitor<N, T, D> extends SerializedMmlVisitor {
...
+   * @override
+   */
+  protected getAttributeList(node: MmlNode): PropertyList {
+    const list = super.getAttributeList(node);
+    if (this.options.filterTex) {
+      delete list['data-latex'], list['data-latexItem'];
+    }
+    if (this.options.filterSRE) {
+      const keys = Object.keys(list).filter(
+        id => id.match(/^(?:data-semantic-.*?|role|aria-(?:level|posinset|setsize))$/)
```
zorkow, line 113: What if we parse from MathML and some of these aria attributes are already in the incoming element?
(I suppose we would have overwritten them at this point anyway.)

zorkow (commented):
One point of discussion.

ts/ui/menu/MmlVisitor.ts
dpvc: This filtering only occurs in the `Show Math As` or `Copy Math As` menus.  Nothing is changed internally.  This filtering is controlled by menu options, so the user can turn off the filtering in the case where retaining those values are desired.  Does that put you more at ease?

dpvc: Note that this `MmlVisitor` is in the `ts/ui/menu` directory, not the one in `ts/core/MmlTree`, so it only affects the menu output.

zorkow: Thanks. Sounds good.

zorkow (approved):
lgtm.

PR #1018, opened 2023-11-13
Proper handling of typesetPromise in defaultPageReady. (mathjax/MathJax#3130)

dpvc:
This PR fixes a problem with `defaultPageReady()` where the typesetting could be done at the wrong time.  With this PR, it should be performed when the `then()` is triggered, whereas before it was done when the `defaultPageReady()` call was made (OOPS!).

Resolves issue mathjax/MathJax#3130.


zorkow (approved):
lgtm.

PR #1019, opened 2023-11-13
Have \unicode check for illegal font name.  (mathjax/MathJax#3129)

dpvc:
This PR adds a check that the font name doesn't include semicolons in the `\unicode` macro.  This prevents extra CSS from being inserted into the styles for the resulting output.

Resolves issue mathjax/MathJax#3129.


zorkow (approved):
lgtm.

PR #1021, opened 2023-11-14
Pnpm move

zorkow:
Changes build system to [`pnpm`](https://pnpm.io/). Major changes are:

* `pnpm` does not run `pre` and `post` hooks for custom scripts. (They still work for lifecycle scripts).
* Scripts are now separated topically into CJS and MJS scripts. There all still some general scripts of `copy`, `clean`, etc.
* Scripts try to make the execution flow as transparent as possible. They are mainly separated into two types
     * Compound scripts that execute multiple. They produce blue logging output.
     * Single scripts that execute a single command or action. They produce brown output.
* All scripts use `:` separation to construct names as meaningful as possible. Exception are some legacy script names - now mainly aliases - like `make-components`.
* The `link:full` does not have to execute explicitly anymore
* `clean` also removes `bundle-cjs`
* `pack` now uses `pnpm` instead of `npx`. However, the former needs to be executed in the root directory of the repository. So the paths need to be given relative to root and not to the directory of the component that is packed.
* Replaces all occurrences of `substr` in the node scripts with `substring`.


zorkow:
Made changes as discussed in PR #1023

package.json
```diff
@@ -65,71 +65,76 @@
...
+    "cjs:components:finalize": "pnpm -s log:comp 'Finalize cjs components'; pnpm -s cjs:components:copy && pnpm -s copy:pkg components/cjs && pnpm -s clean:lib cjs",
+    "cjs:compile": "pnpm -s log:single 'Compiling cjs typescript files' && pnpm tsc --project tsconfig/cjs.json",
+    "cjs:bundle:finalize": "pnpm -s log:single 'Finalize cjs bundle'; echo '{\n  \"type\": \"commonjs\"\n}' > bundle-cjs/package.json;",
+    "cjs:components:make": "make() { pnpm -s log:single 'Making cjs components'; components/bin/makeAll --cjs --terse --bundle-cjs $1 components/cjs; }; make",
+    "cjs:components:src:build": "pnpm -s log:comp 'Building cjs components sources'; pnpm cjs:components:clean && pnpm cjs:components:compile && pnpm cjs:components:finalize",
+    "cjs:components:build": "pnpm -s log:comp 'Building cjs components'; pnpm -s cjs:components:src:build && pnpm clean:dir bundle-cjs && pnpm -s cjs:components:make --build",
+    "cjs:src:build": "pnpm -s log:comp 'Building cjs sources'; pnpm -s link:full && pnpm clean:dir cjs && pnpm -s cjs:compile && pnpm -s copy:assets cjs && pnpm -s copy:pkg cjs",
```
dpvc, line 92: Now I'm wondering if these lines want to be sorted so that all the `cjs.components` are together.  E.g., perhaps move `cjs:bundle:clean` and `cjs:bundle:finalize`to the top followed by `cjs:compile`, and put the other ones in an order that makes sense?  I haven't looked at what order they get used, so if that is the order they are currently in, that's OK, too.

dpvc (approved):
One last potential change, but otherwise looks good.

package.json
zorkow: I've sorted all `:` scripts alphabetically.

PR #1022, opened 2023-11-16
Refactor/parseutil

zorkow:
Refactors the `ParseUtil` module:

* Makes a `Map` out of unit conversion mapping and exports it. This makes it extensible
* Adapts all other constants to use the extensible map. (This is useful for the IEEE where they have a custom unit `pi`).
* *Does not (yet)* combine length functionality with the `util/lengths.ts` module (see my email from yesterday).
* Removes `namespace` and adapts all `import` statements in modules using `ParseUtil`. This is the majority of files!
* Adds some `tests` for this for all the length methods, that I have used during conversion. Note, these already uses `pnpm`. To run the tests, do
```
pnpm -r install
pnpm test
```
The tests are in a separate commit and can be removed.


zorkow:
Since the indentation has changed on `ParseUtil.ts` it is best viewed with `ignore whitespace`.

ts/input/tex/ParseUtil.ts
```diff
@@ -33,668 +33,667 @@ import {entities} from '../../util/Entities.js';
...
+  ['in', emPerInch],
+  ['cm', emPerInch / 2.54], // 2.54 cm to an inch
+  ['mm', emPerInch / 25.4], // 10 mm to a cm
+  ['mu', 1 / 18],
+]);
+
+const num = '([-+]?([.,]\\d+|\\d+([.,]\\d*)?))';
+const unit = () => `(${Array.from(UNIT_CASES.keys()).join('|')})`;
+const dimenEnd = () => RegExp('^\\s*' + num + '\\s*' + unit() + '\\s*$');
+const dimenRest = () => RegExp('^\\s*' + num + '\\s*' + unit() + ' ?');
```
dpvc, line 55: I'm a little concerned about the change of these, from an efficiency standpoint.  Every time a dimension is requests, you have to reconstruct the regular expressions, even when the units haven't changed.  That seems wasteful.

I know you are trying to allow units to be added, but wouldn't it be better to have an `addUnit()` function that, in addition to adding a new unit, also updates the regular expressions once, so they can be reused until a new unit is added?  (And perhaps a `removeUnit()` to balance it?)

ts/input/tex/ParseUtil.ts
```diff
@@ -33,668 +33,667 @@ import {entities} from '../../util/Entities.js';
 import {MmlMunderover} from '../../core/MmlTree/MmlNodes/munderover.js';
 
 
-namespace ParseUtil {
```
dpvc, line 36: I would like to see this replaced by an object `export const ParseUtil {...}` that has the exported functions as methods and the constants as properties (since it is always used as `ParseUtil.method()` anyway.  Then you can use `import {ParseUtil} from '.../ParseUtil.js';` as with all the other objects, rather than needing `* as ParseUtil`.  It would also give a natural place to put the `addUnit()` that I suggest below, for example.

One reason for this is that it makes it possible to override the various functions, if need be.  That is, if there is a bug in a function, say `fixInitialMO()`, then one could replace `ParseUtil.fixInitialMO()` in the startup configuration with a fixed version until a new MathJax version is released.  But with individually exported functions, that can't the done.

We may want to consider similar changes as we remove namespaces from other files during the code cleanup.

ts/input/tex/ParseUtil.ts
```diff
@@ -33,668 +33,667 @@ import {entities} from '../../util/Entities.js';
...
+/**
+ * Convert a dimension string into standard em dimension.
+ * @param {string} dim The attribute string.
+ * @return {number} The numerical value.
+ */
+export function dimen2em(dim: string): number {
+  let [value, unit] = matchDimen(dim);
+  let m = parseFloat(value || '1');
+  let func = UNIT_CASES.get(unit);
+  return func ? func * (m) : 0;
```
dpvc, line 98: Shouldn't this now be called `factor` rather than `func`?

tests/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "type": "module",
+  "devDependencies": {
+    "@babel/preset-env": "^7.23.2",
+    "babel-jest": "^29.7.0",
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "jest": "^29.7.0"
+  },
+  "dependencies": {
+    "@babel/preset-typescript": "^7.23.2",
+    "ts-jest": "^29.1.1"
```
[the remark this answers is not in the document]
zorkow: Moved.

ts/input/tex/ParseUtil.ts
zorkow: Added `addUnit`, `removeUnit`, and `getUnit`. The latter now allows us to always use `pt` if illegal unit is provided, similar to TeX behaviour.

ts/input/tex/ParseUtil.ts
zorkow: True. It used to be a function...

zorkow:
@dpvc I've refactored the `ParseUtil` into a single object. Have a look, if that is really what you want.

The `UNIT_CASES` is now a bespoke class that can be used for similar mappings in lengths.
And there are a number of temporary elements, that should eventually move into lenghts.

zorkow:
Also should this here
https://github.com/mathjax/MathJax-src/blob/8565f9da973238e4c9571a86a4bcb281b1d98d9b/ts/util/lengths.ts#L130
not return '0em'. The rounding function below does that. Then it would be the same as `Em` in `ParseUtil.ts`

dpvc (changes requested):
See my comments below about the `ParseUtil` object.

ts/input/tex/ParseUtil.ts
```diff
@@ -33,668 +33,720 @@ import {entities} from '../../util/Entities.js';
...
+ * @param {TexParser} parser   The current TeX parser.
+ * @param {MmlNode} base       The base node.
+ * @param {MmlNode} script     The under- or over-script.
+ * @param {string} pos         Either 'over' or 'under'.
+ * @param {boolean} stack      True if super- or sub-scripts should stack.
+ * @return {MmlNode}           The generated node (MmlMunderover or TeXAtom)
+ */
+export function underOver(parser: TexParser, base: MmlNode, script: MmlNode, pos: string, stack: boolean): MmlNode {
+  // @test Overline
+  checkMovableLimits(base);
```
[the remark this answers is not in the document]
zorkow: > This should remain `ParseUtil.checkMovableLimits(base);` so that if that method is overridden, it will affect this method.
> 
> Ideally, these functions would **not** be exported except through the `ParseUtil` object, so that all usage of the functions are through `ParseUtil`. If you allows for `import {underOver} from 'ParseUtil.js'` that means that `underOver` can't be properly overridden, because the function with the `import` will not get the overridden version.
> 

That's what I meant to emulate with the object at the end of the file. I just wanted to get feedback first before rewriting all the methods into the structure you indicate below.


> I had meant to have the original `namespace ParseUtil {` be replaced by `export const ParseUtil = {`, and all the constants and functions would be part of this object directory. So you would do
> 
> ```js
> export const ParseUtil = {
> 
>   /**
>    * Matches for a dimension argument.
>    * @param {string} dim The argument.
>    * @param {boolean} rest Allow for trailing garbage in the dimension string.
>    * @return {[string, string, number]} The match result as (Anglosaxon) value,
>    *     unit name, length of matched string. The latter is interesting in the
>    *     case of trailing garbage.
>    */
>   matchDimen(dim: string, rest: boolean = false): [string, string, number] {
>     let match = dim.match(rest ? UNIT_CASES.dimenRest : UNIT_CASES.dimenEnd);
>     return match ?
>       this.muReplace([match[1].replace(/,/, '.'), match[4], match[0].length]) :
>       [null, null, 0];
>   },
> 
>   /**
>    * Transforms mu dimension to em if necessary.
>    * @param {[string, string, number]} [value, unit, length] The dimension triple.
>    * @return {[string, string, number]} [value, unit, length] The transformed triple.
>    */
>   muReplace([value, unit, length]: [string, string, number]): [string, string, number] {
>     if (unit !== 'mu') {
>       return [value, unit, length];
>     }
>     let em = this.Em(UNIT_CASES.get(unit) * (parseFloat(value || '1')));
>     return [em.slice(0, -2), 'em', length];
>   },
> 
>   .
>   .
>   .
> 
> };
> ```
> 
> You could also put `UNIT_CASES` as a property as well, rather than exporting it separately, and use `this.UNIT_CASES` within the functions.

Well, ultimately that will move into lengths (together with the methods under `// Temporary before move to lenghts`). But we can do that for now.

dpvc:
> If I understand correctly, you want to be able to make changes to `ParseUtil` functionality from the configuration options.

In the sense that it can be altered via the `startup.ready()` function, yes.  The idea being that if someone needs to patch one of the routines, they can do so and the modified one will be used everywhere.

I have wanted to be able to do this with other namespaces (like the `startup` namespace in `ts/components/startup.ts`), but have not bee able to.  That is the main reason I have to turning these into actual objects rather than namespaces.

> How should be deal with this when (or if) we combine `ParseUtil` with `lengths.ts`?

I'm thinking that `length.ts` should also be an object as well, rather than individual functions.  That would make everything mutable.  Does that answer your question?

ts/input/tex/ParseUtil.ts
```diff
@@ -64,63 +230,52 @@ namespace ParseUtil {
...
 
 
   /**
    * Convert a dimension string into standard em dimension.
    * @param {string} dim The attribute string.
    * @return {number} The numerical value.
    */
-  export function dimen2em(dim: string): number {
-    let [value, unit] = matchDimen(dim);
+  dimen2em: function(dim: string): number {
```
dpvc, line 247: This could be done as 

``` js
  dimen2em(dim: string): number) {
```

for a savings of 10 characters per function.  Not huge, but worth considering, as it does add up, when you consider doing the same with the other functions below.

dpvc (approved):
Looks good.  I have one suggestion for saving some characters, but otherwise, ready to go.

ts/input/tex/ParseUtil.ts
zorkow: Done.

PR #1023, opened 2023-11-20
pnpm_move update

dpvc:
This PR makes the changes that we discussed face-to-face:

* The `makeAll` script is modified to not cd to the component directories, but rather passes the directory to the various programs as their last argument.  This avoids the problem that you fixed in the `pack` command, in particular.

* The package scripts are modified:
    -  The `\e` escape is replaced by `\x1B` since MacOS ships with an older bash version that doesn't support that.
    - The color was changed from 32 (yellow) to 34 (blue) in `log:single`, and `--` was added to help make it visually distinct (especially when `log:comp` is used before it)`.
    - The `cjs` and `mjs` names were moved to the front of their commands to make the sections each use a prefix.
    - The mml3 XSLT commands were modified to use an `mml3:` prefix like the other sections
    - The cleaning of the `bundle-cjs` directory was moved earlier in the `clean` command so that the log messages are in a more appropriate order.


zorkow (commented):
We could go one step further with our naming convention and do:
```
cjs:components:clean...
cjs:components:build...
...
cjs:bundle:clean ...
```

What do you think?

package.json
```diff
@@ -74,47 +74,47 @@
...
-    "build:cjs": "pnpm -s log:header 'Building cjs'; pnpm -s build:cjs:src && pnpm -s build:cjs:components",
-    "=============================================================================== MJS": "",
-    "make:mjs:components": "pnpm -s log:single 'Making mjs components'; components/bin/makeAll --mjs --terse components/mjs",
-    "finalize:mjs:bundle": "pnpm -s log:single 'Finalize mjs bundle'; echo '{\n  \"type\": \"commonjs\"\n}' > bundle/package.json;",
-    "compile:mjs": "pnpm -s log:single 'Compiling mjs typescript files'; pnpm tsc --project tsconfig/mjs.json",
-    "build:mjs:components": "pnpm -s log:comp 'Compiling mjs component files'; pnpm clean:lib mjs && pnpm clean:dir bundle && pnpm make:mjs:components && pnpm finalize:mjs:bundle",
-    "build:mjs:src": "pnpm -s log:comp 'Building mjs sources'; pnpm -s link:full && pnpm -s clean:dir mjs && pnpm -s compile:mjs && pnpm -s copy:assets mjs",
-    "build:mjs": "pnpm -s log:header 'Building mjs'; pnpm -s build:mjs:src && pnpm -s build:mjs:components",
+    "log:comp": "log() { echo \\\\x1B[32m$1\\x1B[0m\\; }; log",
```
zorkow, line 80: Superfluous `\\;` leads to errors. 
```suggestion
    "log:comp": "log() { echo \\\\x1B[32m$1\\x1B[0m; }; log",
```

zorkow (changes requested):
The `\x1B` change does not work for me. I'll look into it what we could use instead.

zorkow:
> The `\x1B` change does not work for me. I'll look into it what we could use instead.

@dpvc can you try this line, if the `\033` escape code works for you?

```
    "log:test": "log() { echo \\\\033[33m hello world \\\\033[0m; }; log",
```

[Source: Stackoverflow](https://stackoverflow.com/questions/4842424/list-of-ansi-color-escape-sequences)

dpvc:
> We could go one step further with our naming convention

Sounds good.  Go ahead with that.

> can you try this line, if the \033 escape code works for you?

Yes, it does.  That was going to be my suggestion as well.

zorkow (approved):
Thanks. I'll go ahead and merge.

PR #1024, opened 2023-11-24
Don't descend into items that aren't broken when looking for the linebreak node.  (mathjax/MathJax#3135)

dpvc:
This PR fixes a problem with the size of SVG elements for in-lin math when in-line line-breaking is enabled.  It fixes the determination of the node where the line breaking occurs by not descending into unbroken nodes.  The `bbox.start` was not being properly tested, so it would descending into a fraction, for example, and stop at its numerator, giving the wrong scaling factor when the `height` and `width` of the resulting SVG was set.

Resolves issue mathjax/MathJax#3135.


zorkow (commented):
lgtm.

zorkow (approved):

PR #1025, opened 2023-11-24
Update CHTML longdiv output to not require clip-path.

dpvc:
This PR changes how CHTML handles the `longdiv` enclosure so that it doesn't need to use clipping paths.  Apparently Chrome for Android and iOS only implements `clip-path` in very recent versions, and so older versions show a full oval rather than use the right half used for the division symbol.  The solution is to break the half-oval into two parts: the upper and lower halves, and handle each separately.  Each half uses a rounded corner of 100% radius (rather than the 50% radius that was used for the complete oval), with two of the border edges set to `none`.


zorkow (approved):
lgtm.

PR #1026, opened 2023-11-25
Make const enum into a regular object.  (mathjax/MathJax#3114)

dpvc:
This PR changes the `DIRECTION` enum into a regular object, like all other constants.  This was the only enum that we used, and it apparently is a problem when `--isolatedModules` is used.  This also means we don't need the `DirectionVH` object from `ts/common/Wrappers/mo.ts`.

Resolves issue mathjax/MathJax#3114.


zorkow (approved):
lgtm.

PR #1027, opened 2023-11-25
Add cssText() method to adaptor and handle dynamic rules in HTML output.

dpvc:
It turns out that when a stylesheet has dynamically added rules (like the ones MathJax produced for CHTML output when `adaptiveCSS` is true), those rules are not returned by `innerHTML`, `outerHTML`, or `textContent`.  This makes it hard to obtain the CSS needed for HTML output in that case.

The PR adds a `cssText()` method to the adaptor class so that these rules will be included in the output.  This can be used in place of `innerHTML` or `textContent` to get the list of rules when `adaptiveCSS` is being used.

### DIFF
--- a/ts/core/DOMAdaptor.ts
+++ b/ts/core/DOMAdaptor.ts
@@ -344,6 +344,12 @@ export interface DOMAdaptor<N, T, D> {
   insertRules(node: N, rules: string[]): void;
 
   /**
+   * @param {N} node        The stylesheet node whose rules are to be returned
+   * @return {string}       The string versions of the stylesheet rules
+   */
+  cssText(node: N): string;
+
+  /**
    * @param {N} node        The HTML node whose font size is to be determined
    * @return {number}       The font size (in pixels) of the node
    */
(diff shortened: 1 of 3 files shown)


dpvc:
The alternative would be to modify `innerHTML`, `outerHTML`, and `textContent` to include them, but that makes these function less compatible with what the browser actually does, and lead to confusion if these were used on a node *containing* a `style` node, like the `head` node, where the normal HTML behavior would still occur.

ts/core/DOMAdaptor.ts
```diff
@@ -343,6 +343,12 @@ export interface DOMAdaptor<N, T, D> {
    */
   insertRules(node: N, rules: string[]): void;
 
+  /**
+   * @param {N} node        The stylesheet node whose rules are to be returned
+   * @return {string}       The string versions of the stylesheet rules
```
zorkow, line 348:
```suggestion
   * @return {string}       The string version of the stylesheet rules
```

zorkow (approved):
One typo.

PR #1028, opened 2023-12-04
Fixes issue https://github.com/mathjax/MathJax/issues/3120

zorkow:
Fixes [MathJax issue #3120](https://github.com/mathjax/MathJax/issues/3120), by checking for the underscore command case.

### DIFF
--- a/ts/input/tex/TexParser.ts
+++ b/ts/input/tex/TexParser.ts
@@ -572,7 +572,8 @@ export default class TexParser {
       }
       return;
     }
-    if (node.attributes.get(TexConstant.Attr.LATEX) === '_' && str !== '_') {
+    if (node.attributes.get(TexConstant.Attr.LATEX) === '_' &&
+      str !== '_' && str !== '\\_') {
       if (str === '}') {
         this.composeBraces(node.childNodes[1]);
       } else {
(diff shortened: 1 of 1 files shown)


ts/input/tex/TexParser.ts
```diff
@@ -572,7 +572,8 @@ export default class TexParser {
       }
       return;
     }
-    if (node.attributes.get(TexConstant.Attr.LATEX) === '_' && str !== '_') {
+    if (node.attributes.get(TexConstant.Attr.LATEX) === '_' &&
+      str !== '_' && str !== '\\_') {
```
dpvc, line 576: I think you need a similar fix for `^` in line 561 above, in case anyone creates a macro `\^` similar to the `\_` one.

dpvc (changes requested):
This fixes `\_`, but leaves the same potential problem for `\^`, if one were to be defined similarly to `\_`.

ts/input/tex/TexParser.ts
zorkow: At the moment that can't really happen as latex computation happens after the macro replacement.
But I am not really happy with that anyway, so I've added the case.  
PTAL.

dpvc (approved):
Thanks for the change. LGTM.

PR #1030, opened 2023-12-10
Fix problem with let using attributes of {} when they should be null

dpvc:
The current `disassembleToken()` and `assembleToken()` methods produce a new token whose attributes are `{}` when the original had `null` attributes.  This has implications for some tokens where the difference between these are important, as in the `mathchar7` function at

 https://github.com/mathjax/MathJax-src/blob/ae389ded62763bb176f797b6bfbb5a997c42dba1/ts/input/tex/ParseMethods.ts#L156-L165

where line 157 provides a default attribute list when `math.attributes` is `null` (but not if it is an empty object).

This leads to the loss of the default attributes when a new control sequence is `\let` equal to one that uses `mathchar7`, as in 

``` latex
\let\percent=\%
\percent\%
```

where the result of `\percent` will be in italics, while `\%` will not.

This PR resolves the issue by actually eliminating `disassembleToken()` and `assembleToken()` and just using the original token directly in the `method` at new line 137 (via the same closure that preserves `map`).

### DIFF
--- a/ts/input/tex/newcommand/NewcommandMethods.ts
+++ b/ts/input/tex/newcommand/NewcommandMethods.ts
@@ -133,13 +133,9 @@ NewcommandMethods.Let = function(parser: TexParser, name: string) {
       return;
     }
     macro = (map as sm.CharacterMap).lookup(name) as Token;
-    const newArgs = NewcommandUtil.disassembleToken(cs, macro);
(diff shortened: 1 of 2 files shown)


zorkow (approved):
Looks good. Works for all my tests.

PR #1031, opened 2023-12-10
Fix keyvalueOptions to process braces and backslahses better

dpvc:
This PR fixes several issues with the `keyvalueOptions()` utility's handling of braces and control sequences, and adds an extra parameter that controls whether it should do LaTeX3-style processing, or traditional processing.  

The current implementation doesn't handle some control sequences properly, in particular, `\{`, so,

``` js
keyvalueOptions('x = \\{');
```

with throw an error about a missing closing brace or extra open brace.  This PR fixes that issue with the addition of `case '\\'` in the `readValue()` function.

The current implementation doesn't always handle braces properly.  For example

``` js
keyvalueOptions('a = {{x}y}');
```

returns `{ a: "x}" }` rather than `{ a: "{x}y"}`.  This is due to thinking that there are two outer sets of braces to be removed, rather than only one (`start` isn't decremented where it should be, so the `if (start > brace) { start = brace }`  is moved to a location that takes care of that.

Finally, the current implementation doesn't remove an outer set of braces for the entire string, as it should.  That is,

``` js
keyvalueOptions('{x = y}');
```

returns `{ "x = y": true }` rather than `{ x: "y" }`, as expected.  This is corrected via the `dropBraces` flag that allows one level of bracing to be removed.

In addition to these changes, we add a flag to `keyvalueOptions()` that determines whether the LaTeX3 rules are to be used, or the traditional LaTeX2 rules, for parsing the key-value pairs.  The difference is that traditional parsing removes any number of outermost braces, while LaTeX3 only removes one.  So

``` js
keyvalueOptions('x = {{y}}');
```

would return `{ x: "y" }` using traditional rules, while

``` js
keyvalueOptions('x = {{y}}', null, false, true);
```

would return `{ x: "{y}" }` using LaTeX3 rules (the final `true` parameter controls this).

LaTeX3 key-value parsing has the ability to check types of arguments and report errors, but that has not been implemented here.  It might be worth doing so in the future.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -109,16 +109,19 @@ function muReplace([value, unit, length]: [string, string, number]): [string, st
  * Implementation of the keyval function from https://www.ctan.org/pkg/keyval
  * @param {string} text The optional parameter string for a package or
  *     command.
+ * @param {boolean?} l3keys If true, use l3key-style parsing (only remove one set of braces)
  * @return {EnvList} Set of options as key/value pairs.
  */
-function readKeyval(text: string): EnvList {
+function readKeyval(text: string, l3keys: boolean = false): EnvList {
   let options: EnvList = {};
   let rest = text;
   let end, key, val;
+  let dropBrace = true;
   while (rest) {
-    [key, end, rest] = readValue(rest, ['=', ',']);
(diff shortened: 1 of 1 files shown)


zorkow (approved):
lgtm.

PR #1032, opened 2023-12-11
Add ability to do typechecking and conversion on the values of key-value options

dpvc:
This PR adds optional type checking and conversion to the `keyvalueOptions()` function.  Currently, the `allowed` argument gives an object that says what keys are allowed (ones with non-zero values in that object).  Here, we extend the use of the `allowed` object to provide data for type checking and converting the values of the given keys.

This is done through a new `KeyValueType` object that has `validate()` and `convert()` methods (and a name, for potential error reporting).  There is a `KeyValueTypeDefinition()` function that makes creating these objects easier, and a `KeyValueTypes` object that holds predefined definitions for the standard types (boolean, number, string) and some variations (integer, dimen, and `oneof()` that gives a collection of valid option for the string).  These seem to be a good base set of definitions, but you can create your own custom types with more detailed checker and converter, as needed.

We aren't currently using this feature now, but it would be useful in the siunitx extension that I am reviewing for PCC now.

A similar idea may also be useful for making the option checking more robust in the future (in `util/Options.ts`), where we only do key name checking, but not type checking currently.

### DIFF
--- a/ts/input/tex/ParseUtil.ts
+++ b/ts/input/tex/ParseUtil.ts
@@ -109,22 +109,14 @@ function muReplace([value, unit, length]: [string, string, number]): [string, st
 /**
  * The data needed for checking the value of a key-value pair.
  */
-export type KeyValueType<T> = {
-  name: string;
-  verify: (value: string) => boolean;
-  convert: (value: string) => T;
+export class KeyValueType<T> {
+  constructor(
+    public name: string,
+    public verify: (value: string) => boolean,
+    public convert: (value: string) => T
+  ) {}
 }
 
-/**
- * A function for creating a key-value type checker
- */
-export function KeyValueTypeDefinition<T>(
-  name: string,
-  verify: (value: string) => boolean,
-  convert: (value: string) => T,
-) {
-  return {name, verify, convert} as KeyValueType<T>;
-}
 
 /**
  * Predefined value types that can be used to create the list of allowed types.  E.g.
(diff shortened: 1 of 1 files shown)


zorkow (commented):
Would it not be much simpler to make the `KeyValueType<T>` a class and then create objects with new instead of using `KeyValueTypeDefinition`?

dpvc:
You don't need new instances each time they are used.  The definitions are the same, so only one object needs to be used for each type.  (Except the `oneof`, where you have to specify the strings that are valid.)  Also, since the `verify` and `convert` methods differently by the type of `T`, you would either have to have a bunch of different subclasses, one for each `T`, or have the `verify` and `convert` functions have a switch on the type of `T` (which I'm not sure can be done with a template type).

If you have something else in mind, can you be more explicit about it?

zorkow:
I did not mean to create a new object in every call. I was more thinking about having a single class instead of a type and definition function, i.e., something along the line of:

```TypeScript
export class KeyValueType<T> {

  constructor(
    public name: string,
    public verify: (value: string) => boolean,
    public convert: (value: string) => T
  ) {}

}
```

and then rewriting the `KeyValueTypes` constant to:

```TypeScript
export const KeyValueTypes: {[name: string]: KeyValueType<any> | ((data: any) => KeyValueType<any>)} = {
  boolean: new KeyValueType<boolean>(
    'boolean',
    (value) => value === 'true' || value === 'false',
    (value) => value === 'true'
  ),
...
```

This way we still only create one instance object for each entry. But we save the `Definition` method.

dpvc:
OK, I made the changes from the object to a class.  That is a good change, thanks.  Please take a look.

zorkow (approved):
lgtm.

PR #1034, opened 2023-12-18
Clear process bits for findMath and attach-speech, and remove safe bit that isn't needed.

dpvc:
When the document state is reset, the process bits should be reset for the actions that are for higher states.  This wasn't happening for the `attach-speech` bit in the `semantic-enrich` extension.  I did a check to make sure all the other bits are bing properly handled, and the `findMath` bit was not being recent properly, either.  Also, a `safe` bit was being allocated that was never used.  This PR fixes these three problems.

Resolve issues raised in a [user's forum post](https://groups.google.com/g/mathjax-users/c/7vrAnA7SMz0).


zorkow (approved):
lgtm.

PR #1036, opened 2024-01-04
Add Text.splitText() method to linkedom.  (mathjax/MathJax#3134)

dpvc:
This PR adds the missing `splitText()` method to the linkedom `Text` object, which we need since it is used to split the text around TeX and AsciiMath delimiters so that the math can be removed and replaced by the typeset math.


zorkow (approved):
lgtm.

PR #1040, opened 2024-01-15
Introduces a speechgenerator pool

zorkow:
PR introduces a `GeneratorPool` data structure that 
* coordinates the computation of speech. 
* takes care of async locale loading.

This makes the speech computation considerably cleaner. In particular there is no longer speech computation in the semantic enrichment method. Speech can be computed if the `speech="deep"` option is set. 
All Speech is now computed in the `attachSpeech` method, where originally only the `aria-label` was set. 

Since the `attachSpeech` state is called after rerender, this also takes care of computing speech for the collapsed elements etc. 

Another side effect is that we no longer use the walkers provided by SRE, which makes syncing a lot easier. At the moment only the walking and summary functionality is ported. The rest will be ported in a subsequent PR. 

**Note:** Currently only works with SRE branch `new_explorer_changes`. I'll try to merge that asap.

### DIFF
--- a/ts/a11y/explorer.ts
+++ b/ts/a11y/explorer.ts
@@ -101,11 +101,6 @@ export function ExplorerMathItemMixin<B extends Constructor<HTMLMATHITEM>>(
     protected refocus: boolean = false;
 
     /**
-     * Save explorer id during rerendering.
-     */
-    protected savedId: string = null;
-
-    /**
      * Add the explorer to the output for this math item
      *
      * @param {HTMLDocument} document   The MathDocument for the MathItem
@@ -116,10 +111,6 @@ export function ExplorerMathItemMixin<B extends Constructor<HTMLMATHITEM>>(
       if (!this.isEscaped && (document.options.enableExplorer || force)) {
         const node = this.typesetRoot;
         const mml = toMathML(this.root);
-        if (this.savedId) {
-          this.typesetRoot.setAttribute('sre-explorer-id', this.savedId);
-          this.savedId = null;
-        }
         if (!this.explorers) {
           this.explorers = new ExplorerPool();
         }
@@ -132,8 +123,8 @@ export function ExplorerMathItemMixin<B extends Constructor<HTMLMATHITEM>>(
      * @override
      */
     public rerender(document: ExplorerMathDocument, start: number = STATE.RERENDER) {
-      this.savedId = this.typesetRoot.getAttribute('sre-explorer-id');
-      this.refocus = (hasWindow ? window.document.activeElement === this.typesetRoot : false);
+      this.refocus = (hasWindow ?
+        window.document.activeElement === this.typesetRoot?.childNodes[0] : false);
       if (this.explorers) {
         this.explorers.reattach();
       }
--- a/ts/a11y/explorer.ts
+++ b/ts/a11y/explorer.ts
@@ -123,7 +123,7 @@ export function ExplorerMathItemMixin<B extends Constructor<HTMLMATHITEM>>(
      * @override
      */
     public rerender(document: ExplorerMathDocument, start: number = STATE.RERENDER) {
-      let speech = this.explorers.speech();
+      let speech = this.explorers.speech;
       if (speech.attached && speech.active) {
         const focus = speech.semanticFocus();
         this.refocus = focus ? focus.id : null;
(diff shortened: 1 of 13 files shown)


zorkow:
@dpvc I've made the changes as discussed today. PTAL.

ts/a11y/explorer/ExplorerPool.ts
```diff
@@ -295,6 +297,16 @@ export class ExplorerPool {
     this.highlighter.unhighlight();
   }
 
+  /**
+   * Convenience method to return the speech explorer of the pool with the
+   * correct type.
+   *
+   * @return {SpeechExplorer}
+   */
+  public speech(): SpeechExplorer {
```
dpvc, line 306: Would this be better as a getter so you don't need the `()` every time?  I always go back and forth about that.

ts/a11y/speech/GeneratorPool.ts
```diff
@@ -0,0 +1,353 @@
...
+   */
+  public update(options: OptionList) {
+    this.options = options;
+    return this._update(options);
+  }
+
+  private _update(options: OptionList) {
+    let update = false;
+    if (options?.sre?.braille !== this.currentBraille) {
+      this.currentBraille = options.sre.braille;
```
dpvc, line 141: If `options` or `options.sre` don't exist, then the `if` condition will be true, and you will end up with an error from `options.sre.braille` on line 141.  I'm not sure the question marks are needed in the `if` condition, but if you need to be concerned about that, I would suggest something like

``` js
if (!options || !options.sre) return false;
```

at the top.  Similarly for line 145.

ts/a11y/speech/GeneratorPool.ts
```diff
@@ -0,0 +1,353 @@
...
+    return buildLabel(
+      center || this.adaptor.getAttribute(node, 'data-semantic-speech'),
+      this.adaptor.getAttribute(node, 'data-semantic-prefix'),
+      // TODO: check if we need this or if it is automatic by the screen readers.
+      this.adaptor.getAttribute(node, 'data-semantic-postfix'),
+      sep
+    );
+  }
+
+  private copyAttributes(xml: Element, node: N, attr: string) {
```
dpvc, line 314: Do you need jsdoc for this function and the `attrList` below?

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -248,11 +291,61 @@ export class SpeechExplorer extends AbstractExplorer<string> implements KeyExplo
...
+   * @return {HTMLElement} The refocused targeted node.
+   */
+  public nextStyle(node: HTMLElement): HTMLElement {
+    this.generators.nextStyle(node);
+    this.Speech();
+    this.refocus(node);
+    return node;
+  }
+
+  private refocus(node: HTMLElement) {
```
dpvc, line 340: Do you need jsdoc for this function and the one below?

dpvc (changes requested):
I've made some comments about some minor issues.  See what you think.

ts/a11y/explorer.ts
```diff
@@ -132,8 +123,11 @@ export function ExplorerMathItemMixin<B extends Constructor<HTMLMATHITEM>>(
      * @override
      */
     public rerender(document: ExplorerMathDocument, start: number = STATE.RERENDER) {
-      this.savedId = this.typesetRoot.getAttribute('sre-explorer-id');
-      this.refocus = (hasWindow ? window.document.activeElement === this.typesetRoot : false);
+      let speech = this.explorers.speech();
+      if (speech.attached && speech.active) {
+        const focus = speech.semanticFocus();
+        this.refocus = focus ? focus.id : null;
+      }
```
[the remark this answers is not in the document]
zorkow: `this.explorers` can be empty. I've changed both this and the next method to reflect this.

ts/a11y/explorer/ExplorerPool.ts
zorkow: Done.

ts/a11y/speech/GeneratorPool.ts
zorkow: I thought I had caught them all. 
jsdoc added.

ts/a11y/explorer/KeyExplorer.ts
zorkow: ditto

zorkow:
Incorporated the review suggestions.
Also I fixed an issue with the walking when elements are being collapsed. They need to have some `data-semantic-` attributes set explicitly.

PTAL.

ts/a11y/explorer.ts
```diff
@@ -145,11 +139,11 @@ export function ExplorerMathItemMixin<B extends Constructor<HTMLMATHITEM>>(
      */
     public updateDocument(document: ExplorerMathDocument) {
       super.updateDocument(document);
-      this.refocus && this.typesetRoot.focus();
+      this.explorers.speech().restarted = this.refocus;
```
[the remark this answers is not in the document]
zorkow: Done.

dpvc (approved):
lgtm

PR #1042, opened 2024-01-24
Make sure math item have their metrics set before lazy typesetting.  (mathjax/MathJax#3167)

dpvc:
This PR resolves a problem when the lazy typesetter is used with the assistive MathML extension and the initial typesetting is disabled with SVG output.  In that case, any math that is showing initially will get a bad `viewBox` for their `svg` element due to the fact that the metrics for the surrounding font haven't been computed.  We fix this by checking whether the metrics are available and setting the state to before the measuring of metrics in that case.

Resolves issue mathjax/MathJax#3167.


zorkow (approved):
lgtm.

PR #1043, opened 2024-01-26
Fix issues with line breaks and spacing in SVG output. (mathjax/MathJax#3166)

dpvc:
This issue fixes several issues with in-line line breaking in SVG mode that arise when extra spacing is placed around operators, as is done by `\implies`, or `A\;+\;B`, for example.  When in-line breaks are allowed, SVG output will lose the spacing around the operator (both the extra spacing as well as the normal operator spacing).

Both issues are cause primarily by the `computeLineBBox()` function for the SVG `mspace` wrapper.  The size of the space was lost because the width was not being added to the `LineBBox`, and the operator spacing was lost because the `isFirst` property was assuming that the mspace had no width (i.e., was used for line-breaking only).

I also moved `* scale` inside a conditional since there is no need to multiply 0 by the scale.

The `LineBBox.append()` is modified to make sure that `originalL` is not lost if it has already been set earlier on.

Resolves issue mathjax/MathJax#3166.

### DIFF
--- a/ts/output/common/LineBBox.ts
+++ b/ts/output/common/LineBBox.ts
@@ -99,7 +99,7 @@ export class LineBBox extends BBox {
    */
   public append(cbox: LineBBox) {
     if (this.isFirst) {
-      cbox.originalL = cbox.L;
+      cbox.originalL += cbox.L;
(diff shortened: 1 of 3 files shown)


zorkow (approved):
lgtm.

PR #1044, opened 2024-01-27
Fix menu to not rerender if no typesetting has occurred.  (mathjax/MathJax#3167)

dpvc:
If the MathJax configuration includes `startup {typeset: false}` and also has menu settings that require loading extensions (e.g., if assistive-mml is on) or the user has set the menu so that something needs to be loaded (e.g., the renderer has been changed in the contextual menu), then the math will be typeset despite the `typeset: false` setting (because the Menu code will think a pre-render is necessary because of the menu settings).

This PR adds a `rerenderPromise` to the `ts/components/startup.ts` file that is initially `null` (meaning no typesetting has occurred), and is set to the `MathJax.startup.promise` once typesetting has been requested.  That lets the menu code determine if a re-render is needed (i.e., typesetting has already occurred), and gives it a promise to use to queue its `rerender` action so as to synchronize with other typesetting actions.

Resolves issue mathjax/MathJax#3167.

### DIFF
--- a/ts/components/startup.ts
+++ b/ts/components/startup.ts
@@ -96,6 +96,7 @@ export interface MathJaxObject extends MJObject {
     elements: any[];
     document: MATHDOCUMENT;
     promise: Promise<void>;
+    rerenderPromise: Promise<void>;
     /* tslint:disable:jsdoc-require */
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #1046, opened 2024-01-28
Update LiteParser's serializeXML.  (mathjax/MathJax-demos-node#58)

dpvc:
This PR fixes two problems with the LiteDOM `serializeXML()` method.  It protects `<` and `>` in attributes, which is not needed for HTML, but is for XHTML, and it adds namespaces for `html`, `svg`, and `math` elements, if they don't already have one.

Resolves an issue from mathjax/MathJax-demos-node#58.


zorkow (approved):
lgtm.

PR #1048, opened 2024-02-06
Fix processing of \U{...}.

dpvc:
This PR fixes the index being used when `\U{...}` has been found so that it is at the correct location to continue processing.  Without this, two occurrences of `\U` in a row will not process the second one.  E.g., `\text{\U{61}\U{62}}` would produce `a\U{62}` rather than `ab`.


zorkow (approved):
lgtm.

PR #1049, opened 2024-02-06
Move \divsymbol to physics package and add \divisionsymbol. (mathjax/MathJax#3173)

dpvc:
This PR moves the `\divsymbol` mapping from the `base` package to the `physics` package, where it comes from in actual LaTeX, and adds `\divisionsymbol` as well.  It looks like the LaTeX `physics` package uses `\divisionsymbol`, while `physics2` uses `\divsymbol`.

Resolves issue mathjax/MathJax#3173.


zorkow (approved):
lgtm.

PR #1051, opened 2024-02-08
Refactor/explorer expansion

zorkow:
The PR 

* adds keyboard support for collapsing and expanding elements.
* fixes the problem that on collapse  the explorer might restart on the wrong collapsed expression.

When a node is collapsed during the exploration, the explorer restarts on that node if it is still in the tree. If the node is no longer in the tree it should restart on the collapsed element that contained the original node in its subtree. Previously we would only look for a collapsed node (`dummy` type), which could be the wrong one if there are multiple in the expression. Now we do a path to root analysis in the original enriched node to find the right maction element in the typeset node to restart on.


dpvc (approved):
This looks fine, though I didn't actually run it as you said a new version of SRE is needed.

PR #1052, opened 2024-02-08
Refactor/explorer depth

zorkow:
The PR
* adds depth message output to the explorer
* generalises the handling of `InPlace` moves

`InPlace` moves are those, that generate a different speech output without actually walking in the DOM, like summary or depth. They need to be copied into the `aria-label`, the node needs to be refocused to alert a screen reader and they need to be cleaned up on the next move (ie., the original speech needs to be copied back into the `aria-label`).

This needs some changes in SRE that I still need to push.

### DIFF
--- a/ts/a11y/explorer/KeyExplorer.ts
+++ b/ts/a11y/explorer/KeyExplorer.ts
@@ -28,7 +28,7 @@ import type { ExplorerMathItem } from '../explorer.js';
 import {Explorer, AbstractExplorer} from './Explorer.js';
 import {ExplorerPool} from './ExplorerPool.js';
 import {MmlNode} from '../../core/MmlTree/MmlNode.js';
-import { honk } from '../speech/SpeechUtil.js';
+import { honk, InPlace } from '../speech/SpeechUtil.js';
 import {Sre} from '../sre.js';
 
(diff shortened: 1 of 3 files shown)


ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -295,6 +295,7 @@ export class SpeechExplorer extends AbstractExplorer<string> implements KeyExplo
     ['<', this.nextStyle.bind(this)],
     ['x', this.summary.bind(this)],
     ['-', this.expand.bind(this)],
+    ['d', this.depth.bind(this)],
```
dpvc, line 298: I'm wondering if this should be a symbol rather than a letter, like `=` or `#`, instead of a letter that ties to a word in a particular language.

dpvc (approved):
This looks fine, other than question about key to use for depth.

ts/a11y/explorer/KeyExplorer.ts
zorkow: The problem with these symbols is that on many keyboard layouts they are only reachable via shift.
I will merge for now and then change it in the next PR.

PR #1053, opened 2024-02-08
Fix and catch speech computation errors, add missing braille option

zorkow:
PR contains some corrections.

* Builds the correct speech string from the SSML also for the `mjx-container` element's `aria-label`.
* No speech is computed when `enableEnrichment` is set to `false`. Also catches error on speech computation when `enableEnrichment` is `true` but the element is actually not enriched. I think this currently only happens in the lab or when an enrichment error is caught.
* Adds missing `enableBraille` option.

### DIFF
--- a/ts/a11y/semantic-enrich.ts
+++ b/ts/a11y/semantic-enrich.ts
@@ -244,10 +244,8 @@ export function EnrichedMathItemMixin<N, T, D, B extends Constructor<AbstractMat
      */
     public attachSpeech(document: MathDocument<N, T, D>) {
       if (this.state() >= STATE.ATTACHSPEECH) return;
-      if (this.isEscaped || !document.options.enableEnrichment) {
-        this.state(STATE.ATTACHSPEECH);
-        return;
-      }
+      this.state(STATE.ATTACHSPEECH);
+      if (this.isEscaped || !document.options.enableEnrichment) return;
       let [speech, braille] = this.existingSpeech();
       let [newSpeech, newBraille] = ['', ''];
       if ((!speech && document.options.enableSpeech) ||
@@ -262,10 +260,7 @@ export function EnrichedMathItemMixin<N, T, D, B extends Constructor<AbstractMat
       }
       speech = speech || newSpeech;
       braille = braille || newBraille;
-      if (!speech && !braille) {
-        this.state(STATE.ATTACHSPEECH);
-        return;
-      }
+      if (!speech && !braille) return;
       const adaptor = document.adaptor;
       const node = this.typesetRoot;
       if (speech) {
@@ -279,7 +274,6 @@ export function EnrichedMathItemMixin<N, T, D, B extends Constructor<AbstractMat
       }
       this.outputData.speech = speech;
       this.outputData.braille = braille;
-      this.state(STATE.ATTACHSPEECH);
     }
 
   };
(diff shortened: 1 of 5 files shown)


ts/a11y/semantic-enrich.ts
```diff
@@ -244,12 +244,21 @@ export function EnrichedMathItemMixin<N, T, D, B extends Constructor<AbstractMat
      */
     public attachSpeech(document: MathDocument<N, T, D>) {
       if (this.state() >= STATE.ATTACHSPEECH) return;
+      if (this.isEscaped || !document.options.enableEnrichment) {
+        this.state(STATE.ATTACHSPEECH);
```
dpvc, line 248: Do any of the calls below cause a MathJax retry error?  (That is, do any call `mathjax.retryAfter()`?)

If not, then you can do `this.state(STATE.ATTACHSPEECH)` above line 247, and then just return here and the other places where the state is set.  That would avoid having to set the state in several places.  But it only works if the rest of the code will always complete and not restart the typesetting via `retryAfter()`.

ts/a11y/semantic-enrich.ts
```diff
@@ -244,12 +244,21 @@ export function EnrichedMathItemMixin<N, T, D, B extends Constructor<AbstractMat
...
+        return;
+      }
       let [speech, braille] = this.existingSpeech();
       let [newSpeech, newBraille] = ['', ''];
       if (!speech || !braille ||
         document.options.enableSpeech || document.options.enableBraille) {
```
dpvc, line 254: I didn't catch this earlier, but I think this condition is wrong.  That is, if `document.options.enableSpeech` is false, you should not generate speech even if `!speech` is true, and similarly for braille.  So I think this should be

``` js
if ((!speech && document.options.enableSpeech) || (!bralle && document.options.enableBraille)) {
```

dpvc (changes requested):
There are a couple of items identified here for your consideration.

Also, it looks like the function 

https://github.com/mathjax/MathJax-src/blob/799ad89bcfc2dd750c7e970214d87065117a4c01/ts/a11y/semantic-enrich.ts#L290-L306

is no longer used, so can be eliminated.

Also, I'm wondering if

``` js
if (this.options.enableSpeech || this.options.enableBraille) {
...
}
```

should be placed around lines

https://github.com/mathjax/MathJax-src/blob/799ad89bcfc2dd750c7e970214d87065117a4c01/ts/a11y/semantic-enrich.ts#L417-L419

like is done at line 430 below.  That would save having to loop through all the math when we aren't adding speech or braille.

ts/a11y/semantic-enrich.ts
zorkow: It should not. The only place that triggers a `retryAfter` is in line 183:
https://github.com/mathjax/MathJax-src/blob/799ad89bcfc2dd750c7e970214d87065117a4c01/ts/a11y/semantic-enrich.ts#L183
But that should have run before `attachSpeech` at all times.

ts/a11y/semantic-enrich.ts
zorkow: Thanks. Good catch.

zorkow:
I've made changes except the `attachSpeech` one and I've made a change in `package.json` to the new SRE release.

dpvc (approved):
OK, this all looks good.  If you want to do the `STATE.ATTACHSPEECH` change, that is great.  Approved either way.

zorkow:
> OK, this all looks good. If you want to do the `STATE.ATTACHSPEECH` change, that is great. Approved either way.

I've tested with the states and I've made the simplification.

PR #1054, opened 2024-02-10
Feature/better honking

zorkow:
One more PR. Fixes error in expressions that already contain an `maction` (i.e., not one introduced by `collapsible` option), where `visitMactionNode` can introduce a second duplicate `data-maction-id` attribute. This breaks XML parsing. Previously this was not an issue as we used HTML parsing, but now we are using SRE's XML parsing. 

Also some improvements to expansion/collapse/action:
* Only honks when no action can be taken on a node.
* Uses the `Enter` key for action as this is what the aria tree pattern expects for triggering actions.


dpvc (approved):
This looks good.

PR #1058, opened 2024-02-15
Fixes issues with firefox and safari.

zorkow:
These are the workarounds for the issues regarding voice events on Firefox.
@dpvc can you have a look if this works for Safari as well?

### DIFF
--- a/components/bin/build
+++ b/components/bin/build
@@ -39,7 +39,7 @@ const EXPORTPATTERN =
       /(^export(?:\s+default)?(?:\s+abstract)?\s+(?:[^ {*}]+\s+(?:enum\s+)?[a-zA-Z0-9_.$]+|\{.* as .*\}))/m;
 
 const EXPORT_IGNORE = ['type', 'interface'];
-const EXPORT_PROCESS = ['let', 'const', 'var', 'function', 'class', 'namespace', 'as'];
+const EXPORT_PROCESS = ['let', 'const', 'var', 'function', 'class', 'namespace', 'enum', 'as'];
 
 /**
  * The module type to use ('cjs' or 'mjs')
@@ -159,7 +159,7 @@ function processParts(parts) {
   for (let i = 1; i < parts.length; i += 2) {
     const words = parts[i].split(/\s+/);
     const n = words.length;
-    const type = (words[n - 2] === 'enum' ? words[n - 3] : words[n - 2]);
+    const type = (words[n - 2] === 'enum' && n > 3 ? words[n - 3] : words[n - 2]);
     const name = words[n - 1].replace(/\}$/, '');
 
     if (words[1] === 'default' || type === 'default') {
--- a/components/mjs/a11y/explorer/config.json
+++ b/components/mjs/a11y/explorer/config.json
@@ -6,7 +6,6 @@
   "webpack": {
     "name": "a11y/explorer",
     "libs": [
-      "components/src/ui/menu/lib",
       "components/src/a11y/semantic-enrich/lib",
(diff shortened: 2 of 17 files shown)


dpvc (changes requested):
This is still not working with Firefox, but that seems to be due to the fact that the `role` attributes are missing.  I think you did something in SRE to fix this, but I don't have that updated SRE, so my Firefox isn't working.

Safari is nearly working.  I can click to activate the explorer, and use keyboard to move around the equation, and the speech and braille show up properly.  But if the explorer is active, then clicking does not select the clicked item.  I tracked this down in the code, and the difference between Safari and Chrome is in how this line

https://github.com/mathjax/MathJax-src/blob/d17d8f0d0010c78160dcebeee296c8df3c89eafe/ts/a11y/explorer/KeyExplorer.ts#L180

is handled. In Safari (and presumably Firefox for you), when the `tabindex` is removed, the browser fires a focus-out even immediately, and the KeyExplorer's `FocusOut` method is called immediately before the rest of the `Click` handler is run.  In particular, that means that the `Stop()` method will be run before the rest of the `Click` method, and so `this.active` will be set to `false` before `this.Start()` is run at line 184.

In Safari, moving the `tabindex` does **not** fire a focus-out event, so `this.active` is still `true` when `this.Start()` is called, and because `Start()` returns if `this.active` is true on this line

https://github.com/mathjax/MathJax-src/blob/d17d8f0d0010c78160dcebeee296c8df3c89eafe/ts/a11y/explorer/KeyExplorer.ts#L452

in Safari, the rest of the `Start()` doesn't run, and the selected node is never switched.

One solution would be to add

``` js
this.FocusOut(null);
```

right after line 180 shown above.  That will force the focus-out action in Safari, and for browsers that fire a focus-out event at this point, it should still be OK since `FocusOut()` exists if `this.active` is false, as it should be after the first `FocusOut()` is run.  Of course, you might have a better idea for this.

I'll have to check Firefox again once I have the SRE update that allows the `role` attributes to be inserted.

dpvc:
Here are the list of continuing issues from our chat today:

* Hover highlighting leads to an error.
* While explorer is active, all key events are blocked (including reload, etc.)
* Can't start explorer if both speech and braille are unchecked.  I believe that this is due to the fact that there are no role attributes in that case, because those seem to be added during the speech/braille creation at
* 
* , which isn't being called

dpvc:
Here are the list of continuing issues from our chat today:

* Hover highlighting leads to an error.
* While explorer is active, all key events are blocked (including reload, etc.)
* Can't start explorer if both speech and braille are unchecked.  I believe that this is due to the fact that there are no role attributes in that case, because those seem to be added during the speech/braille `creationSpeech()` function in semantic-enrich, which isn't being called.
* After a change in language, you must do a move to get new language (but you know this one already).

dpvc:
I figured out the third issue above, and have fixed it in my menu update branch (not yet pushed).

dpvc:
> That was actually an issue related to how explorers were attached.

I fixed it in the `sre-menus` branch, as I mentioned above, so you don't need to worry about that one.

zorkow:
> After a change in language, you must do a move to get new language (but you know this one already).

This is now fixed. However, we still have an issue when changing rule sets/preference settings. I will tackle this in a new PR to simplify review.

zorkow:
@dpvc as discussed today, this should be ready to merge.

dpvc (approved):
This all seems to work now, except that I still can't get hover or flame highlighting to work.  But I may not know what I'm doing, so that may be the problem.  We can chat about it in our next meeting.

PR #1059, opened 2024-02-17
Make stretchy paried delimiters not be class INNER.  (mathjax/MathJax#3183)

dpvc:
This PR fixes an issue with the `mathtools` package's paired delimiter implementation.  The used to have the default class of  `INNER`, but now they have class `OPEN` and are followed by an empty item of class `CLOSE`, so the stretchy delimiters now act like regular open and close delimiters, but are stretchy.

Resolves issue mathjax/MathJax#3183.


zorkow (approved):
lgtm.

PR #1061, opened 2024-02-18
Make \symbf and \symsf handle Greek italics as in LaTeX.

dpvc:
This PR allows `\symbf` and `\symsf` to put lower-case Greek letters into italics, as they do in actual LaTeX.  These seem to be the only two of the `\sym...` macros that do (and the only two for which there are mathvariants that would work).

See this [forum post](https://groups.google.com/g/mathjax-users/c/ZnZ_hGPwioo).


pkra:
I realize that these have been in MathJax already despite being from unicode-math (and xelatex/luatex land). We don't have those macros in https://github.com/AmerMathSoc/mathjax-unicode-math but I'm happy to add them there if that's helpful.

dpvc:
> I'm happy to add them there if that's helpful.

Thanks for the offer.  I think we will leave them here for now.  The base configuration does have a hodgepodge of pieces from various packages.  At some point it might want to be straightened out, but not just now.

zorkow (approved):
lgtm.

PR #1062, opened 2024-02-21
Add ability for extensions to set the SMP mappings for new variants

dpvc:
This PR provides a mechanism for font extensions to set the SMP remapping data for new variants that the extension defines.  It also adds a new options section to the `SmpData` type that allows additional mappings, and uses that to set the extra mappings for the bold variant that were hard-coded into the `remapSmpChars()` method.  Finally, it allows the SMP mapping for a new variant to be defined in terms of previous SMP mappings by specifying the string name of a variant whose mapping is the same.  That way, fonts like `dsfonts` that is a double-struck font can simply specify that the `double-struck` mapping is to be used rather than duplicating the data for that.

### DIFF
--- a/ts/output/common/FontData.ts
+++ b/ts/output/common/FontData.ts
@@ -184,7 +184,7 @@ export type SmpMap = {
  * Data for Math Alphanumeric conversion:  starting positions for
  *  [Alpha, alpha, Greek, greek, Numbers]
  */
-export type SmpData = [number, number, number?, number?, number?];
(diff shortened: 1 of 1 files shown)


dpvc:
Viewing the file difference with whitespace changes hidden makes it easier to see the differences, since there are indentation changes due to an earlier `return` (I know you link those).

zorkow (approved):
lgtm.

PR #1063, opened 2024-02-23
Update config file and dependencies (bboldx)

dpvc:
This PR fixes the config file for bboldx, and adds this and the other fonts to the dependencies file (since `bboldx` depends on `textmacros`).


zorkow (approved):
lgtm.

PR #1066, opened 2024-02-26
Fix error on empty vec. (mathjax/MathJax#3195)

tsutaj:
This PR resolves https://github.com/mathjax/MathJax/issues/3195

Applied https://github.com/asciimath/asciimathml/commit/36b3cb0a1bac755b4f66a0b49e2e7c9e85cf55bb.


dpvc:
We don't make individual adjustments to the AsciiMath code, but rather, when a new version is released, we update the AsciiMath code to the current version of AsciiMath all at once.  The AsciiMath input jax needs to be updated to be an actual v3 one rather than the hack that patches in the legacy v2 version.  In any case, as we are not updating the v2 legacy code here, as the v2 version is the primary version for this, and we take the legacy version here from that.

dpvc:
Closing this PR, since it will be taken care of when AsciiMath is updated in the next v2 version, or when an actual v3 input jax is created.

PR #1068, opened 2024-03-01
Update MathJax contextual menus for assistive tools

dpvc:
This PR implements the new menu structure that we discussed over the last few developer meetings.

While updating the menus, we also normalize how the settings are handled, so that the `enableSpeech`, `enableBraille`, `enableEnrichment` and `enableExplorer` document options control whether those actions are taken, and are controlled automatically by the menu settings.  When enrichment is enabled, the accessibility submenus are made active (and the `enable...` options are set to true).  This allows the explorer to run, and whatever settings are present in the `Speech`, `Braille`, and `Explorer` menus to take effect.  Those submenus control the `a11y` settings of the document's options (and the `menuOptions.settings` section, which is what initializes the `a11y` values when the menu code is loaded, and are what are stored in the menu's `localStorage` cache).

All of this means that, for example, for speech to be added, both the document's `enableSpeech` and `a11y.speech` options must be true, and for the subtitles to be shown, the `a11y.substitles` option must also be true.  Similarly for the Braille settings.  This generated the changes to `KeyExplorer.ts` for example.

The `setA11yOptions()` function in `explorer.ts` is simplified by moving the handling of `locale` to the `setA11yOption()` function where all the other special keys are processed.

The `ExplorerPool.ts` file is modified to attach explorers when speech isn't enabled but one of Braille or magnification is (there were not separate explorer keys for these, as the speech explorer handles all three).

In `semantic-enrich.ts`, we add code to remove enrichment if the state drops below the enrichment state, and we add an error function to process enrichment errors that can be overriden by the user.

The `MJContextMenu` is modified to:
* Add a `Braille Code` item to the `Show As` and `Copy to Clipboard` submenus (enabled when Braille generation is active).
* Modify the `DynamicSubmenus` code to allow enabling of a non-empty menu to be controlled by a menu setting (so that, for example, the `language` submenu does not get enabled when speech is not being generated). 
* The `findID()` method has been rewritten to not use `find()`, which descends into submenus too soon for `findID()`.

The `SpeechMenu.ts` file uses the new `DynamicSubmenus` parameter to handle the language and clearspeak submenus.

The `MenuHandler` adds the `enableSpeech` and `enableBraille` options, and handles the transfer of the menu settings to the various `enable...` options.  Note that `enableExplorer ` and `enableEnrichment` are initially the same, since enrichment is all that is needed for the explorer at this point.

Finally, the `Menu.ts` file has significant changes to implement the new menu structure and loading of the needed components when the settings change.  The `explorer` menu setting has been removed, since there is no longer an explicit menu setting for it (instead, it is controlled by enrichment setting).

It may be useful to turn off white-space difference when looking at the changes.

Note that you will need to use the `lab-menu-update` branch in the `MathJax-dev` repository to get the needed changes to the lab in order for the lab to work with this branch.

### DIFF
--- a/components/bin/build
+++ b/components/bin/build
@@ -39,7 +39,7 @@ const EXPORTPATTERN =
       /(^export(?:\s+default)?(?:\s+abstract)?\s+(?:[^ {*}]+\s+(?:enum\s+)?[a-zA-Z0-9_.$]+|\{.* as .*\}))/m;
 
 const EXPORT_IGNORE = ['type', 'interface'];
-const EXPORT_PROCESS = ['let', 'const', 'var', 'function', 'class', 'namespace', 'as'];
+const EXPORT_PROCESS = ['let', 'const', 'var', 'function', 'class', 'namespace', 'enum', 'as'];
 
 /**
  * The module type to use ('cjs' or 'mjs')
@@ -159,7 +159,7 @@ function processParts(parts) {
   for (let i = 1; i < parts.length; i += 2) {
     const words = parts[i].split(/\s+/);
     const n = words.length;
-    const type = (words[n - 2] === 'enum' ? words[n - 3] : words[n - 2]);
+    const type = (words[n - 2] === 'enum' && n > 3 ? words[n - 3] : words[n - 2]);
     const name = words[n - 1].replace(/\}$/, '');
 
     if (words[1] === 'default' || type === 'default') {
--- a/components/mjs/a11y/explorer/config.json
+++ b/components/mjs/a11y/explorer/config.json
@@ -6,7 +6,6 @@
   "webpack": {
     "name": "a11y/explorer",
     "libs": [
-      "components/src/ui/menu/lib",
       "components/src/a11y/semantic-enrich/lib",
       "components/src/a11y/sre/lib",
       "components/src/input/mml/lib",
--- a/components/mjs/a11y/explorer/explorer.js
+++ b/components/mjs/a11y/explorer/explorer.js
@@ -1,17 +1,7 @@
 import './lib/explorer.js';
 
-import {combineDefaults} from '#js/components/global.js';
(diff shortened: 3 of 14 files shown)


dpvc:
I've made the name change to the `Semantic Enrichment` menu item.

I also made changes to some of the component `config.json` files to update them to handle the current interdependencies better.  E.g., the explorer component no longer relies on `ui/menu`, so that has been removed from its `config.json` file and from the `dependencies.js` file, and since uses the `a11y/speech/SpeechUtil.ts` file and so does `semantic-enrich`, this is now shared between them rather than both including separate copies of it.

I've looked at the conflicts with the updates to the `fix/ff_safari_issues` branch, and there are four to be dealt with.

* In `explorer.ts`, I think you can remove the new lines that set the enable options, as these are handled by the menu (when you check or uncheck `Generate` in the `Speech` submenu, `enableSpeech` is set or unset (so `enableSpeech` tracks `options.a11y.speech` automatically.  Similarly for Braille.  Note that since these lines are in the `explorable()` function, and that runs *after* the `attachSpeech()` function, setting these options here is too late anyway.  If you want to have the a11y options for the enable options, it might be best to do that in the constructor rather than in `explorable()`.

* In `ExplorerPool.ts`, I had handled the speech/braille/keyMagnifier issues by modifying line 239 in the conflict editor.  You can decide which method to use, but note that I made an `a11y` variable at line 220 of the conflict editor, and that is used later, so you should keep that line.  My lines 221 and 222 should be replaced by your line 224, though.  If you keep your changes, you can use `a11y` in line 231 to simplify that (and probably move it onto the previous line).  And of course you would remove my changes from line 239.

* In `KeyExplorer.ts`, I think you want to keep my lines, for when the menu isn't being used.

* In `semantic-enrich.ts`, my changes are equivalent to yours, but I have set up an `options` variable earlier that I use here, where you use the full path to the options.

So I think there is not much problem with the conflicts, and they are easily sorted out.

PR #1069, opened 2024-03-06
Handle primes in munderover as in msubsup.  (mathjax/MathJax#3202)

dpvc:
This PR resolves the issues raised in mathjax/MathJax#3202, namely that primes are not placed properly on operators.  E.g., `\sum_n'` should act like `\sum_n^\prime` but doesn't.  The reason is that the `PrimeItem` and `BaseMethods.Prime` function don't treat `munderover` the same as `msubsup`, so this PR fixes that.

This also fixes the problem with `\int\limits'` not working like `\int\limits^\prime`.

It also fixes the problem with `\sum''` and larger numbers of primes not using the proper size in display mode.

Resolves issue mathjax/MathJax#3202.


zorkow (approved):
lgtm.

PR #1072, opened 2024-03-09
Make non-shown regions be display:none

dpvc:
This PR makes the various regions be `display: none` when not in use, so that there are not small white "artifacts" floating around when they are not being used.

I've made the non-show versions be just `display: none`, and moved any needed CSS from there to the `_show` version.


zorkow (approved):
lgtm.

PR #1073, opened 2024-03-09
Make textcomp package register with textmacros automatically.

dpvc:
This PR causes the `textcomp` package to add itself to the `textmacros` packages automatically.  In order to allow it to work with `\require{textcomp}` (and to work better with the Lab), we do this through the `config()` function of the configuration rather than setting it via `options`.  This is because when `\require{textcomp}` is processed, its dependecies are loaded and added to the configuration first, then `textcomp` is added.  Because `textmacros` is a dependency, it will be added and configured *before* `textcomp` adds to the `textmacros.packages` if it is placed in the `options.textmacros.packages` configuration option.  Here, using `config()`, we know that `textmacros` will already be configured, so we can add to the `textConf` configuration directly.

It may be possible to have `\require` collect the `config()` methods from the dependencies and run them *after* all the initialization is done, but I didn't want to make that big a change here.  This should the job for now.

### DIFF
--- a/ts/input/tex/textmacros/TextMacrosConfiguration.ts
+++ b/ts/input/tex/textmacros/TextMacrosConfiguration.ts
@@ -39,6 +39,7 @@ import './TextMacrosMappings.js';
  */
 export const TextBaseConfiguration = Configuration.create('text-base', {
   parser: 'text',
+  priority: 1,
   handler: {
     character: ['command', 'text-special'],
     macro: ['text-macros']
@@ -96,6 +97,7 @@ function internalMath(parser: TexParser, text: string, level?: number | string, 
 //  The textmacros package configuration
 //
 export const TextMacrosConfiguration = Configuration.create('textmacros', {
+  priority: 1,
   /**
    * @param {ParserConfiguration} config   The configuration object we are being configured within
    * @param {TeX<any,any,any>} jax         The TeX input jax in which we are running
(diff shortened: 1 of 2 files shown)


ts/input/tex/textmacros/TextMacrosConfiguration.ts
```diff
@@ -39,6 +39,7 @@ import './TextMacrosMappings.js';
  */
 export const TextBaseConfiguration = Configuration.create('text-base', {
   parser: 'text',
+  priority: 1,
```
zorkow, line 42: Is there a reason for the priority?

zorkow (approved):
Question on priority. Otherwise looks fine.

ts/input/tex/textmacros/TextMacrosConfiguration.ts
dpvc: It was so that the base configuration is done first and then others that override anything in it don't need to specify their own priorities.  But I think that the packages may be processed in reverse order, which means I should make this 10 instead.

dpvc: Just checked, and this does mean text-base will have its `init()` and `config()` methods called before the other configurations.  So that means the default priority of 5 will put others after this, without having to set the priority.

PR #1074, opened 2024-03-10
Fix dsfont name in dependencies, and add new fonts to sources

dpvc:
This PR fixes a typo in the `dsfont` name in the dependencies file, and adds the source locations to the source file.

Note that the new font packages are listed in the lab checkboxes, but won't actually work in the lab, since the component files that load the fonts themselves aren't run by the lab when the box is checked.  It might be possible to make that work, but I will need to look into it further.


zorkow (approved):
lgtm.

PR #1075, opened 2024-03-14
Make tex font packages work with \require

dpvc:
This PR makes the changes needed to the font extension infrastructure to allow a TeX font package to be loaded via `\require` (e.g., `\require{dsfont}`) and have it load the needed font.

The `components/mjs/input/tex/extension.js` file exports a function that does the work of loading and adding a font extension, simplifying the `dsfont`, `bbm` and `bboldx` component files.  This sets up the needed path for the extension, and the load hook that loads the font extension and adds it into the output jax when it is loaded.

To do this, the common output jax gets a new `addExtension()` function that looks up the font extension and adds it to the current font for that output jax.  The chtml output jax overrides this so that it can add the needed CSS rules to load the fonts needed for the extension, if any.  These come from a modified `addExtension()` in the `ChtmlFontData` object, which produces the needed `@fontface` rules and passes them back so that the output jax's `addExtension()` can insert the rules into the chtml stylesheet.

The common FontData object now saves the complete `FontExtensionData` object so that it can be used in the common output jax's `addExtension()` to add the extension to its existing font.

Finally, the `bboldx` extension is modified to configure the `textmacros` packages in its `config()` method rather than via its `options` setting.  This is because when you use `\require{bboldx}`, this forces the `textmacros` package to be loaded and configured before `bboldx` is, and so the `textmacros` packages have already been processed before `bboldx` adds its package.  So this allows `bboldx` to adds the package after the fact.  You need to merge the `textcomp-textmacros` branch in order for this to work, since it includes changes to `textmacros` that are needed for the `config()` method to work (the `textConf` object).

I'm sorry to be adding yet another PR for beta.5, but since we are including the font extensions, it is probably best to have them work with `\require`.

### DIFF
--- a/ts/output/chtml/FontData.ts
+++ b/ts/output/chtml/FontData.ts
@@ -25,7 +25,7 @@ import {CharMap, CharOptions, CharDataArray, VariantData,
         DelimiterData, FontData, FontExtensionData, DIRECTION} from '../common/FontData.js';
 import {Usage} from './Usage.js';
 import {StringMap} from './Wrapper.js';
-import {StyleList, StyleData} from '../../util/StyleList.js';
+import {StyleList, StyleData, CssStyles} from '../../util/StyleList.js';
 import {em} from '../../util/lengths.js';
 
 export * from '../common/FontData.js';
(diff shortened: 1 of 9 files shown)


ts/output/chtml/FontData.ts
```diff
@@ -188,6 +188,24 @@ export class ChtmlFontData extends FontData<ChtmlCharOptions, ChtmlVariantData,
...
+  ): string[] {
+    super.addExtension(data, prefix);
+    if (data.fonts && this.options.adaptiveCSS) {
+      const css = {};
+      const styles = new CssStyles();
+      (this.constructor as typeof ChtmlFontData).addDynamicFontCss(css, data.fonts, data.fontURL);
+      styles.addStyles(css);
+      return styles.getStyleRules();
+    }
+    return [];
```
zorkow, line 206: Do that case first with the negated `if` condition and safe indentation for the main logic.

zorkow (approved):
One cosmetic rewrite of an `if` condition.
Otherwise looks good.

PR #1076, opened 2024-03-19
Fix use of textmacros in textcomp/bboldx, and update source.js and dependencies.js

dpvc:
This is a small PR that does several things:

1.  It fixes the `config()` functions in `textcomp` and `bboldx` that use `textmacros` to check that `textmacros` has been configured before trying to use the `textmacros` package data (avoiding an error message when it is not available).

2.  It updates the `source.js` and `dependencies.js` files to include some missing packages, and sorts the lists for easier maintenance.

3. Fixes the configuration file for `texhtml` so that the version check is correct.

4. Updates the `extensions.js` so that font extensions have an output jax as a dependency, so that if components are loaded separately rather than through a combined component, the output jax will be loaded before the font extension.


zorkow (approved):
lgtm.

PR #1077, opened 2024-03-24
Handle pnpm node_module paths better in [node] prefix in messages

dpvc:
This is a small PR to improve the output from the `components/bin/pack` command for reporting files included from the `node_modules` directory.  Since `pnpm` uses symbolic links, the filenames reported are not as concise as they could be.  This alters the regex pattern used for the `[node]` prefix to collapse `.pnpm` links for shorter file names.


zorkow (approved):
lgtm.

PR #1078, opened 2024-03-27
Filter aria-owns attributes from SRE

dpvc:
This is a small PR to remove the `aria-owns` attributes from SRE when SRE attributes are being filtered from the MathML output in the "Show Math As" menu.


zorkow (approved):
lgtm.

PR #1079, opened 2024-03-29
Fix/explorer rules

zorkow:
Reenables the change of rule sets and preferences for speech generation.
* Uses a standard variable definition for`speechRules`.
* Switches to clearspeak as default, thereby also fixing a type `mathspeek` that lead to incorrect speech computation.
* Fixes an error when the selection box is closed, as it is not attached to a `mathitem`.

Still two issues:
* I still need to make an SRE release. 
* Currently one can select rule sets that do not always exist for every locale. There are two solutions for this:
1. disactivating the menu items that should not be chosen, which would require quite a bit of refactoring of the current speech menu code,
2. on the other hand I always wanted to ensure that SRE produces reasonable output even with a non-existing rule set is chosen. I go with this solution for now.

### DIFF
--- a/ts/ui/menu/Menu.ts
+++ b/ts/ui/menu/Menu.ts
@@ -146,7 +146,7 @@ export class Menu {
       speech: true,
       braille: true,
       brailleCode: 'nemeth',
-      speechRules: 'mathspeek-default'
+      speechRules: 'clearspeak-default'
     },
     jax: {
       CHTML: null,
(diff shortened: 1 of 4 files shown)


ts/ui/menu/Menu.ts
```diff
@@ -507,11 +507,20 @@ export class Menu {
...
+          },
+          setter: (value: string) => {
+            const [domain, style] = value.split('-');
+            this.settings['speechRules'] = value;
+            this.document.options.sre.domain = domain;
+            this.document.options.sre.style = style;
+            this.rerender(STATE.COMPILED);
+            this.saveUserSettings();
+          }
+        },
```
dpvc, line 523: You should be able to use `this.variable()` rather than `this.a11yvar()` to define this, so you don't have to do the `getter` and setter by hand (and won't need to include lines 517 and 521, which are part of what `this.variable()` does.  So

``` js
          this.variable<string>('speechRules', value => {
            const [domain, style] = value.split('-');
            this.document.options.sre.domain = domain;
            this.document.options.sre.style = style;
            this.rerender(STATE.COMPILED);
          }),
```

should do it (that is, changing `a11yvar` to `variable` and adding the `rerender()` call to the original lines 510 to 514 should be enough).

dpvc (approved):
One change suggestion to simplify the `speechRules` definition, but other than that looks good.

ts/ui/menu/Menu.ts
zorkow: I tried that and it not work. And it drove me crazy, so I decided to stick with the pedestrian version. 
But I will try again. Maybe I was just too thick the other day.

dpvc: I just tried this out, and it seemed to work fine for me.  Give it another try, and see if you have any problems.

PR #1080, opened 2024-04-01
Allow tex packages that load font extensions to work better with \require and renderer changes.

dpvc:
This PR is for beta.6.

The addition of the new TeX packages that load various double-struck fonts has broken the previous separation of input and output jax.  There are some assumptions that relied on this separation, so these new TeX packages complicate things somewhat.  In particular, switching renderers when one of these packages has been loaded via `\require` would not necessarily retain the font in the new renderer.  Furthermore, because of the way the extensions hooked into the renderer changes, leading a second font extension would interfere with the loading of the first one when a renderer change occurred.

The component packages for these extensions tried to set things up overcome some of these complications, but were not completely successful.  So this PR includes changes to properly handle switching renderers and fonts, both in regular pages and in the v4-lab.  In addition, it implements generic font extensions that are not tied to a particular base font, so that the extensions like `mathjax-dsfont` can be applied to any font.

One of the important changes is to move the code that adds an extension to a font from the TeX component file (where it currently is) into the font extension's component file instead.  So all the code controlling the extension being added to a font is now in the extension itself, rather than in the TeX package.  The TeX package file now simply has to set up the loading of the extension and that is it.  This is done via the changes to the `components/mjs/input/tex/extension.js` file.

To facilitate this, and to overcome the problem with multiple font extensions, the loader has been extended to include a new function `addPackageData()` that can be used to update the loader package information (like the `checkReady()` function), and that data can now include a list of additional packages to load after the initial one is loaded.  So we can now set up several font extensions to be loaded when the output jax is changed, for example.

There is also an array of extensions that need to be installed when the renderer changes that is set when one of the font extensions is loaded by `\require` (in case the new renderer has already been loaded, and so the file-based hooks will not be triggered).  This is referenced by the changes in the `\require` macro, which sets a corresponding array in the document's `Menu` object, and that is used by the menu code to install the needed extensions when the output jax changes.

Because of the `extraLoads` array, the `Package` object gets a new method `loadPromise()` that returns a promise that is resolved when all the extra files are loaded, and the `checkReady()` method (if any) resolves.  This is also used in the `components/mjs/output/util.js` file to make sure these files are loaded when a new output jax component is loaded.

Some other cleanup is also performed.  The `addExtension()` functions in the common and chtml output jax files now have the optional `prefix` argument that the `FontData` version has (though it is not currently used), and the instance versions of these functions now take a `FontExtensionData` object rather than an extension name, so that we don't have to store that data in the dynamic extension data.  (Since these `addExtension()` calls have been moved to the font extension components themselves, and those have access to the font data, that means we don't need to keep that data in the saved dynamic extension object any longer).  The dynamic extension data now uses a default value for the `prefix`, rather than requiring it to be set in the font extension component, as in the past.

Finally, some of the config files have been updated to remove `exclude` arrays that are no longer needed.

This branch requires changes in the font components, as well as the font tools, and also the v4-lab.  You will need to check out the `generic-extensions` branches in the font-tools and fonts repositories, then compile the font-tools, and rebuild the extension fonts (all three).  Then check out the `lab-extension-update` in the MathJax-dev repository.  Finally, link the newly built fonts into MathJax-dev's `node_modules` directory next to the other fonts.

### DIFF
--- a/ts/input/tex/require/RequireConfiguration.ts
+++ b/ts/input/tex/require/RequireConfiguration.ts
@@ -29,11 +29,12 @@ import {ParseMethod} from '../Types.js';
 import TexError from '../TexError.js';
 import {TeX} from '../../tex.js';
 
-import {MathJax} from '../../../components/global.js';
+import {MathJax} from '../../../components/startup.js';
 import {Package} from '../../../components/package.js';
 import {Loader, CONFIG as LOADERCONFIG} from '../../../components/loader.js';
 import {mathjax} from '../../../mathjax.js';
 import {expandable} from '../../../util/Options.js';
+import {MenuMathDocument} from '../../../ui/menu/MenuHandler.js';
 
 /**
  * The MathJax configuration block (for looking up user-defined package options)
@@ -147,11 +148,12 @@ export function RequireLoad(parser: TexParser, name: string) {
   if (!allowed) {
     throw new TexError('BadRequire', 'Extension "%1" is not allowed to be loaded', extension);
   }
-  if (Package.packages.has(extension)) {
-    RegisterExtension(parser.configuration.packageData.get('require').jax, extension);
-  } else {
+  if (!Package.packages.has(extension)) {
     mathjax.retryAfter(Loader.load(extension));
   }
+  const require = LOADERCONFIG[extension]?.rendererExtensions;
+  (MathJax.startup.document as MenuMathDocument)?.menu?.addRequiredExtensions?.(require);
+  RegisterExtension(parser.configuration.packageData.get('require').jax, extension);
 }
 
 /**
--- a/components/mjs/input/tex/extension.js
+++ b/components/mjs/input/tex/extension.js
@@ -1,24 +1,16 @@
 import {combineDefaults} from '#js/components/global.js';
 
 export function fontExtension(id, name, pkg = `@mathjax/${name}`) {
-  if (MathJax.config?.loader) {
-    const FONTPATH = (typeof document === 'undefined' ? pkg :
-                      `https://cdn.jsdelivr.net/npm/${name}`);
+  if (MathJax.loader) {
+    const FONTPATH = (typeof document === 'undefined' ? pkg : `https://cdn.jsdelivr.net/npm/${name}`);
     const path = name.replace(/-font-extension$/, '-extension');
     const extension = name.replace(/-font-extension$/, '');
+    const jax = (MathJax.config?.startup?.output || 'chtml');
     combineDefaults(MathJax.config.loader, 'paths', {[path]: FONTPATH});
-    combineDefaults(MathJax.config.loader, 'dependencies', {
-      [`[${path}]/chtml`]: ['output/chtml'],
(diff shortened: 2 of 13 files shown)


zorkow (commented):
When trying this out in the lab it installed the `mathjax-X-font-extension` packages but it was looking for the `mathjax-X-font` packages in `node_modules`. Is that correct? In other words, will the updated font extensions be published under the latter package name?

dpvc:
You're right, they should have `-extension`.  I'll fix that.  It is in the `v4-lab.js` file (setting the `loader.source` to redirect to the `node_modules` versions rather than the CDN versions.

zorkow:
After updating the lab, I am now getting the following error:

> Invalid option "enrich" (no default value).

I have not yet had a chance to look into it in detail why. I think it comes from the mml3 module.

zorkow:
I have been going back and forth with this a few times and I believe that I can rule out a caching error. It happens in multiple browsers, etc. The issue seems to be that some of the elements of `menuOptions.settings` are not in the named options.

Are you sure you have not forgotten to push a commit?

dpvc:
I'm using a test branch from `develop` that merges in the `font-extensions-support`.  It looks like the `font-extensions-support` branch by itself is not enough.  It looks like it was the `fix/ff_safari_issues` branch that includes the `sre-menus` changes that are what are needed to resolve this problem.  So try merging `font-extensions-support` into a branch off of `develop`.  Sorry for the bad instructions.  I knew the `sre-menu` branch had been merged, but didn't realize it wasn't in develop until after the `font-extensinos-support` branch was split from `develop`.

dpvc:
Now that the needed branches have been merged into `develop`, we should be able to test and merge this branch.  Can you give it a look?  The instructions about using the `generic-extensions` branches in the fonts and font-tools repositories still apply.  You may also have to make a `@mathjax` directory in the MathJax-src node_modules and link fonts and font-tools into it in order to build the extensions, and you probably have to build the generic font in the fonts directory.

zorkow:
I've been trying to build the `generic` font but I am getting an `esm` error. I have to admit that one has defeated me, even when I try to build things manually. Maybe we could have a f2f for that.

ts/input/tex/require/RequireConfiguration.ts
```diff
@@ -147,11 +148,12 @@ export function RequireLoad(parser: TexParser, name: string) {
...
     throw new TexError('BadRequire', 'Extension "%1" is not allowed to be loaded', extension);
   }
-  if (Package.packages.has(extension)) {
-    RegisterExtension(parser.configuration.packageData.get('require').jax, extension);
-  } else {
+  if (!Package.packages.has(extension)) {
     mathjax.retryAfter(Loader.load(extension));
   }
+  const require = LOADERCONFIG[extension]?.rendererExtensions;
+  (MathJax.startup.document as MenuMathDocument)?.menu?.addRequiredExtensions?.(require);
```
zorkow, line 155: Will this work even if `require` is `undefined`?

zorkow (approved):
Just one concern. O/w lgtm.

PR #1084, opened 2024-04-18
Correct width for unknown fonts in scaled text in CHTML (mathjax/MathJax#3194)

dpvc:
This PR fixes a problem with the width of text that is taken from a non-MathJax font when it is in a context where it is scaled in CHTML output.  For example, when `mtextInteritFont` is true,

``` latex
x_{\bbox[red]{\text{xxxxxxx}}}
```

will have an incorrectly sized red background in CHTML.  Other examples are

``` latex
\Large \bbox[red]{\text{xxxxx}}
```

``` latex
\textstyle \frac{1}{\bbox[red]{\text{xxxxx}}}
```

``` latex
\sqrt{\scriptstyle\bbox[red]{\text{xxxxx}}}
```

Resolves issue at the bottom of mathjax/MathJax#3194.

### DIFF
--- a/ts/output/chtml.ts
+++ b/ts/output/chtml.ts
@@ -291,7 +291,7 @@ CommonOutputJax<
     //    and call to getBBox().w in TextNode.ts)
     //
     if (width !== null) {
-      styles.width = this.fixed(width * this.math.metrics.scale * rscale) + 'em';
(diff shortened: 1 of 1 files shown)


zorkow (approved):
lgtm.

PR #1085, opened 2024-04-26
Fix issues with GeneratorPool init() looping, and move some a11y options to semantic-enrich.  Add SRE dependencies.

dpvc:
This PR makes the changes we talked about face-to-face today.  I will make the version change to beta.6 when we are ready to release the new version.


zorkow (commented):
lgtm.

PR #1086, opened 2024-04-26
Refactor Units computation into separate Utils

zorkow:
The PR 
* refactors utility functions for units into a separate module. This resolves a circular dependency between `TexParser` and `ParseUtils`. Therefore also the `trimSpaces` function is moved to `UnitUtils` although it does not "topically" fit.
* rewrites `UnitMap` from an extension of `Map` to using an private `Map` property, to avoid problems with ES5.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,319 +1,196 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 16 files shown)


zorkow:
For some reason there is an update in the `pnpm-lock.yaml` file. There don't seem to be many changes, though.

dpvc:
Looks good. lgtm.

dpvc (approved):

PR #1087, opened 2024-04-28
Corrects static method calls

zorkow:
Rewrites method calls `this.makeProcessor` to `Configuration.makeProcessor` as `makeProcessor` is actually a static method. It was probably pure luck that this never failed and terser did not uniquely renames static methods.

### DIFF
--- a/ts/input/tex/Configuration.ts
+++ b/ts/input/tex/Configuration.ts
@@ -77,12 +77,12 @@ export class Configuration {
                                   parser?: string,
                                  } = {}): Configuration {
     let priority = config.priority || PrioritizedList.DEFAULTPRIORITY;
-    let init = config.init ? this.makeProcessor(config.init, priority) : null;
-    let conf = config.config ? this.makeProcessor(config.config, priority) : null;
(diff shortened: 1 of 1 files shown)


dpvc:
I don't understand the need for this.  The `_create()` method where these calls are made is also a static method, so `this` will be the `Configuration` object already.  I don' think terser can rename the static methods (any more than it can rename the instance methods), and if it did, it would need to change these references as well.

It is actually valuable to keep these as `this`, so that if `Configuration` is subclassed and overrides the `makePorcessor()` method, for example, you won't have to also override `_create()` to call the subclass instead of `Configuration`.  I have run into this difficulty in the past when subclassing objects that call their class name explicitly rather than using `this` is static methods and `this.constructor` in instance methods.  So I'd prefer that these stay as `this`.

zorkow:
As discussed f2f, this came out of experiments with [tsickle](https://www.npmjs.com/package/tsickle) and [closure-compiler](https://www.npmjs.com/package/google-closure-compiler), which do not like the references to static methods via `this` nor our way of using `this.constructor`.

Currently, we are not going for that. PR closed.

PR #1089, opened 2024-05-03
Fix handling of unknown characters in SVG output. (mathjax/MathJax#3224)

dpvc:
This PR fixes an issue with the handling of the width of characters that are not in the MathJax fonts in SVG mode that could lead to overlapping characters.  The SVG wrapper's `toSVG()` code has been reordered to make this work properly.  The handling of unknown characters has been moved to the top, and `if-then-else` has been broken into separate `if-then` with returns, with the width of previous unknown characters being taken properly into account (which was what wasn't being done).

This only affects text that is all in the same MathML node, so you can test using `\text{aaa 中文 bbbb}` for the `unknown` and `p` sections of the code.  To get a `c` test, you need to use the `mathjax-tex` font with `\text{aaa\U{220C}bbb}`.

Resolves issue mathjax/MathJax#3224.

### DIFF
--- a/ts/output/svg/Wrapper.ts
+++ b/ts/output/svg/Wrapper.ts
@@ -489,21 +489,25 @@ export class SvgWrapper<N, T, D> extends CommonWrapper<
     }
     const C = n.toString(16).toUpperCase();
     const [ , , w, data] = this.getVariantChar(variant, n);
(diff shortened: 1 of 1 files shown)


zorkow (approved):
lgtm.

PR #1090, opened 2024-05-03
Use XML serialization not HTML serialization for SVG images. (mathjax/MathJax#3226)

dpvc:
This PR changes the serialization of SVG elements to use XML rather than HTML rules (so that the xml in the `data-semantic-speech` attribute is properly quoted).

Resolves issue mathjax/MathJax#3226.


zorkow (approved):
lgtm.

PR #1093, opened 2024-05-19
Prevent linebreaking from crashing on empty containers. (mathjax/MathJax#3232)

dpvc:
This PR fixes a problem in the LinebreakVisitor object where `malignmark` or `maligngroup` elements (which have no children) cause a "Math output error" due to an attempt to use `childNodes[0]` when that is null.

Resolves issue mathjax/MathJax#3232.


zorkow (approved):
lgtm.

PR #1094, opened 2024-05-24
Fix regression from #1069 (mathjax/MathJax#3233, mathjax/MathJax#3234)

dpvc:
This PR fixes a regression introduced in #1069 in which primes are misplaced for `\underline{x}` and give an error about double superscripts for `\overline{x}` and similar macros.  This was du to not taking the `subsupOK` property in account for these `munder` and `mover` constructs.  We add those checks here.

Also, this fixes some non-critical number/boolean issues in the base mappings and macros.  It works the way it is, but the types are wrong, so we straighten that out here.

Resolves issue mathjax/MathJax#3233 and mathjax/MathJax#3234.

### DIFF
--- a/ts/input/tex/base/BaseItems.ts
+++ b/ts/input/tex/base/BaseItems.ts
@@ -208,8 +208,11 @@ export class PrimeItem extends BaseItem {
    */
   public checkItem(item: StackItem): CheckType {
     let [top0, top1] = this.Peek(2);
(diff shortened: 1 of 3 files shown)


zorkow:
Two points:
* The underlying `BaseMethods` in `develop` has now changed significantly, so it might be easier to "replay" the changes in a new branch and create a new PR.
* It is also likely that the current changes break some of the tests.

dpvc:
OK, I have made a new branch and will make a new PR for it, so closing this one.

The tests suggested some additional changes to improve the results, and after that the tests all pass, so don't need modification.

PR #1096, opened 2024-05-30
refactors and simplifies macro mappings

zorkow:
* Refactors macro, command and environment maps in the  packages to directly use methods provided by the `MacroMap` classes. This 
    1. makes our code less dependant on string references
    2. should allow bundlers that minimize code (like Closure) to more efficiently replace method calls
* Retains legacy functionality by keeping an optional `functionMap` argument and checking if the `ParseMethod` reference is a string. This should ensure backward compatibility with third party packages.


dpvc (approved):
lgtm.

PR #1097, opened 2024-05-30
Refactors the Treatment of Namespaces for Methods of TeX Packages

zorkow:
PR removes the old namespacing using a `Record` type and instead uses a single object of type `{[key: string]: ParseMethod` that contains all the named methods that we need in the `MacroMap` and `EnvironmentMap` objects. Layout-wise this has the effect:
1. All methods belonging to the single `PackageMethods` object are indented by one extra layer.
2. All auxiliary functions (private to the module or exported) that used to be interspersed are now in the preamble of the module.

In addition the PR introduces `enum` types for `ConfigurationType` and `HandlerType` moving away from pure string based naming in configurations.


dpvc (approved):
Looks good.  I requested one type of change concerning how methods from previous method objects are included in new ones.  Take it or leave it.  In any case, I approve.

zorkow:
> Looks good. I requested one type of change concerning how methods from previous method objects are included in new ones. Take it or leave it. In any case, I approve.

Done the change as requested. Looks better this way.

PR #1098, opened 2024-06-02
`\sqrt` with index of radical exceeds two digits display incorrect

azurechen:
When using the `sqrt` syntax but the `index of radical exceeds two digits`, the offset of expression is incorrect.

For example
```
aa\sqrt[33333]{a^{15}} + \sqrt[33333]{a^{15}}aa
```

image

It is very easy to reproduce by using the [demo page](https://mathjax.github.io/MathJax-demos-web/input-tex2svg.html) provided by [MathJax-demos-web](https://github.com/mathjax/MathJax-demos-web)

The correct result should be displayed like the following image (the output is HTML, which means the issue is only on SVG)

image

After investigation, this is a bug caused by the following line. Seems like adding an extra `dx` value accidentally.

https://github.com/mathjax/MathJax-src/commit/5f6cc8a78092f6b7b0849deb0e57002a61e7953d#diff-21adda599dc0790184ec7dfa4cb615b263b9a91917f3386bc62d8730460548acR356
(Downgrading to v3.2.0 can fix this issue because the pr was not merged yet)

Remove this line, and the expression is displayed correctly.

### DIFF
--- a/ts/output/svg/Wrapper.ts
+++ b/ts/output/svg/Wrapper.ts
@@ -364,7 +364,6 @@ CommonWrapper<
    * @param {N} element  The element to be placed
    */
   public place(x: number, y: number, element: N = null) {
-    x += this.dx;
     if (!(x || y)) return;
     if (!element) {
       element = this.element;


azurechen:
link to issue mathjax/MathJax#3238

dpvc:
Thanks for being willing to contribute to MathJax!

Unfortunately, the line you are removing is needed in other situations, so this fix is not sufficient.  The issue is [fixed in v4](https://github.com/mathjax/MathJax-src/blob/2185f0b7e36436cb8b52a7f361fcaf0582d40684/ts/output/svg/Wrapper.ts#L416-L418) (currently in beta), and after v4 is officially released, some fixes (like this) may be back ported to v3.  

Also, we use a [git-flow](https://www.gitkraken.com/learn/git/git-flow) approach to releases, so fixes are done on the `develop` branch, not `master` (only releases are merged into master).  So you would need to make your patch from and to the `develop` branch.  But since that is now the v4 development, you would need to branch from the 3.2.2 tag and there currently isn't a branch to merge to (we will need to create a `v3-latest` branch for changes to v3 when we start back-porting v4 fixes).

So I'm going to close this PR for these reasons, though we do appreciate the effort you put into it.

PR #1099, opened 2024-06-03
Allow jest and tests to be in a separate directory

dpvc:
This PR moves the jest configuration into a `jest` subdirectory, so they don't need to be in the main directory, and removes the dependencies from the `package.json`, moving them into the `package.json` file in the subdirectory.  This means that people installing the `mathjax-full` package don't need to get all the testing dependencies if they are just using the compiled sources in their own projects.  It also keeps the top-level directory a bit cleaner.

The trick to getting the coverage to work is to use `rootDir: '..'` in the configuration, so that the `ts` directory is in the root directory.  A complication is that the `ts-jest` preset has to be loaded out of the `jest/node_modules` directory, so the `jest.config.ts` file is modified to handle that.

The `jest/package.json` file maps the `#js` files to the main directory's `mjs` directory (the other `#...` definitions are handled by the main `package.json` file, as they are called from within the files in the `mjs` directory).

The `tsconfig.json` file points to the main `tsconfig/mjs.json` file (or we could point it to the main `tsconfig.json` file, which in turn points to `tsconfig/mjs.json`).

Finally, the imports for the main MathJax files now use `#js` rather than `../../ts` or whatever (the current `tsconfig.json` doesn't allow extenionless imports).  Also, isn't it better to test the actual .js files that are being used rather than having to have ts-jest recompile them?

Since we now have a separate `jest` directory, the `tests/src` files could be moved to the jest directory, as `jest/lib` or something like that, rather than being int he tests directory (I didn't make this change, but think it is a good idea).  if you don't like the name `jest` for the subdirectory, it could be `testsuite` instead.

If you don't like the subdirectory idea, we can close this PR, but I think it is a good idea to separate out the dependencies.  What do you think?


dpvc:
> For some reason we still have a `tape` dependency in the main `package.json` file.

Yes, I had meant to remove that before making the PR bug forgot.  I had used it back in the early days of v3 for the tests that I had written (I had tests for a number of the files in the `ts/util` folder).  I was trying to resurrect those tests in order to convert them over to jest, so had left `tape` in place while I was trying them out.

> I am in general not a fan of the `.js` extensions in our `ts` files.

While Typescript could handle leaving them out, there were issues with some of the other places where MathJax could be used.  I don't remember the details any longer, but I think it was `System.js` that we used originally for the lab that was what required that, though it might have been some node or webpack thing.  I think they are still needed for the `<script type="module">` usage that we have for the current lab, as I don't think the browser will add missing extensions, and the Typescript seem adamant about not having `tsc` add them itself.  So unless you want to have a post-processing step that adds them, I think they have to stay.  But we can try to find another solution that works in all our use-cases.
That was originally needed

PR #1100, opened 2024-06-04
Renames testsuite, moves and simplifies helper import

zorkow:
Followup on PR #1099:
* Renames `jest` to `testsuite`
* Removes unused dependency `tape` from root `package.json`
* Moves `tests/src` to `testsuite` root
* Exports all elements in `src` via a single `index.ts` file and points to them via a `#helpers` path. This should encourage us to keep auxiliary functions for tests simple. I.e., we want to test our code, not the functions that test our code.


dpvc (approved):
LGTM.

PR #1102, opened 2024-06-06
Allow columns to shrink during line breaking.  (mathjax/MathJax#3235)

dpvc:
During line breaking, when a table column has breaks, the computed with was not always being reduced, leaving the column width the original size, leading to the output bing too wide despite the line breaking.

This PR resolves the problem by allowing columns that are `fit`, `auto`, or percentage-width to shrink to the width of the newly broken cells rather than their original widths.

A test case is

``` html
<div style="width:20em; border:1px solid black">
\begin{equation*}
\begin{cases}
\def\div{\operatorname{div}}-\div(A(x)Du)=c_\lambda(x)u+( M(x)Du,Du)+h(x)\\
u\in H_0^1(\Omega)\cup L^\infty(\Omega)
\end{cases}
\tag{$P_\lambda$}\label{$P_lambda$}
\end{equation*}
</div>
```

with `overflow` set to `linebreaks` and `displayAlign` set to `left`.  Without this PR, the tag will be outside the `div`, and with it, the tag should be properly placed.

Resolves issue mathjax/MathJax#3235.

### DIFF
--- a/ts/output/common/Wrappers/mtable.ts
+++ b/ts/output/common/Wrappers/mtable.ts
@@ -209,10 +209,11 @@ export interface CommonMtable<
   /**
    * Do linebreaking on the cells of a column
    *
-   * @param {number} i   The number of the column to break
-   * @param {number} W   The width to break to
+   * @param {number} i      The number of the column to break
+   * @param {number} W      The width to break to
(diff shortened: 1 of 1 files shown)


zorkow (approved):
Works for me.

PR #1103, opened 2024-06-10
Improve performance by detaching speech computation

zorkow:
Detaches the speech computation by starting it after the typeset elements are inserted:
* Introduces a set of timeouts that allow us to 
    * to initially delay the computation by 100ms
    * loop through mathitems for ~250ms before breaking 
    * for 10ms to allow other processes to come in.
* Similarly the explorer restarts in case the speech is not yet attached


zorkow:
@dpvc This should be ready to review now.

dpvc (approved):
lgtm.

PR #1104, opened 2024-06-10
Fixes incorrect lifecycle script for test

zorkow:
Fixes an oversight from PR #1100, that retains the `jest` directory name in the `test` lifecycle script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "use-cjs": "echo '{\n  \"extends\": \"./tsconfig/cjs.json\"\n}' > tsconfig.json",
     "use-mjs": "echo '{\n  \"extends\": \"./tsconfig/mjs.json\"\n}' > tsconfig.json",
     "=============================================================================== aliases": "",
-    "test": "cd jest && pnpm test",
+    "test": "cd testsuite && pnpm test",
     "clean": "pnpm -s clean:mod cjs && pnpm -s cjs:bundle:clean && pnpm -s clean:mod mjs && pnpm -s clean:dir bundle",
     "compile-cjs": "pnpm -s cjs:compile",
     "compile-mjs": "pnpm -s mjs:compile",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -108,7 +108,7 @@
     "use-cjs": "echo '{\n  \"extends\": \"./tsconfig/cjs.json\"\n}' > tsconfig.json",
     "use-mjs": "echo '{\n  \"extends\": \"./tsconfig/mjs.json\"\n}' > tsconfig.json",
     "=============================================================================== aliases": "",
-    "test": "cd jest && pnpm test",
+    "test": "cd testsuite && pnpm test",
```
dpvc, line 111: I was going to make a PR for this after you merged the test suite, but you beat me to it.

I would ask, however, that you make this

``` suggestion
    "test": "cd testsuite && pnpm test --",
```

instead.  The reason for this is so that you can pass options on to the test script in the testsuite directory.  For example, in that directory, you can do

``` shell
pnpm -s test -- --collectCoverageFrom 'mjs/input/tex/ams/*.js' tests/input/tex/Ams.test.js
```

in order to restrict the coverage listing to be only the files in the ams package.  With the addition of the `--` in the package script, you can do the same from the top-level directory.  Without it, you would have to use

``` shell
pnpm -s test -- -- --collectCoverageFrom 'mjs/input/tex/ams/*.js' testsuite/tests/input/tex/Ams.test.js
```

with two `--`.  Adding one into the script allows the same format to be used in both directories.

dpvc (changes requested):

package.json
dpvc: Actually

``` suggestion
    "test": "cd testsuite && pnpm -s test --"
```

is even better.

zorkow:
Changes made. PTAL.

dpvc (approved):
Looks good.  I'll merge.

PR #1105, opened 2024-06-12
Add tests for all ts/util files and update some utils to fix issues found from tests

dpvc:
This PR adds tests for all `ts/util` files and updates the length tests to 100% coverage.  While writing the tests, some issues were found with the util files, and these are also fixed here (they are at the bottom of the list).

I've also added some coverage exclusions to the jest configuration file to prevent some files from showing own the coverage that aren't needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ coverage
 /mjs
 /bundle
 /bundle-cjs
-
+/testsuite/js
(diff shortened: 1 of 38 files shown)


dpvc:
I've also updated the `tsconfig.json` file to compile to a `js` directory so that `.js` files aren't produced in the `tests` directories when editing the files in emacs, and added the `js` directory to the `.gitignore` file.

zorkow (commented):
Well done: 100%!
I've only skim read the tests, but reviewed the actual MathJax code.
lgtm.

PR #1107, opened 2024-06-20
Improve speech performance, especially for lazy typesetting

dpvc:
This is the PR that we discussed today in our meeting.  I have reworked the handling of the speech-generation loop using the idea I mentioned toward the end:  I record the return value from the `setTimeout()` command and use `clearTimeout()` to prevent the next iteration of the loop from running if the `attachSpeech()` render-action is called (e.g., the page is retypeset) while there are still MathItems waiting to be processed.  That turns out to be much cleaner, and you can even eliminate the `attachSpeechStart()` method along the way.  So I think these are good changes.

Just to have it in the record, this PR:

* Moves the `Timing` values to document options, where a page author can adjust them if needed.
* Changes the `setTimeout()` in the KeyExplorer to a direct call to `attachSeech()` if there is no speech already attached.
* Changes `private` to `protected` in case any of this needs to be subclassed.
* Adds `speechTimeout` to record the value of the `setTimeout()` calls for the speech loop.
* Uses that value to clear any pending loop calls if `attachSpeech()` is called while there are still MathItems in the `awaitingSpeech` array.
* Then directly starts the loop using `setTimeout()`.
* Sets the process bit in `attachSpeech()`, indicating that the render-action has done its job (of starting the asynchronous loop).  Without this, every typeset call will run `attachSpeech()`, even if it is not needed.
* The loop is made into a do-while loop so that at least one MathItem will be processed (avoiding a potential infinite loop if the threshold is set too small, like 0)
* If there are still more MathItems to process, we use `setTimeout()` to schedule the next loop and save the return value, otherwise, we clear the timer indicator and stop looping.
* The lazy handler is modified to override the MathItem's `attachSpeech()` method so that it is not performed until we actually typeset the math.
* We also increase the margin where the lazy loader will trigger typesetting so that we get a slightly larger head start on the speech processing.

### DIFF
--- a/ts/a11y/explorer/KeyExplorer.ts
+++ b/ts/a11y/explorer/KeyExplorer.ts
@@ -29,7 +29,7 @@ import type { ExplorerMathItem } from '../explorer.js';
 import {Explorer, AbstractExplorer} from './Explorer.js';
 import {ExplorerPool} from './ExplorerPool.js';
 import {MmlNode} from '../../core/MmlTree/MmlNode.js';
-import { honk, InPlace, Timing } from '../speech/SpeechUtil.js';
+import { honk, InPlace } from '../speech/SpeechUtil.js';
 import {Sre} from '../sre.js';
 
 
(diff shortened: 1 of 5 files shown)


dpvc:
There is an issue with compiling the cjs version of the changes I made to the lazy handler.  I will work out a fix.

dpvc:
Ok, I made the fix.

zorkow (approved):
lgtm.

PR #1108, opened 2024-06-20
Jest custom reporter for macro coverage of tex package

zorkow:
Title says it all. A custom reporter that outputs the coverage of how many macros per TeX package are tested.
Since we cannot store information globally and pass them to the reporter  (Some discussions on this are here: https://github.com/jestjs/jest/issues/7421) I have employed a hack outputting JSON to the FS, currently a file `/tmp/test.json`.
Then we output one table per TeX package, not has nicely coloured as the general coverage table.

This is a mix of `.js`, `.ts.d` and `.ts` files, as I could not get the reporter to run as `.ts` file, since we do not compile our test sources.

I am sure it could be done better, but I think spending two says on this is sufficient, as all we really want it for is to get to 100% coverage for our macros.


testsuite/src/texReporter.js
```diff
@@ -0,0 +1,108 @@
...
+
+  onRunStart(
+    aggregatedResults, options
+  ) {
+    fs.writeFileSync(tmpJsonFile, '[{}');
+  }
+
+  onRunComplete(testContexts, results) {
+    fs.appendFileSync(tmpJsonFile, ']');
+    let coverage = JSON.parse(fs.readFileSync(tmpJsonFile));
```
dpvc, line 24: Should the temp file get removed here?

testsuite/src/setupTex.ts
```diff
@@ -34,3 +41,104 @@ export function tex2mml(tex: string) {
...
+        table: handler,
+        size: exp,
+        actual: act,
+        missing: Array.from(diff)
+      })
+    }
+  });
+  fs.appendFile(tmpJsonFile, ',' + JSON.stringify(outJSON, null, 2), (_err) => {
+    console.warn(`could not write file ${tmpJsonFile}`);
+  });
```
dpvc, line 134: Should this be using `fs.appendFileSync`?  I got the error message when I was running just a single file, e.g.

``` shell
pnpm -s test testsuite/tests/input/tex/Ams.test.ts
```

and got no table output from there reporter (the error message shows up as a warning above the coverage table).

If I change these three lines to 

``` js
  try {
    fs.appendFileSync(tmpJsonFile, ',' + JSON.stringify(outJSON, null, 2));
  } catch (err) {
    console.warn(`could not write file ${tmpJsonFile}: ${err.message}`);
  };
```

then it works as expected.

dpvc: As an aside, using

``` shell
pnpm -s test -- --collectCoverageFrom 'mjs/input/tex/ams/*.js' testsuite/tests/input/tex/Ams.test.ts
```

limits the coverage table to just the ams files, which makes for a nicer result when trying to test just one extension.

dpvc (changes requested):
I had a problem when running just a single test file (see comments below), which seems to be due to the asynchronous `appendFile()` call.  I suspect that the asynchronous append hasn't completed when the final lines are trying to be appended to the file.

dpvc:
> since we do not compile our test sources.

Actually, they should be compiled into the `js` directory when you edit them in emacs, and you could do a `tsc` command in the test suite directory if you wanted to, and if that helps.  But not really necessary since you have it working, just if it made things cleaner.

testsuite/src/setupTex.ts
zorkow: Changed to use `appendFileSync` everywhere. I did get the occasional non-deterministic error on this, but rerunning would get rid of it (probably due to the change in test order with the jest cache) but overall async seemed to be faster. But I don't think the speed gain is worth it.

zorkow:
Changes made. PTAL.

testsuite/tests/input/tex/Bussproofs.test.ts
```diff
@@ -3209,3 +3209,5 @@ describe('BussproofsRegProofs', () => {
 </math>`
     ));
 });
+
+afterAll(() => getTokens('bussproofs'));
```
dpvc, line 3213: I think you need to import `afterAll` and `getTokens` for this one.

dpvc (approved):
Any thoughts on the removal of the temp file that I mentioned previously?

Other than the missing imports, this looks fine.

testsuite/tests/input/tex/Bussproofs.test.ts
zorkow: Done.

testsuite/src/texReporter.js
zorkow: Done.

zorkow:
Took care of the two missing comments and merged.

PR #1110, opened 2024-06-25
Remove the AllPackages file, the all-packages TeX extension and the tex-full component

dpvc:
This PR removes the `AllPackages.ts` file, the `all-packages` TeX extension, and the `tex-full` component.  Since these no longer include all packages anyway, and we no longer make `-full` versions of the components, these are misleading and problematic in other ways.

The dependencies and source files are modified to remove these packages as well.

As an aside, it used to be that `AllPackages` was used in the v4-lab to get the TeX packages, but that is now being done using the `source.js` file, so AllPackages is not used anywhere any more.


zorkow (approved):
lgtm.

PR #1111, opened 2024-06-27
Towards completion of TexError tests 

zorkow:
Adds tests for errors in TeX packages in preparation of the l10n `TexError` rewrite. Note, that this is not fully complete coverage for all occurring `TexError`s, but should be sufficient for the time being to actually work on the `TexError` refactoring.

In addition the PR adds:
* An Emacs `.el` script for ease of working with jest test output in Emacs. This is partially lifted from the old script in `MathJax-Tests` repository.  
* Some tests to achieve complete coverage for packages: `braket`, `unicode`
* The old `ConfigMacros` tests.
* Updates tests that failed after merging PR #1106 .


ts/input/tex/unicode/UnicodeConfiguration.ts
```diff
@@ -87,6 +90,7 @@ const UnicodeMethods: {[key: string]: ParseMethod} = {
         }
       }
     } else if (variant) {
+      console.log(12);
```
dpvc, line 93: The `console.log()` calls in this file need to be removed.

dpvc (changes requested):
The `console.log()` calls need to be removed, but other than that, it looks like a good start.

ts/input/tex/unicode/UnicodeConfiguration.ts
zorkow: Removed.

zorkow:
As discussed yesterday, I've added a few more tests for completeness of `ConfigMacros` as well as for the `sideset` macro in AMS. Also one correction to the `jest-replace-expected-for-actual` function in `test-diff.el`.

PTAL.

PR #1113, opened 2024-06-28
make typesetPromise() wait for speech to be attached, add synchronous option, and copy ARIA labels to internal MathML

dpvc:
This PR implements the ideas we discussed in our meeting yesterday:

* A new `renderPromises` array is added to the MathDocument that holds promises that are created by the render actions.
* The `typesetPromise()` function now waits for those promises to resolve before its own promise resolves.
* The `attachSpeech()` action now creates a promise that resolves when the speech loop completes so that `typesetPromise()` will wait for it to complete.
* A new `asynchronous` option is added to the `speechTiming` configuration object, initially set to `true`, that controls whether the speech is added synchronously or asynchronously.  (When set to `false`, the speech is added during the `renderAction` processing directly, which will be useful for node-based page preprocessing.)

In addition, a bug with the `defaultPageReady()` call is fixed, so that it will resolve the initial typesetting promise.  That way, if a user provides a `pageReady()` configuration function, it can call the promise-based converters, which now wait on that promise before they run.  That was causing a catch-22 where the `pageReady()` call never completed because the initial typeset promise was being resolved only in the `ready()` function (which calls `pageReady()`).

### DIFF
--- a/ts/a11y/semantic-enrich.ts
+++ b/ts/a11y/semantic-enrich.ts
@@ -285,9 +285,11 @@ export function EnrichedMathItemMixin<N, T, D, B extends Constructor<AbstractMat
       const node = this.typesetRoot;
       if (speech && options.enableSpeech) {
         adaptor.setAttribute(node, 'aria-label', speech as string);
+        this.root.attributes.set('aria-label', speech);
       }
       if (braille && options.enableBraille) {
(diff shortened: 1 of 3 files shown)


zorkow:
When experimenting with this PR, I ran into a build error. Please see my email on it.

zorkow (approved):
Tested with latest develop merged. Works fine.

PR #1117, opened 2024-07-02
Fix inlkine breaking in Safari. (mathjax/MathJax#3252)

dpvc:
This PF works around an issue with Safari where in-line line breaks aren't being honored.  It appears that Safari doesn't break at spaces that are in `content` CSS like all other browsers, so we add the space manually instead.

Resolves issue mathjax/MathJax#3252.


zorkow (approved):
lgtm

PR #1120, opened 2024-07-10
Tests cleanup after AllPackage removal.

zorkow:
PR updates tests to work again after PR #1110 and #1115
* Configurations are included in the test files directly, that limits the actual code that is loaded (and not used) during tests.
* Busproofs tests are updated mainly wrt. to new spacings.

### DIFF
--- a/testsuite/tests/input/tex/Bussproofs.test.ts
+++ b/testsuite/tests/input/tex/Bussproofs.test.ts
@@ -1,5 +1,7 @@
 import { beforeEach, describe, it } from '@jest/globals';
 import { toXmlMatch, setupTexWithOutput, tex2mml } from '#helpers';
+import '#js/input/tex/bussproofs/BussproofsConfiguration';
+import '#js/input/tex/ams/AmsConfiguration';
 
 beforeEach(() => setupTexWithOutput(['base', 'ams', 'bussproofs']));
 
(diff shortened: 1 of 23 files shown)


testsuite/tests/input/tex/Bussproofs.test.ts
```diff
@@ -310,140 +312,140 @@ describe('BussproofsRegInf', () => {
...
+                <mstyle displaystyle=\"false\" scriptlevel=\"0\">
+                  <mtext>B</mtext>
+                </mstyle>
+                <mspace width=\".5ex\"></mspace>
+              </mrow>
+            </mtd>
+          </mtr>
+        </mtable>
+      </mrow>
+    </math>`
```
dpvc, line 353: Are all the backslashes needed for the quotation marks?  I guess they don't hurt, but they make the expected output a little harder to read.

dpvc (approved):
Looks good.  One question about backslashes, but not critical.

testsuite/tests/input/tex/Bussproofs.test.ts
zorkow: That comes from the little Emacs function that allows for copying received values from the test log.
And they actually do hurt, at least in the last three tests, where I had to manually copy things over from the test output as the slashed quotes ruined the tests.

I'll remove them and then merge.

PR #1121, opened 2024-07-16
Better validation of style values

dpvc:
This PR improves the handling of CSS styles, in particular the use of quotation marks and backslashes in CSS style values, and sanitizes CSS values when output as a string, avoiding a potential for CSS insertion.  It also replaces newlines by spaces to avoid some unusual parsing by browsers that newlines can cause.

The parsing of a CSS string now handles quoted material properly, and in the case that a value includes an unmatched quotation mark, one is appended to make sure they are matched.  When outputting a CSS value, the value is trimmed at the first unquoted semi-colon to avoid the possibility of having the value insert additional unexpected CSS rules.

I will make some tests for this in a separate PR.

### DIFF
--- a/ts/util/Styles.ts
+++ b/ts/util/Styles.ts
@@ -319,7 +319,9 @@ export class Styles {
    * Patterns for style values and comments
    */
   public static pattern: {[name: string]: RegExp} = {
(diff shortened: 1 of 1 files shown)


zorkow (approved):
lgtm. 
I did some simple manual tests in node to check the patterns. They appear to work as expected. But some detailed tests will be good.

PR #1122, opened 2024-07-27
Initial implementation of Locale framework

dpvc:
This is a draft for the Locale framework.  I will add a better description here later, but wanted to make this available in case you wanted to look at it.

The `bbox` extension is an example of this in use (it changes one error message so you can see that it is working).

It is probably best to look at the `BBoxConfiguration.ts` file with what space being ignored, as it was not properly indented originally.

This should work both in components and direct calls, both in the browser and in node applications.  The interactive lab needs minor additions because of the unusual way that it is handling the TeX extensions.  I will make a branch for that in the dev repo.

More details to come.


dpvc:
I have made the changes that we talked about in our meeting today.  This now tries to load localization files from any component that registers with the `Local` object, and the error functions no longer include the English messages.  Instead, the message look-up tries to find the value from the current locale, and if that fails, looks in the default locale, which is `en`, but can be changed if desired.

dpvc:
I'm closing this in favor of #1407, which corrects the merge conflicts, improves some features, and adds tests.

PR #1124, opened 2024-08-07
prettier with comma options all

zorkow:
For comparison, `prettier` with the `trailing-commas: "all"` option. 

Personally, I don't think it is a good idea to have trailing commas at every, single parameter list. If we change parameter lists, we need to go through a proper review anyway, so a single comma will not make such a difference.


zorkow:
Agreed in f2f that we don't want these trailing commas and stay with `es5` option.

PR #1130, opened 2024-09-11
Add git hooks for commit and push

zorkow:
PR adds `git` hooks for `commit` and `push`. The hooks are run by [husky](https://typicode.github.io/husky/).  Currently the hooks are setup as follows:
1. `pre-commit` runs [lint-staged](https://www.npmjs.com/package/lint-staged) that applies lifecycle scripts to staged files, restages the files and commits. It blocks the commit if the lifecycle scripts fail or would result in an empty commit. Right now I have only added `pnpm format:fix`. Once the code base is fully linted we should also add `pnpm lint:fix`.
2. `pre-push` should run the test suite. But I've committed them out for the time being as this can be pretty annoying...

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-pnpm pre-push-lint
+pnpm lint-staged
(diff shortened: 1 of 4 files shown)


dpvc (approved):
I haven't tested it, but it looks fine.

PR #1133, opened 2024-09-30
rewrite occurrences of hasOwnProperty

zorkow:
* Rewrites all occurrences of `hasOwnProperty` to use `Object.hasOwn`
* Adds `ES2022` library support

### DIFF
--- a/ts/core/MathDocument.ts
+++ b/ts/core/MathDocument.ts
@@ -411,7 +411,7 @@ export interface MathDocument<N, T, D> {
   /**
    * Rerender the MathItems on the page
    *
-   * @param {number} [start]    The state to start rerendering at
+   * @param {number} start      The state to start rerendering at
    * @returns {MathDocument}    The math document instance
    */
   rerender(start?: number): MathDocument<N, T, D>;
--- a/ts/core/MathDocument.ts
+++ b/ts/core/MathDocument.ts
@@ -411,7 +411,7 @@ export interface MathDocument<N, T, D> {
   /**
    * Rerender the MathItems on the page
    *
-   * @param {number=} start    The state to start rerendering at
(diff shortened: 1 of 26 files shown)


ts/core/MathDocument.ts
```diff
@@ -411,7 +411,7 @@ export interface MathDocument<N, T, D> {
   /**
    * Rerender the MathItems on the page
    *
-   * @param {number=} start    The state to start rerendering at
+   * @param {number} [start]    The state to start rerendering at
```
dpvc, line 414: Are the brackets right?  We don't do this for any other variables.  Should we?

tsconfig/common.json
```diff
@@ -11,7 +11,7 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "baseUrl": ".",
-    "lib": ["ES6", "DOM", "ES2020"],
+    "lib": ["ES6", "DOM", "ES2020", "ES2022"],
```
dpvc, line 14: Does that mean `ES2020` can be removed?

dpvc (approved):
Looks good.  Two questions, but approved.

ts/core/MathDocument.ts
zorkow: There's multiple ways to indicate optional parameters:
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns
Coming from ChromeVox I used the *Google closure style* with `=`. I am not sure why I changed it here, maybe because there is no default value.

Yes, we should probably come to uniform way. For now I remove the brackets until we decide on **our** style.

tsconfig/common.json
zorkow: I would assume we can. I'll try.

PR #1134, opened 2024-09-30
Remove superfluous slashes in reg exps

zorkow:
Remove superfluous slashes from all regular expressions. There is still one bizarre `escape character` error here:

```
ts/input/mathml/mml3/mml3.ts:691:13: Unnecessary escape character: \<. [Error/no-useless-escape]
```


dpvc (approved):
Looks good.  I think making the `\` into `\\` will fix the other message.

zorkow:
> Looks good. I think making the `\` into `\\` will fix the other message.

Turns out that the entire case does not work. See PR #1136 for a fix.

PR #1136, opened 2024-09-30
Fixes the Dutch longdiv issue

zorkow:
Ensures that the `left/\right` case works correctly. See:
https://www.w3.org/TR/MathML3/chapter3.html#presm.mlongdiv.ex


dpvc (approved):
Looks good.  Thanks for fixing this.

dpvc:
PS, I changed the target from `master` to `develop`.

PR #1140, opened 2024-10-12
Change default font to mathjax-newcm

dpvc:
This PR changes the default front from mathjax-modern to mathjax-newcm, which has more coverage and is slightly darker.  The lightness of mathjax-modern was one of the complaints with v4.

You need to do `pnpm install` to get the new font installed, and need the `lab-extension-update` branch in the `MathJax-dev` repository, and do `npm install` there as well, then relink `node_modules/mathjax-full` to your current `MathJax-src` repository.


zorkow (approved):
Looks indeed slightly darker.

Issue #1142, opened 2024-10-21
Newline in @  column leads to infinite loop


zorkow:
```latex
\begin{array}{r@{a\\b}l}a&b\end{array}
```

PR #1144, opened 2024-10-24
Add a template substitution count to avoid infinite loop (#1142)

dpvc:
This PR adds a counter to record array template substitutions and an option `maxTemplateSubtitutions` that is used to detect an infinite loop due to a template that includes something like `@{\\}`.  Because the count is incremented for any cell in a column that has a `@` or other similar substitution, the maximum count is set pretty high to accommodate large arrays, but can be configured to be larger in the document options if necessary.  I would have liked to have been able to detect this more directly, but I wasn't able to come up with another reliable method.

Resolves issue #1142.


zorkow (approved):
lgtm.
I'll add a test once this and PR #1148 are merged.

PR #1145, opened 2024-10-24
Have \let look up macros first then delimiters, and removes \\ delimiter

dpvc:
This PR switches the order in which the `\let` command looks up macros and delimiters, so that something like `\let\xyz=\\` will make `\xyz` work as `\\` rather than produce a backslash (which is the delimiter tied to `\\`).  Also removes `\\` as a delimiter, as LaTeX doesn't actually allow that.

### DIFF
--- a/ts/input/tex/base/BaseMappings.ts
+++ b/ts/input/tex/base/BaseMappings.ts
@@ -348,7 +348,6 @@ new sm.DelimiterMap('delimiter', ParseMethods.delimiter, {
   '/': '/',
   '|': ['|', { texClass: TEXCLASS.ORD }],
   '.': '',
-  '\\\\': '\\',
   '\\lmoustache': '\u23B0', // non-standard
   '\\rmoustache': '\u23B1', // non-standard
   '\\lgroup': '\u27EE', // non-standard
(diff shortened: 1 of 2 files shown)


dpvc:
Thinking about this just now, I'm wondering if using `\let` on a delimiter should add the new control sequence to the delimiter list?  And if redefining a delimiter should remove it from the list (if it is not redefined as a potential delimiter)?  Or maybe there needs to be a non-standard macro to manipulate the delimiter list?

zorkow:
Just testing what the LaTeX behaviour is in those two cases:

```LaTeX
\documentclass{article}

\begin{document}
\let\lpar(
\let\rpar)
\[
  \lpar\frac{a}{b}\rpar
  \]
\[
  \left\lpar\frac{a}{b}\right\rpar
\]
\[
  \left\vert \frac{a}{b} \right\vert
\]
\let\vert A
\[
  \vert \frac{a}{b} \right
\]
\[
  \left\vert \frac{a}{b} \right\vert
\]

\end{document}
```

In MathJax, the behaviour of the first two lets is the sam (stretching still works). However, for `\vert` only the first expression is as expected, while the second one yields the vertical bars (and no error), probably because it is still available in the delimiter subhandler.

zorkow (approved):
lgtm.

dpvc:
I'll make a separate PR to handle the delimiter issue.

PR #1147, opened 2024-10-28
Convert style values to strings before passing them to setStyle.

olafes:
Fixes mathjax/MathJax#3303.


dpvc:
This is the solution I was going to suggest as well.  The `as string` was certainly an error.  Thank for the PR.

PR #1148, opened 2024-10-30
Complete base tests as much as possible

zorkow:
This PR completes the tests for the TeX base package as much as possible. In particular, it also ensures that every single command, macro, and environment has at least one test. 

Initially this was only aimed at adding tests before removing the `expected assignment` linting issues of the form `p && x = n`. It took on a life of its own. It also turns out that rewriting these assignment lines will add additional test branches and thus extra tests will need to be added. Therefore I made this a PR on its own now, before the next linting step.

As mentioned, tests are filled "as much as possible". In particular, 
1. I have added mocks for macros/environments that are only used from other packages outside base. These are bundled in the `User Defined Macros` and `User Defined Environments` test series. These  has the advantage, that when commented out we can see the  lines only used by non-base packages.
2. There are two skipped tests, that should normally produce an error but currently go into an infinite loop. See issue #1142.
3. There are a number of unreachable lines or lines only reachable from other packages, for which I failed to write a mock. For a list see below:

`BaseConfiguration.ts`

| line number | possible reason |
|------|-----------------------------------------------------------------|
| 214: | Needs `document` element. Should we really set the option here? |

`BaseMethods.ts`

| line number | possible reason |
|------------|-----------------------------------------------------------------------------------------|
| 1897-1899: | Used for environments defined with newenvironment. I failed to do a `base` only mock up. |

`BaseItems.ts`

| line number | possible reason |
|------------|------------------------------------------------------------------|
| 284-285:   | Unreachable fail case. Maybe never used?                         |
| 363-366:   | toString method. Probably for debugging. I think that should go? |
| 600-601:   | Unreachable fail case that could go?                             |
| 1277-1278: | Unreachable fail case that could go?                             |
| 1309-1310: | Cell alignment. I could not find a way to trigger it. @dpvc ?     |

Run with:
```
pnpm -s test -- testsuite/tests/input/tex/Base.test.ts --collectCoverageFrom="./mjs/input/tex/base/*.js" &> /tmp/out
```


dpvc:
> Needs `document` element. Should we really set the option here?

This option is only used in the base methods, so this is the place for it, and it has to be in the parser options object, so must be an option in a configuration.  We do some tests for the existence of the `window` or `document` objects in several places.  It would probably be a good idea to do that once and store the result, say in the `mathjax` object, so we don't have to do multiple tests.  One could even store the `window` and/or `document` there for easier reference.  That could then be used here.

> 1309-1310: Cell alignment. I could not find a way to trigger it.

You are right, that isn't currently reachable.  It was intended to allow the `\breakAlign{c}{...}` macro to override the vertical alignment for a cell in a column that has alignment set from `m`, `c`, or `p` templates, but because these are wrapped in `\text{...}`, the `\breakAlign` macro can't be used in those columns.  It may be worth fixing `\breakAlign` to work in those cases, so I would leave this in place for now.

> 1277-1278: Unreachable fail case that could go?

This one can be reached, but only in a sneaky way that reveals problems with how the column processing is handled.  You can do

``` latex
\let\err=&
\begin{array}{l@{X}r}
x\err y
```

and this will fall through to the `return fail`, but it will also cause the loop that is finally stopped by the substitution count that was introduced in #1144.  The problem is that the column processing is looking for `&`, `\\`, etc., and the new `\err` macro that is equivalent to `&` isn't noticed, and so the processing gets messed up.  I'm afraid that this means the column processing needs to be changed, but this is an edge case, and it can wait for now.

> 600-601: Unreachable fail case that could go?

It does look like the `end` property check is not needed, so that lines 596, 601, and 602 could be removed.  But the use of `end` should be cleaned up elsewhere as well.  I will make a PR for that.  The `cases` and `empheq` packages use this, so a little care needs to be taken.

> 363-366: toString method. Probably for debugging. I think that should go?

Yes, it is for debugging, and there are other `toString()` methods that are also used for that (e.g., `TexParser`, `MapHandler`, the `Stack` and `StackItem` objects, along with `Node` and `MmlNode` objects).  I use the `MmlNode` `toString()` frequently during debugging, and the `Stack` and `StackItem` ones less frequently.  But I suppose these could be removed from the main code and patched in by the Lab or via a debug extension.

> 284-285: Unreachable fail case. Maybe never used?

It does seem that this is not currently reachable, and that if it were, it would cause an error at

https://github.com/mathjax/MathJax-src/blob/844ad7513005fa096a07d2468bf59c7f6a6c3a50/ts/input/tex/Stack.ts#L92-L93

where the return value is being deconstructed.  The return value probably should be `BaseItem.fail`.  It was probably added so that there was an explicit return value to avoid typescript complaining.  But it does look like you don't need the `super.checkItem()` test and could just do the error message regardless.

zorkow:
Thanks for the explanations. I will take care of the missing tests in a later PR.
I will merge for now so I can create the remaining linting PRs against develop.

PR #1149, opened 2024-11-06
Linting of "expected assignments" error

zorkow:
PR rewrites all expressions that use an implicit condition of the form `P && f(a)` and rewrites them into explicit conditional of the for `if (P) { f(a); }`.
However, we do allow for standalone ternary expressions as in `C ? f(a) : f(b);` and add a rule to `esllint.conf.mjs` for that.

While this adds somewhat to the length of code it should make the conditions somewhat more readable. (As an old LISP hacker I never had an issue with conjunctive side effects...). But the real advantage is that those lines are now actually testable and produce useful coverage information, as previously the line would be considered as covered, even if the first condition never evaluated to `true`. As a consequence, some addition tests are added to `Base.test.ts`.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -25,6 +25,7 @@ export default tseslint.config({
                                             "caughtErrorsIgnorePattern": "^_" }
                                          ],
     "@typescript-eslint/no-empty-object-type": ["error", {"allowInterfaces": "with-single-extends"}],
+    "@typescript-eslint/no-unused-expressions": ["error", { "allowTernary": true }],
     "prefer-const": ["error", {"destructuring": "all"}],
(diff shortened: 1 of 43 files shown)


dpvc (approved):
This is fine.  I think there might be one or two `||` conditions that work this way as well.  Can your code locate those?

dpvc:
> I'll have a browse around to see if I can locate them.

I don't think there are many, just one or two, and they may already have been removed.  So there may not be any left.  Don't spend too much time on this.  I wish I could remember where they were.

zorkow:
> > I'll have a browse around to see if I can locate them.
> 
> I don't think there are many, just one or two, and they may already have been removed. So there may not be any left. Don't spend too much time on this. I wish I could remember where they were.

I had a brief look, but could not find any.

PR #1150, opened 2024-11-06
Rewrite general Function type to concrete types

zorkow:
Rewrites all (but one) occurrences of the `Function` type to the correct types to resolve the [no-unsafe-function-type](https://typescript-eslint.io/rules/no-unsafe-function-type) error. Some of them are a catch all function types, of course.

### DIFF
--- a/ts/util/Retries.ts
+++ b/ts/util/Retries.ts
@@ -26,7 +26,11 @@
  *  The legacy MathJax object  (FIXME: remove this after all v2 code is gone)
  */
 
-declare const MathJax: { Callback: { After: Function } };
+declare const MathJax: {
+  Callback: {
+    After: (cb1: () => void, cb2: () => void) => Promise<any>;
+  };
+};
 
 /*****************************************************************/
 /**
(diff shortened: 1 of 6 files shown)


ts/util/Retries.ts
```diff
@@ -26,7 +26,11 @@
  *  The legacy MathJax object  (FIXME: remove this after all v2 code is gone)
  */
 
-declare const MathJax: { Callback: { After: Function } };
+declare const MathJax: {
+  Callback: {
+    After: (cb1: () => void, cb2: () => void) => Promise<any>;
```
dpvc, line 31: This isn't actually the correct argument list.  A callback can be `Function`, `[Function]`, `[Function, ...args: any[]]`, `[Object, Function]`, `[Object, Function, ...args: any[]]`, `[string, Object]`, `[string, Object, ..args: any[]]`, or `{hook: Function, data: any[], object: Object}`, and the arguments to `After()` is a list of these, potentially more than 2.

But I guess this is sufficient for the usage here.

dpvc (approved):
Looks good.

ts/util/Retries.ts
zorkow: Indeed, but since we are never actually using it in the code base, I thought I go for the *arbitrary function* type as discussed in https://www.typescriptlang.org/docs/handbook/2/functions.html#function

PR #1151, opened 2024-11-06
Fix processing of latex attributes for user-defined environments thatwas broken by #1139

dpvc:
This PR fixes a problem where the `data-latex` attributes were missing or incomplete when a user-defined macro was used.  This caused some tests to fail (so kudos for the tests).

All the tests should pass with this update.

### DIFF
--- a/ts/input/tex/newcommand/NewcommandMethods.ts
+++ b/ts/input/tex/newcommand/NewcommandMethods.ts
@@ -263,8 +263,7 @@ const NewcommandMethods: { [key: string]: ParseMethod } = {
         if (edef) {
           // Parse the commands in the end environment definition.
           let rest = parser.string.slice(parser.i);
-          parser.string = edef;
-          parser.i = 0;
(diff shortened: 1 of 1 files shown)


zorkow (approved):
Fixes the currently failing tests:
```
    ✕ Newenvironment Optional
    ✕ Newenvironment Arg Optional
    ✕ Let Brace Delim
    ✕ Let Brace Equal Stretchy
    ✕ Let Paren Circular
    ✕ Let Angle Circular
    ✕ Newenvironment Empty
    ✕ Newenvironment Content
    ✕ Newenvironment Nested Double
    ✕ Newenvironment Nested Double 2
    ✕ Newenvironment Nested Triple
    ✕ Newenvironment Nested Triple Text
    ✕ Let Bar
    ✕ Let Bar Stretchy
    ✕ left right /
    ✕ Moveright
    ✕ Moveleft
    ✕ Move Left Negative
    ✕ Move Right Negative
    ✕ Environment Simple
```

PR #1152, opened 2024-11-06
Lints the trivial Rest

zorkow:
PR lints all issues with the exception of 
* rewriting the ES2015 module syntax
* `Unexpected aliasing of 'this'`

In particular it
* irons out the remaining missing jsdocs `@returns` errors,
   * adding missing `return` statements
   * adding missing `return` types
* adds a `eslint-disable` statements where really necessary
* removes empty blocks
* avoids linting in the `cjs` folder
* adds some tests for `amscd`, `braket`, `physics` as a consequence of the above
* adds complete set of tests for all macros in the `physics` package

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -475,6 +475,7 @@ export class CommonWrapper<
     inline: boolean;
     width: string;
     lineleading: number;
+    LinebreakVisitor: null;
   } {
     return this.jax.options.linebreaks;
   }
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -21,6 +21,8 @@
  * @author dpvc@mathjax.org (Davide Cervone)
  */
 
+import { DOMAdaptor } from '../../core/DOMAdaptor.js';
+import { Metrics } from '../../core/MathItem.js';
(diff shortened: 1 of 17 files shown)


ts/output/common/Wrapper.ts
```diff
@@ -450,18 +453,29 @@ export class CommonWrapper<
...
-  get linebreakOptions() {
+   * @returns {{inline: boolean,
+   *   width: string,
+   *   lineleading: number,
+   *   LinebreakVisitor: null}} The linebreak options
+   */
+  get linebreakOptions(): {
+    inline: boolean;
+    width: string;
+    lineleading: number;
```
dpvc, line 477: Is `LinebreakVisitor` left out here for a reason?

dpvc (approved):
Looks good.

ts/output/common/Wrapper.ts
zorkow: Yes, because its generic type is a bit of a nightmare: https://github.com/mathjax/MathJax-src/blob/aa3c017e317a4ce768bccd6a0385c7f15694bbf1/ts/output/common/LinebreakVisitor.ts#L145
However, it turns out leaving it as `null` actually works, too. So I am adding that.

PR #1153, opened 2024-11-06
Add context.ts to hold hasWindow flag and browser's window and document objects.

dpvc:
This PR creates a `context` object that holds the browser `window` and `document` values, if they exist, simplifying tests for when we are in a Node application versus the browser.  It then converts uses of `typeof window === 'undefined'` and `typeof document === 'undefined'` to use this new object.  This reduces redundant code, and puts the browser detection into one location.


ts/ui/menu/MenuUtil.ts
```diff
@@ -1,15 +1,38 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @file  Implements a functino to handle copying to clipboard
```
zorkow, line 19:
```suggestion
 * @file  Implements a function to handle copying to clipboard
```

zorkow (approved):
One typo. Otherwise fine.
Makes things look a lot cleaner!

PR #1154, opened 2024-11-07
Remove unneeded lab and pack versions of a11y/util.js

dpvc:
This PR removes the `util-lab.js` and `util-pack.js` files from the `components/mjs/a11y` directory, as current versions of node and webpack are able to process the json files with the proper type declarations.  If used directory in node, there is a warning, which could be overcome by using `--no-warnings=ExperminentalWarning`, but the only uses of the `util.js` file are in the combined components, like `tex-chtml.js`, which only work in a browser (they load the `browswerAdaptor`), so really can't be used very easily in node anyway.  So this file is really only used in the webpacked combined components.


zorkow (approved):
lgtm.

PR #1155, opened 2024-11-07
Prevent attempts to set aria-hidden attributes on text nodes

zorkow:
Puts in a test to see if the child element of the container node is not a text node. This prevents issues, when a render actions inserts some spaces, as in the Euro Braille example for selecting and copying:
https://mathjax.github.io/MathJax-demos-web/euro-braille/

### DIFF
--- a/ts/a11y/semantic-enrich.ts
+++ b/ts/a11y/semantic-enrich.ts
@@ -319,8 +319,10 @@ export function EnrichedMathItemMixin<
         adaptor.setAttribute(node, 'aria-braillelabel', braille as string);
         this.root.attributes.set('aria-braillelabel', braille);
       }
-      for (const child of adaptor.childNodes(node) as N[]) {
-        adaptor.setAttribute(child, 'aria-hidden', 'true');
+      for (const child of adaptor.childNodes(node)) {
+        if (adaptor.kind(child) !== '#text') {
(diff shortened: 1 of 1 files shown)


dpvc (approved):
I'm OK with the change, but I don't see where these text elements could be coming from.  The `node` is coming from the MathItem's `typesetRoot`, and that should be a `mjx-container` element.  It's children should not be the `mjx-math` and the assistive math element (if any).  I don't see where a text element could be coming from.

Can you provide an example of an expression that produces this?

zorkow:
> I'm OK with the change, but I don't see where these text elements could be coming from. The `node` is coming from the MathItem's `typesetRoot`, and that should be a `mjx-container` element. It's children should not be the `mjx-math` and the assistive math element (if any). I don't see where a text element could be coming from.
> 
> Can you provide an example of an expression that produces this?

Here's our example code for the direct LaTeX copying:
https://mathjax.github.io/MathJax-demos-web/euro-braille/config.js

dpvc:
OK, thanks for the pointer.  I remembered doing something like that, but couldn't find it in the codebase.  I see it was a custom `renderAction`.  So when that U+200C is included in the math that is copied and pasted into math delimiters, it causes plain text elements to be left at the top level in the output generated by MathJax?  That seems like something that shouldn't happen.

PR #1156, opened 2024-11-10
Remove need for 'end' property in handling of environments, and refactor numcases and empheq items

dpvc:
This PR was started as a way to remove the original line 600 from `BaseItems.ts`, but includes changes to `empheq` and `cases` as well.

The changes to `ParseMethods.ts` are because the `end` property is no longer needed (and never really was).  The `BaseItem.ts` changes remove the test for `end` that isn't needed.

The `EmpheqBeginItem` subclass of `BeginItem` is not needed, as it doesn't actually do anything, and so it is removed, as is the `environment()` function in `EmpheqIUtil` that was used to create it.  The latter has been moved to the `CasesConfiguration.ts` environment, where it is still used.  The `end` property is used in `numcases` environments because it wraps the `empheq` environment, and `end` its used to properly handle producing the end for both environments.  A missing `return` was added to `CasesConfiguration.ts` as well.

Finally, the `end` property was removed from `NewcommandMethods.ts`, as it is no longer used.

### DIFF
--- a/ts/input/tex/ParseMethods.ts
+++ b/ts/input/tex/ParseMethods.ts
@@ -203,14 +203,14 @@ const ParseMethods = {
    *
    * @param {TexParser} parser The current tex parser.
    * @param {string} env The name of the environment.
-   * @param {Function} func The parse method for the environment.
+   * @param {ParseMethod} func The parse method for the environment.
    * @param {any[]} args A list of additional arguments.
    */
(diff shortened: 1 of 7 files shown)


dpvc:
> Question is of course: Do all the tests still run?

I had to merge develop to get rid of merge conflicts, and that brings in the changes that broke a lot of tests.  I have the fixes for those tests ready to make into a PR, and if those changes are included, the tests run.

PR #1157, opened 2024-11-11
Remove latest.js as jsdelivr can do this automatically.

dpvc:
This PR removes the `ts/components/latest.js` and the `latest` component.  CDNs like `jsdelivr` and `unpkg` already have mechanisms for this, so we don't need to maintain one ourselves.


zorkow (approved):
lgtm. No review.

PR #1159, opened 2024-11-12
Removes all but one namespace

zorkow:
Removes all `namespace` elements except `Startup`.
* The majority of namespaces are rewritten into `const` objects containing 
* There are two notable exceptions:
1. `Sre` is rewritten into a regular ES5 module and it's imports are changed to ```import * as Sre from '...'`. This is mainly due to Sre importing and exporting SRE types. I also assume that we do not want to overwrite SRE methods from inside MathJax.
2. The `PATTERNS` namespace contains calls to previously defined constants in the `RegExp`, which does not work inside an object. Therefore the `PATTERNS` object only exports the regexp (those are the only ones used elsewhere), while the pure string expressions are in an `enum` and a `const` object, respectively.

Also adds two missing tests for the `newcommand` package.


ts/adaptors/lite/Parser.ts
```diff
@@ -32,23 +32,30 @@ import { LiteAdaptor } from '../liteAdaptor.js';
...
 }
 
+const Complex = {
+  ATTRIBUTE: `${Simple.ATTNAME}(?:${Simple.OPTIONALSPACE}=${Simple.OPTIONALSPACE}${Simple.VALUE})?`,
+  ATTRIBUTESPLIT: `(${Simple.ATTNAME})(?:${Simple.OPTIONALSPACE}=${Simple.OPTIONALSPACE}${Simple.VALUESPLIT})?`,
+  TAG: '',
+};
+Complex.TAG =
+  `(<(?:${Simple.TAGNAME}(?:${Simple.SPACE}${Complex.ATTRIBUTE})*` +
+  `${Simple.OPTIONALSPACE}/?|/${Simple.TAGNAME}|!--[^]*?--|![^]*?)(?:>|$))`;
```
dpvc, line 51: Since these are not being exported, can they just be local `const` variables rather than packaging them up in arbitrary objects?  That should make the handling of `TAG` a bit less cumbersome.

ts/components/loader.ts
```diff
@@ -150,26 +150,26 @@ export const PathFilters: { [name: string]: PathFilterFunction } = {
...
+const VERSION = MJGlobal.version;
 
-  /**
-   * The versions of all the loaded extensions.
-   */
-  export const versions: Map<string, string> = new Map();
+/**
+ * The versions of all the loaded extensions.
+ */
+const versions = new Map<string, string>();
```
dpvc, line 160: I would like this to be exported as well, as I plan to use this in the future to make a better list of components in the "About MathJax" menu item.  Or it could be in the `Loader` constant below, in which case, you need `this.versions` in the `load()` and `saveVersion()` methods below.

ts/input/tex/NodeUtil.ts
```diff
@@ -33,62 +33,62 @@ import { Property, PropertyList } from '../../core/Tree/Node.js';
...
+  ['movesupsub', true],
+  ['subsupOK', true],
+  ['texprimestyle', true],
+  ['useHeight', true],
+  ['variantForm', true],
+  ['withDelims', true],
+  ['mathaccent', true],
+  ['open', true],
+  ['close', true],
+]);
```
dpvc, line 48: It looks like this could be made into a `Set` rather than a `Map`.  Also, should this be a proper of the `NodeUtil` object so that the list can be augmented if needed?

dpvc (approved):
A few comments, but otherwise, looks good.

ts/adaptors/lite/Parser.ts
zorkow: Done.

ts/input/tex/NodeUtil.ts
zorkow: Done. Ultimately, all these strings should become string enums, so we don't run into problems with expressions like `{open: true, stretchy: true}` etc. when we are using different bundlers (rollup, tscc).

ts/components/loader.ts
zorkow: Included `versions` in `Loader`, but referenced as `Loader.versions` as v8 can not optimize objects with `this` references. (I learned that at Google IO 2013, so this might be outdated knowledge.)

PR #1160, opened 2024-11-12
Remove startup namespace

zorkow:
Refactors the `Startup` namespace into an `abstract class`. It can't be turned into a `const` object as it contains a number of uninitialised elements. 
I followed the trickery at:
https://stackoverflow.com/questions/13212521/typescript-static-classes


dpvc:
Nope, I'm wrong.  `Startup` gets minified to a single letter during web packing, it seems.  So leave as is.

zorkow:
> Nope, I'm wrong. `Startup` gets minified to a single letter during web packing, it seems. So leave as is.

Also, since it is effectively a `static` class, it feels wrong to use `this`.

PR #1161, opened 2024-11-12
Removes occurrences of aliased `this`

zorkow:
Removes all aliasing of `this` in the code.
Note, that the function in `Factory.ts` works because arrow functions preserve lexical scope.


dpvc (approved):
This all looks good.

PR #1163, opened 2024-11-15
Lints new issues

zorkow:
PR does three things:

* Lints a few issues that have been introduced during the latest PR merges
* Adds the linting rule to prefer `includes` over `indexOf` for array membership test, as discussed f2f
* Adds mandatory linting for every commit. Note,
    * this means that commit will fail if there are lint errors, but not when there are only warnings 
    * it also adds a bit of extra time when committing. If this is annoying we can take it out again.


dpvc (approved):
Looks good.

PR #1165, opened 2024-11-16
Update newcommand to handle delimiters vs macros better, and add tests

dpvc:
This PR fixes some problems with the `newcommand` package where definitions for macros and delimiters don't always stay synchronized (e.g., `\let\|=\sqrt \left\| X \right\|` should produce an error but doesn't currently).

In order to manage this, the `addMacro()` and `addDelimiter()` methods in `NewcommandUtil` are modified to keep the macro and delimiter mappings in sync.  One approach would have been to delete a delimiter when a corresponding macro is defined, for instance; but we don't want to remove delimiters from the main `delimiter` mapping, as that is global, and might be in use by other TeX input jax.  Had we done this, changes in one TeX input jax would affect all others.  In particular, with the jest tests, where the TeX input has is re-instantiated for each test, would see previous tests affect later ones, for example.

In order to avoid that problem, we the ability to have definitions that force the parsing of macros and delimiters to terminate and drop directly to the fallback method.  For macros, if the parse function returns a new special symbol, that causes the parse loop to terminate early and go to the fallback method.  For delimiters, if the character is `null`, that terminates the lookup in the `applicable()` method.  These are the changes in the `MapHandler.ts` file.

The structure of the `Let()` method in `NewcommandMethods.ts` is reorganized a bit to break out the various cases for the defining control sequence (macro vs. character vs. delimiter), and to include a new case where the macro becomes undefined (in the past, `\def\test{\text{Wrong!}} \let\test=\undefined \test` would produce `Wrong!` rather than an undefined macro error, as it should). 

The case for a CharacterMap now excludes DelimiterMap, since something like `\let\test=<` would find the delimiter definition and use that to create `\test` as a macro rather than a delimiter.  For example, `\def\test{X}\let\test=< \test` used to produce `X` rather than `⟨`.

Finally, 17 new tests are added to check for the various interactions of macros and delimiters.  With this PR, together with `fix/3300-tests` all the tests should now pass.

### DIFF
--- a/testsuite/tsconfig.json
+++ b/testsuite/tsconfig.json
@@ -3,7 +3,6 @@
   "compilerOptions": {
     "rootDir": ".",
     "outDir": "./js",
-    "moduleResolution": "node",
     "paths": {
       "#js/*": ["../mjs/*"],
       "#source/*": ["../components/mjs/*"],
--- a/ts/input/tex/newcommand/NewcommandUtil.ts
+++ b/ts/input/tex/newcommand/NewcommandUtil.ts
@@ -21,6 +21,8 @@
  * @author v.sorge@mathjax.org (Volker Sorge)
  */
 
+import { HandlerType } from '../HandlerTypes.js';
+import { SubHandler } from '../MapHandler.js';
 import { UnitUtil } from '../UnitUtil.js';
 import TexError from '../TexError.js';
 import TexParser from '../TexParser.js';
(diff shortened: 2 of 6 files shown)


ts/input/tex/newcommand/NewcommandUtil.ts
```diff
@@ -320,4 +325,48 @@ export const NewcommandUtil = {
...
+   */
+  undefineDelimiter(parser: TexParser, cs: string) {
+    const handlers = parser.configuration.handlers;
+    const delims = handlers.retrieve(NewcommandTables.NEW_DELIMITER);
+    (delims as sm.DelimiterMap).remove(cs);
+    if (handlers.get(HandlerType.DELIMITER).applicable(cs)) {
+      //
+      // This will hide the delimiter that is in a later mapping
+      // by forcing the parser to jump directly to the fallback
+      // handler.
```
zorkow, line 367: It does not really jump to the fallback handler.

zorkow (approved):
A clarification on a comment. Otherwise fine.

testsuite/tsconfig.json
```diff
@@ -3,7 +3,6 @@
   "compilerOptions": {
     "rootDir": ".",
     "outDir": "./js",
-    "moduleResolution": "node",
```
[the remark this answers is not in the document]
dpvc: OK, I put it to `nodenext`, but apparently that means that we should be using `.js` on the import names in `src/index.js` and other `src` files (at least emacs tells me this).  Doing that, however, cases the tests to fail, but leaving the files as they are doesn't seem to cause the tests any problems, so I'm not changing them.

I don't remember why I originally removed it, but I'm OK with putting this as `nodenext`.

ts/input/tex/newcommand/NewcommandUtil.ts
dpvc: OK, I've adjusted the comment.

PR #1167, opened 2024-11-16
Update lint and format scripts to allow specifying the files to process

dpvc:
This PR modifies the `lint` and `format` scripts to allow you to specify the files to be processed.  It also makes `mjs` the default second parameter for `make-one`.


zorkow (approved):
I was not aware of the optional syntax. Nice.

PR #1168, opened 2024-11-19
Move to scoped packages for fonts, and use woff2 format

dpvc:
This PR moves to using WOFF2 formatted fonts, and moves the font npm packages into the `@mathjax` scope.  I've already published the font packages in that scope.  A later step before beta.8 is release will be to move `mathjax-full` into the scoped package as well, which will require additional changes, both to the fonts and to MathJax source (`mathjax-full` will need to be replaced by `@mathjax/src`, for example).  That will be done in a separate PR later.

The `fontURL` and `dynamicPrefix` values in these files aren't actually used, as they are overridden by the fonts, which are subclasses of the `FontData` objects and provide their own value for these.

This PR also corrects the tsconfig files to point to `mathjax-newcm` rather than `mathjax-modern`, which were missed in the PR that swapped the default font.

### DIFF
--- a/components/bin/package.json
+++ b/components/bin/package.json
@@ -7,6 +7,6 @@
     "#menu/*": "mj-context-menu/cjs/*",
     "#sre/*": "speech-rule-engine/cjs/*",
     "#mhchem/*": "mhchemparser/dist/*",
-    "#default-font/*": "mathjax-newcm-font/cjs/*"
+    "#default-font/*": "@mathjax/mathjax-newcm-font/cjs/*"
   }
 }
--- a/components/mjs/input/tex/extension.js
+++ b/components/mjs/input/tex/extension.js
@@ -4,7 +4,7 @@ import {hasWindow} from '#js/util/context.js';
 export function fontExtension(id, name, pkg = `@mathjax/${name}`) {
   if (MathJax.loader) {
     const FONTPATH = hasWindow ? `https://cdn.jsdelivr.net/npm/${name}` : pkg;
-    const path = name.replace(/-font-extension$/, '-extension');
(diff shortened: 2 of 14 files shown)


dpvc:
I forgot to say that you need to use the `lab-woff2` branch in the MathJax-dev repository to test this.  Do a `ppm install` first to get the new versions of the fonts.

zorkow (changes requested):
Finally tested it full. Works initially great in the lab. It also loads the font files for `newcm` correctly, e.g., `mjx-ncm-ds.woff2` and `mjx-ncm-c.woff2` for `mathbb` and `mathcal`, respectively. 

However, when I try to change the output font to, say, `asana` I get the following error:
```
Failed to fetch dynamically imported module: ￼http://localhost/MathJax/MathJax-dev/node_modules/@mathjax/mathjax-asana-font/chtml/woff2/chtml.js
```
* When changing the `fontURL` then I `newcm` does not load anymore on startup.
* When I copy the `chtml.js` from the toplevel of `mathjax-asana-font` to `woff2`, it works for the basic font, but fails for doublestruck and caligraphic, as it tries to find the sources at `woff2/chtml/dynamic` rather than `woff2/dynamic`
* When using the SVG renderer, I run into the same problem. 

Is there something wrong in my setup? Just in case, I installed and built everything from scratch.

dpvc:
It looks like a path is incorrectly set.  I'll look into it.

dpvc:
Ok, I've updated both the `lab-woff2` branch and the `woff2` branch in MathJax-src to fix these issues.  You will need to do `pnpm install` again when you get the new `lab-woff2` branch in order to get the mhchem font (for use with the `mhchem-stretchies` branch).

zorkow (approved):
lgtm.

PR #1170, opened 2024-11-24
Fix more problems from #1146

dpvc:
The changes in #1146 caused an unintended consequence of not making some tables be in display style, and I accidentally "fixed" the tests to match the wrong results during the massive update I did on the test output.  This PR fixes that problem with tables and returns the tests to their proper values.


zorkow (approved):
lgtm.

PR #1171, opened 2024-11-24
Handle clipping of characters in CHTML better and avoid vertical scroll bars for overflow="scroll"

dpvc:
This PR introduces clipping paths for the CHTML characters so that when selected, they highlight region will match the character bounding box (rather than having excessive depth, for example).  It also clips the spaces used for in-line breaking so that they don't show selection, just as the spacing for non-breakpoints doesn't show.

This clipping makes the focus outline match the bounding box better, particularly for the explorer on in-line expressions (the outline used to be excessively below the expression).  This necessitated some changes to the CSS for inline breaking, and requires one more rule to be in the font-specific CSS, so I had to rebuild all the fonts.  You will need to use the woff2 version of the lab to test this (as this is build on top of the woff2 branch).

I also normalized the size of the explorer outline so that the HTML and SVG versions are more consistent (so slightly smaller outline in CHTML and slightly larger in SVG).  I always thought the CHTML outline was too heavy anyway.

I also moved some CSS that was common to both CHTML and SVG to the `common.ts` file, for easier maintenance.

Finally, when `overflow` is either `scroll` or `truncate`, the `overflow-y` CSS is set to clip, to avoid unwanted vertical scroll bars.

### DIFF
--- a/ts/output/chtml.ts
+++ b/ts/output/chtml.ts
@@ -97,7 +97,15 @@ export class CHTML<N, T, D> extends CommonOutputJax<
       'white-space': 'nowrap',
     },
 
-    'mjx-container[jax="CHTML"] :focus': { outline: 'solid 3px' },
+    //
+    // Clip the token elements' character content,
+    //   to remove excessive height and depth of ZERO font
+    //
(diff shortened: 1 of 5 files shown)


zorkow (approved):
lgtm.

PR #1172, opened 2025-01-13
Add textmacros to the input/tex components

dpvc:
This PR adds the `textmacros` package to the list of default packages for any component that includes `input/tex`.  That means that macros will be processed within `\text{}` and other text-mode macros.

This could affect existing expressions that include backslashes that used to be displayed verbatim, but now will try to process the macros.  Sites can use

``` js
MathJax = {
  tex: {
    packages: {'[-]': ['textmacros']}
  }
};
```

to return to the previous behavior if they are loading a configuration that includes TeX input.


zorkow (approved):
lgtm.

PR #1174, opened 2025-01-13
Add stretchy characters needed by mhchem

dpvc:
This PR adjusts the `mhchem` package to provide several stretchy characters that it needed but that weren't in the original MathJax TeX fonts (and two that don't have Unicode equivalents).  This is done by using a separate `mathjax-mhchem` font that has the needed arrows and pieces needed for stretchy assemblies for these characters, based on the `mathjax-newcm` arrows.  In order to make the arrows used in mhchem consistent with these, all the arrows uses in mhchem are included in this new font, so that even when mhchem is used in a setting where a different font is selected, the arrows will not change their look between ones that could have been found in the selected font and ones that are in the `mathjax-mhchem` font.

The new font also includes character used for the bonds that have in the past been constructed from multiple characters.

In order to use these characters and arrows, we end up scanning the TeX string constructed by mhchem and replacing the TeX code that creates the arrows and bonds by new mhchem-specific macros.  That is a bit ugly, but gets the job done, without overriding existing macros.  After v4.0 is released, I can approach Martin about chaing his code to make this easier, but didn't to do that until everythig was ready for him.

### DIFF
--- a/components/mjs/input/tex/extensions/mhchem/mhchem.js
+++ b/components/mjs/input/tex/extensions/mhchem/mhchem.js
@@ -1,1 +1,4 @@
 import './lib/mhchem.js';
+import {fontExtension} from '../../extension.js';
+
+fontExtension('[tex]/mhchem', 'mathjax-mhchem-font-extension');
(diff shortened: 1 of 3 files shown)


dpvc:
PS, you will need the `woff2` branch to test this, since the new `mathjax-mhchem` font is in that form.  For best results, also include `chtml-stretchies` as the results are much nicer.  I still need to push the new versions of all the fonts, including this one, and will make another post when that is done.

dpvc:
The `mathjax-mhchem` font is now in place.

zorkow:
Code looks good. I'll test it once all the other PRs are merged.

PR #1175, opened 2025-01-15
Make bboldx delimiters be plain characters, as in the actual package.

dpvc:
This PR makes the bboldx delimiters into plain characters, as they are in the actual TeX package.

Since the characters used are actual parentheses, brackets, and angles, but just in a different variant, they obtain their TeX class (and other properties) from the operator dictionary.  So the open ones have TeX class OPEN, and the closes have TeX class CLOSE, as expected.

I'm not sure whey the `fence` property was being set to `false`, as the operator dictionary says they should be `true`, so I've removed that.  If there is a reason for it, I can put that back.


zorkow (approved):
lgtm.

PR #1176, opened 2025-01-15
Use Info's keydown() method to handle ctrl-A (it wasn't working this way).

dpvc:
This PR changes from overriding `addEvents()` to overriding `keydown()` to handle the select-all action.  It wasn't working (at least for me in Firefox), and the lack of a `super.addEvents(element)` was preventing other handlers from being added, like the one for ESC to close the dialog.


zorkow (approved):
lgtm.

PR #1177, opened 2025-01-15
Rename loader.preLoad() to loader.preLoaded() to make it clear tht this doesn't do any loading itself.

dpvc:
This PR renames the `loader.preLoad()` method to `loader.preLoaded()` to make it clear that this function doesn't do any actual loading of files itself, but rather just tells the loader that you have preloaded some files.  There is no functionality difference, but the name has always been more confusing than it should be.

It would be possible to have `loader.preLoad()` call `loader.preLoaded()` for backward compatibility.  Do you thin that is necessary?


zorkow (approved):
> It would be possible to have loader.preLoad() call loader.preLoaded() for backward compatibility. Do you thin that is necessary?

No, there is nothing wrong with a breaking change in 4.0.

PR #1178, opened 2025-01-16
Allow better configuration for patterns for numbers and identifiers

dpvc:
This PR makes it possible to more fully configure the patterns used for number recognition and multi-letter identifiers.  In the past, you could set the pattern used for a number, but not the characters that initiate the check for one, and you could specify the pattern to use for multi-letter itednifiers in `\mathrm{}` and the other similar macros, but could not set the letter pattern that initiated the check.  That makes it harder to configure MathJax for use in other languages, that have different number characters, such as Persian and Arabic, or if you wanted to use accented letters directly in identifiers.

Here, we move the `digits` pattern to `numberPattern` in the `base` configuration and add an `initialDigit` pattern.  Similarly, we add an `initialLetter` pattern for identifiers.  Then we create the `RegExpMap`s for `digit` and `letter` in the base `config()` method (where the options have been initialized) and add these to the handler list at this point, so these are removed from the original handler lists because we don't have the options available then to set up the regular expressions.  (I also added a backward-compatibility `digits` that maps onto `numberPattern` if specified, so that old configurations can still work.  But I could be persuaded to remove that, since v4 is breaking anyway.)

The `ParseMethods.digit()` method is modified to drop out without doing anything if the number pattern doesn't match (e.g., if it is just a decimal point not followed by a digit), so that in case there is another handler that can process it, that will be used instead of just putting it in an `mo`.  (I considered doing something similar for `ParseMethods.varaiable()`, but don't think it is necessary.)

One question is whether we should include more letters in the letter patterns, like should they include the Latin accented character by default?

### DIFF
--- a/ts/input/tex.ts
+++ b/ts/input/tex.ts
@@ -68,8 +68,6 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
     ...AbstractInputJax.OPTIONS,
     FindTeX: null,
     packages: ['base'],
-    // Digit pattern to match numbers.
-    digits: /^(?:[0-9]+(?:\{,\}[0-9]{3})*(?:\.[0-9]*)?|\.[0-9]+)/,
     // Maximum size of TeX string to process.
     maxBuffer: 5 * 1024,
     // Maximum number of array template substitutions (avoids infinite loop from @{\\} for example)
--- a/ts/input/tex/ParseMethods.ts
+++ b/ts/input/tex/ParseMethods.ts
@@ -26,9 +26,8 @@ import { Token } from './Token.js';
 import TexParser from './TexParser.js';
 import NodeUtil from './NodeUtil.js';
 import { TexConstant } from './TexConstants.js';
-import { MmlNode } from '../../core/MmlTree/MmlNode.js';
 import { ParseUtil } from './ParseUtil.js';
-import { ParseMethod } from './Types.js';
+import { ParseMethod, ParseResult } from './Types.js';
 import { StackItem } from './StackItem.js';
 
 const MATHVARIANT = TexConstant.Variant;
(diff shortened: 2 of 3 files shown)


dpvc:
> Since the mappings are now explicitly defined in the configuration, can we get rid of the definitions in `BaseMappings.ts`.

Yes, I will remove them.

> Can we think of a way to *a posteriori* add number and letter regExps?

Sure.  One could do something like

``` js
const handler = config.handlers.get(HandlerType.CHARACTER);
handler.remove(['letter', 'digit'], null);
new RegExpMap('digit', ParseMethods.digit, options.initialDigit);
new RegExpMap('letter', ParseMethods.variable, options.initialLetter);
handler.add(['letter', 'digit'], null, 4);
```

where `options.initialDigit` and `options.initialLetter` are the new patterns that you want to use.

dpvc:
On looking at this again now, I'm wondering if we really need a separate pattern for the initial versus the complete one.  Couldn't the two `RegExpMap` objects take the complete pattern, provided they also match a single letter or number?  The defaults do match a single letter or number, but there is no guarantee that a user's patterns will.  But having the redundant patterns for the first letter or number is also a potential source of inconsistency, so it would be nice to only use one pattern for each.

zorkow:
> On looking at this again now, I'm wondering if we really need a separate pattern for the initial versus the complete one. Couldn't the two `RegExpMap` objects take the complete pattern, provided they also match a single letter or number? The defaults do match a single letter or number, but there is no guarantee that a user's patterns will. But having the redundant patterns for the first letter or number is also a potential source of inconsistency, so it would be nice to only use one pattern for each.

We discussed this last week. Do you want to make that change in this PR or in a separate one?

dpvc:
It turns out that this doesn't work as I had hoped, as you are right that numbers starting with a decimal point don't get parsed properly; the pattern for the `RegExpMap` needs to match a single character at the beginning of a number, but a single decimal point doesn't match since the pattern is `\.[0-9]+`.  We can't match a single decimal place in the number pattern, since that would cause a single period to be considered a number.  So I think this needs to stay as is, with two separate patterns, at least for the numbers.  We could use a single one for identifiers, though, if that seems useful.

dpvc:
(So you were right!)

zorkow:
> (So you were right!)

Thank you. Not sure if it's helpful though...
Anyway, I'll approve now.

zorkow (approved):

PR #1180, opened 2025-01-20
Rename StyleList to StyleJson, etc.

dpvc:
This PR implements the changes we discussed last week to disambiguate the two uses of `StyleList`, but not quite as I suggested at the time.  Instead, I'm changing `StyleList` to `StyleJson`, `StyleData` to `StyleJsonData`, and `CssStyles` to `StyleJsonSheet`, as this more accurately describes what it is.  The `StyleJson` type gives style definitions as eseentially JSON objects (though not as strict about quotation marks), and `StyleJsonSheet` is the object that holds `StyleJson` data (not my favorite name, but the best I could come up with), whereas the `Styles` object in `util/Style.ts` is an actual style implementation (for use with parsing and managing styles that are applied to the internal MathML elements).  Calling the class `CssStyles` still makes it unclear which is which.


zorkow (approved):
lgtm.

So the remaining `StyleList` type is really just a string to string map. Sometimes it is used as an `OptionList`, which can be a bit confusing.

dpvc:
> Sometimes it is used as an `OptionList`, which can be a bit confusing.

I don't see that.  Can you point out where?  Or do you mean that styles can be included as part of an option specification?

dpvc:
As discussed face-to-face, I'm merging this.

PR #1184, opened 2025-01-21
Add ability to filter SRE and LaTeX attributes from SVG image

dpvc:
This PR allows you to filter attributes from the SVG image generated by the contextual menu "Show As" or "Copy to Clipboard" submenus.  It uses the settings from the "MathML includes" menu, renamed as "MathML/SVG has" (to not make it wider than it was).

This makes it easier to make cleaner SVG images.


zorkow (approved):
lgtm.

PR #1185, opened 2025-01-22
Sanitize during set rather than cssText, and add tests for sanitizing

dpvc:
This PR moves the sanitization of CSS values to when they are set, rather than when they are output.  This should prevent problems if the values are used internally, not just when they are output to actual DOM elements.

I also added tests for the sanitization.


zorkow (approved):
lgtm.

PR #1187, opened 2025-02-04
Update test framework

dpvc:
This is the first in a series of PRs that I'm going to be making that add or updates tests for the TeX input jax and its packages.  I have broken down the PRs into separate pieces by tex package, to make them easier to review.  But some rely on some of the open PRs (like `rename-stylelist`, or `woff2`, or `begingroup`) in order to run, and I have not merged those into these branches, other than the `mhchem-stretchies` branch, which is merged into the `test-mhchem` PR that I will be making.  So you may not be able to test these individually without merging some of the other branches.  Many of these PRs include changes to the ts/input/tex package files to either fix bugs that I found while making the tests, or to remove unreachable code that was identified by jest during testing.  I will try to explain those changes in the PRs themselves.

In addition, the `test-tex` PR includes changes to the main ts/input/tex files along with miscellaneous tests to fill out the coverage for lines from those files that weren't run by other tests.  But the changes in these files may be needed by other PRs, so yo umight want to merge that into a test branch, along with this PR, and use that as a basis for runing the tests in the other PRs, if you want to do that.

To get the complete coverage, you will need to merge all these test PRs together, and merge in all the pending PRs of mine in order to for them all to run properly.  Note that the testsuite/package.json file has been updated, so you should to an `pnpm install` in the `testsuites` directory to get the added font package that is needed for one of the tests.

Merging the pending PRs will cause merge conflicts in several cases.  If you want to merge them into your test branch, you should always keep the changes from the `test-...` branches, and discard the changes in the older PRs, as the test branches are the most up-to-date.  I will update either these or the older PRs to resolve conflicts once one or the other is merged.

-----

This PR adds new functionality to the test framework.  In particular:

* New functions are added in `setupTex.js` that provide for component-based tests, like those needed for the `require` and `autoload` packages.  You use `setupComponents()` (once per file) to specify the needed MathJax configuration (loading the needed components), and then `setupTexTypeset()` or `setupTexPage()` to configure the TeX input jax on a per-test basis.  The tests then use `async` functions along with `await typeset2mml()` or `await page2mml()` to perform the tests.  Traditional tests and component-based tests can both be used int he same file.
* A new `render2mml()` function and associated `setupTexRender()` use an `HTMLDocument.render()` call rather than `HTMLDocument.convert()` call to do the typesetting.  This is needed for the `texhtml` tests, in order to have actual `<tex-html>` DOM elements to work with rather than just LaTeX text.
* New functions `trapErrors()`, `trapAsyncErrors()`, and `trapOutput()` are provided that can be used to test for TeX syntax errors, or for `console` messages.  The first two are accomplished by overriding the `formatError()` option of the TeX input jax and the `compileError()` option of the `MathDocument` to throw errors rather than generate their ususal `merror` elements, and trapping those errors to retrieve their messages.  The latter is done by temporarily overriding the `console.log` or `console.warn` methods to capture their output so it can be tested.
* Error trapping is made even easier by new functions `expectTexError()` and `expectTypesetError()`, which wrap the appropriate typesetting calls inside `expect()` calls.
* Configuration snippets for the error trapping are also made available, in case they are needed in a more sophisticated testing setting.

* A new `toXmlArrayMatch()` is added to `xmlMatch.ts` that matches an array of XML strings to an array of expected strings.  This is for use with `page2mml()`, which produces an array of XML, one for each MathItem in the page.

* The `texReporter` output is now colorized like that in the main jest file output, and all the tables are combine into a single one, similar to the jest output.  The "Missing" column is clipped to the width of the terminal window, as it is in the jest output.

Finally, the tsconfig file is modified and `.js` is added to the imports to avoid typescript errors.

### DIFF
--- a/testsuite/src/setupTex.ts
+++ b/testsuite/src/setupTex.ts
@@ -63,7 +63,7 @@ export function trapOutput(method: string, code: () => void) {
 }
 
 /**
- * Trap errors produces while running code.
+ * Trap errors produced while running code.
  * @param code The code to run.
  * @return The error message produced.
  */
@@ -76,7 +76,7 @@ export function trapErrors(code: () => void) {
 }
 
 /**
- * Trap errors produces while running code.
+ * Trap errors produced while running code.
  * @param code The code to run.
  * @return The error message produced.
  */
@@ -89,7 +89,7 @@ export async function trapAsyncErrors(code: () => Promise<void>) {
 }
 
 /**
- * When true, errors with throw rather than poroduce merror elements.
+ * When true, errors will throw rather than produce merror elements.
  */
 let reportErrors = false;
 
--- a/testsuite/src/setupTex.ts
+++ b/testsuite/src/setupTex.ts
@@ -326,12 +326,8 @@ export async function setupComponents(config: any) {
   MathJax.config.loader.require = (file: string) => {
     return new Promise((ok, fail) => import(file).then(ok).catch(e => fail(e)));
   }
-  if (!config.startup) {
-    config.startup = {};
-  }
-  if (config.startup.typeset === undefined) {
-    config.startup.typeset = false;
-  }
+  config.startup ??= {};
+  config.startup.typeset ??= false;
   config = Object.assign({}, throwCompileErrors, config);
   componentPromise = init(config);
 }
--- a/testsuite/src/setupTex.ts
+++ b/testsuite/src/setupTex.ts
@@ -1,43 +1,342 @@
-import {TeX} from '#js/input/tex';
-import {AbstractParseMap, RegExpMap} from '#js/input/tex/TokenMap';
-import {ConfigurationHandler} from '#js/input/tex/Configuration';
+import {TeX} from '#js/input/tex.js';
+import {AbstractParseMap, RegExpMap, CommandMap} from '#js/input/tex/TokenMap.js';
+import {ConfigurationHandler} from '#js/input/tex/Configuration.js';
 import {HandlerType, ConfigurationType} from '#js/input/tex/HandlerTypes.js';
 import {MapHandler} from '#js/input/tex/MapHandler.js';
-import {HTMLDocument} from '#js/handlers/html/HTMLDocument';
-import {liteAdaptor} from '#js/adaptors/liteAdaptor';
-import {STATE} from '#js/core/MathItem';
-import {SerializedMmlVisitor} from '#js/core/MmlTree/SerializedMmlVisitor';
-import {MmlNode} from '#js/core/MmlTree/MmlNode';
-import {tmpJsonFile} from './constants';
+import {HTMLDocument} from '#js/handlers/html/HTMLDocument.js';
+import {RegisterHTMLHandler} from '#js/handlers/html.js';
+import {liteAdaptor} from '#js/adaptors/liteAdaptor.js';
+import {MathItem, STATE} from '#js/core/MathItem.js';
+import {SerializedMmlVisitor} from '#js/core/MmlTree/SerializedMmlVisitor.js';
+import {MmlNode} from '#js/core/MmlTree/MmlNode.js';
+import {mathjax} from '#js/mathjax.js';
+import {OptionList} from '#js/util/Options.js';
+import {tmpJsonFile} from '#src/constants.js';
 import * as fs from 'fs';
+import {init} from '#source/node-main/node-main.mjs';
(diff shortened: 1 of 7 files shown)


testsuite/src/setupTex.ts
```diff
@@ -1,44 +1,343 @@
...
+let render: (text: string, display: boolean) => string;
+let typeset: (text: string, display: boolean) => Promise<string>;
+let page: (text: string) => Promise<string[]>;
 
-export function setupTex(packages: string[] = ['base'], options = {}) {
-  const parserOptions = Object.assign({}, {packages: packages}, options);
+/**
+ * A promise that resolves when the components are loaded and set up.
+ */
+let componentPromise: Promise<string[]>;
```
zorkow, line 34: I wonder if this is reset every time we add components in a single test fun.
For example, if I have a test file with a number of `describe` elements, each containing a setup command in a `beforeEach`, will the promise be renewed every time?

testsuite/src/setupTex.ts
```diff
@@ -1,44 +1,343 @@
...
+ */
+export function setupTexTypeset(packages: PackageList = ['base'], options: OptionList = {}) {
+  MathJax.config.tex = Object.assign(
+    {},
+    {packages: packages, inlineMath: {'[+]': [['$', '$']]}},
+    throwTexErrors,
+    options
+  );
+  typeset = async (text: string, display: boolean) => {
+    await componentPromise;
```
zorkow, line 205: See question above.

testsuite/src/setupTex.ts
```diff
@@ -1,44 +1,343 @@
...
+ */
+export function setupTexPage(packages: PackageList = ['base'], options: OptionList = {}) {
+  MathJax.config.tex = Object.assign(
+    {},
+    {packages: packages, inlineMath: {'[+]': [['$', '$']]}},
+    throwTexErrors,
+    options
+  );
+  page = async (text: string) => {
+    await componentPromise;
```
zorkow, line 233: ditto.

zorkow (approved):
There is mainly one question on the `componentPromise`, which should be resolved. 
Otherwise  a number of little things.

And the tex reporter list looks really good now! Well done!

testsuite/src/setupTex.ts
dpvc: The `componentPromise` is the one obtained from the `init()` function from the importing `mathjax-full` (which load the `node-main` component).  That allows you to pass a configuration that loads components, etc.  That should only be done once, loading all the components that may be needed for a particular file.  For TeX tests, it should load whatever component packages are going to be used.

Once that is done, the individual tests or `beforeEach` command can use `setupTexTypeset()` or `setupTexPage()` to set the configuration for the TeX input jax that these create, using any (or none) of the packages that were loaded.  The purpose of this promise is simply to wait for the components to be fully loaded before starting any test that uses components.  The `setupTexTypeset()`, `setupTexPage()`, or any bespoke code can create and configure the TeX input jax (or other objects) as needed.

This is done in several of the test files that come in later PRs.

testsuite/src/setupTex.ts
dpvc: Same answer.

testsuite/src/setupTex.ts
dpvc: ditto.

testsuite/src/texReporter.js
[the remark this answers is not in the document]
dpvc: Thanks!  I thought it looked good, too.

testsuite/src/setupTex.ts
dpvc: See the [Require.test.ts](https://github.com/mathjax/MathJax-src/pull/1203/files#diff-9f1f26be14e215c76170a04afb208f026d6b01e6092cebbac97b12858a9c64d5) file, for example, which has multiple different `setupTexTypeset()` calls.

testsuite/src/setupTex.ts
```diff
@@ -1,44 +1,343 @@
...
+  //
+  MathJax.config.loader.require = (file: string) => {
+    return new Promise((ok, fail) => import(file).then(ok).catch(e => fail(e)));
+  }
+  if (!config.startup) {
+    config.startup = {};
+  }
+  if (config.startup.typeset === undefined) {
+    config.startup.typeset = false;
+  }
```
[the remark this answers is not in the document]
dpvc: done

PR #1188, opened 2025-02-04
Normalize format for tests and convert error tests to trapped errors

dpvc:
I have found it difficult to read the test files, as it is hard to see where one starts and other stops, and the indenting of the XML is not consistent and gets in the way.  Consequently, I have reformatted all the tests to add separators between the tests and the `describe()` calls, and have normalized the indentation of the XML data.  This makes it easier for me to read the files, and I hope it helps you as well.

In addition, I changed the tests that look for errors to using the new `expectTexError()` calls.

This PR just does that reformatting to the test files that I didn't otherwise modify, except that one test was added to the `Colorv2.test.ts` file.

This PR is probably best viewed with spacing changes ignored, since most of the changes are indentation ones.

### DIFF
--- a/testsuite/tests/input/tex/Action.test.ts
+++ b/testsuite/tests/input/tex/Action.test.ts
@@ -4,38 +4,59 @@ import '#js/input/tex/action/ActionConfiguration';
 
 beforeEach(() => setupTex(['base', 'action']));
 
+/**********************************************************************************/
+/**********************************************************************************/
+
 describe('Action', () => {
-  it('TextTip', () =>
+
+  /********************************************************************************/
+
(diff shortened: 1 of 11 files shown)


zorkow (approved):
The indentation issue stems from the `jest-replace-expected-for-actual` function in `scripts` folder. That simply copies the actual element from the output buffer, which has only the XML indentation but not the additional JS indentation.

Do you have a corresponding script for the cleanup, which we could add for future use?

dpvc:
I do, but it does some other stuff, too.  I can see about isolating the indentation part and submit that.

PR #1190, opened 2025-02-04
Miscellaneous tests for core tex files, and update code

dpvc:
This PR adds a `Tex.test.ts` file that is intended to complete the coverage of the TeX input jax main files for lines that aren't run by other package-based tests.  Therefore, it is a bit of a hodge-podge of miscellaneous tests.

It also includes modifications to a number of the core TeX files in order to either correct bugs, or remove unreachable code, as described below:

* In `tex.ts`, the `globalEnv` is no longer used.  It was part of handling the `multline` environment, but that was changed some time ago in the `ams` package, while this code was never removed.
* in `ColumnParser.js`: 
   - the `codePointAt()` function never returns undefined. (line 142)
   - the `getBraces()` function always returns a string, so no need for `|| ''`. (line 265)
   - we want to stop at the end of the string, not one past it (bug). (line 301)
* In `Configuration.ts`,
  - an error message is improved (line 440)
  - and a new warning added (line 443) for when a package is used that isn't defined,
  - using a new `warn()` method like the one in `MapHandler.ts`. (lines 483-491)
* In `FilterUtil.ts`
  - the `data-latex` and `data-latex-item` attributes were interfering with the combining or relations, so the code has been updated to take those into account. (lines 65-70 and 248-254)
  - because `getList()` checks that the items it returns are children of the `MathItem.root` node, their `parent` is always defined, except for the top-level `math` element; since these are `msubsup` elements, they always have a `parent`, so the checks for that are not needed. (lines 107-111 and 140-144)
  - because `walkTree()` doesn't descend into token nodes, the only nodes that it receives will always have an `attributes` property (lines 192-194)
  - an `mstyle` node always has an inferred row as its 0th child, and that always has a `childNode` array (line 329)
* A new `getMoAttribute()` function is added to `NodeUtil.ts` in order to look up an `mo` element's attribute, using the operator dictionary if it is not explicitly given (note that this is used during the construction of the internal MathML tree, before inheritance is put in place, so surrounding `mstyle` changes aren't taken into account); if no operator-dictionary entry exists, the default value is used.
* In `ParseUtil.ts`,
  - using the `KeyValueTypes` caused type errors in `.ts` files, so I have reorganized that a bit to make it work without having to cast anything.  I had to move the `oneof()` method to the `KeyValueDef` class (renamed from `KeyValueType`, which is now a `type`).  So a type definition object would look like:
    
    ``` js
    const allowed = {
      compact: KeyValueTypes.boolean,
      direction: KeyValueDef.oneof('up', 'down'),
      'open-brace': KeyValueTypes.string
    };
    ```
    
    with `KeyValueDef.oneof` rather than `KeyValueTypes.oneof`.  Not as nice, but I wasn't able to get Typescript to accept it any other way.
  - the checking for erroneous nesting of environments was updated some time ago, and that broke the `empheq` package, which nests AMS environments inside an `empheq` environment.  This needed a new `nestStart` property to resolve the issue. (line 818)
* In `StackItem.ts`, all stak items inherit the `error` property of the base item, so it is never undefined. (line 564)
* In `TexError.ts`, the `parts` are never `null`, as the separators always have to match at least one character or they wouldn't be matched.
* In `TexParser.ts`,
  - we add a check for the the character being undefined in order to get the `null` branch coverage. (line 266)
  - The changes in 318 to 523 are just cosmetic: assign `false` as default rather than leaving it undefined, and change `parens` to `braces`, since that is what is really being counted.
  - line 709 prevents `undefined` from being used in things like `^a` that have no base.
  - the node passed to `composeBraces()` is never emtpy.
  - the `!node.childNodes[pos1]` in line 707 is really not needed, either, since the base node is always defined (it is an empty `mi` in `^a` for example), but I've left this in just in case, as I don't really know how this function is used.

### DIFF
--- a/ts/input/tex/FilterUtil.ts
+++ b/ts/input/tex/FilterUtil.ts
@@ -52,6 +52,7 @@ function _copyExplicit(attrs: string[], node1: MmlNode, node2: MmlNode) {
  * - lspace attribute of node1 is ignored.
  * - rspace attribute of node2 is ignored.
  * - stretchy=false attributes are ignored.
+ * - data-latex and -data-latex-item are ignored.
  *
  * @param {MmlNode} node1 The first node.
  * @param {MmlNode} node2 Its next sibling.
@@ -61,7 +62,12 @@ function _compareExplicit(node1: MmlNode, node2: MmlNode): boolean {
   const filter = (attr: Attributes, space: string): string[] => {
     const exp = attr.getExplicitNames();
     return exp.filter((x) => {
-      return x !== space && (x !== 'stretchy' || attr.hasExplicit('stretchy'));
+      return (
+        x !== space &&
+        (x !== 'stretchy' || attr.getExplicit('stretchy')) &&
+        x !== 'data-latex' &&
+        x !== 'data-latex-item'
+      );
     });
   };
   const attr1 = node1.attributes;
@@ -104,11 +110,7 @@ function _cleanSubSup(options: ParseOptions, low: string, up: string) {
           children[mml[up]],
         ]);
     NodeUtil.copyAttributes(mml, newNode);
-    if (parent) {
-      parent.replaceChild(newNode, mml);
-    } else {
-      options.root = newNode;
-    }
+    parent.replaceChild(newNode, mml);
     remove.push(mml);
   }
   options.removeFromList('m' + low + up, remove);
@@ -137,11 +139,7 @@ function _moveLimits(options: ParseOptions, underover: string, subsup: string) {
     ) {
       const node = options.nodeFactory.create('node', subsup, mml.childNodes);
       NodeUtil.copyAttributes(mml, node);
-      if (mml.parent) {
-        mml.parent.replaceChild(node, mml);
-      } else {
-        options.root = node;
-      }
+      mml.parent.replaceChild(node, mml);
       remove.push(mml);
     }
   }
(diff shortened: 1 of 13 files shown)


ts/input/tex/TexParser.ts
```diff
@@ -263,7 +263,7 @@ export default class TexParser {
    */
   public convertDelimiter(c: string): string {
     const token = this.lookup(HandlerType.DELIMITER, c) as Token;
-    return token ? token.char : null;
+    return token && token.char ? token.char : null;
```
zorkow, line 266:
```suggestion
    return token?.char ?? null;
```

ts/input/tex/FilterUtil.ts
```diff
@@ -324,7 +326,7 @@ const FilterUtil = {
     const options = arg.data;
     const remove: MmlNode[] = [];
     for (const mml of options.getList('mstyle')) {
-      if (mml.childNodes?.[0]?.childNodes?.length !== 1) {
+      if (mml.childNodes[0].childNodes.length !== 1) {
```
zorkow, line 329: Can we be absolutely certain these children exist?

zorkow (approved):
Two small issues.
Otherwise I am getting a number of test failures, mainly due to multi relation and its corresponding latex-item. I assume these will be fixed in the subsequent PRs.

PR #1191, opened 2025-02-04
Update ams tests and code

dpvc:
This PR updates the tests for the `ams` package, and adjusts `AmsMethods.ts` as follows:

* The `!taggable` at line 192 is always true, since if `taggable` is true, the block at lines 167-177 will run and `return` a value before this.
* Split up the `let` at lines 390-391.
* You are right, `MultiIntegral` is only used for `\idotsint`, so lines 532-538 can be simplified.
* The `xArrow` function gets a new parameter for the minimum width (so that `xlongrightarrow` can be implemented better, for example).

If you look at the `Ams.test.ts` file, it is probably best to hide whitespace changes.


dpvc:
I'm afraid the diff gets messed up with the `Ams.test.ts` file, as some of the new tests have output that match the following original test output in part, and so the determination of the new and old gets out of sync.

zorkow (approved):
lgtm.

PR #1192, opened 2025-02-04
Update tests for amscd

dpvc:
This PR updates the tests for the `amscd` package and its code.

Here, the `if (a || b)` on line 127 is always true due to lines 124 to 126 that set `a` when it is not set.  The `if` is removed, and the block indentation is adjusted.

The change at line 204 is because the table and row arrays are always defined.

Again, viewing without white-space changes is useful, here.

### DIFF
--- a/ts/input/tex/amscd/AmsCdMethods.ts
+++ b/ts/input/tex/amscd/AmsCdMethods.ts
@@ -126,16 +126,14 @@ const AmsCdMethods: { [key: string]: ParseMethod } = {
         } // minsize needs work
         const pad: EnvList = { width: '+.67em', lspace: '.33em' };
         mml = parser.create('node', 'munderover', [mml]) as MmlMunderover;
-        if (a) {
-          const nodeA = new TexParser(
-            a,
-            parser.stack.env,
-            parser.configuration
-          ).mml();
-          const mpadded = parser.create('node', 'mpadded', [nodeA], pad);
-          NodeUtil.setAttribute(mpadded, 'voffset', '.1em');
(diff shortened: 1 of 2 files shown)


ts/input/tex/amscd/AmsCdMethods.ts
```diff
@@ -124,37 +124,35 @@ const AmsCdMethods: { [key: string]: ParseMethod } = {
...
-          }
-          if (parser.configuration.options.amscd.hideHorizontalLabels) {
-            mml = parser.create('node', 'mpadded', mml, {
-              depth: 0,
-              height: '.67em',
-            });
-          }
+        const pad: EnvList = { width: '+.67em', lspace: '.33em' };
+        mml = parser.create('node', 'munderover', [mml]) as MmlMunderover;
+        if (a) {
```
zorkow, line 129: So why do we need that `if` here if we already know that `a` is always set?

zorkow (changes requested):
I think we can get rid of the conditional in line 130 alttogether.

ts/input/tex/amscd/AmsCdMethods.ts
dpvc: True enough.  I didn't notice it, because I was just responding to the fact that the `|| b` portion of the `if statement above was not being covered, and didn't look further to see if anything else could be eliminated.

dpvc: I've made the change.

zorkow (approved):
lgtm.

PR #1193, opened 2025-02-04
Add tests for autoload Package

dpvc:
This PR adds tests for the `autoload` package, and converts the configuration to use the `ConfigurationType` values rather than explicit names.


zorkow (approved):
lgtm.

PR #1194, opened 2025-02-04
Update tests for the base package and update code

dpvc:
This PR updates and adds to the `base` package tests.  Again, there are many indentation changes, so hide whitespace is a good idea.

Also added is a `Base-browser.test.ts` file that tests the automatic configuration of the `baseURL` option when there is a `<base>` element in the page.  It does so by making a fake window that returns a fake `<base>` element.

The changes to `BaseItems.ts` include:
* Checking for `msup` and `mover` elements in lines 194 and 197.
* The `checkItem()` call always returns true, so the `if` is not needed at line 279 (and the original `return null` is never reached).
* At line 1195, we check that `term` is deifined before testing whether it doesn't start with `\end`
* It is not currently possible to have `breakAlign.cell` be set when `ralign` is set, as `ralign` is only produced from the `p`, `m`, `b`, and `w` (or their capitalized versions), and those operate by inserting `\vtop`, `\vcenter`, and `\vbox` macros into the cell template.  Since `\breakAlign` can only be used as the first thing in a cell, those box macros prevent it from being used in those cells, so this code is unreachable currently.
* In line 1544, the `columnalign` is always set, currently.  All the environments that use the `EqnArrayItem` have `columnalign` set.

In `BaseMethods.ts`,
* We use the new `getMoAttribute()` function in several places to get an attribute of an `mo` element, in order to take the operator-dictionary defaults into account.  You'll need the `test-tex` branch to run the tests, as `getMoAttribute()` is defined there.
* In line 812, the `moNode` is never used, so it is removed.


ts/input/tex/base/BaseItems.ts
```diff
@@ -276,12 +276,10 @@ export class SubsupItem extends BaseItem {
       const result = this.factory.create('mml', top);
       return [[result], true];
     }
-    if (super.checkItem(item)[1]) {
-      // @test Brace Superscript Error, MissingOpenForSup, MissingOpenForSub
-      const error = this.getErrors(['', 'sub', 'sup'][position]);
-      throw new TexError(error[0], error[1], ...error.splice(2));
-    }
-    return null;
+    super.checkItem(item);
```
zorkow, line 279: I know that I never managed to reach the `return null` case. 
But are you sure, that `super.checkItem` can never return `BaseItem.fail`, the final case in `BaseItem`? Then the second element would be `false`.

zorkow (commented):
One question regarding the removed case in `SubsupItem` before we can merge.

ts/input/tex/base/BaseItems.ts
dpvc: Yes, I'm sure.

The `SubsupItem` is a subclass of `BaseItem`, so its `super.checkItem()` is the one from there.  There are only two ways to get a return value of `BaseItem.fail` from that:  either the item being checked is a `cell` and the class doing the checking has `isOpen` true, or if `item.isFinal` is true (because of the `if (!item.isFinal) return BaseItem.success` at the bottom of the super `checkItem()`).

But the first case can't happen, since `SubsupItem` has `isOpen` being false, and the latter can't happen because the only stack item with `isFinal` being true is the `mml` item, and that case is already handed in the `if (item.isKind('mml'))` block just before this, which ends with a `return`, so in that case, the `super.checkItem()` call is never made.

So the check will always succeed, and there is no need for it.  Also, a return value of `null` would cause an error higher up, so we don't want that in any case.

(For anything I removed like this, I did go through an argument like the one above to justify the removal.  It was not just that we never got it from our tests; it was that you couldn't get it.)

zorkow: OK, makes sense.

PR #1195, opened 2025-02-04
Add tests for the bbm, bboldx, and dsfonts packages, and update code

dpvc:
This PR adds tests for the three black-board bold font extensions.  It also includes the changes from `bboldx-delimiters` to change the delimiters to plain character macros.  In `BboldxMethods.ts`, the `parser.options.bboldx` object is set up during the package initialization, so will always be available at line 135.


zorkow (approved):
lgtm.

PR #1196, opened 2025-02-04
Update tests for the boldsymbol package and update code

dpvc:
This PR updates the tests for the `boldsymbol` package, and its configuration file.

In the configuration file, the `NodeUtil.getAttribute(node, 'mathvariant')` on line 99 never is `null` because it will return the default value if it is not explicitly set, so the conditional is removed.


zorkow (commented):
lgtm.

zorkow (approved):

PR #1197, opened 2025-02-04
Add tests for the cases and empheq packages and update code

dpvc:
This PR adds tests for the `cases` and `empheq` packages, and updates their code.

In `CasesConfiguration.ts`,
* The `subnumber` is non-zero only in `subnumcases`, which is handled in the condition above, so technically, this `if` should always be true and could be removed, but we don't really want increment the tag counter during the processing of the left-hand material, so we leave it with that condition only.
* In order to use the `\empheqlbrace` at line 105, we would need to load the `empheq` extension and add it to the `package` list.  Rather than force the user to do that, we use `\mmlToken` to create a stretchy brace instead.
* In line 200, we check that the `cs` is defined, which it might not be if the `\\` is at the end of the line.

In `EmpheqConfiguration.ts`,
* Since `GetArgument()` always returns a string, there is no need for `|| ''`. (line 62)
* The error message at line 65 is adjusted.
* The property `nestStart` is used instead of `nestable` in order to avoid the erroneous nesting message.  This requires the changes to `ParseUtil.ts` that are in the `test-tex` branch.

In `EmpheqUtil.ts`,
* Since `attributes.get()` will return the default value (always a string), there is no need for the `|| ''` (lines 167, 171, and 225).
* We create an empty row if there are no rows at line 173-175.
* At line 210, we skip over the tag cell of an `mlabeledtr` row, and we add braces around the loop.

### DIFF
--- a/testsuite/tests/input/tex/Cases.test.ts
+++ b/testsuite/tests/input/tex/Cases.test.ts
@@ -1,0 +1,510 @@
+import { afterAll, beforeEach, describe, test } from '@jest/globals';
+import { getTokens, toXmlMatch, setupTex, tex2mml, expectTexError } from '#helpers';
+import '#js/input/tex/cases/CasesConfiguration';
+import '#js/input/tex/empheq/EmpheqConfiguration';
+import '#js/input/tex/ams/AmsConfiguration';
+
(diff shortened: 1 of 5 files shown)


zorkow (approved):
lgtm.

PR #1198, opened 2025-02-04
Add tests for the colortbl package, and update its code

dpvc:
This PR adds tests for the `colortbl` package.  The `ColortblConfiguration.ts` file adds lines 98-100 to skip over an `mstyle` that may have been added to set the `scriptlevel` for the table.  The comments from the configuration object itself are moved to before it so that prettier can be run on the configuration.


zorkow (approved):
lgmt.

PR #1199, opened 2025-02-04
Update tests for mhchem, and update code

dpvc:
This PR updates the tests for the `mhchem` package.  It also includes the `mhchem-stretchies` branch, so the changes to the configuration file are really from there.

Again, best to view with whitespace hidden.


zorkow (approved):
lgtm.
I am a bit confused why there is a merge conflict for `MhchemConfiguration.ts`.

dpvc:
> I am a bit confused why there is a merge conflict for `MhchemConfiguration.ts`.

yes, I don't understand that, either.  It ends up with a duplicate command map, which can be removed.  Not sure whey it didn't recognize it as the same thing.  I'll fix it.

PR #1200, opened 2025-02-04
Update tests for the newcommand package, and update code

dpvc:
This PR updates the test for the `newcommand` extension.  In the `NewcommandUtil.ts`, the case where `j < 0` does not occur, as `j` is reduced only on line 213, and that is only when `hasBraces` is true.  For that to be the case, there would have had to have been an initial open brace, in which case line 206 would have called `GetArgument()`, which would find a matching close brace, so there must be at least two characters (the two braces), and `j` would be at least 2 before it is decremented.  So it never gets below 0.


zorkow (approved):
lgtm.

PR #1202, opened 2025-02-04
Update tests for the physics package and update code

dpvc:
This PR updates the pyhsics tests, and adjusts the `PhysicsMethods.ts` file:
* Lines 286-289 are never reached because on line 245, if `arg` is true and next is not an open brace, an error is thrown.  So to get this far, we know `arg` is false or `next` is an open brace.  Since 28 checks for `arg` being true, that would only occur if `next` is an open brace.  But in that case, the condition for line 271 will be true, and that block ends with a `return`.  So lines 287-288 can never be reached.
* In original line 714, `starKet` can only be true if `hasKet` is true, but the `if` at original line 707 means that `hasKet` must be false for line 714, and so `starKet` is false, and the `|| starKet` is superfluous.
* Lines 1046-1049 are not needed.  This is because `endI` is the string index for the character that follows the argument found by the `GetArgument()` call on original line 1032, and because original line 1028 guarantees that the argument starts with a brace, this means that `endI` is after the closing brace.  Note that `parser.i` changes in the loop only by the `GetUpTo()` on line 1040 succeeds, and that only happens when it finds a `,`.  that would put `parser.i` just after that, so it is at most pointing to the closing brace, and `endI` is *after* that, so `parser.i` is never greater-than-or-equal-to `endI` on line 1046.  So I removed that block.


zorkow (approved):
lgtm.

PR #1204, opened 2025-02-04
Add tests for the setoptions package and fix code

dpvc:
This PR adds tests for the `setoptions` package, and updates its configuration file.

The changes to the configuration include:
* Breaking the error message at 82-86 into two messages, one for the TeX options, and one for package options, mirroring the messages in (original) lines 91-103.
* To do that, the `isTex`variable is moved earlier to line 78.
* The `filterValue()` function is extended to handle RegExp values (like the one for the ams parameter `operatornamePattern`).
* The `setoptions` command map is now created within the `setoptionsConfig()` function.  This means that the map will be recreated for each instance of the `setoptions` package, so if there are two TeX input jax created, one with `[tex]/require` includes and one without it, the first will get the updated `\require` macro that allows options, while the second will have the original `\require`.  The original version would have shared the `setoptions` map between the two, so both would get the updated `\require`.

### DIFF
--- a/testsuite/tests/input/tex/Setoptions.test.ts
+++ b/testsuite/tests/input/tex/Setoptions.test.ts
@@ -1,0 +1,338 @@
+import { beforeEach, describe, test } from '@jest/globals';
+import {
+  toXmlMatch,
+  setupTex,
+  setupTexTypeset,
+  tex2mml,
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #1208, opened 2025-02-05
Add tests for the textmacros package and update code

dpvc:
This PR adds tests for the `textmacros` package, and makes two small adjustments to the code:

* In `TextMacrosMethods.ts`, the `GetNext()` method already skips spaces, so there is no need for the loop here.
* In `TextParser.ts`, the addition of the TeXAtom is moved to after the checking for size and color attributes, otherwise they will be hidden inside the TeXAtom and never seen.


zorkow (approved):
lgtm.

PR #1209, opened 2025-02-05
Update unicode tests and code

dpvc:
This PR updates the format of the tests for the `unicode` and fixes one issue in the configuration file:  since `GetBrackets()` always returns a string, there is no need for the `|| ''`.


zorkow (approved):
lgtm.

PR #1210, opened 2025-02-05
Add tests for the units package and update code

dpvc:
This PR adds tests for the `units` package, and adjusts the configuration file:

* The `GetBrackets()` method always returns a string, so no need for `|| ''`.
* The `GetBrackets()` method has an optional parameter for the default value, so we use that in line 74.


zorkow (approved):
lgtm.

PR #1211, opened 2025-02-05
Update UnitUtil tests and code

dpvc:
This PR updates the tests for `UnitUtil.ts` and adjusts the code:

* In original line 98, since we know (from line 94) that we have `mu` units, that means `value` will not be an empty string, and so the `|| '1'` is not needed.
* In original line 146, we know `factor` is defined, since the `UNIT_CASES.get()` function in the previous line will return the value for 'pt' (line 65) if the unit itself is not found.  So unless someone deletes the 'pt' value, the `factor` should always be defined, and if it turns out to be 0, then `factor * m` will be 0, so there is no need for the test.

### DIFF
--- a/testsuite/tests/input/tex/UnitUtil.test.ts
+++ b/testsuite/tests/input/tex/UnitUtil.test.ts
@@ -1,6 +1,8 @@
 import { describe, it, expect, beforeEach } from '@jest/globals';
 import { UnitUtil } from '#js/input/tex/UnitUtil.js';
 
+/**********************************************************************************/
+
 // These methods will be rewritten into non-ParseUtil ones.
 function convertLength(unit: string, num: number) {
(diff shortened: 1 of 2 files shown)


dpvc:
> one could replace the `num` field in `UNIT_CASES` with a regexp that would allow for an empty value

Is that a reasonable thing to do?  (That is, is this a use-case that we need to support?)  Or would this regex be considered a mistake, and we would *want* something to go wrong so that the user knows it should be fixed?  Because this is being used to determine when a dimension is the first (or only) thing in a string, and dimensions in TeX require numbers, it seems to me that such a regexp would lead to all kinds of trouble.

I'll change it and make a test for it if you want, but I just want to make sure that it is actually a reasonable regex to use for `num`.

zorkow (approved):
as per f2f we don't worry about that case now.

PR #1212, opened 2025-02-05
Add tests for util/context.ts and fix coverage for asyncLoad

dpvc:
This PR adds a `Context-browser.test.ts` file that tests the `ts/util/context.ts` file in a browser by including a fake `window` object, and a `Context-node.test.ts` to test `ts/util/context.ts` when used in node.

Also, two `asyncLoad` files are adjusted so that the coverage is properly computed.

### DIFF
--- a/ts/util/asyncLoad/esm.ts
+++ b/ts/util/asyncLoad/esm.ts
@@ -31,9 +31,7 @@ let root = new URL(import.meta.url).href.replace(
 if (!mathjax.asyncLoad) {
   mathjax.asyncLoad = async (name: string) => {
     const file = name.charAt(0) === '.' ? new URL(name, root).pathname : name;
-    return import(file).then((result) =>
-      result.default === undefined ? result : result.default
-    );
+    return import(file).then((result) => result.default ?? result);
   };
 }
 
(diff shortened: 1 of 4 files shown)


ts/util/asyncLoad/esm.ts
```diff
@@ -31,7 +31,9 @@ let root = new URL(import.meta.url).href.replace(
 if (!mathjax.asyncLoad) {
   mathjax.asyncLoad = async (name: string) => {
     const file = name.charAt(0) === '.' ? new URL(name, root).pathname : name;
-    return import(file).then((result) => result?.default || result);
+    return import(file).then((result) =>
+      result.default === undefined ? result : result.default
```
zorkow, line 35: Can `result.default` be `null`? If so, do we want `null` returned here?
I think this can only happen if the file has an `export default null;` entry. But even then we will probably want to take the content of `result`. 

In case the above reasoning holds, `result.default ?? result` would be more concise.

ts/util/asyncLoad/system.ts
```diff
@@ -33,7 +33,9 @@ if (!mathjax.asyncLoad && typeof System !== 'undefined' && System.import) {
     const file = (
       name.charAt(0) === '.' ? new URL(name, root) : new URL(name, 'file://')
     ).href;
-    return System.import(file).then((result: any) => result?.default || result);
+    return System.import(file).then((result: any) =>
+      result.default === undefined ? result : result.default
+    );
```
zorkow, line 38: ditto

zorkow (changes requested):
Please see my comments.

ts/util/asyncLoad/esm.ts
dpvc: I've made the change, and agree with your analysis.

zorkow (approved):
lgtm.

PR #1213, opened 2025-02-05
Add tests for the mathtools package, and update code

dpvc:
This PR adds tests for the `mathtools` package and updates its code.

It turns out that `mathtools` got an update in 2022, and some things have changed, so I've updated the code to reflect those changes.  In particular, there are some breaking changes to `\coloneq` and three other colon macros.  A new `legacycolonsymbols` option controls which set to use (jsut as in actual `mathtools`).  I also added support for `\MakeAboxedCommand` which was missing from the current package.  This includes a non-standard starred version that handles box commands whose contents are in math-mode rather than text-mode, like `\bbox` and `\boxed` versus `\fbox` and `\fcolorbox`.  Finally, several new extensible arrows were added in 2022 and 2024, and I've added those as well.

To handle `legacycolonsymbols`, we use a new command map (lines 179-185 of `MathtoolsMappings.ts`) and dynamically add it when the option is selected.  This is done in the `mathtoolsConfig()` method at lines 62 to 64 of `MathtoolsConfiguration.ts` if the option is set in the MathJax configuration, and lines 1011-1017 of `MathtoolsMethods.ts` if set with `\mathtoolsset{}`.  The data needed for that is stored in `LEGACYCONFIG` and `LEGACYPRIORITY` in `MathtoolsMethods.ts` and imported into `MathtoolsConfiguration.ts`.

In `dependencies.js`, the `boldsymbol` extension is added for `mathtools`, since the `\bigtimes` macro uses that.

In `MathtoolsConfiguration.ts`:
* Lines 77-84 are removed because the empty prescripts are already converted `none` nodes by `MathtoolsUtil.getScript`, so this check is not needed.
* The options at line 108-131 have had their indentation corrected, and the `legacycolonsymbols` option added.  Also, `multlinegap` is changed to `multlined-gap` to be more consistent with the other options (this was a non-standard option in any case), and its use in `MathtoolsItems.ts` is updated.

In `MathtoolsMappings.ts`:
* The new extensible arrows are added (lines 64-67)
* The `\MakeAboxedCommand` is added (line 102)
* The name for `\vcentercolon` is corrected (line 121)
* The 2022 corrections are made to the `\centercolon` and other macros (lines 127-132) and the new ones added (138-145)
* The legacy map is added (lines 179-185)

In `MathtoolsMethods.ts`:
* We export the legacy configuration and map priority (lines 50-53)
* We use the default multlined options properly (lines 123-124, 130, and 157)
* We use `nestStart` rather than `nestable` in order to deal with the erroneous nesting of environments error properly (you will need to include the `test-tex` branch to run the tests)
* We add parameters to the `Aboxed()` method to allow `\MakeAboxedCommand` to provide the box command and whether it is math-mode or not (line 415-420, 448, 452-453)
* Add the `MakeAboxedCommand` method to implement `\MakeAboxedCommand` (lines 459-479)
* Remove the `isFlush` testing, since this doesn't work well with the new fonts (lines 475-477, 493-497)
* Handle the `legacycolonsymbols` option in `\mathtoolsset` (lines 1011-1017)

In `MathtoolsUtil.ts`:
* Since `attributes.get()` will return the default value if it is not explicitly set, `rowspacing` is always defined, so there is no need for the test at line 116 so it and its `else` clause are removed and the block is re-indented.
* The creation of `none` blocks is now performed in `getScript()` at lines 156-163.

### DIFF
--- a/components/mjs/dependencies.js
+++ b/components/mjs/dependencies.js
@@ -44,7 +44,7 @@ export const dependencies = {
   '[tex]/extpfeil': ['input/tex-base', '[tex]/newcommand', '[tex]/ams'],
   '[tex]/gensymb': ['input/tex-base'],
   '[tex]/html': ['input/tex-base'],
-  '[tex]/mathtools': ['input/tex-base', '[tex]/newcommand', '[tex]/ams'],
+  '[tex]/mathtools': ['input/tex-base', '[tex]/newcommand', '[tex]/ams', '[tex]/boldsymbol'],
   '[tex]/mhchem': ['input/tex-base', '[tex]/ams'],
   '[tex]/newcommand': ['input/tex-base'],
   '[tex]/noerrors': ['input/tex-base'],
--- a/testsuite/tests/input/tex/Mathtools.test.ts
+++ b/testsuite/tests/input/tex/Mathtools.test.ts
@@ -1,0 +1,6440 @@
+import { afterAll, beforeEach, describe, test, expect } from '@jest/globals';
+import { getTokens, toXmlMatch, setupTex, tex2mml, expectTexError, trapErrors } from '#helpers';
+import '#js/input/tex/mathtools/MathtoolsConfiguration';
+import '#js/input/tex/ams/AmsConfiguration';
+import '#js/input/tex/boldsymbol/BoldsymbolConfiguration';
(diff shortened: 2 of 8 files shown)


zorkow (approved):
lgtm.

PR #1215, opened 2025-02-15
Add ability in config.json to specify mutiple copy configurations

dpvc:
This PR provides a method for specifying multiple copy operations in a component's `config.json` file.  The `copy` property can now be an array of copy configurations (or still a single one).  This will allow the sre component to copy both its map files and the webwork files using a single configuration file.


zorkow (approved):
Works fine. I tried it out in the PR, which is now obsolete due to PR #1219

PR #1217, opened 2025-02-17
initial refactor of enriched document into speech document

zorkow:


zorkow:
Subsumed by #1219, similar to #1214 .

PR #1218, opened 2025-02-18
Some fixes for the separation of speech from enrichment

dpvc:
Here are some changes that I would make to your separation of enhancement from speech.

* I've removed the `SpeechMenu.js` file from the speech component, as that would cause the menu code to be included in the speech component, and you could get two conflicting copies of it if both the speech and menu components are loaded (though a combined component should be OK).

* I removed the check for `hasWindow` from the speech component, as that would prevent it from working in node applications.  (Of course, you will need my node updates in order for that to work.)

* I added the semantic-enrich dependency for the speech component, and made the explorer dependent on the speech extension.  These might have been the source of the problem you were having, but I'm not sure.

* I removed the generator pool from the semantic-enrich code, since it is not used there, and I think it would have caused unneeded duplicate creations of the pools.

* I moved the `a11y` options to the speech code, as that is where they are used.

* I removed the duplicate `sre` options from the speech code, as they are already in the `EnhanceDocument` on which the `SpeechDocument` is based, so don't need to be set again.  These are used in the enhancement process, so need to remain in the enhanced document, but if there are any that aren't used there, they could be moved to the speech document.  But since you can specify deep speech during enhancement, I think they are all needed there, right?

* I removed the visitor and `toMathML` code from the speech file, and use the `toMathML` that is already in the enhanced MathItem.

* Finally, I did remove the asynchronous speech code.

After these changes, it all seems to work for me, both in the lab, and in stand-alone pages with lots of math using laxy loading.  See if that works better for you.

### DIFF
--- a/components/mjs/a11y/speech/config.json
+++ b/components/mjs/a11y/speech/config.json
@@ -1,7 +1,8 @@
 {
   "build": {
     "component": "a11y/speech",
-    "targets": ["a11y/speech.ts", "a11y/speech"]
+    "targets": ["a11y/speech.ts", "a11y/speech"],
+    "exclude": ["a11y/speech/SpeechMenu.ts"]
   },
   "webpack": {
     "name": "a11y/speech",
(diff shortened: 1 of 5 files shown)


dpvc:
Closing this in favor off #1219.

PR #1219, opened 2025-02-24
Updates for running workers in node

dpvc:
# Overview

This PR adds the ability to use the new web-worker-based speech code in node applications.  It looks like a lot of files, but most are either changes to component configurations, some new component files, some deleted files, or changes to the the adaptors to support the worker code.  Only 9 files contain changes to the worker code itself.  So it is not as bad as it looks.

Because this builds on both the `feature/webworkers` and `refactor/speech_document` branches, I've made a new, temporary branch `feature/speech-document` that merges these two so that the differences here are against the combined branches.  That way, this shows the differences from your code, otherwise the differences would have included your changes for one or the other branch.

I've included the change from my earlier `refactror/speech-update` branch here, so will close that PR.

There are a few significant changes here, including:

* New components for the workerpool and worker code so that they get webpacked to reduce their size.  This also allows webpacking the SRE code into the worker to eliminate one additional file transfer as well.

* In order to make the speech computations to work in convert operations (like `MathJax.tex2chtml()` and `MathJax.tex2svg()`), we can't use the `data-worker` id to locate the DOM node where the speech is to be attached since the MathItems for the conversions are not in the document DOM itself.  So I've changed the task list to hold the MathItem itself rather than its id.

* In the lab, the worker code uses the compiled versions from the `mjs/a11y/sre` directory rather than the bundled versions, so it is easier to debug (you don't have to remake the components).  There is a new `lab` directory that holds a special compiled version of `sre.js` that is used in the worker in the lab, but that isn't included in the bundles or compiled directories (so the npm package doesn't include duplicate code).

* Better handling of setting the worker path so that it works in node and browsers.

* The asynchronous speech code and options are removed.

* The duplicate options in `semantic-enrich.ts` and `speech.ts` are removed from one or the other of the two files.

* The handling of the `SREfeature` and importing of `sre.js` in the worker is now done via straight import calls, and code is added at the top of the worker to accommodate node (it is a bit complicated because of the need to handle both node worker-threads and web workers).  Code is added at the top of the workerpool to accommodate node as well.

* Because the worker has to wait until SRE is set up, the worker now signals the client when it is ready to run (rather than the pool doing so, which may happen too soon).  In order to accommodate compiling into CJS files, where `await` can only use used in a function (not at the top level of the module as in ESM), the main worker code setup is placed in an `async` function to allow `await` and avoid complicated promise chains that would have been needed for the node-based code.  (Unfortunately, this has caused the diff for `speech-worker.ts` to be badly displayed, so it is not so easy to see what the changes are.)

* The handling of the DOM in `WebWorker.ts` has been abstracted and moved to the DOMadaptor.

-----

# In more detail

## The config files

Since the explorer now depends on the speech component, its library is added to the `libs` property in its configuration.

The `semantic-enrich` configuration now uses the `[sre]` path (which now maps to `[mathjax]/sre` rather than `[mathjax]/sre/mathmaps`) to look up the location of the worker path.  This needs to use `Packate.resolvePath()`, whcih is part of the loader files, so the `loader/lib` is added to its `lib` in its configuration.  

The `speech` component's build configuration now excludes the `SpeechMenu.ts` file, as that was bringing in all the menu code, which doesn't belong there.  The speech component no longer checks for `hasWindow`, since it is supposed to work in node as well.

The `sre` configuration doesn't include the `a11y/sre` directory, since the worker code now has its own components, and the copy section is removed, since that is done in a new speech-worker component.  The `sre_congif.js` file adds the `mathmaps` directory to `[sre]` that is removed in the `source.js` file.

New configuration files for the `speech-workerpool` and `speech-worker` have been added.  These include the copy configurations to move the needed HTML and other files to the bundle directory.  These have `webpack.cjs` files to customize the webpack configurations for these components.  In particular, they are built as ESM modules so that their exports are able to be accessed (the node code needs to access the workerpool exports).  The worker component also specifies that a single chunk should be produced, as the inclusion of SRE was causing it to be split into several chunks.  

The math maps are removed from the `a11y/util.js` component file.

The `node-main` file is changed to move the `asyncLoad` so that it is used in both the bundled and source versions.

The dependencies file is modified to have the explorer be dependent on the speech component, and an entry for speech is added (dependent on semantic-enrich).  The `sre` path is modified to remove the `mathmaps` directory, so that it can be used as the path to the worker code.

The speech component is added to the `source.js` file so that the lab file will use the source rather than the webpacked version (and node applications that use the source versions will as well).

The `sre.js` file is removed from the copy configuration in the `sre` component, since it is now compiled into the worker component.

A `lab` directory is added to house a special version of SRE for use in the lab, together with the component build files for it.  This is only used by the lab, and is needed because the lab now uses the worker files from `mjs/a11y/sre` rather than the the webpacked bundle files.  the alternative would be to put the `sre.js` into `mjs/a11y/sre`, but it seems wasteful to put such a large file into a directory that is part of the npm package (already quite large) when it is only needed for the lab.

The `package.json` file is modified to add some commands to make the CJS compilation work (it was broken before this),  as there are some cjs files that need to be moved before the typescript compilation can be done.  Also, the `speech-worker.ts` needs to be compiled without comment-removal to allow some webpack control comments to be retained, so a special `tsc` call is made to compile that file, using a new custom `tsconfig` files, one for mjs and one for cjs.  Finally, a command is added to compile the `sre.js` file needed for the lab.

## The Speech files

In the key explorer, because the worker tasks now use the `MathItem` directly, we pass that rather than its `typesetRoot` and `outputData.mml` values.

In `semantic-enrich.ts` the generator pool is removed (as it is now in `speech.ts` and not used here), the `a11y` options are removed (as they are not used here and are in `speech.ts`), and the `attach-speech` process-bit is removed (as it is created in `speech.ts`).

In `speech.ts`, we use the `toMathML` function that is part of the `EnrichMathItem` rather than having to get a new one here.  We use the `enableSpeech` and `enableBraille` options rather than `enableEnrichment` to decide if we are attaching speech.  The `webworker` property is moved to be above the functions in the interface for the `SpeechMathDocument`, and a new `getWebworker()` method is added, so that the `SpeechMathItem` can call it (this is needed for convert actions, which use a `MathItem` that may be processed individually without the document's render actions being called first).  The `Import()` action is not needed, since the imoprting is already done in the worker directly (see below).

The worker `path` is obtained from `sreRoot()`, which comes from a new mjs/cjs-specific file similar to the `mjxRoot()` command that gets the root directory for MathJax itself.  Since the `sre` directory is added there already, there is no need for `basedir` here.   Because `sre.js` is compiled into the speech-worker component, there is no need for the `sre` configuration property.

The timing options are removed, and the sre options are duplicates of the ones from `semantic-enrich`, so are removed as well.

The asynchronous speech code is removed, and the synchronous code is moved into the `attachSpeech()` method.

In the generator pool, the `node` and `mml` arguments have been replaced by the single `MathItem` argument, since we need that for handing single-equation conversions.

In `WebWorker.ts`, the task now stores the `MathItem` rather than the node id.  That means the counter is no longer needed, as the id is never used.  The `iframe` is made an `N` rather than `HTMLIFrameElement` so it can be handled by the adaptor.  The `pool` becomes a `D` (document) rather than a window, and the adpator will use the document's `defaultView` when needed to get at the window.  The `url` is not needed, as the `path` now handles that directly.

To create the iframe, we now get the `src` value here (rather than in `computeDomain()`), and create the iframe with the `src` property already in place using the DOM adaptor.  If we are in node (i.e., no window), we extend the iframe to have an `options` property with the worker options (this is used by the `liteAdaptor` to be able to fake the loading of the html file and run the code that it would have performed).  It's a bit of a hack, but we can work out something better later, if need be.

For `computeSrc()` we encode all the pieces, as they include arbitrary data from the options, so may need escaping.

We no longer need `computeDomain()` as that has been moved to the adaptor, which is used directly in the`Start()` method below.

The `Start()` method now uses the adaptor to add the event listener, append the iframe to the document, and get the domain.  It is important that the domain is looked up after the iframe is appended, as the adaptor needs the iframe's document to be loading before it can get the correct domain (the adaptor uses the document's `location`, which isn't set until the iframe is in the page).

The `Post()` method now takes the `MathItem` rather than a worker id, and the `postMessage()` call is moved into the adaptor in the `postNext()` call.

The `Import()` method is no longer needed, as the feature-setting and importing is handled directly in the worker.  The `Speech()`, `nextRules()`, and `nextStyle()` methods now use the `MathItem` rather than a worker id.  The `Terminate()` and `Ready()` functions now use adaptor functions (and remember that `pool` is a document).

The `Attach()` method looks up the `MathItem` from the current task rather than trying to find it in the document (where it may not be).  We use the adaptor to locate the internal `maction` elements by id.  We also have to deal with a complication that the data structure passed by the worker isn't transferable when created in a node worker-thread, so it is JSON stringified in that case, and we need to parser it here.

In `sre.ts`, we no longer have preloaded maps, and since we don't need maps at all in the main thread, we just return an empty JSON string for now.  This will all change when SRE is separated into APIs that don't need the maps.

The `require` files are here because the speech-worker needs them when used in node (we use them to load the SRE map files).  The `package.json` is there so that it can be copied into the `bundle/sre` directory (the main `bundle` directory has a `package.json` that specified commonjs modules, but since the speech worker and pool are bundled as modules, we need a `package.json` file here to indicate that).

In the `speech-worker.ts` file, the diff is not very helpful, as it is not recognizing the matches properly, so it is a bit hard to read.  The section between the star-comments is the code that is used to set things up properly for either node use or web use.  Because the node setup code uses `await` in several places, and because that can't be used at the top level in cjs files, everything is placed within an `async` function that is called immediately, which avoids that issue, but makes the diff hard to read.

Because this function uses `import()` to dynamically load some code needed by node, and we don't want these webpacked into the final worker component, we need to mark these imports using webpack magic comments.  But our tsconfig removes comments, so we need to use a separate tsconfig to compile the speech-worker file that doesn't do comment removal.  But the `import` commands for `Options` and `MessageTypes` woudl cause those files to be compiled with those options as well, which we don't want to do, so we make explicit `type` calls to prevent that.

Next comes the code that sets up what is needed special for worker threads versus webworkers.

In the browser web-worker, `self` is the global namespace, but in the node worker-threads, `self` is not defined, so we use that to distinguish between the two cases.  When undefined, we define it using `global`, which is available in the worker thread, but use global throughout (since `self` is typed as `Window & typeof globalThis`).  We add a `DedicatedWorkerGlobalScope` so that SRE will think we are in a webworker.  Then we define `copyStructure()` to JSON stringify the structure (in worker-threads the structure ends up not being transferable, so we use a string instead, and parse the string in the `Attach` method of the WorkerHandler if it come in as a string).

Then we load the `worker_thread` code and set up the event-listener and posting code in the form that it would have in a webworker.

Then we load the require command (if needed), and use that to load the SRE map files from the proper location (from SRE itself, in this case, since we will be running this code from the `mjs/a11y/sre` directory in node or the lab, and that doesn't have the `mathmaps` directory).

Otherwise, we are not in node and are in a webworker, so we set up `global` to be equal to `self`, make the `copytStructure()` function just return the actual structure (it can be passed in webworkers unlike worker-threads), and set the map directory for SRE.

The rest of the code is common to both node and webworkers.

We import SRE, and if that fails, as it will in the lab, since the `sre.js` file in the `mjs/a11y/sre` directory will try to import the SRE source code, and not be able to locate `@xmldom/xmldom` (and you can't have an importmap in a worker file), we try to load the lab version, we then try to load the lab version of SRE.  This is only needed in the lab, as in actual node applications, the initial import will succeed (provided the dependencies have been installed), and in actual browser use, the webpacked components are used, so sre will be included directly.  It seems unecessary to keep an extra copy of SRE in `mjs/a11y/sre` just for the lab, as that would also become part of the mathjax-full npm package for no reason.

The command functions are basically untouched, except for the `import()` function, which now uses actual `import()` as it is available to both worker-threads and webworkers.  The `feature()` and `setup()` functions are not currently used, and since SRE is imported at the top, I don't think `feature` will work, in any case.   So I think they can be removed, but I didn't do that just in case you have a use for them now.  The `speech()`, `nextrule()`, and `nextstyles()` functions no longer use the `workerId`.

The `Finished()` function now calls `Client()` rather than `Pool('Client', ...)` since you have that shorthand.

The `Speech()` function doesn't add the `id` to the structure, since that is no longer used.

The `imported` array and the `Import()` function are removed, since we are using direct `import()` commands now.

Finally, at the end, we wait for `SRE.engineReady()` and then singal the WorkerPool that we are ready.  (Without this, it is possible for the WorkerPool to send us commands before SRE is ready, and even before the event listener has been installed, so this is an important hand-shaking action).

The `speech-workerpool.ts` file has a new section at the top for setting things up for both node and webworker use.  We set up a `context` variable that holds the values that differ between the two settings.  The node versions are left unset initially, and there is a function `setContext()` that is exported that can be used to set the values needed by node.  When used in node, the LiteDOM doesn't actually load the HTML file and parse the script file it contains.  Instead, it loads the iframe code directly into the main thread and uses `setContext()` to set it up to give it the fake window, parent, location hash, and the replacement for the Worker class.  We then use the `context` versions of these four values through the rest of the file.

The `Create()` function gets the locate hash from the `context` and decodes all three parts, since they were all encoded originally.

The `Start()` command no longer sends the ready signal to the client, since we need to wait for the worker to set up SRE before it is ready to process commands.  A new `Ready()` worker command is added to respond to the ready signal sent by the worker when it is ready, which is when the ready signal is sent to the client.

The `sre-lab.js` file is used to access the lab version of SRE.  The `sre.ts` file loads SRE without the CLI (this eventually gets packed into the speeck-worker in the bundle directory).  A definition file is given for `worker_threads` so that typescript will be happy.

## The DOM Adaptors

The remaining changes are for the DOM adaptors to handle the new commands needed for the workers, and changes to teh tsconfig files.

For the HTML adaptor, we add the needed properties to the minimal interfaces, and set some defaults for a few methods to make it easier to get the head, body, and so on.  We add the new `domain()`, `listener()`, and `post()` functions, and a new `getElement()` that is neededd to handle `querySelector()` calls.

For the LiteDOM, the LiteDocument is extended to hold a pointer to the associated LiteWindow (so we can access that from the LiteIFrame), and we implement a list of message listeners (the only event listeners that we support).  We add the `addEvenetListener()` and `postMessage()` methods needed for the worker message passing.

In `LiteElement.ts`, we make a `LiteIFrame` subclass of `LiteElement` that has the needed infrastructure for handling the worker code.  It is not really a full IFrame implementation, but just enough to handle the worker sitution.  In particular, we have it create its own `LiteWindow`, and add a `loadWorker()` function that is used to similate the webworker.  This is called when the iframe is appended to the document body (see below).

It works by first loading the `Worker` class from `worker_threads` node module, and then creating a `LiteWorker` class that has the needed interface for a webworker as given at the top of the file.  (I wanted to make `LiteWorker` a subclass of `Worker`, but whioe that was find for the mjs files, typescript didn't produce the correct code for the cjs ones (since it didn't know that `Worker` was a class rather than a function, and I couldn't find a way to tell it that).  Then it sets up the hash value needed by the WorkerPool.  Then it loads the WorkerPool code an sets its context to use the `LiteWorker`, the `LiteIFrame` window, the client document, and the constructed hash value.  Finally, it starts the worker pool, just as the HTML would have.  (I supose it would have been possible to use `fs` to load the HTML file and parse it, then lok for the script it contains and then run that, but that seemed unnecessary, as the script there is very simple and we can just implement it here.)

Note that this all relies on knowing the worker options like the worker path and the debug value.  That is why the `LiteIFrame` has an `options` property, which is initialized by the `WorkerHandler` when a node adaptor is used.

In `LiteWindow.ts`, we add the needed event-handling methods, which push these down to the document, where the actual processing is done.  (The processing is in done in the document because not every document in the LiteDOM has an associated window .  In particular, the main document from `mathjax.document()` doesn't, only the iframes do.)

In the lite adaptor itself, when a LiteElement is created, we check if it should be a LiteIFrame, and create that instead.  We add the `domain()` method and just return `file://` regardless (we are not working on the web in a node aplication),   The `listener()` and `post()` methods call the document-level commands.

The `tag()` method gets a new parameter `stop` to indicate how many tags to locate (so that `getElement()` will stop after finding the first one).  This is added to the other element-finding methods as well, and a new `elementsByAttribute()` function is added to handle `[attr="value']` selectors.  (This was initially to handle the request to locate the `data-worker` id, but that is no longer being done, so this function is not actually used, but I have left it in so that the LiteDOM can process that in `getElements()` and `getElement()`.)

Finally, the `append()` method checks if the appended child is a `LiteIFrame` with an id attribute of  `WorkerHandler-n`, and if so, it calls its `loadWorker()` method to start the worker code.

The `sreRoot.ts` files implement the `sreRoot()` function for mjs and cjs files.

The core DOM adaptor file has the definitions for the new methods implemented above, plus new `setProperty()` and `getProperty()` values for accessing ode properties directly (llke the IFrame `src`).

A new `node-import` asyncLoad file is added that can be used from a CommonJS node application to use `import()` rather than `require()` to load modules.  Since the speech-workerpool file is an ES module, you can't use `require()` to load it, so this asyncLoad method is necessary in that case.

The tsconfig files are updated and new ones added for the speech-worker (whicih need to have the comments preserved, and to be compiled as ESM).

Finally, the old top-level `worker` directory is removed.


dpvc:
Here are some test files for running in node:

**speech.mjs**
``` js
#! /usr/bin/env node

//
//  Load the packages needed for MathJax
//
import 'mathjax-full/components/require.mjs';
import {TeX} from 'mathjax-full/js/input/tex.js';
import {CHTML} from 'mathjax-full/js/output/chtml.js';
import {mathjax} from 'mathjax-full/js/mathjax.js';
import {SpeechHandler} from 'mathjax-full/js/a11y/speech.js';
import {RegisterHTMLHandler} from 'mathjax-full/js/handlers/html.js';
import {liteAdaptor} from 'mathjax-full/js/adaptors/liteAdaptor.js';
import {STATE} from 'mathjax-full/js/core/MathItem.js';
import {MathML} from 'mathjax-full/js/input/mathml.js';
import 'mathjax-full/js/util/asyncLoad/esm.js';
//
//  The packages to include
//
const packages = ['base'];

//
//  Create the input and output jax
//
const tex = new TeX({packages});
const chtml = new CHTML();

//
//  Register the HTML handler with liteDOM adaptor

SpeechHandler(RegisterHTMLHandler(liteAdaptor()), new MathML());

//
//  Create an HTML document with the input and output jax
//
const html = mathjax.document('', {InputJax: tex, OutputJax: chtml});

//
//  Convert the math from the command line to serialized HTML
//
const node = await mathjax.handleRetriesFor(() => html.convert(process.argv[2] || '', {display: false}));
await tex.parseOptions.mathItem.generatorPool.promise;
console.log(html.adaptor.outerHTML(node))

html.webworker.Terminate();
```

and 

**speech.cjs**
``` js
#! /usr/bin/env node

//
//  Load the packages needed for MathJax
//
const {TeX} = require('mathjax-full/js/input/tex.js');
const {CHTML} = require('mathjax-full/js/output/chtml.js');
const {mathjax} = require('mathjax-full/js/mathjax.js');
const {SpeechHandler} = require('mathjax-full/js/a11y/speech.js');
const {RegisterHTMLHandler} = require('mathjax-full/js/handlers/html.js');
const {liteAdaptor} = require('mathjax-full/js/adaptors/liteAdaptor.js');
const {STATE} = require('mathjax-full/js/core/MathItem.js');
const {MathML} = require('mathjax-full/js/input/mathml.js');
require('mathjax-full/js/util/asyncLoad/node-import.cjs');
//
//  The packages to include
//
const packages = ['base'];

//
//  Create the input and output jax
//
const tex = new TeX({packages});
const chtml = new CHTML();

//
//  Register the HTML handler with liteDOM adaptor

SpeechHandler(RegisterHTMLHandler(liteAdaptor()), new MathML());

//
//  Create an HTML document with the input and output jax
//
const html = mathjax.document('', {InputJax: tex, OutputJax: chtml});

//
//  Convert the math from the command line to serialized HTML
//
mathjax.handleRetriesFor(() => html.convert(process.argv[2] || '', {display: false}))
  .then((node) => {
    tex.parseOptions.mathItem.generatorPool.promise
      .then(() => console.log(html.adaptor.outerHTML(node)))
      .then(() => html.webworker.Terminate());
  });
```

dpvc:
I will be making another PR to handle some additional tasks (like canceling pending tasks when clearing math items from a container, etc.)

dpvc:
PS, to run this in the lab, you need to do

``` shell
pnpm -s lab:sre
```

that only needs to be done once, and then you won't have to do it again.

ts/a11y/speech.ts
```diff
@@ -156,7 +163,7 @@ export interface SpeechMathDocument<N, T, D>
   attachSpeech(): SpeechMathDocument<N, T, D>;
 
   /**
-   * Attach speech to the MathItems in the MathDocument
+   * Dettach speech from the MathItems in the MathDocument
```
zorkow, line 166:
```suggestion
   * Detach speech from the MathItems in the MathDocument
```

ts/a11y/sre/speech-worker.ts
```diff
@@ -27,224 +27,266 @@
...
+    ) => {
+      parentPort.on(kind, listener);
+    };
+    global.postMessage = (msg: any) => {
+      parentPort.postMessage({ data: msg });
+    };
+
+    //
+    // SRE needs require(), and we use it to load mathmaps
+    //
```
zorkow, line 72: For future reference: This should go away in the next version of SRE.

ts/a11y/sre/speech-worker.ts
```diff
@@ -27,224 +27,266 @@
...
+    global.SREfeature = {
+      custom: (locale: string) => {
+        const file = 'speech-rule-engine/lib/mathmaps/' + locale + '.json';
+        return Promise.resolve(JSON.stringify(global.require(file)));
+      },
+    };
+  } else {
+    global = (self as any).global = self; // for web workers make global be the self object
+    global.copyStructure = (structure: any) => structure;
+    global.SREfeaure = { json: './mathmaps/' };
```
zorkow, line 89:
```suggestion
    global.SREfeature = { json: './mathmaps/' };
```

ts/a11y/sre/speech-worker.ts
```diff
@@ -27,224 +27,266 @@
...
-    };
-  },
+  const Commands: {
+    [id: string]: (data: Message) => void | Promise<void>;
+  } = {
+    /**
+     * This loads one or more libraries.
+     *
+     * @param {Message} data The data object
+     * @returns {Promise} A promise the completes when the imports are done
```
zorkow, line 147:
```suggestion
     * @returns {Promise} A promise that completes when the imports are done
```

zorkow (approved):
I've made my way through the code. It's really great. In particular the node integration. 
Using the `MathItem` makes life a lot easier as well!

There are a couple of minor things, mainly typos and one *major* typo in a misspelled `SREfeature`.
Do you want to fix those and then simply merge the whole thing into develop?

components/mjs/a11y/util.js
```diff
@@ -3,22 +3,13 @@ import '../input/mml/init.js';
...
   'a11y/sre',
   'a11y/semantic-enrich',
   'a11y/explorer'
 );
 
-MathMaps.set('base', base);
-MathMaps.set('en', en);
-MathMaps.set('nemeth', nemeth);
-
 export function checkSre(startup) {
```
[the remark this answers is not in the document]
dpvc: I will make a separate PR for this, as it touches all the combined configuration files.

zorkow:
> > Do you want to fix those and then simply merge the whole thing into develop?
> 
> Yes, I think that's the way to go. If you are OK with the `sreReady()` change going into a separate PR, then I think this is ready to merge. If you agree, I will target develop and merge there, then remove the other branches that have been merged into this one.

Yes, I am fine with that.
There's also an issue that now crops up in the lab, which might go away, once redundant code is culled.

dpvc:
> There's also an issue that now crops up in the lab, which might go away, once redundant code is culled.

I've merged it, but I'm not having any problems.  Can you say what is going wrong for you?

PR #1220, opened 2025-02-27
Update tests for the html pacakge and fix code

dpvc:
Apparently I pushed this branch but forgot to make the PR for it.

This PR updates the tests for the HTML package and adjust the code.

* The changes to `oldStyle` are because the `;` was being added in the wrong place.
* Lines 154-157 are removed because they can't be reached in our current code.  The reason is that `arg` comes from `ParseArg()`, which returns the `mml()` value of the parsed argument.  That in turn returns the top stack item's first `mml` node.  But `mml` nodes are created by the StackItem's `toMml()`, and that already removes top-level rows if they contain only one child node, expect if the `forceRow` argument is true.  The only call where that is the case is in `BracketItems.ts` line 62, where the resulting inferred mrow is unpacked immediately by the spread operator on its children.  So there is no case where the `if` condition at 154 will be met in the current code.  One could create a fake `StackItem` that would call `toMml(true, true)` to generate a condition where this occurred, but in that case the only consequence is that it will produce an `mrow` with a single entry, which is what the seconf `true` is requesting, so that is probably the correct result in any case.

### DIFF
--- a/testsuite/tests/input/tex/Html.test.ts
+++ b/testsuite/tests/input/tex/Html.test.ts
@@ -1,118 +1,229 @@
 import { afterAll, beforeEach, describe, it } from '@jest/globals';
-import { getTokens, toXmlMatch, setupTex, tex2mml } from '#helpers';
+import { getTokens, toXmlMatch, setupTex, tex2mml, expectTexError } from '#helpers';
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #1221, opened 2025-02-28
Fix position adjustment for scaled svg items with ids

dpvc:
This PR fixes an issue where superscripts (or other scaled items) that have IDs are not positioned properly in SVG output.  For example, `10^4` has the 4 too high.  This is due to the correction being done to try to allow elements with IDs to be targets of a link and have the browser scroll to them properly.  Unfortunately, the position adjustement wasn't being properly scaled.


zorkow (approved):
lgtm.

PR #1223, opened 2025-03-01
Error for colors that include semicolons

dpvc:
This PR adds a check that the color value doesn't contain any semi-colons.  (Technically, we already took care of this with the change to how CSS styles are sanitized, but this at least gives people a message if they use a color that will be discarded for that reason.)


zorkow (approved):
lgtm.

PR #1225, opened 2025-03-01
Fix incorrect character for \char`\x (was incorrectly set to \char'\x)

dpvc:
This PR fixes an error in the `\char` command where it was using `\char'\x` where it should have been using ``\char`\x``.  It also updates the tests accordingly.


zorkow (approved):
lgtm.

PR #1226, opened 2025-03-04
Remove checkSre() calls in combined configurations

dpvc:
This is the PR that I promised that removes the `cherckSre()` function from the `a11y/util.js` file.  It removes its usage in all the combined components.

In addition, it remove the SpeechUtil, GeneratorPool, and WebWorker targets from the semantic-enrich build configuration, since they are not used there (they are loaded by the speech component), and removes the `sreReady()` call from there.  Instead, that is being made in the `sre` commponent itself (just to make sure that SRE is ready to go when the components are finished loading).


zorkow (approved):
lgtm.

PR #1227, opened 2025-03-04
Fix miscellaneous issues with components

dpvc:
This PR makes miscellaneous updates to various component configurations and related changes.

* A `ui/menu` dependency on `a11y/sre` is added (since `SpeechMenu.ts` uses it).
* The `output/chtml` and `output/svg` config files now target `common.ts`, so that that will be included in the `MathJax._` tree, and the `components/src/startup/lib` is added so that `ts/components/package.ts` will be shared properly.  (Otherwise this leads to two separate copies of the `Package` class that don't know about what the other one has loaded.)
* The `webpack.common.js` and `package.json` are modified to allow `compoments/js` to obtain `compoments/mjs` or `components/cjs` depending on whether it is used in an `import` or a `require()`.  (The older `components/src` is still there for backward compatibility with v3, before we had the mjs/cjs split).
* The `global.ts` file is modified to check that the `MathJax` variable is actually an object literal, rather than a string or array or some other thing.
* The `all-packages` component is remove from the `\require` list, since that component was removed in a previous PR.

### DIFF
--- a/components/mjs/dependencies.js
+++ b/components/mjs/dependencies.js
@@ -60,7 +60,8 @@ export const dependencies = {
   '[tex]/unicode': ['input/tex-base'],
   '[tex]/units': ['input/tex-base'],
   '[tex]/upgreek': ['input/tex-base'],
-  '[tex]/verb': ['input/tex-base']
+  '[tex]/verb': ['input/tex-base'],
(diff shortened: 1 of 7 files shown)


zorkow (approved):
lgtm.

PR #1205, opened 2025-02-05
Update begingroup tests

dpvc:
Because this PR requires both the `begingroup` branch and the `update-test-framework` branch, I've merged `begingroup` into it and targeted it to `update-test-framework`.  Unfortunately, that means that all the changed from `begingroup` are listed her as well.  You can ignore all the files other than the `Begingroup.test.ts` file, as that is the only one actually changed here, and that is just some formatting and changes to use `expectTexError()` from the new `setupTex.ts` file rather than the home-grown version here (which is actually what inspired the one in `setupTex.ts`).

PR #1228, opened 2025-03-04
Add texClass REL when combining mo nodes so they round-trip through SRE properly, and adjust tests to match

dpvc:
This PR makes sure that when adjacent `mo` elements with TeX class `REL` are combined into a single `mo`, they are given the `data-mjx-texclass="REL"` attribute, so that when they pass through SRE, they will remain TeX class REL.  Because the multiple-character `mo` will not be in the operator dictionary, without this attribute, they will come back as BIN class operators, so get the wrong spacing.

Most of the changes here are to adjust the tests to include the new attributes, so you can ignore them.  Only the last file is the code change, and that is just one line.


dpvc:
There are some tests in #1205 that will also need to be updated when those tests are merged.

PR #1229, opened 2025-03-04
Miscellaneous small updates

dpvc:
This PR makes a few miscellaneous changes that I found while updating the documentation.

* In `MathDocument.ts`, you only need to set the `mtextFamily` and `merrorFamily` if we are inheriting the them.
* In `HTMLDomStrings.ts`, the `skipHtmlTags` array is simplified (`math` is added and annotation nodes are removed, since they shoudl be inside `math` nodes).
* In `MathMLCompile.ts`,  any `data-mjx-` attribute that isn't otherwise handled is added as an attribute (e.g., `data-mjx-hdw`, `data-mjx-error`, etc., for example).
* In `FindTeX.ts`, escaped MathItems now get a `\` as their opening delimiter, so that if the Math is inserted back into the page (e.g., when setting `state()` back to 0), the escaped character will be replaced properly.
* In `common.ts`, the `display` value is now included in the metrics from `getMetricsFor()`.
* `Menu.ts`, the `data-mjx-pseudoscript` attribute is now filtered along with the other TeX attributes.

### DIFF
--- a/ts/core/MathDocument.ts
+++ b/ts/core/MathDocument.ts
@@ -822,10 +822,10 @@ export abstract class AbstractMathDocument<N, T, D>
     const mitem = new this.options.MathItem(math, jax, display);
     mitem.start.node = this.adaptor.body(this.document);
     mitem.setMetrics(em, ex, containerWidth, scale);
-    if (this.outputJax.options.mtextInheritFont) {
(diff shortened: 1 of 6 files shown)


zorkow (approved):
lgtm.

PR #1233, opened 2025-03-24
Don't allow inline breaks at negative spaces

dpvc:
This PR prevents negative spaces from being considered as valid breakpoints.  This is particularly important for in-line breaks, but affects both in-line and display breaking.


zorkow (approved):
lgtm.

PR #1238, opened 2025-04-06
Fix Lookbehind assertions was not widely supported in browser

WizardMeow:
The "Lookbehind in JS regular expressions" feature is not widely supported in browsers, and this lack of support is causing issues when using MathJax in certain scenarios.

Resolve issue https://github.com/mathjax/MathJax/issues/3352


dpvc:
I think your fix is more complicated than is necessary.  It should be sufficient to simply replace

``` js
dynamic.file.replace(/(?<!\.js)$/, '.js');
```

with

``` js
dynamic.file.replace(/(\.js)?$/, '.js');
```

so that if there is already a `.js` it is removed and then put back, but if there isn't, one is added.

PR #1239, opened 2025-04-06
Add more log print when pack failed

WizardMeow:
The error messages emitted by spawned webpack were not printed, which caused difficulties in adapting the script. 

Simply print stderr.

Before
image

After
image


dpvc:
This is quite helpful.  Thanks!

Would you consider amending the PR to record the cat that an error has occurred and in that case, skip the JSON parsing in the `close` handler for `stdout`?  That way, we can avoid the "unexpected end of JSON" error that occurs above the `stderr` output.

WizardMeow:
It should be good to use the exit code to detect the pack result.

dpvc:
Looks good.  Thanks!

PR #1240, opened 2025-04-08
Use actual space rather than \s to trim mtext content (to not match U+00A0). (mathjax/MathJax#3353)

dpvc:
This PR changes the regexp used to trim spaces from the content of `mtext` elements generated by the `internalText()` command.  The `\s` pattern used originally would match non-breaking spaces (U+00A0), which should not be collapsed into other spaces.  So this uses just the actual space character, not all other space-like characters matched by `\s`.

Resolves issue mathjax/MathJax#3353.


zorkow (approved):
lgtm.

PR #1241, opened 2025-04-18
Implement new explorer paradigm

dpvc:
This PR implements a new paradigm for the expression explorer.  The main goals are:

* Work across the major browser/OS/screen-reader combinations.
* Properly read the full expression when reading the whole page or stepping through the page sentence-by-sentence (or by other units).
* Automatically enter "focus mode" when the expression is focused via tabbing or clicking
* Allow control over the description of the math (e.g., "clickable math") used during reading and focusing.
* A "press H for help" message should be spoken when the math is first focused (but can be turned off by a menu preference).

I tested with 11 combinations (Chrome, Firefox, and Safari on MacOS with VoiceOver; Chrome, Firefox, and Edge on Windows with NVDA and JAWS; and Chrome and Firefox on Linux with Orca) and I believe the all these goals are met.  I spent several weeks experimenting with these 11 configurations to find out what worked the same among them and what worked differently in order to try to find a common approach that worked for all the combinations.  Unfortunately, there didn't seem to be a single setup that worked consistently across all 11, but fortunately, within an OS, there was a setup that worked for all the browser/screen-reader combinations, so it is possible to make it work without users having to change settings by hand.

The prior approach (in the current beta.7 version) is to add aria-labels and `role="treeitem"` to the sub-nodes of the DOM tree and move the focus among these as the expression is explored.  This had several problems.  First, `treeitem` nodes must be in a container with `role="tree"`, which was not the case for us.  Second, it did not enter "focus mode" automatically.  Third, it did not read the full expression when reading the full page in some of the combinations.  Fourth, it basically did not work at all in VoiceOver.

In the new approach, rather than focusing the various node in the math display tree, when you walk to a new node that should be spoken, the explorer inserts a new temporary node that has the needed aria-label and focuses that while visually highlighting the proper display node.  The temporary node is removed when you navigate to the next node and a new one is added for the new speech.  The whole display tree has `aria-hidden="true"` and only the new speech node is visible to the AT.  This part works consistently across all the browser/OS/screen-reader combinations tested.  The only point of contention is how the top-level container element is handled.

In MacOS, there is no "focus mode" versus "browse mode" dichotomy, but in Windows and Unix, we need trigger the screen reader to enter "focus mode" when the expression is focused.  That requires the top-level `mjx-container` to have a role that does that, such as `application` or `tree`.  While `tree` could be used to enter focus mode, this role is designed as a means of making selections, and screen readers speak extra information about where you are in the tree and whether the current item is selected or not, which interferes with the reading of the math itself.  (E.g., NDVA would say things like "item 1 of 8, not selected" when moving within the expression.)  So it turned out to be impractical to use the `tree` role.  The `application` role was the only one that worked consistently across browsers and screen readers in Windows.  Fortunately, that role also works with Orca in Unix.  This role also has the advantage that its `aria-label` is spoken when reading through the page, whereas other roles would end up not speaking anything, so the math was skipped when reading through the page.

For MacOS, the handling of the various top-level roles varied widely across browsers, and there was not a role that would be read properly in all browsers when the page is being read as a whole.  The solution used here is to insert an extra node in the container that holds the speech for the entire equation and give it `role="img"`, since that seems to be one of the few roles whose `aria-label` will be spoken when the page is being read (this is also needed for JAWS to read the expression while reading the full page, if I recall correctly).  The top-level container gets no `role` or `aria-label` in this case, and when the expression is focused, the explorer descends to the `math` node and produces its speech (for Windows and Linux, the focus must remain on the container in order for the screen reader to enter focus mode).

In order to handle the "press H for help" message, MacOS includes it in the speech string when the expression is first focused (it is on the temporary node created for the speech).  Because the other OSs must keep the top-level expression focused in order to enter focus mode, the help message is added to the `aria-roledescription` on the fly (technically not legal, but works; this is the only illegal thing being done in this new explorer, though the use of `img` role is not ideal).

The timing of when the temporary images are created and removed, and when the `tabindex` is changed is important, and has been carefully managed, here.

I will add more comments about the details of the changes below soon, but wanted to get the PR out so that you can be trying it out in the meantime.

### DIFF
--- a/components/mjs/a11y/semantic-enrich/semantic-enrich.js
+++ b/components/mjs/a11y/semantic-enrich/semantic-enrich.js
@@ -3,20 +3,6 @@ import './lib/semantic-enrich.js';
 import {combineDefaults} from '#js/components/global.js';
 import {EnrichHandler} from '#js/a11y/semantic-enrich.js';
 import {MathML} from '#js/input/mathml.js';
-import {Package} from '#js/components/package.js';
-import {hasWindow} from '#js/util/context.js';
-
-if (MathJax.loader) {
-  let path = Package.resolvePath('[sre]', false);
-  if (!hasWindow) {
-    // In Node get the absolute path to the pool file.
-    try {
-      const require = MathJax.config.loader.require;
-      path = require.resolve(path, MathJax.config.options.worker.pool).replace(/\/.*?$/, '');
-    } catch(_err) { }
-  }
-  combineDefaults(MathJax.config, 'options', { worker: { path } });
-}
 
 if (MathJax.startup) {
   MathJax.startup.extendHandler(
--- a/components/mjs/a11y/speech/speech.js
+++ b/components/mjs/a11y/speech/speech.js
@@ -1,6 +1,26 @@
 import './lib/speech.js';
 
+import {combineDefaults} from '#js/components/global.js';
+import {Package} from '#js/components/package.js';
+import {hasWindow} from '#js/util/context.js';
 import {SpeechHandler} from '#js/a11y/speech.js';
+
+if (MathJax.loader) {
+  let path = Package.resolvePath('[sre]', false);
+  if (!hasWindow) {
+    if (MathJax.config.loader.require?.resolve) {
+      const require = MathJax.config.loader.require;
+      const pool = MathJax.config.options?.worker?.pool || 'speech-workerpool.js';
+      path = path.replace(/\/bundle\/sre$/, '/cjs/a11y/sre');
+      path = require.resolve(`${path}/${pool}`).replace(/\/[^\/]*$/, '');
+    } else {
+      path = '';
+    }
+  }
+  if (path) {
+    combineDefaults(MathJax.config, 'options', { worker: { path } });
+  }
+}
 
 if (MathJax.startup) {
   MathJax.startup.extendHandler(handler => SpeechHandler(handler));
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "clean:mod": "clean() { pnpm -s log:comp \"Cleaning $1 module\"; pnpm -s clean:dir $1 && pnpm -s clean:lib $1; }; clean",
     "=============================================================================== copy": "",
     "copy:assets": "pnpm -s log:comp 'Copying assets'; copy() { pnpm -s copy:mj2 $1 && pnpm -s copy:mml3 $1 && pnpm -s copy:html $1; }; copy",
-    "copy:html": "copy() { pnpm -s log:single 'Copying worker html file'; pnpm copyfiles -u 1 'ts/a11y/sre/*.html' 'ts/a11y/sre/require.*' $1; }; copy",
+    "copy:html": "copy() { pnpm -s log:single 'Copying sre auxiliary files'; pnpm copyfiles -u 1 'ts/a11y/sre/*.html' 'ts/a11y/sre/require.*' $1; }; copy",
     "copy:mj2": "copy() { pnpm -s log:single 'Copying legacy code AsciiMath'; pnpm copyfiles -u 1 'ts/input/asciimath/legacy/**/*' $1; }; copy",
     "copy:mml3": "copy() { pnpm -s log:single 'Copying legacy code MathML3'; pnpm copyfiles -u 1 ts/input/mathml/mml3/mml3.sef.json $1; }; copy",
     "copy:pkg": "copy() { pnpm -s log:single \"Copying package.json to $1\"; pnpm copyfiles -u 2 components/bin/package.json $1; }; copy",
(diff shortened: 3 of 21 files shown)


dpvc:
# More Details

For `region.ts`, the speech defaults to U+00A0 so that the region will stay full sized if it contains no speech (rather than collapsing to a thin line).

In `speech-enrich.ts`, these are the changes I mentioned that convert the SRE attributes to the ones used in the new explorer.  In particular, `role="treeitem"` is removed and replaced by `data-speech-node="true"`, and the `aria` attributes (other than `aria-level`) are removed.  These changes could be made in SRE and removed from here.

In `speech.ts`, we add a `detachSpeech()` that calls the web-worker to remove the speech attributes that were added by `attachSpeech()` when the state is changed to below when speech is added.  In `attachSpeech()` we save the promise for when the speech is attached so that the explorer can find it in its overridden `attachSpeech()` method.  In the SpeechMathDocument, a couple of properties left over from the no-longer used asynchronous speech code are removed.    The `detachSpeech()` call is made in the `state()` function, when needed.

In `GeneratorPool.ts`, the `MathItem` type is replaced by `SpeechMathItem` in order to access it properties.  We add the `enableSpeech` and `enableBraille` properties to the options so that we can tell whether speech or Braille should be added to the expression tree.  The `summary()` processing has been simplified and moved to `KeyExplorer`.  The `CleanUp()` code is no longer needed, since we don't modify any of the tree node's attributes.  The last move and last speech code is no longer used (since we don't change the expression nodes' attributes) and is removed.  The `updateRegions()` method is simplified put pushing most of the work into `getLabel()` and a new `getBraille()` method.  Finally, the `depth()` action has been simplified and moved to the `KeyExplorer`.

In `SpeechUtil.ts`, we add a `BRAILLE` attribute name for use in obtaining the braille label, similar to the ones for getting the speech.

in `WebWorker.ts`, we replace the `MathItem` type with `SpeechMathItem` in order to access its properties.  We pass the `enableSpeech` and `enableBraille` options on to the `Attach()` method so that only the needed attributes are attached to the DOM tree.  Since the new explorer paradigm manages `aria-label` and `aria-braillelabel` in the `KeyExplorer`, we don't add those attributes here; instead we add these using the data-semantic attributes where the explorer will pick them up.  The `setSpeechAttribute()` and `setSpeechAttributes()` methods now add speech and braille only when the options allow it.  Finally, we add a `Detach()` method that undoes the `Attach()` method using a recursive `detachSpeech()` method to remove any aded attributes from the expression DOM tree.  This is called by the SpeechMathItem when the state changes to below where the speech has been added.

The `:focus` CSS has been removed from `output/chtml.ts`, since the focus is not placed on the expression nodes themselves, but on the `mjx-speech` nodes.

The `Menu.ts` file now includes menu items to control the role description to use (e.g., 'math', 'clickable math', no description, etc.), and a checkbox for whether to add the 'press H for help' message.  In addition, in `applySettings()`, the settings are updated *before* setting the renderer, so that things like the scaling factor are properly set when the pre-rendering occurs.  (Prettier also adjust the line breaks in the `render()` method.)  Finally, in the `contextemenu` event handler, we retain the focused node (if the explorer exists and is active) so that if the context menu is used while an expression is being walked, it will refocus the correct node when the menu exits.

In `MenuHandler.ts`, we ad a `getMenus` render action that causes the menu store to be rebuilt.  This was being done in the document's `updateDocument()` method, but if a single expression was rerendered (e.g., but an maction toggle), the menu store didn't properly include the newly rendered expression, and that caused problems with reactivating the explorer after a menu closed on an expression that includes an maction that was toggled.  Because of the new `getMenu` action, the `updateDocument()` method no longer needs to be overridden.

Finally, `MenuUtil.js` now uses the new `context.os` value to determine whether we are in MacOS or not.

That completes the details about these changes.

zorkow (approved):
Some typos. 

I'll send you an email regarding testing.

zorkow:
There are some issues that I have now discovered (in the `update/explorer-misc` branch):
1. Highlight is left on changing rules/styles. Easy fix by adding an `this.pool.unhighlight();` before the `restartAfter`.
2. The removal of the `lastMove` and `cleanup` leads to summary and depth no longer toggling the speech. That is, we can no longer get back to the full speech, without making a move.  Theres effectively two ways to solve this issue:
    1. Reinstate keeping (and clearing) the last move. But that does not really fit well with the current way keydown is handled.
    2. Have a `repeat` key. Previously (in 2.7), we had used the `tab` key for repeat, but that does not make sense since we still want to tab out of the expression. We could use `r`/`R`
3. Collapse does no longer seem to work as expected. I am still investigating this one.

zorkow:
I've partially solved point (3) by ensuring that `data-speech` attribute is copied. Exploration over the collapsed element
does not work yet, but I am sure has a similar solution.

These will be my next steps: 
* PR that removes SRE related functionality no longer needed in MathJax main.
* PR with the corrections I have mentioned above.

zorkow:
Great. I am just in the process of making an SRE release for the new
SRE PR. It took me all morning to get the tests sorted out and the
github actions running again.
Then I make a PR of the fixes and you can have a look at it.

Best,
Volker

PR #1242, opened 2025-04-23
Refactor Tree Colorer

zorkow:
Refactors code for tree coloring from SRE to the actual explorer. This reduces the footprint of code that is pulled in from SRE. I will remove more dependencies once PR #1241 is merged.

Note, that commit 04775ebbe35a00344dc28eab0d2220637554d450 only contains changes from automatic formatting and linting.


zorkow:
> This looks good, but the changes in your second commit for prettier changes are already in #1241. I don't know if that will cause a conflict or not (#1241 does have other changes to `Menu.ts`).

Good point. I'll try to remove the last commit with `pick` and `push`. I have never done that, so it should be interesting.

zorkow:
Removed last commit following https://stackoverflow.com/questions/36168839/how-to-remove-commits-from-a-pull-request

PR #1243, opened 2025-04-25
Change to using @mathjax/src rather than mathjax-full

dpvc:
This PR moves `mathjax-full` to a scoped package, `@mathjax/src`, as we have discussed.  The fonts are already scoped, and this completes the move to `@mathjax`.


zorkow (approved):
Looks a lot cleaner.

PR #1244, opened 2025-04-25
Misc fixes

dpvc:
This is a collection of miscellaneous fixes that I have accumulated while updating the documentation for v4.

The `makeAll` and `pack` changes are to handle building of fonts and third-party extensions better.

The `assistive-mml` change is to prevent the assistive MathML representation from being pushed outside the bounding box of the MathJax output.

The `lite/Element.ts` changes are to prevent the version mismatch error when a node application uses MathJax components rather than direct calls, and the loader uses the `MathJax.loader.load()` command.

The `components/package.ts` change is to get the proper result returned when `require()` is used in `MathJax.loader.load` (e.g., when `MathJax = {loader: {require: require}}` is used in CommonJS applications).

The `components/startup.ts` change is to add a `MathJax.whenReady()` command to access the document's `whenReady()` method.

The `RequireConfiguration.ts` change is to prevent an error when the menu code isn't loaded (e.g., when components are loaded by hand rather than through a combined component, and `ui/menu` isn't included).

The `mmultiscripts.ts` change is to avoid a bug in WebKit that causes incorrect placement of scripts when `<none>` is used as a superscript in `<mmultiscripts>`.

### DIFF
--- a/components/bin/makeAll
+++ b/components/bin/makeAll
@@ -125,10 +125,9 @@ function fileRegExp(name) {
 }
 
 /**
- * Get and change the current working directory
+ * Get the current working directory
  */
 const root = process.cwd();
-process.chdir(path.dirname(path.dirname(__dirname)));
 
 /**
  * Regular expression for the components directory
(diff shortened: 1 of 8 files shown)


zorkow (approved):
lgtm.

PR #1245, opened 2025-04-28
Update refocus from menu actions, and adjust some options

dpvc:
This PR is an update to `update/explorer` that fixes some problems with refocusing the selected item when the menu is use from within the explorer, particularly when the menu selection opens a dialog.  It also moves the `roleDescription` option to the explorer's `a11y` options, so it can be configured even when the menu component is not loaded, and removes the `internalSpeechTitles` option, which is no longer supported since the speech is added to the DOM elements after the output is created.

The changes in `explorer.ts` make `roleDescription` an `a11y` option for the document.

In `KeyExplorer.ts`, we clear the `nofocus` setting after it is used, and change the button checking from `button !== 0` to `button === 2` when checking for right mouse buttons that will cause context-menu events.

In `GeneratorPool.ts` the setting of the options in the `init()` function is always done, so that if the menu changes `enableSpeech` or `enableBraille` and does a re-render of the output (which calls `init()` again), the options will be updated.

The changes to the SVG output jax are to remove the `internalSpeechTitles` option and the code that implements it.

The changes to the menu code is to better set the `MathItem.outputData.nofocus` and to set the speech explorer's `refocus` value to the node that was selected before the menu opened.  Because there are focus changes that clear the selection before the `contextmenu` event is triggered, we use a `mousedown`event to record the current selection before it is lost.

When a dialog is opened by a menu action, we need to keep the `nofocus` value, so introduce a new `postInfo()` method that records the `nofocus` value before posting an `Info` or `SelectableInfo` dialog box.  This is used for all the `.post()` commands in the menu.

The `roleDescription` setting now uses `a11yVar<string>()` so we don't need to call `setA11yVar()` by hand in `setRoleDescrition()`.

We add some checks to make sure `options.displayOverflow` and `options.linebreak` are available (since these are defined in `output/common.ts`, and may not be in all output jax).

We check for `MathJax._.a11y.explorer` rather than just `MathJax._.a11y.speech` or the other extensions, since we load `explorer`.  This makes sure the explorer is available, not just the lesser extension.

We add code to tell the explorer where to refocus when the scale dialog box is open (it uses `prompt()` rather than an `Info` instance, so `postInfo()` isn't being called for this).

Finally, we remove the `data-speech-node` attributes that where introduced by the explorer update.

### DIFF
--- a/ts/a11y/explorer.ts
+++ b/ts/a11y/explorer.ts
@@ -287,6 +287,7 @@ export function ExplorerMathDocumentMixin<
         viewBraille: false,                // display Braille output as subtitles
         voicing: false,                    // switch on speech output
         help: true,                        // include "press h for help" messages on focus
+        roleDescription: 'math',           // the role description to use for math expressions
       }
     };
 
@@ -429,6 +430,7 @@ export function ExplorerMathDocumentMixin<
         options.a11y.speechRules = `${options.sre.domain}-${options.sre.style}`;
       }
       options.MathItem = ExplorerMathItemMixin(options.MathItem, toMathML);
+      options.MathItem.roleDescription = options.roleDescription;
       this.explorerRegions = new RegionPool(this);
       if ('addStyles' in this) {
         (this as any).addStyles(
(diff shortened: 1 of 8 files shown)


zorkow (approved):
lgtm.

With the exception mentioned in PR #1241, we solve those in the next PRs.

PR #1246, opened 2025-04-28
Corrects worker issues

zorkow:
Corrects following issues:
* Updates `package.json` to the correct SRE package. Also removes an unused linting option that crashes the `lint:fix` lifecycle script.
* Makes cancel calls on webworker more secure, to avoid error that occurs, for example, when using lazy typesetting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
   "scripts": {
     "=============================================================================== code hygene": "",
     "lint": "check() { eslint ${1:-ts/}; }; check",
-    "lint:fix": "fix() { eslint --fix ${1:-ts/}; -- format unix}; fix",
+    "lint:fix": "fix() { eslint --fix ${1:-ts/}; }; fix",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
   "scripts": {
     "=============================================================================== code hygene": "",
     "lint": "check() { eslint ${1:-ts/}; }; check",
-    "lint:fix": "fix() { eslint --fix ${1:-ts/}; -- format unix}; fix",
```
dpvc, line 75: I think it may have been intended as `--format unix` (no space after `--`), but I'm OK with removing it.

dpvc (approved):
LGTM

package.json
zorkow: I've pivoted twice on this one. Regardless of the space, the order would have been incorrect anyway. But we really only need the unix formatting, when in Emacs. On the console the regular one looks better.

PR #1247, opened 2025-05-14
Update to the SRE API 

zorkow:
This removes everything from MathJax that we don't need anymore. Primarily,
*  `MathMaps` handling, 
* customised locale loading,
* `Promise` handling for SRE to be ready. We don't load locales anymore, hence this is no longer necessary.

In addition we now no longer go through the top-level SRE API, which still would require the promise, insteady we only go through the API of the `enrich` module, which is a lot slimmer. Hence we shave off some file sizes as well.
Options are now set explicitly in the new SRE `Options` object. Consequently this only works with SRE v5.0.0-alpha.2.


zorkow:
> This looks good. Does the updated `pnpm-lock.yaml` file also need to be included here?

Done.

PR #1248, opened 2025-05-14
Fix/explorer issues

zorkow:
This fixes three issues:
* highlighting issue on rule change 
* remembering explorer state
* walking issue on collapsed elements

### DIFF
--- a/ts/a11y/speech/WebWorker.ts
+++ b/ts/a11y/speech/WebWorker.ts
@@ -393,6 +393,7 @@ export class WorkerHandler<N, T, D> {
       node = adaptor.childNodes(node)[0] as N;
       adaptor.setAttribute(node, 'data-semantic-type', 'dummy');
       this.setSpecialAttributes(node, sid, '');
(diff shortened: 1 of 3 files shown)


ts/a11y/speech/WebWorker.ts
```diff
@@ -393,6 +393,7 @@ export class WorkerHandler<N, T, D> {
       node = adaptor.childNodes(node)[0] as N;
       adaptor.setAttribute(node, 'data-semantic-type', 'dummy');
       this.setSpecialAttributes(node, sid, '');
+      adaptor.setAttribute(node, 'data-speech-node', 'true');
```
dpvc, line 396: Although this works, I've handled it in a different location just to be sure it is always set when speech or Braille is available, rather than relying on the origin `treeitem` node, as currently done in semantic-enrich.

dpvc (commented):
As mentioned in the comment, I think there are other solutions to these, and I've used them in my own PR.  So I would prefer that PR to this one.

zorkow:
Subsumed by #1251

PR #1249, opened 2025-05-15
Fix a number of explorer and menu issues.

dpvc:
# Overview

This PR fixes the issues [identified in review](https://github.com/mathjax/MathJax-src/pull/1241#issuecomment-2876224008), plus a number of other explorer problems that I found while updating the documentation, and in particular, some interactions between the menu and explorer components.

In particular:
1. The updating of ARIA attributes after the DOM nodes are in place was not working with Orca, and so the `mjx-container` node is now cloned and reinserted in order to update the accessibility tree properly.
2. This lead to some timing issues with the handing of maction re-rendering, which is fixed here.
3. There is also a problem with refocusing the proper element when an maction from the complexity extension triggers a re-render.
4. The `d` and `x` keys are now toggles, as they were before (I didn't realize that they worked that way before).
5. The depth speech now correctly uses `expandable` when the first toggle is showing.
6. The highlighting is removed when style or rules are changed.
7. The `data-speech-node` attribute is placed properly for complexity maction nodes.

Other issues or features include:
1. A new double-click handler is added so that double clicking on an equation focuses the complete equationn rather than the clicked node.
2. The `roleDescription` menu option is now mirrored in the `a11y` options so that it is possible to set this via configuration in a situation where the menu component is not loaded.
3. Problems with autoloading the complexity and explorer components when sticky menu options selected by the user are in effect at startup have been fixed.

-----

# Details

## `explorer.ts`

New types are added to allow access to the document's menu and the new MathItem `addListeners()` method when they are available (the menu component may not be loaded, and `addListener()` is part of the `HTMLMathItem`, not the base `MathItem`, and we don't know which we may be extending).

The functionality that was in `updateDocument()` is moved to the `attachSpeech()` method so that we can wait on the speech promise to do it (without this, we end up focusing the old nodes before the `mjx-container` is cloned, and so lose the focus and end up focusing on the top level rather than the proper element).  The `updateDocument()` method is no longer needed.

We include an `addListener()` method that is used by the KeyExplorer to reattach event listeners to the cloned `mjx-container` element.  Since the child nodes are moved from the old container to the clone, we only need to worry about the listeners for the top-level container.

## `KeyExplorer.ts`

A new `speechType` properly is added to indicate what speech is currently being presented (so that we can toggle the `d` and `x` values).

A new `dblclick` event is added to allow double-click to select the entire equation rather than just the clicked node.  This makes it easy to navigate a complete equation by double-clicking on it rather than clicking and arrowing to the top.  This is facilitated by changes in the `MouseDown()` method to not refocus on the mouse down for the double click (when a range is already in effect), and to not avoid saving the click when the container node is focused after the first click.

The `depth()` function now acts as a toggle using the `speechType`, and `expandable` is used at the proper time (the old criterion always caused `collapsable` to be used).  Note that this is not appropriate for `maction` elements that don't come from the complexity extension, like those from the TeX `\toggle` command, which probably should have `clickable` or `toggle` instead, but there is no translation for that at the moment.

Similarly the `summary()` method now acts as a toggle.

The `speechType` is cleared when `setCurrent()` is used (since that means speech is from a math tree node).

In the `speak()` method, a blank string is converted to `blank` (this needs to be localized), so that something is spoken (e.g., for invisible times).  This might not be the right thing to do.

An `updateAT()` method is added that does the cloning and replacement of the `mjx-container` element once the ARIA attributes have been added.  This is the key to getting things to work properly in Orca.

The un-highlight operation is added to the `restartAfter()` method so as not to need to repeat it in `depth()` and `summary()`.

An `addListeners()` method is added so that the `SpeechMathItem.addListeners()` method can call it to reattach the listeners to the cloned container.

Finally, the `semanticFocus()` method is modified to return the proper selectort string for maction nodes from the complexity extension.

## WebWorker.ts

The `data-speech-node` attribute is set for any node that has speech or Braille, fixing the navigation issues with mactions.

## Menu.ts

The `roleDescription` is transfered to the a11y settings on initialization so that the user's sticky setting is used when it is not the default setting.

The `checkLoadableItems()` method is modified to properly load the explorer and complexity components, if the user's sticky items are not the defaults and they are needed.

The `loadComponent()` callback is moved to after the `_loadingOK()` call since some callbacks need to know that the promise is resolved.

When a new document is created after loading a component, the `webworker` is transferred to the new document.  (there may be a need for a more general solution to copying data from one document to another, but it can wait for now.)

The creation of the event listeners in `addMenu()` has been moved to a separate `addEvents()` method so that it can be called by `addListeners()` (and lines 1711 and 1712 are moved up from after event creation in the diff listing).

## MenuHandler.ts

New types are added to simplify some of the later code (though prettier has complicated things a bit).

A new `addListeners()` method is created to add the menu event listeners (so they can be added again when the `mjx-container` is cloned by the KeyExplorer).  To allow access to the `super.addListeners()` method, the `MenuMathDocument` interface extends `ExplorerMathDocument` in addition to the `ComplexityMathDocument`, since the former has `addListener`.


zorkow (commented):
I have not reviewed the code in details yet. However, I have done a couple of experiments: collapsing with navigation and speech works now fine. There are is one issue when collapse/expansion is triggered by mouseclick the query selector is incorrect, pointing to `null` rather than the semantic id. I assume this is a timing issue, that the node is already cleared by the time `semanticFocus` is called.

Second issue is that the state of the explorer is not kept for restart. My solution from PR #1248 [Line 767ff](https://github.com/mathjax/MathJax-src/pull/1248/files#diff-8e540e55d7d22d4ba84e9bf07da8db96b0cee069dc2fc46954decdcc79749dc6L767) would still work.

dpvc:
> Second issue is that the state of the explorer is not kept for restart

If you are doing this in the lab, I need to make an update to the lab for proper handling of the `MathItem.rerender()` calls that mactions use, as the lab overrides that and doesn't take the new code into account.  I have a version of the lab that uses the `renderAction` path to handle the typesetting updates instead, which avoids that problem.  It is the lab, not the new branch, that is at fault for this.

Also, I have worked out the problem with the hover and flame issues (reattaching the events properly, and getting the bounding box in SVG), and will make a PR for that, but I want to test with this new branch first.

dpvc:
That is, your new highlighter branch.

zorkow:
> > Second issue is that the state of the explorer is not kept for restart
> 
> If you are doing this in the lab, I need to make an update to the lab for proper handling of the `MathItem.rerender()` calls that mactions use, as the lab overrides that and doesn't take the new code into account. I have a version of the lab that uses the `renderAction` path to handle the typesetting updates instead, which avoids that problem. It is the lab, not the new branch, that is at fault for this.
> 

I tried both ways: In the lab and in a local test file (using `tex-chtml.js`. Neither kept the state. With my little fix, it did.

> Also, I have worked out the problem with the hover and flame issues (reattaching the events properly, and getting the bounding box in SVG), and will make a PR for that, but I want to test with this new branch first.

Very good. 

I have fixed the auto voicing issue in branch `fix/auto_voicing`. But I need to make new SRE release for that. I want to fix the clearspeak preference menu first and then make a release. I think I have a way to easily do this via the worker.

dpvc:
> I tried both ways: In the lab and in a local test file (using tex-chtml.js. Neither kept the state.

Perhaps I'm misunderstanding what you are referring to.  Can you give the sequence of actions needed to reproduce the issue?

> With my little fix, it worked.

That is going to cause other problems.  The `current` value needs to be `null` when `setCurrent(null)` is called.

> I have fixed the auto voicing issue

Great!  

> But I need to make new SRE release for that. I want to fix the clearspeak preference menu first and then make a release

OK, sounds good.

zorkow:
> > I tried both ways: In the lab and in a local test file (using tex-chtml.js. Neither kept the state.
> 
> Perhaps I'm misunderstanding what you are referring to. Can you give the sequence of actions needed to reproduce the issue?
> 

Take as example $f(x)$. Tab to the expression and exploration starts at the top level $f(x)$, then you press down - right - right and you should be at $(x)$. Now press tab and you leave exploration. Go back with shift tab and exploration restarts:
At the moment it restarts at the top level, but it should restart at $(x)$, where we left it.

> > With my little fix, it worked.
> 
> That is going to cause other problems. The `current` value needs to be `null` when `setCurrent(null)` is called.
> 

I see. Maybe we can compute the query  selector using `semanticFocus` before restarting.

zorkow:
> I see. Maybe we can compute the query selector using semanticFocus before restarting.

How about this as a fix:

``` Typescript
    if (this.current) {
      this.current.classList.remove('mjx-selected');
      this.pool.unhighlight();
      if (!node) {
        this.restarted = this.semanticFocus();
        this.removeSpeech();
      }
      this.current = null;
    }
```

This will also help out with the cs preference menu.

zorkow:
I am getting the following error on changing rules/styles  (keys '<' and '>') during exploration in the `updateAT` method:

```
Uncaught (in promise) NotFoundError: Failed to execute 'append' on 'Element': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?
```

dpvc:
> At the moment it restarts at the top level, but it should restart at $(x)$, where we left it.

Ah, now I see what you are talking about.  That was actually intentional.  I don't agree with that it should go back to where it was, and believe that tabbing to an expression should always start at the top.  It would be confusing to move to an expression and have it read only part of it (unless there were some indication that it is not at the top level).  And the expression can receive focus in a number of ways, not just tabbing.  For example, when reading through the page, some AT will focus the element when it is read, and if focus returned to a sub-expression, then reading the page would give the wrong reading (instead of saying "the function f of x" one would hear "the function x", which I think you'll agree is not ideal.

Just because you had moved to a sub-expression doesn't mean it should restart at that sub-expression, just as if you open the menu using space, arrow to a submenu, then press escape to close the menu, hitting space again only opens the initial menu, not the submenu that you had open when you left the menu.

dpvc:
> I am getting the following error on changing rules/styles

It looks like different browsers fire events differently during the copying of nodes to the cloned container and Chrome is performing a focusout during the `append()` call.  I'll need to look into that further.

zorkow:
Somehow speech stopped working altogether now. But that could be a messed up setup. I'll try again later.

dpvc:
I've tried with Windows and Ubuntu, and they both worked for me.  (I am using a branch that merges all the remaining v4 branches.)  Are you still having troubles?  It may be that the `fix/explorer-issues` branch needs to use the older SRE version?

dpvc:
Or that you need to re-run `pnpm -s lab:sre` to get it built with the right version?

dpvc:
Closing in favor of #1262

PR #1250, opened 2025-05-15
Add highlighter module

zorkow:
*Note, that this PR includes PR #1247. Only the four files in the `explorer` directory are relevant.*

This PR adds the highlighter as module directly to MathJax, rather than outsourcing it to SRE. The module combines (roughly) the functionality of these SRE sources: [ColorPicker](https://github.com/Speech-Rule-Engine/speech-rule-engine/blob/develop/ts/highlighter/color_picker.ts), [Highlighter Interface](https://github.com/Speech-Rule-Engine/speech-rule-engine/blob/develop/ts/highlighter/highlighter.ts), [Abstract Highlighter Class](https://github.com/Speech-Rule-Engine/speech-rule-engine/blob/develop/ts/highlighter/abstract_highlighter.ts), [CHTML Highlighter](https://github.com/Speech-Rule-Engine/speech-rule-engine/blob/develop/ts/highlighter/chtml_highlighter.ts), [SVG(v3) Highlighter](https://github.com/Speech-Rule-Engine/speech-rule-engine/blob/develop/ts/highlighter/svg_v3_highlighter.ts) and the [Highlighter Factory](https://github.com/Speech-Rule-Engine/speech-rule-engine/blob/develop/ts/highlighter/highlighter_factory.ts)

* Highlighter handles adding and removing highlighting of single nodes, set of nodes and all maction nodes (e.g. for the flamer). Most basic functionality is implemented in the `AbstractHighlighter` class. The dedicated SVG and CHTML highlighters implement:
    * `highlightNode` for highlighting single nodes
    * `unhighlightNode` for removing highlights from a single nodes
    * and how to test for and get all `maction` nodes in an expression.
* Highlighters have a unique id and maintain an internal state in terms of the nodes they highlighted and their original colors. Both is important for correctly unhighlighting in particular if more than one highlighter operates on an expression (e.g., in the case of auto voicing).
* Colors are set as named colors plus an optional alpha value and are transformed into `rgba` color strings for both fore and background colors.

The module has a slim interface, only exposing the `Highlighter` interface and the factory method `getHighlighter`, where the latter picks and returns the correct highlighter for the current renderer.


ts/a11y/explorer/Highlighter.ts
```diff
@@ -0,0 +1,485 @@
...
+      node.getElementsByClassName(this.mactionName)
+    ) as HTMLElement[];
+  }
+
+  /**
+   * @override
+   */
+  public isMactionNode(node: Element): boolean {
+    const className = node.className || node.getAttribute('class');
+    return className ? !!className.match(new RegExp(this.mactionName)) : false;
```
dpvc, line 275: Note that MathJax v4 does not use the class name to identify maction nodes (that was a v2 approach).  In version 4, you could use the tag name being `'mjx-' + this.mactionName` or having an attribute `data-mjx-node` be `this.mactionName`.  Because of this, both the CHTML and SVG highlighters have to override this.   It might be better to either make this an abstract method that the subclass must implement, or make it check for the two approaches that are used.

ts/a11y/explorer/Highlighter.ts
```diff
@@ -0,0 +1,485 @@
...
+    const transform = node.getAttribute('transform');
+    if (transform) {
+      rect.setAttribute('transform', transform);
+    }
+    rect.setAttribute('fill', this.background);
+    node.setAttribute(this.ATTR, 'true');
+    node.parentNode.insertBefore(rect, node);
+    info = { node: node, foreground: node.getAttribute('fill') };
+    if (node.nodeName !== 'rect') {
+      // We currently do not change foreground of collapsed nodes.
```
dpvc, line 362: I'm not sure I understand how the condition in the previous line relates to this.  And in my testing, the foreground color of collapsed nodes does get changed.

ts/a11y/explorer/Highlighter.ts
```diff
@@ -0,0 +1,485 @@
...
+  public unhighlightNode(info: Highlight) {
+    info.node.style.backgroundColor = info.background;
+    info.node.style.color = info.foreground;
+  }
+
+  /**
+   * @override
+   */
+  public isMactionNode(node: HTMLElement) {
+    return node.tagName?.toUpperCase() === this.mactionName.toUpperCase();
```
dpvc, line 447: How about using `toLowerCase()` and not having to uppercase `this.mactionName` every time?

dpvc (approved):
This works for me in my test cases.  I had some typos, and a few minor suggestions.  Take them or leave them.

ts/a11y/explorer/ExplorerPool.ts
```diff
@@ -1,6 +1,6 @@
 /*************************************************************
  *
- *  Copyright (c) 2022-2024 The MathJax Consortium
+ *  COPYRIGHT (c) 2022-2024 The MathJax Consortium
```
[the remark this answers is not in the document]
zorkow: Must have been a random `M-u`. Fixed.

ts/a11y/explorer/ExplorerPool.ts
```diff
@@ -29,7 +29,8 @@ import { SpeechExplorer } from './KeyExplorer.js';
 import * as me from './MouseExplorer.js';
 import { TreeColorer, FlameColorer } from './TreeExplorer.js';
 
-import * as Sre from '../sre.js';
+import { Highlighter, getHighlighter } from './Highlighter.js';
+// import * as Sre from '../sre.js';
```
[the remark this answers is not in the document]
zorkow: Removed.

ts/a11y/explorer/Highlighter.ts
zorkow: That's the old default from SRE, which worked on a number of the 2.7 renderers. I thought I had caught them all and made them abstract.

ts/a11y/explorer/Highlighter.ts
dpvc: > But there is still the timing issue wrt. highlighting `maction` backgrounds

That is resolved in one of my PRs.

PR #1251, opened 2025-05-16
Fix problems with flame/hover, re-attaching after < and >, and with the menu loading extensions at startup

dpvc:
# Overview

This PR fixes several more problems with the explorer:

1. Fixes the flame and hover highlighting by making sure the event handlers are re-attached.
2. Makes sure that flame and hover can be turned on and off properly, and that they start up when sticky menu selections are on.
3. Fixes flame display in SVG output.
4. Removes the outline from flame and hover highlighting in SVG output.
5. Fixes problems with `<` and `>` in Chrome.
6. Adds a missing `Log()` command in the web-worker code.
7. Adjusts the menu code for loading extensions so that the complexity extension in particular will load properly on startup when its menu item is selected.

-----

# Details

## explorer.ts

The `EXPLORER` state is move to after the `INSERTED` state, so that the output is actually in the page when the flame highlighter tries to measure the size of SVG output nodes.

The `addListeners()` method now has the new container node as a parameter so that the explorers can update their `node` properties using a new `updateNode()` method of the explorer pool

## Explorer.ts

We add the `UpdateNode()` method to the base `Explorer` class.  It updates the saved `node` and adds any needed event handlers, when requested.

## ExplorerPool.ts

We add an `updateNode()` method that calls `UpdateNode()` on all the explorers, indicating whether event handlers should be added (by whether the explorer is attached or not).

## KeyExplorer.ts

The `updateAT()` method is updated to sop and restart the explorer if it has a current selection (as is the case when the style or ruleset is changed).  This resolves the problems with `<` and `>`.

The unhighlightin and setting of the `current` value are removed from the `restartAfter()` method, as the changes to `updateAT()` now cause the unhighlighting and restarting the speech to occur when that is called by the `attachSpeech()` method.

The `addListeners()` method has been replaces by the `UpdateNode()` method, and we override it here to unset `eventsAttached` so that the super-class `UpdateNode()` will re-add the events.

## WebWorker.ts

The `Log()` command is added, as it was missing and caused error messages during debugging.

## svg.ts

The CSS selector for adding a border to the highlighter rectangles has been made more specific so that flame and hover highlighting are not outlined.

## MenuHandler.ts

The `addListeners()` method now has the `node` argument.

The `checkLoading()` method has been modified to return a boolean that is used to stop the renderActions when loading is still occurring.  This is needed because in that case the `MathJax.startup.document` is replaced when the component loads (since there will be a new handler that needs to be included); in the past, the `restartAfter()` calls would cause the `renderActions` to be restarted on the **original** document, not the new one.  Now, the return value of `true` will cause the render actions to stop, but the callback from the `loadComponents()` call already includes a `rerender()` call, and so it will start the render actions will be run again on the new document automatically.  These changes were the key to getting the complexity extension to load properly on startup.

The `checkLoading` render action now calls `doc.checkLoading()` by hand, since that is the only way to return the `true` value that stops the render actions.

The old `checkLoading()` method is now `_checkLoading()`, and the new `checkLoading()` calls it and traps the retry errors to determine wherther to stop the render actions or not.

### DIFF
--- a/ts/a11y/explorer.ts
+++ b/ts/a11y/explorer.ts
@@ -52,16 +52,19 @@ export type HTMLDOCUMENT = SpeechMathDocument<HTMLElement, Text, Document> & {
   menu?: any;
 };
 export type HTMLMATHITEM = SpeechMathItem<HTMLElement, Text, Document> & {
-  addListeners?(doc: SpeechMathDocument<HTMLElement, Text, Document>): void;
+  addListeners?(
+    node: HTMLElement,
+    doc: SpeechMathDocument<HTMLElement, Text, Document>
+  ): void;
 };
 export type MATHML = MathML<HTMLElement, Text, Document>;
 
 /*==========================================================================*/
 
 /**
- * Add STATE value for having the Explorer added (after TYPESET and before INSERTED or CONTEXT_MENU)
+ * Add STATE value for having the Explorer added (after INSERTED and before CONTEXT_MENU)
  */
-newState('EXPLORER', 160);
+newState('EXPLORER', STATE.INSERTED + 30);
 
 /**
  * The properties added to MathItem for the Explorer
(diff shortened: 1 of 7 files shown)


dpvc:
Note:  this PR targets the `fix/explorer-issues` branch, and should not be merged until that branch is reviewed.

Also note: you can merge the `feature/highlighter` branch into this one and that works as well.  Since the highlighter branch includes the new SRE branch, that means this works with that as well.

dpvc:
OK, I'll check into it.

PR #1252, opened 2025-05-17
Fix cssText creation to handle border-radius and similar values. (mathjax/MathJax#3365)

dpvc:
This PR fixes a problem where `border-radius` CSS was being dropped.  The fix is to make sure that when the parent name is found, it has the child name as an actual child (e.g., the `border` data doesn't have `border-radius` as a child).

Resolves issue mathjax/MathJax#3365.


zorkow (approved):
lgtm.

PR #1253, opened 2025-05-17
Add an `insertStyles()` method to update styles when an extension is loaded

dpvc:
This PR adds support for an extension having styles.  E.g., the explorer has styles, and if it is loaded by the menu after the page is already running, those style wouldn't get set.  This PR fixes that.

### DIFF
--- a/ts/handlers/html/HTMLDocument.ts
+++ b/ts/handlers/html/HTMLDocument.ts
@@ -349,6 +349,9 @@ export class HTMLDocument<N, T, D> extends AbstractMathDocument<N, T, D> {
    */
   public addStyles(styles: StyleJson) {
     this.styles.push(styles);
+    if ('insertStyles' in this.outputJax) {
+      (this.outputJax as any).insertStyles(styles);
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


ts/handlers/html/HTMLDocument.ts
```diff
@@ -349,6 +349,9 @@ export class HTMLDocument<N, T, D> extends AbstractMathDocument<N, T, D> {
    */
   public addStyles(styles: StyleJson) {
     this.styles.push(styles);
+    if ('insertStyles' in this.outputJax) {
```
zorkow, line 352: Would this be better?
```suggestion
    if (Object.hasOwn(this.outputJax, 'insertStyles') {
```

zorkow (approved):
One comment. Otherwise fine.

ts/handlers/html/HTMLDocument.ts
dpvc: This won't work, because 'inserStyles` is in the prototype, not a direct property of the object.  That means that `Object.hasOwn()` will not find it, but `in` will.

PR #1254, opened 2025-05-17
Remove unneeded configuration, and simplify replacements

dpvc:
This PR cleans up some replacements (they don't need to use functions), and removes unneeded configuration options.


zorkow (approved):
lgtm.

PR #1255, opened 2025-05-17
Report failure of \require rather than throw an error

dpvc:
This PR makes a failed `\require{}` produce a TeX error rather than throwing a general error that causes MathJax to stop running.  A new test is added to cover the new code.


zorkow (approved):
lgtm.

PR #1259, opened 2025-05-22
Fix script spacing when MathML spacing is used. (mathjax/MathJax#1224)

dpvc:
This PR fixes an issue with MathML spacing where an `mo` whose content is not in the operator dictionary and when it is used with `scriptlevel` greater than 0.  In that situation, it should use the reduced space size of `SMALLSIZE.


zorkow (approved):
lgtm.

PR #1260, opened 2025-05-26
Fix/cs preferences

zorkow:
**Only works with the latest versions of SRE, v5-alpha.3 and above. Make sure to run `pnpm install` first!**

The PR does 4 main things:
1) Fixes the auto voicing option
2) Reinstates the functionality of the Clearspeak submenu
3) Ensures that consistent ruleset/preference changes on an element. 
4) Removes `ChromeVox` menu entry

## Auto Voicing Option

Auto voicing with synchronised highlighting uses the `SemAttr.SPEECH_SSML` attributes to compute speech, pauses and the highlighting marks. It also depends on these being available at all levels of the expression.
* The container gets the ssml attribute form the speech structure from SRE (latest version!)
* We compute SSML not just for summary but also for regular speech in `SsmlAttributes`.
* Hiding the `SpeechRegion` cancels the speech.

## Clearspeak Submenu

Since some of the functionality of the Clearspeak submenu depends on locales being loaded we now need to route this through the worker. In particular we need to call the worker in two functions:
* `getLocalePreferences` computes the clearspeak preferences for a locale; they need to be computed only once are stored in a dedicated `localePreferences` map
* `getRelevantPreferences` computes the category of preferences that is useful for the currently focused node. E.g., if it is a fraction then the fraction related preferences are computed (this is similar somewhat similar `nextStyle`). Results are passed back via the `relevantPreferences` map.
 
To support this I have added two methods in `GeneratorPool`, `WebWorker`, and `speech-worker`, respectively.
In addition 
* `DynamicSubmenus` now work asynchronously.
* SRE previously kept the global state of styles/preferences for `MathSpeak` and `ClearSpeak`. These are now kept in `previousPrefs`. There is still an issue when a locale has no clearspeak, going back to one that does still leaves it on Mathspeak.

## Conistent rule and preference changes

Now items also have a `data-semantic-domain2style` that remembers the last style used when swapping styles/rules with the `>` and `<` keys. This avoids bleed through from one expression to the other, as SRE no longer maintains a global state for elements.

## ChromeVox Menu

I removed the `ChromeVox` menu entry as this only exists in English and is no longer maintained.

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,7 +167,7 @@
     "@xmldom/xmldom": "^0.8.10",
     "mhchemparser": "^4.2.1",
     "mj-context-menu": "^0.9.1",
-    "speech-rule-engine": "^5.0.0-alpha.3",
+    "speech-rule-engine": "^5.0.0-alpha.4",
     "wicked-good-xpath": "^1.3.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -167,7 +167,7 @@
     "@xmldom/xmldom": "^0.8.10",
     "mhchemparser": "^4.2.1",
     "mj-context-menu": "^0.9.1",
-    "speech-rule-engine": "^5.0.0-alpha.2",
+    "speech-rule-engine": "^5.0.0-alpha.3",
     "wicked-good-xpath": "^1.3.0"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,8 +21,8 @@ importers:
         specifier: ^0.9.1
         version: 0.9.1
       speech-rule-engine:
-        specifier: ^5.0.0-alpha.3
-        version: 5.0.0-alpha.3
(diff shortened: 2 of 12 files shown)


dpvc:
Sorry, I guess my removal of the fix/explorer-issues branch closed this PR automatically.  Didn't mean for that to happen.  It looks like this branch is a little more complicated to rebase than I thought (I'm working on it now).

zorkow:
Indeed, it looks like the only clean way to get this back is to replay the commits in this PR on top of the `update/explorer` branch. I'll do that tomorrow.

dpvc:
I've done it, and need to do some checking, and will make a new branch once I'm convinced that it is an accurate rebasing.  I did have to do several separate rebasings of subranges of the branch in order to do it.

zorkow:
Great, thanks for doing this. I did test and it has the functionality as expected. Will make a PR #1263

PR #1262, opened 2025-05-29
Fix a number of explorer and menu issues, and update to work with Orca 46

dpvc:
This PR implements some changes to the explorer that makes it work better with Orca in Linux, particularly with Orca 46.1 in Ubuntu 24.04.  It removes the need to clone the container element that was part of #1249 and #1251, and makes the process used more browser and OS independent (the only OS dependency now is the role used on the temporary speech elements).  I will close the two older PRs in favor of this one.

The details about the changes will be added below, but I wanted to get it out for testing and not wait for that.

### DIFF
--- a/ts/a11y/complexity/collapse.ts
+++ b/ts/a11y/complexity/collapse.ts
@@ -616,6 +616,7 @@ export class Collapse {
     for (const name of Object.keys(attributes)) {
       if (
         name.substring(0, 14) === 'data-semantic-' ||
+        name.substring(0, 12) === 'data-speech-' ||
         name.substring(0, 5) === 'aria-' ||
         name === 'role'
       ) {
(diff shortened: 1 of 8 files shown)


zorkow:
I can verify that it works now with Orca 46.1 as well as the latest 48 version from the ppa.

dpvc:
> I can verify that it works now with Orca 46.1 as well as the latest 48 version from the ppa.

Awesome!  Thanks for checking.

I have worked out a solution for the refocusing issues that I [mentioned yesterday](https://github.com/mathjax/MathJax-src/pull/1260#issuecomment-2920866812), and will make one more PR for that.  I think that will be the last of them, unless you find anything else problematic.

> lgtm

I will merge.  But I never wrote up the details of this PR.  Do you still want me to do that?

PR #1268, opened 2025-06-03
Fix problem with CD environment cell ending with a prime. (mathjax/MathJax#3373)

dpvc:
This PR reverts a change from #1192 that turns out to have been needed.  One case is when the cell entry ends with a prime, so the top stack item is the prime item not the table item.  There are other cases where this can occur as well.  A test for this situation is included.

Resolves issue mathjax/MathJax#3373.


zorkow (approved):
lgtm.

PR #1269, opened 2025-06-03
Fix a problem with substyles generating undefined base styles.

dpvc:
This PR fixes a problem with the `util/Styles.ts` file where the use of a sub-style could cause a parent style to be created with an undefined value.  For example, defining `border-radius: 3px` causes `border` to be created but be undefined.  That leads `Styles.cssText` to produce `border-radius: 3px; border: undefined;`.

The fix checks that the sub-style is one of the child styles of the parent style before trying to connect it to the parent (which creates the parent style).


zorkow (approved):
lgtm.

PR #1274, opened 2025-06-06
Add publishing action

zorkow:
This PR 
* adds a publishing action
* renames the test action from `test.js.yml` to `test.yml` as the `js` bit does not make any sense!

I have tested the publishing action for the build. I need an `npm` authentication token for the final step.


zorkow (commented):
I've done the changes and it runs locally with `act`. Yet to be tested is the `publish` job.

zorkow:
> Should the `.travis.yaml` file be removed with this PR? I'm not sure how that is hooked in. I think there is a setting that should be removed for the repository as well.
> 

I had forgotten to remove the `.travis.yaml` file, but that's gone now.

> I also was wondering about whether the tests need to be run here, but I guess if they are run on all the PRs, they don't have to be done here.

Tests will run for every PR to `develop` branch, regardless whether or not test relevant files have changed. I have never worked out to restrict this to say "only changes in the `ts` directory". But I am also not sure if it is worth it, because you still want  tests running when changing things like packages in `package.json` or changes in the `components` directory, etc.

PR #1275, opened 2025-06-17
Action to push to the `MathJax` repository

zorkow:
PR adds an action that pushes the new release to the `MathJax` repository, creating a branch with the new release name and making pull request into develop.

* This action could be combined with the `publish` action. For ease of testing I kept it separately.
* I have tested with `act` up to the point of actually pushing the branch as this needs a token that is automatically created on Github but cannot be simulated in `act`.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,68 @@
+NAME    : Publish Release to MathJax
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dpvc (changes requested):
A couple of suggestions for this.

One additional one is that the `package.json` version should be adjusted to match the source version, as well.

Do we still need the three non-mathjax dependencies in the MathJax repo's `package.json`?  We needed them in the past for people using the `mathjax` npm package in node applications that include speech generation.  Is that still the case?

PR #1276, opened 2025-06-17
Update build and ui/menu so its builds with new highligher

dpvc:
This PR fixes a problem with building the `ui/menu` component.  The `components/bin/build` command is modified to handle `export async function` which was causing an error during the build phase.  The `ts/a11y/sre.ts` function also was causing problems when webpacking the `ui/menu` component.  The the SRE functions for the preferences were not being found by webpack for some reason that I could not fully determine.  I don't see anything wrong in any of the files, but it seems to be connected to the direct `export` of those methods in `ts/a11y/res.ts`, so I've change those to an import followed by exports, which seems to do the trick.

In addition, I've added a generic highlighter in `ts/a11y/explorer/Highlighter.ts` so that it can work in the absense of an output jax (I have an example in the documentation of a a role-you-own output jax for native MathML that causes an error otherwise).

I also add a new named `STATE` value in `ts/handler/html/HTMLDocument.ts` just for convenience, and so that all the default render actions have a named state value.

### DIFF
--- a/components/bin/build
+++ b/components/bin/build
@@ -36,7 +36,7 @@ const INDENT = '  ';
  * The pattern to use when looking for explort commands to process
  */
 const EXPORTPATTERN =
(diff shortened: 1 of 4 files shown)


zorkow (approved):
lgtm.

PR #1278, opened 2025-06-18
Fix incorrect removal of braces in template macros

dpvc:
This PR fixes a problem with parsing the arguments of a macro with template where braces could be removed incorrectly.  For example, with `\def\text#1\stop{#1}`, if used with `\text{a}{b}\stop`, you currently will get an error about missing open braces.  That is because the parameter `#1` will end up containing `a}{b` rather than `{a}{b}` as expected.

The solution is to set `hasBraces` to false when a second braced argument is used.  I also changed `hasBraces` to a boolean rather than a number.

Two new tests are added to check for this issue.  There should probably be some more tests for when `hasBraces` gets unset, for completeness.

### DIFF
--- a/testsuite/tests/input/tex/Newcommand.test.ts
+++ b/testsuite/tests/input/tex/Newcommand.test.ts
@@ -552,6 +552,28 @@ describe('Newcommand', () => {
 
   /********************************************************************************/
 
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #1280, opened 2025-06-19
Fix problem with MathJax reference in liteDOM

dpvc:
This PR fixes a problem with the changes in b7f1942 that cause an error in node applications that use direct loading of the MathJax modules, where the `MathJax` variable is not defined.  I thought the `MathJax?.loader` would check for that, but it turns out that isn't the case, and  a more explicit check for being defined is needed.

### DIFF
--- a/ts/adaptors/lite/Element.ts
+++ b/ts/adaptors/lite/Element.ts
@@ -172,7 +172,7 @@ export class LiteIFrame extends LiteElement {
       this.options.debug,
     ];
     const pool = `${this.options.path}/speech-workerpool.js`;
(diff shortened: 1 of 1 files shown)


dpvc:
This was part of my testing to see that wicked-good-expath is not needed in node applications that use SRE, and with this fix, everything looks good to remove that dependency.

dpvc:
> the tests are complaining about the new line not being covered.

That file never had complete coverage.  We only have complete coverage for the TeX input jax and the `ts/util` folder.  Everything else still needs work.

PR #1284, opened 2025-06-21
Fix problem with sre-root

dpvc:
This PR resoves a problem with how the `worker.path` option is computed.  It needs to be a full path when used in node applications, and the `ts/components/mjs/sre-root.ts` file uses `import.meta.url` to determine it.  But webpack converts that to the full unix path of where it was compiled, and so that path was being included in the component files like `tex-chtml.js` which is not desired.  (It would get replaced when used in a web setting, but not in a node setting.)

The `ts/components/mjs/root.ts` file works similarly, but is replaced during webpacking by `components/root-pack.js`, and that needs to be done for `sre-root.js` as well.  So I've added the needed `sre-pack.js` file and a webpack rule to do the replacement.

In addition, there was still a problem with producing a full path to the webworker file when the mathjax root is set to a node packlage like `@mathjax/src/bundle`, which is the most convenient thing to do in node applications that use MathJax components.  This works OK when loading the components from source, but not when loading the bundled versions.  So I've modified `components/mjs/a11y/speech/speech.js` to work harder to resolve the path for the worker.  This now works in all the use-cases that I have tried (loading of bundled combined components, loading of combinded components from source, loading of individual bundled components, loading of individual compoments from source, direct loading of modules).  So I think this takes care of all the worker path problems.

Now that the speech is in the worker, not MathJax proper, the `components/mjs/a11y/sre/sre_config.js` file, which had been setting up the path for the JSON files, is no longer needed, so I've removed that.

Finally, I changed the version number to 4.0.0-rc.2.


zorkow (approved):
lgtm.

PR #1286, opened 2025-06-23
Improve mathmap configuration

dpvc:
This PR adds a `worker.maps` configuration option that specifies where the mathmaps are located.  This allows node applications that are using direct loading of MathJax or loading the MathJax components from source to get the correct location for the maps.  It also also gives the ability to specify an alternate location for the maps during testing and development, if needed.

The defaults are set so that even when loading from the `mjs/a11y/sre` or `cjs/a11y/sre`, the mathmaps will be taken from the `bundle/sre` directory, since they don't exist in the `[mc]js/a11y/sre` directories.  For the web, or when components are used, the `[sre]` and the new `[mathmaps]` prefixes determine these.  When loading components from source (with the `source.js` file), the `[mathmaps]` are directed to the bundle directory again.  So this works without special configuration in all the use cases that I have tried.  Also, it means that the `lab:sre` build doesn't need its `init.js` file any more, since that was doing the work of redirecting the maps, which is now done in MathJax proper.

I've also moved the setup for `SREfeature` into the `speech-worker.ts`; I hope that addresses your concerns about that.  So all the SRE stuff is in that one file now.  The adaptors that set up the workers pass the mathmaps location on to the worker (for the web, that is done in the blob script that first sets the `mathmaps` value and then loads the worker script, and for node, it is passed in the `workerData` object).

Since both the web and node versions now use a custom map loader, that is moved to the common section of `speech-worker.ts`, with a new `getMap()` function in the web/node-specific code that uses `fetch()` for the web and `require()` for node.

Finally, in the HTML adaptor, I do some extra work to make sure that `` ` `` and `\` characters are properly quoted before inserting them into the `content` script in order to avoid any problems if the paths use these characters for some reason (like Windows paths, for example).

I think this should be it, and can be used for an rc.3 release.

### DIFF
--- a/components/mjs/a11y/speech/speech.js
+++ b/components/mjs/a11y/speech/speech.js
@@ -7,18 +7,21 @@ import {SpeechHandler} from '#js/a11y/speech.js';
 
 if (MathJax.loader) {
   let path = Package.resolvePath('[sre]', false);
+  let maps = Package.resolvePath('[mathmaps]', false);
   if (hasWindow) {
     path = new URL(path, location).href;
+    maps = new URL(maps, location).href;
   } else {
     const REQUIRE = typeof require !== 'undefined' ? require : MathJax.config.loader.require;
     if (REQUIRE?.resolve) {
-      path = REQUIRE.resolve(`${path}/require.mjs`).replace(/\/[^\/]*$/, '');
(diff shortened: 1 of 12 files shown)


dpvc:
PS, the clearspeak menu is working fine after all.  Must have not re-build everything when I was doing that testing.

zorkow (approved):
lgtm.

PR #1290, opened 2025-07-02
Allow Chrome to load speech worker from CDN in a page from a file:// URL

dpvc:
This PR fixes a problem with importing the worker code from a page the was loaded from a `file://` URL with Mathjax coming from the CDN.  The problem only seems to affect Chrome, which won't allow the cross-origin load, even though the CDN marks the script as allowed.

The solution is to use `importScripts()` rather than `import()` to get the worker code.  That means the worker can't be loaded as a module, as `importScripts()` isn't allowed there.  

The `package.json` file is no longer needed, and the worker doesn't need to be a module anymore (that could have been removed when the worker-pool was removed, as that was only needed for the iframe hack to the liteDOM).

The v4-lab needs a new way to detect the need for the `sre-lab.js` file, as the `importScripts()` can't load an ES6 module, and the compiled `speech-worker.js` is such a module.  So we need to use `import()` in the lab.   The `ww-v4-lab` branch in the MathJax-dev repository has the needed changes:  the HTML adaptor is subclassed and the `createWorker()` method is overridden with one that uses `import()`, and also sets `self.isLab = true` to mark the fact that the worker is running int he lab.  The worker now uses that value to determine whether to load `sre-lab.js` or `sre.js`, so that is a bit cleaner than having to test for the failure of SRE loading.

### DIFF
--- a/components/mjs/a11y/speech/speech.js
+++ b/components/mjs/a11y/speech/speech.js
@@ -14,7 +14,7 @@ if (MathJax.loader) {
   } else {
     const REQUIRE = typeof require !== 'undefined' ? require : MathJax.config.loader.require;
     if (REQUIRE?.resolve) {
-      path = REQUIRE.resolve(`${path}/package.json`).replace(/\/[^\/]*$/, '');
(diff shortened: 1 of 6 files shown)


zorkow (approved):
lgtm.

PR #1292, opened 2025-07-02
Don't try to create the clearspeak menu if enrichment is disabled

dpvc:
This PR prevents the menu code from trying to create the clearspeak setting menu when enrichment is off (and so the webworker may not even be created yet).  When enrichment is off, the speech menu is disabled, so the clearspeak menu will not be available in that case, so it is OK to not generate it.


zorkow (approved):
lgtm.

PR #1295, opened 2025-07-05
Add explorer key actions and improve auto-voice handling

dpvc:
This PR adds the missing explorer keys and improves auto-voice handling.


dpvc:
Thanks for letting me know about the table limitations.  I've pushed a new commit that I hope addresses the situation.  See a16ca58 for the differences.

The changes take into account the fact that some tables don't have rows, so the methods that handle table cell positions now check for that.  Also, finding the table cell itself is no longer by a search, but instead uses the semantic data to move up the tree to the semantic parents.

See what you think.

zorkow (approved):
Works for all my use cases.

PR #1296, opened 2025-07-06
Backward compatibility for mathjax-full package

dpvc:
This branch implements a backward-compatibility definition for the `mathjax-full` package that could be used to make `mathjax-full` to `@mathjax/src`.  This is the idea I had mentioned to Peter in mathjax/MathJax#3390.  It works well, and would make it easier to upgrade to MathJax v4 for those already using `mathjax-full`.  But perhaps we don't want to support that, and should force everyone to change to `@mathjax/src` to use v4.  This could remain as a separate branch and not be merged, as a means of keeping this separate.  Comments?

### DIFF
--- a/mathjax-full/link-mathjax.cjs
+++ b/mathjax-full/link-mathjax.cjs
@@ -1,0 +1,6 @@
+const fs = require('fs');
+const path = require('path');
+
+const link = path.resolve(path.dirname(process.argv[1]), '@mathjax');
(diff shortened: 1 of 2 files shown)


dpvc:
In face-to-face meeting, we decided not to make this package, so I'm closing the PR.  But I'm leaving the branch for now as I may want to remember the package.json that does this.

Issue #1297, opened 2025-07-07
this match is poor


treeandgrass:
[https://github.com/mathjax/MathJax-src/blob/844ad7513005fa096a07d2468bf59c7f6a6c3a50/ts/adaptors/lite/Parser.ts#L38](https://github.com/mathjax/MathJax-src/blob/844ad7513005fa096a07d2468bf59c7f6a6c3a50/ts/adaptors/lite/Parser.ts#L38)

The regular expression matching is too weak and fails to match in many cases, such as --name='name' and ..name='name', which causes an error: "Cannot read properties of undefined (reading 'parent')". The regular expression in this code can be replaced with:

`export const ATTNAME = '[^\\s"\'<>/=]+';`

treeandgrass:
During usage, I compiled a package myself. After testing, it passed and resolved a bug in the business. I suspect others may encounter this issue as well.

dpvc:
Thanks for reporting your issue.  The LiteDOM is pretty minimal and has a number of rough edges like this.

I assume these are being used on custom tags, not actual HTML tags, as only the defined attributes are allowed on standard HTML tags.  Custom attributes should be handled through `data-*` attributes, not invalid ones like `--name` or `..name`, as those are invalid on HTML tags according to the spec.

I will say more in your PR.

dpvc:
I've made a PR for the changes.

dpvc:
Fixed in v4.0.

PR #1298, opened 2025-07-07
Expand the attribute name matching scope

treeandgrass:
The following [issue](https://github.com/mathjax/MathJax-src/issues/1297) needs to be addressed:

For example, cases like ---name='value' and __name='value' can be normally handled by browsers, but the Parser in MathJax itself cannot process such situations.

Significance: When rendering in non-browser environments, it can handle more scenarios, reducing the code error "Cannot read properties of undefined (reading 'parent')". I guess others may encounter similar issues in use, because in practical scenarios, HTML is often not perfectly standardized.

### DIFF
--- a/ts/adaptors/lite/Parser.ts
+++ b/ts/adaptors/lite/Parser.ts
@@ -35,7 +35,7 @@ import {LiteAdaptor} from '../liteAdaptor.js';
  */
 export namespace PATTERNS {
   export const TAGNAME = '[a-z][^\\s\\n>]*';
-  export const ATTNAME = '[a-z][^\\s\\n>=]*';
+  export const ATTNAME = '[^\\s"\'<>/=]+';
   export const VALUE =  `(?:'[^']*'|"[^"]*"|[^\\s\\n]+)`;
   export const VALUESPLIT =  `(?:'([^']*)'|"([^"]*)"|([^\\s\\n]+))`;
   export const SPACE = '(?:\\s|\\n)+';


treeandgrass:
If the official code is used to enhance the matching of HTML tag attributes, users won't need to compile a version themselves in such scenarios.

dpvc:
Thanks for contributing to MathJax.  We appreciate your efforts in trying to improve it.  There are several issues with your PR, however.

First, we follow the git-flow model, which means that merges to the master branch only occur for releases (or when the files changed don't alter the code, like updates to the README file).  All other PRs should be for branches that original from the develop branch and target the develop branch.  So you would need to redo this PR in order for it to be accepted.

Second, the changes you made to the pattern do not yet make it correct.  In particular, the removal of the `\n` from the pattern would allow incorrect attribute names to be accepted (ones that break across lines).

According to the [spec](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2):

> Attribute names must consist of one or more characters other than [controls](https://infra.spec.whatwg.org/#control), U+0020 SPACE, U+0022 ("), U+0027 ('), U+003E (>), U+002F (/), U+003D (=), and [noncharacters](https://infra.spec.whatwg.org/#noncharacter). In the HTML syntax, attribute names, even those for [foreign elements](https://html.spec.whatwg.org/multipage/syntax.html#foreign-elements), may be written with any mix of [ASCII lower](https://infra.spec.whatwg.org/#ascii-lower-alpha) and [ASCII upper alphas](https://infra.spec.whatwg.org/#ascii-upper-alpha).

so the pattern you use gets this incorrect (as does our original one).  In particular, you have not filtered out the control characters (with `\n` being one of them), nor the non-characters (there are quite a few of these), but you incorrectly remove `<`, which apparently is allowed, and we both ruled out U+00A0 (NON-BREAKING SPACE), which, according to this, is allowed, so `\s` should not be used.

Apparently, [custom elements](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-core-concepts) have a slightly more restrictive definition, as they need to match the [XML-compatible](https://html.spec.whatwg.org/multipage/infrastructure.html#xml-compatible) standard, which means it matches the [XML name](https://www.w3.org/TR/xml/#d0e804) standard 

> [4]   	NameStartChar	   ::=   	":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
> [4a]   	NameChar	   ::=   	[NameStartChar](https://www.w3.org/TR/xml/#NT-NameStartChar) | "-" | "." | [0-9] | #xB7 | [#x0300-#x036F] | [#x203F-#x2040]
> [5]   	Name	   ::=   	[NameStartChar](https://www.w3.org/TR/xml/#NT-NameStartChar) ([NameChar](https://www.w3.org/TR/xml/#NT-NameChar))*

and not include a colon.  Since the LiteDOM doesn't distinguish between standard tags and custom tags, it is probably best to use the XML convention instead.

Since I'm going to close this PR for the first reason above, I'll make a PR based on the XML requirements, but without a colon.

PR #1301, opened 2025-07-08
Make 'Semantic Info' menu active (not sure why it was disabled)

dpvc:
This PR enables the "Semantic Info" sub-menu in the "Explorer" menu.  I'm not sure why it was disabled before.  This fixes the attribute used for the "Prefix" selection, and adjusts the CSS for the tooltips so that the inner `div` has the same rounded corners as the outer container.


zorkow (approved):
lgtm.

PR #1302, opened 2025-07-09
Changes the TAG and ATTR patterns to be more in line with HTML5 (#1297)

dpvc:
This PR makes the parsing of tags and attributes more compliant with he HTML5 spec, which has much looser parameters for tag and attribute names.  XML parsing is more restrictive, but since the parser must work for both (it is not very sophisticated), it needs to use the less restrictive version (HTML5).

The main change is that attribute names don't have to start with a letter.  The other restrictions remove the control characters and the other disallowed characters.  See #1297 for some discussion of the specs.


zorkow (approved):
lgtm.

PR #1303, opened 2025-07-09
Make the retry error message more descriptive

dpvc:
This PR adds more detail to the "MathJax retry" error message so that programmers who use the synchronous functions and don't trap this error will get more details about what is happening.  (There is also a completely new section in the v4 documentation about this error, so I'm hoping this PR and that new page will reduce the questions about this.)

The error message is never actually used internally (just the `.retry` property), so the actual message is immaterial from the standpoint of the code.


zorkow (approved):
lgtm.

PR #1304, opened 2025-07-12
Improve loading of @mathjax/src by import and require

dpvc:
This PR updates `node-main.*` to export not just `init()` but the `MathJax` object as the default object with the `init()` function added to it.  This allows for more meaningful imports, particularly for ES6 modules.

Now you can do

``` js
import MathJax from '@mathjax/src';
await MathJax.init({...});
...
```

rather than

``` js
import {init} from '@mathjax/src';
await init({...});
```

(though both will work).  It also allows

``` js
const MathJax = require('@mathjax/src');
MathJax.init({...}).then(() => {...});
```

while still supporting

``` js
require('@mathjax/src').init({...}).then(() => {...});
```

so this is backward compatible.

This needs to work in several settings:

``` js
import MathJax from '@mathjax/src';               // the bundled mjs version
import MathJax from '@mathjax/src/source';        // the source mjs version
const MathJax = require('@mathjax/src');          // the bundled cjs version
const MathJax = require('@mathjax/src/source');   // the source cjs version
const MathJax = (await import('@mathjax/src')).default;         // the bundled mjs version;
const MathJax = (await import('@mathjax/src/source')).default;  // the source mjs version;
```

with each followed by  `await MathJax.init({...});` or `MathJax.init({...}).then(() => {...});` as appropriate.  Also,

``` js
require('@mathjax/src').init({...}).then(() => {...});
require('@mathjax/src/source').init({...}).then(() => {...});
(await import('@mathjax/src')).init({...}).then(() => {...});
(await import('@mathjax/src/source')).init({...}).then(() => {...});
```

should work.

This gives a more natural definition for the `MathJax` variable to be used later on in the application.

### DIFF
--- a/components/mjs/node-main/node-main.cjs
+++ b/components/mjs/node-main/node-main.cjs
@@ -2,4 +2,4 @@ if (!global.MathJax) global.MathJax = {};
 
 global.MathJax.__dirname = __dirname;
 
-module.exports = require('./node-main.js');
+module.exports = require('./node-main.js').MathJax;
--- a/components/mjs/node-main/node-main.js
+++ b/components/mjs/node-main/node-main.js
@@ -27,8 +27,10 @@ import '../core/core.js';
 import '../adaptors/liteDOM/liteDOM.js';
 import {source} from '../source.js';
 
-const path = eval('require("path")');          // get path from node, not webpack
-const dir = global.MathJax.config.__dirname;   // set up by node-main.mjs or node-main.cjs
+const MathJax = global.MathJax;
+
+const path = eval('require("path")');   // get path from node, not webpack
+const dir = MathJax.config.__dirname;   // set up by node-main.mjs or node-main.cjs
 
 /*
  * Set up the initial configuration
@@ -78,7 +80,7 @@ MathJax._.mathjax.mathjax.asyncLoad = function (name) {
  * The init() function returns a promise that is resolved when MathJax is loaded and ready, and that
  * is passed the MathJax global variable when it is called.
  */
-const init = (config = {}) => {
+const init = MathJax.init = (config = {}) => {
   combineConfig(MathJax.config, config);
   return Loader.load(...CONFIG.load)
     .then(() => CONFIG.ready())
@@ -87,6 +89,6 @@ const init = (config = {}) => {
 }
 
 /*
- * Export the init() function
+ * Export MathJax (with its init() function)
  */
-export {init};
+export {MathJax};
--- a/components/mjs/node-main/node-main.mjs
+++ b/components/mjs/node-main/node-main.mjs
@@ -1,2 +1,4 @@
 import './node-main-setup.mjs';
-export * from './node-main.js';
+import {MathJax} from './node-main.js';
+export default MathJax;
+export const init = MathJax.init;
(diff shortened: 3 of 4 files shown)


zorkow:
I can't get the `import` to work. Somehow node still tries to load `cjs` files. 
This is the error I get after packaging `@mathjax/src` and installing it in `/tmp/`:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'loader')
    at 680 (/tmp/node_modules/.pnpm/@mathjax+src@file+mathjax-src-4.0.0-rc.4.tgz/node_modules/@mathjax/src/bundle/node-main.js:1:141204)
    at __webpack_require__ (/tmp/node_modules/.pnpm/@mathjax+src@file+mathjax-src-4.0.0-rc.4.tgz/node_modules/@mathjax/src/bundle/node-main.js:1:154895)
    at /tmp/node_modules/.pnpm/@mathjax+src@file+mathjax-src-4.0.0-rc.4.tgz/node_modules/@mathjax/src/bundle/node-main.js:1:155521
    at /tmp/node_modules/.pnpm/@mathjax+src@file+mathjax-src-4.0.0-rc.4.tgz/node_modules/@mathjax/src/bundle/node-main.js:1:158014
    at Object.<anonymous> (/tmp/node_modules/.pnpm/@mathjax+src@file+mathjax-src-4.0.0-rc.4.tgz/node_modules/@mathjax/src/bundle/node-main.js:1:158062)
    at Module._compile (node:internal/modules/cjs/loader:1734:14)
    at Object..js (node:internal/modules/cjs/loader:1899:10)
    at Module.load (node:internal/modules/cjs/loader:1469:32)
```

dpvc:
Hmmm..  It works for me.  Can you be more explicit about what you are trying?

I did the following from the MathJax-src directory:

``` shell
git checkout update/node-main
pnpm -s build-all
pnpm pack
mkdir /tmp/mjx
cd /tmp/mjx
pnpm install path-to-MathJax-src/mathjax-src-4.0.0-rc.4.tgz
pnpm install @mathjax/mathjax-newcm-font
cat > test.mjs
import MathJax from '@mathjax/src';

await MathJax.init({
  loader: {
    load: ['input/tex', 'output/chtml']
  }
})

console.log(MathJax.startup.adaptor.outerHTML(MathJax.tex2chtml('x')));
^D
node test.mjs
```

and got

```
<mjx-container class="MathJax" jax="CHTML" overflow="overflow" display="true"><mjx-math data-latex="x" display="true" style="margin-left: 0; margin-right: 0;" class=" NCM-N"><mjx-mi data-latex="x"><mjx-c class="mjx-c1D465">𝑥</mjx-c></mjx-mi></mjx-math></mjx-container>
```

as expected.  It also worked with `@mathjax/src/source`.  I didn't try the `require()` versions.

zorkow:
It turns out we cannot bind to `MathJax`, but any other constant name.

dpvc:
> It turns out we cannot bind to `MathJax`, but any other constant name.

That only applies to *interactive* use of MathJax within `node`.  If you use a file (either `.cjs` or `.mjs`), you can use `MathJax`.  That is because the interactive `node` is working in the global namespace, and so there is a conflict when you do `const MathJax = (await import('@mathjax/src'))` or `let MathJax = require('@mathjax/src')`.  But if you do `MathJax = (await import('@mathjax/src'))` or `MathJax = require('@mathjax/src')` without the `const` or `let`, then it is OK.

PR #1307, opened 2025-07-14
Move toEntry to be a method so that it can be override, if necessary

dpvc:
This PR moves the `toEntry()` call to a method so that it can be overridden if subclassed (I need the in one of the examples in the demos-node update).


zorkow (changes requested):
Could you then also remove the function in line 29/30? Or is that the one you need to have in the demos?

dpvc:
It could be removed if the definition is moved to the `toEntry()` method instead.  But it seemed like a potentially useful function, so I left it exported.  But I suppose it could be moved to `ts/util/string.ts` instead.  I'll make another commit to do that.

zorkow (approved):
lgtm.

PR #1308, opened 2025-07-14
Place the hidden svg/defs elements at the beginning of the document, not the end.

dpvc:
This PR changes where the `MathDocument` puts additional elements into the page.  Currently, only the SVG output has any of these, and they are the `svg` element used to store the global font cache (in a `defs` element).  This should be at the top of the document, not the bottom, so that the definitions are in place before the elements that use them (important when the document is preprocessed on the server).


zorkow (approved):
lgtm.

PR #1312, opened 2025-07-17
Update CSS used to clip stretchy characters in CHTML. (mathjax/MathJax#3032)

dpvc:
The CSS for stretchy characters was not correct (it used the `mjx-c` clip-path, which includes some slop on all sides, and that means the stretches aren't clipped well).  This uses proper widths for horizontal stretches and proper heights for vertical ones.  It may be that we need to put a little slop in the stretch direction, but I haven't seen it needed yet.

Resolves issue mathjax/MathJax#3032.


zorkow (approved):
lgtm.

PR #1313, opened 2025-07-18
Don't alter the typeset option lists in case they are reused

dpvc:
This PR changes the handling of the `options` object in the typesetting and conversion commands created by the Startup module so that they use a copy of the options rather than the original object.  This prevents the changes we make to `options` from being made to the original object passed to the functions (in case they are used again later by the calling program).


zorkow (approved):
lgtm!.

PR #1314, opened 2025-07-18
Add configuration options for pre- and post-filters

dpvc:
This PR adds the ability to have pre- and post-filters for the input and output jax be included in their configuration objects, rather than having to use the `startup.ready()` function to add them.  It also adds a pre-filter to the output jax, for symmetry (and because it may be easier to do filtering there than to have to add a post-filter onto each of the input jax if there are multiple input formats).

To facilitate this, a new `addList()` method is added to the `FunctionList` object, and for good measure this is used in the TeX input jax to add its filters.  You can also now pass a function list to the `FunctionList` constructor.  A new test is added to the `FunctionList` tests to cover the new code.

### DIFF
--- a/ts/core/OutputJax.ts
+++ b/ts/core/OutputJax.ts
@@ -47,12 +47,12 @@ export interface OutputJax<N, T, D> {
   options: OptionList;
 
   /**
-   * Lists of pre-filters to call after typesetting the math
+   * List of pre-filters to call after typesetting the math
    */
   preFilters: FunctionList;
 
   /**
-   * Lists of post-filters to call before typesetting the math
+   * List of post-filters to call before typesetting the math
    */
   postFilters: FunctionList;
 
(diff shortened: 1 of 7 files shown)


ts/core/OutputJax.ts
```diff
@@ -47,7 +47,12 @@ export interface OutputJax<N, T, D> {
   options: OptionList;
 
   /**
-   * Lists of post-filters to call after typesetting the math
+   * Lists of pre-filters to call after typesetting the math
```
zorkow, line 50: This is a single list not a list of lists?

ts/core/OutputJax.ts
```diff
@@ -47,7 +47,12 @@ export interface OutputJax<N, T, D> {
...
+   */
+  preFilters: FunctionList;
+
+  /**
+   * Lists of post-filters to call before typesetting the math
```
zorkow, line 55: ditto.

zorkow (approved):
A query on the comments. O/w lgtm.

PR #1317, opened 2025-07-22
Fix addClass() and similar to not add extra blank in class

dpvc:
This PR fixes a problem where the LiteDOM incorrectly adds a space at the beginning of the `className` of LiteElements.  That is because when the `className` is empty, the split produces `['']`, not an empty list, so when more classes are added or removed, and the joined with a space, there is an empty name at the beginning, giving a space at the front of the list.

This also uses `indexOf()` and `includes()` rather than `find()` and `findIndex()`, as these should be more efficient.


zorkow (approved):
lgtm.

PR #1318, opened 2025-07-26
Have the explorer look harder for the associated anchor node.  (mathjax/MathJax#3398)

dpvc:
In the key explorer, when enter is pressed on a node with a link, we currently send a click event to the parent node, which is expected to be the anchor taking you to the linked file.  It is possible, however, that the parent may be some other node that the CHTML output needs for the layout, and so the click goes to the wrong node. 

This PR has the explorer look higher up until it finds the associated anchor rather than assume it is the parent element.

Resolves issue mathjax/MathJax#3398.


zorkow (approved):
lgtm.

PR #1319, opened 2025-07-26
Handle mml spacing for mo with no parent (like surd for root).

dpvc:
Currently, when `mathmlSpacing` is used, an error can be thrown for an `mo` that doesn't have a parent node.  Such elements occur when special characters are needed, like the surd for a square root.  Changing original line 1079 to use `(!n)` rather than `n !== 0` would prevent the error message, but in that case the default spacing for `mo` elements would give the surd spacing on the left and right, which throws off the display.  Instead, we move the test for no parent to before when the MathML spacing is added to avoid both the error message and the wrong display.


zorkow (approved):
lgtm.

PR #1322, opened 2025-07-30
Update `replaceUnicode()` to handle two `\U{}` calls in a row.

dpvc:
This PR fixes an issue with the string utility's `replaceUnicode()` function that prevented it from processing two adjacent `\U{}` calls, so `replaceUnicode('\U{A0}\U{A0}')` would only replace the first `\U`, not the second.  That was due to the requirement of having a non-`\` character preceding the `\U` which blocked it from matching the second `\U`, as that character can't be part of a previous replacement.  That requirement was part of an attempt to avoid escaped backslashes like `\\U`.  But, in addition to the problem it caused for replacing the `\U{}`, the escaped backslashes were never converted to single backslashes.  The new code now properly does that as well.

The tests are adjusted to accommodate the translation of `\\` to `\`, and to add a pair of adjacent `\U`.


zorkow (approved):
lgtm.

PR #1323, opened 2025-07-30
Fix `registerTex()` to handle `tex.packages` that is already an array

dpvc:
This PR fixes the component utility that marks preloaded TeX packages so that it will properly merge the new packages into the existing package list if the old list was an array (rather than `{'[+]': [...]}`).


zorkow (approved):
lgtm.

PR #1324, opened 2025-07-30
Don't increase the spacing size for MathML spacing if it is 0

dpvc:
This PR fixes a problem with `mathmlSpacing` when an `mo` element has a default spacing of 0, a with `<mo>&#x2062;</mo>`.  In the past, this would get extra spacing that it should not have.  Now, if the default spacing is 0, we take away the extra `+ 1` that is being added to get the proper number of 1/18th ems from the operator dictionary.


zorkow (approved):
lgtm.

PR #1325, opened 2025-08-01
Fix navigation issues when semantic tree is not aligned with syntactic tree

zorkow:
The PR fixes a number of issues for the explorer, when the semantic structure is not aligned with the syntactic tree. As example consider the following equation with the `empheq` package:

```latex
\begin{empheq}[left=\empheqlbrace, right=\empheqrbrace]{align} 
  E&=mc^2 \\ 
  Y&= \sum_{n=1}^\infty \frac{1}{n^2} 
\end{empheq}
```

In the current version the left brace is a "child" of the first row. Conversely, the right brace has the first row as "parent". The issue is that semantic children do not necessarily have to live in the DOM tree rooted in their semantic parent. Likewise, the semantic root is not necessarily the first speech element in the DOM. We now compute:

* `rootNode` by using the root id given in the `data-semantic-structure` element
* `firstChild` by iterating throw the `data-semantic-owns` list until we find a node which has `data-speech-node`
* `parentNode` (in `moveUp`) by looking for the actual parent id in the tree.

In addition the `nextSibling` and `prevSibling` methods are fixed to actually ignore elements that are not speech nodes. As an example consider the `mc^2` above in MathSpeak, where the elided times is now ignored.

### DIFF
--- a/ts/a11y/explorer/KeyExplorer.ts
+++ b/ts/a11y/explorer/KeyExplorer.ts
@@ -582,7 +582,7 @@ export class SpeechExplorer
       this.FocusOut(null);
     } else {
       this.stopEvent(event);
-      this.refocus = this.firstNode(this.node);
+      this.refocus = this.rootNode();
       this.Start();
     }
   }
@@ -649,7 +649,7 @@ export class SpeechExplorer
    * Select top-level of expression
    */
   protected homeKey() {
-    this.setCurrent(this.firstNode(this.node));
+    this.setCurrent(this.rootNode());
   }
 
   /**
@@ -673,7 +673,7 @@ export class SpeechExplorer
   protected moveUp(shift: boolean): boolean | void {
     return shift
       ? this.moveToNeighborCell(-1, 0)
-      : this.moveTo(this.current.parentElement.closest(nav));
+      : this.moveTo(this.getParent(this.current));
   }
 
   /**
@@ -760,7 +760,7 @@ export class SpeechExplorer
   protected prevMark() {
     if (this.currentMark < 0) {
       if (this.marks.length === 0) {
-        this.setCurrent(this.lastMark || this.firstNode(this.node));
+        this.setCurrent(this.lastMark || this.rootNode());
         return;
       }
       this.currentMark = this.marks.length - 1;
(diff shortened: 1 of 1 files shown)


zorkow:
As an aside, here is the "native" rendering of the above equation

$$
\begin{empheq}[left=\empheqlbrace, right=\empheqrbrace]{align} 
  E&=mc^2 \\ 
  Y&= \sum_{n=1}^\infty \frac{1}{n^2} 
\end{empheq}
$$

Github clearly still uses MathJax to convert to MathML, as the result is still exactly our structure including `data-mjx` attributes. But the rendering simply looks atrocious. What a terrible implementation.

dpvc:
> What a terrible implementation.

Yes, and because they are now using the browser's native MathML that means they are relying on MathML-Core, so things like equation numbers won't work:

$$E=mc^2\tag{1}$$

 and the various table attributes aren't supported, so you get things like:

``` math
\begin{align}
a+b &= c + d\\
e &= f + g + h
\end{align}
```

which has two much space before the equal signs, and (depending on the browser), the left-hand side can be misaligned.

The worst, however, is that since MathML-Core doesn't allow `mathvariant` attributes (except for `<mi mathvariant="normal">` for a single-letter content to prevent it from being italic), things like `\mathbb{A}` will be $\mathbb{A}$, which in some browsers will be just $A$.

dpvc:
The math experience on GitHub is now very much browser and font dependent, which is a shame.  It has been a pretty problematic implementation from the very beginning.

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -1293,13 +1293,48 @@ export class SpeechExplorer
...
+      return node.querySelector(nav) as HTMLElement;
+    }
+    const ownsList = owns.split(/ /);
+    for (const id of ownsList) {
+      const node = this.getNode(id);
+      if (node?.hasAttribute('data-speech-node')) {
+        return node;
+      }
+    }
+    return null;
```
dpvc, line 1314: Should this return `node.querySelector(nav)` rather than `null`?  I'm not sure.

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -1311,10 +1346,16 @@ export class SpeechExplorer
...
+      .getAttribute('data-semantic-owns')
+      ?.split(/ /);
     if (!owns) return null;
-    const i = owns.indexOf(this.nodeId(node));
-    return this.getNode(owns[i + 1]);
+    let i = owns.indexOf(this.nodeId(node));
+    let next = this.getNode(owns[++i]);
+    while (next && !next.hasAttribute('data-speech-node')) {
+      next = this.getNode(owns[++i]);
+    }
```
dpvc, line 1357: Could this be

``` js
let next;
do {
  next = this.getNode(owns[++i]);
} while (next && !next.hasAttribute('data-speech-node'));
```

to avoid some redundancy?

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -1326,10 +1367,16 @@ export class SpeechExplorer
...
+      .getAttribute('data-semantic-owns')
+      ?.split(/ /);
     if (!owns) return null;
-    const i = owns.indexOf(this.nodeId(node));
-    return this.getNode(owns[i - 1]);
+    let i = owns.indexOf(this.nodeId(node));
+    let prev = this.getNode(owns[--i]);
+    while (prev && !prev.hasAttribute('data-speech-node')) {
+      prev = this.getNode(owns[--i]);
+    }
```
dpvc, line 1378: Similarly here?

dpvc (approved):
This looks great.  Thanks for fixing these issues.  I left a possible alternative loop structure to reduce redundancy, and one question, but have approved as is.

ts/a11y/explorer/KeyExplorer.ts
zorkow: Done.

ts/a11y/explorer/KeyExplorer.ts
zorkow: Done

ts/a11y/explorer/KeyExplorer.ts
zorkow: That works, too.

PR #1326, opened 2025-08-01
Include the menu loading promises in the default startup page ready promise chain. (mathjax/MathJax#3401)

dpvc:
This PR adds the menu's `loadingPromise` into the promise chain for the initial typeset (and therefore the `MathJax.startup.promise`).  This makes sure that everything is loaded before MathJax's initial typesetting occurs.  This should only affect those loading components by hand rather than through a combined component.

Resolves issue mathjax/MathJax#3401.


zorkow (approved):
lgtm.

PR #1328, opened 2025-08-02
Move the node createWorker() method to the NodeMixin.

dpvc:
This PR moves the node-based `createWorker()` function from the `liteAdaptor` into the generic `NodeMixin` that is used by the `liteAdaptor` as well as the `jsdomAdaptor` and the `linkedomAdaptor`, so that they can support speech generation in node as well.

The code is moved unchanged, but the `liteBase` class gets a stub for `createWorker()` that is overridden in the mix-in since one is required for the `AbstractDOMAdaptor` class that `liteBase` creates before using the mixin.

### DIFF
--- a/ts/adaptors/NodeMixin.ts
+++ b/ts/adaptors/NodeMixin.ts
@@ -23,8 +23,18 @@
  * @author dpvc@mathjax.org (Davide Cervone)
  */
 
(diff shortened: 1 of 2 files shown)


zorkow (approved):
lgtm.

PR #1334, opened 2025-08-08
Update key explorer to use dialog element for help dialog. (mathjax/MathJax#3405)

dpvc:
This PR allows the KeyExplorer to use a `<dialog>` element when the browser supports that, and uses the current method when that isn't available.  In the latter case, we still create a `dialog` element, but it acts like a custom element in that case, since it isn't implemented directly.

This resolve the `z-index` issue reported in mathjax/MathJax#3405.

### DIFF
--- a/ts/a11y/explorer/KeyExplorer.ts
+++ b/ts/a11y/explorer/KeyExplorer.ts
@@ -933,18 +933,23 @@ export class SpeechExplorer
    * Displays the help dialog.
    */
   protected help() {
+    const isDialog = !!window.HTMLDialogElement;
     const adaptor = this.document.adaptor;
-    const helpBackground = adaptor.node('mjx-help-background');
+    const helpBackground = isDialog ? null : adaptor.node('mjx-help-background');
     const close = (event: Event) => {
-      helpBackground.remove();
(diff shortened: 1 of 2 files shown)


dpvc:
Just for the record, the reason for the fallback is that `dialog` support is fairly recent in some browsers (2023), so we may not want to rely on that.

ts/a11y/explorer/KeyExplorer.ts
```diff
@@ -933,40 +933,59 @@ export class SpeechExplorer
...
+    helpDialog.addEventListener('keydown', (event: KeyboardEvent) => {
       if (event.code === 'Escape') {
         close(event);
       }
     });
-    const [title, select] = helpData.get(context.os);
-    (help.childNodes[1] as HTMLElement).innerHTML = helpMessage(title, select);
-    document.body.append(helpBackground);
-    help.focus();
+    if (isDialog) {
```
zorkow, line 977: Can we add a comment, that we can simplify this in two years.

zorkow (approved):
One comment. O/w lgtm.

dpvc:
Closing in favor of #1357.

PR #1337, opened 2025-08-16
Add  `fontExtensions` configuration option. (mathjax/MathJax#3410)

dpvc:
This PR adds a `fontExtensions` configuration that makes it easier to specify a font extension to load (e.g., `mathjax-euler`).  It sets up the `extraLoads` package data to make sure the extension loads after the font itself.  The current method of including the extension in the `loader.load` array doesn't work properly as it can be loaded before the font arrives.  That can be handled by setting up dependencies and paths in the loader configuration, but it is harder than it should be, so this makes it easy.

Resolves issue mathjax/MathJax#3410.

The original section that was used to configure a font has been moved to a `configFont()` function, and a new `configExtensions()` function uses it to set up the font extensions.

### DIFF
--- a/components/mjs/output/util.js
+++ b/components/mjs/output/util.js
@@ -2,9 +2,29 @@ import {combineDefaults, combineWithMathJax} from '#js/components/global.js';
 import {Package} from '#js/components/package.js';
 import {hasWindow} from '#js/util/context.js';
 
-export const FONTPATH = hasWindow ?
-                        'https://cdn.jsdelivr.net/npm/@mathjax/%%FONT%%-font':
-                        '@mathjax/%%FONT%%-font';
+export const FONTPATH = hasWindow
+  ? 'https://cdn.jsdelivr.net/npm/@mathjax/%%FONT%%-font'
+  : '@mathjax/%%FONT%%-font';
+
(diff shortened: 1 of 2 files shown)


dpvc:
I added a commit that includes support for generic font extensions that have dynamic font ranges, like `mathjax-euler`.  It turns out that they reference the `generic` font, which was only meant as a placeholder for the actual font, but that wasn't being replaced.   This commit fixes that, but unfortunately, it requires a static property on the output jax because the dynamic files don't have access to the output jax instance.  So the `typeset()` method sets up that static value, so that any dynamic loads during the typesetting will go to the right font.  We do it here in case there are several output jax instances with different fonts.  We also set it during the output jax constructor, in case the dynamic files are loaded up front (like some of the documentation examples illustrate).

dpvc:
I just realized there is a potential problem, here.  If there is a retry during the `toDOM()` call (for loading dynamic font ranges), then the `genericFont` will not be reset.  So I need to made another commit to handle that.

dpvc:
I've added the error checking.  Do you want to re-review?

zorkow (approved):
lgtm.

PR #1339, opened 2025-08-29
Use node to call makeAll rather than assuming hash-bang works

dpvc:
This PR uses `node` to call `components/bin/makeAll` in the `package.json` scripts rather than assuming the file is executable.


zorkow (approved):
lgtm.

PR #1341, opened 2025-09-01
Don't add invisible funciton apply after an empty mathop

dpvc:
This PR prevents the invisible function apply from being added after an empty `\mathop`.   See Speech-Rule-Engine/speech-rule-engine#817.  I include two new tests for this situation.

### DIFF
--- a/testsuite/tests/input/tex/Base.test.ts
+++ b/testsuite/tests/input/tex/Base.test.ts
@@ -12198,6 +12198,33 @@ describe('Character Class Changes', () => {
 
   /********************************************************************************/
 
+  it('Mathop Apply', () => {
+    toXmlMatch(
+      tex2mml('\\mathop{F} x'),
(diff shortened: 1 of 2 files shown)


pkra:
As mentioned at https://github.com/Speech-Rule-Engine/speech-rule-engine/issues/817#issuecomment-3852245914 I'm still seeing the problem when using mathjax v4.1.0 and sre v5-beta.3.

dpvc:
Ok, because `\mathop{\mathrm{}}` actually has an `mrow` (from the `\mathrm{}`) in the mathop, it ends up not being counted as empty.  I will need to make the check be more careful.

pkra:
Thanks, Davide! (It's obviously bad code, actually due to a bug on the AMS end, but I still thought I should mention it.)

dpvc:
I've made a new PR to resolve the issue.

PR #1345, opened 2025-09-04
Feature/tag texts

dpvc:
This PR changes the handling of tags to make three separate `mtext` nodes: one for the left delimiter, one for the tag number, and one for the right delimiter.

The `formatTag()` and `formatRef()` functions now take either a string or a triple of strings to use for the tag format.  If it is just a string, then parentheses, brackets, and braces are removed from the beginning and ending of the thing to make the triple, otherwise, the string is used as is (and only a single `mtext` will be created).

The tests are updated to take these changes into account.

### DIFF
--- a/ts/input/tex/Tags.ts
+++ b/ts/input/tex/Tags.ts
@@ -571,7 +571,7 @@ export class AbstractTags implements Tags {
       ? format
       : format.match(/^(\(|\[|\{)(.*)(\}|\]|\))$/)?.slice(1) || [format];
     const mml = new TexParser(
-      tag.map((part) => `\\text{${part}}`).join(''),
+      tag.map((part) => part ? `\\text{${part}}` : '').join(''),
       {},
       this.configuration
     ).mml();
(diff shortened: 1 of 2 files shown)


dpvc:
> Is there a reason why we would not want that?

That would be a potential breaking change, which we aren't supposed to make in minor versions.  So this is for backward compatibility, mainly with the `tagformat` extension, which currently has the tag format being a string not an array of strings.  This allows either one. 

I suppose the breaking up of the string could be moved to the `tagformat` extension rather than here, if you like that better.  But it will need to be done somewhere, so I put it at the lowest level so that in case there are other subclasses of `Tag` that are in use, they will still work.  Probably not much chance of that, however, so moving the complications into `tagformat` is probably sufficient.  Would you prefer that?

zorkow:
> > Is there a reason why we would not want that?
> 
> That would be a potential breaking change, which we aren't supposed to make in minor versions. So this is for backward compatibility, mainly with the `tagformat` extension, which currently has the tag format being a string not an array of strings. This allows either one.
> 

I had not thought about the potential breaking change. Makes sense.

PR #1350, opened 2025-09-10
Change CSS for help icon so that when scroll or truncate is used, it is not clipped.  (mathjax/MathJax#3432)

dpvc:


zorkow:
> * puts the icon next to a displayed expression when there isn't an equation tag (rather than at the far right as currently).

I find this too close to the expression; see below. Can we somehow have it on the edge of the top-level highlighter box?

image

dpvc:
This is the same positioning use for inline expressions.  The overlap is intentional to help link the info icon to the expression itself.  I suppose it could be moved to the right for display expressions, but right now it is consistent with the inline placement.

PR #1351, opened 2025-09-10
Provide hasOwn polyfill configuration option.

dpvc:
This PR provides the option if installing a polypill for the `Object.hasOwn()` method, since that seems to be the only significant thing that we are using that isn't in ES2017 or earlier.  The default is false, since this modifies global objects, and the page author should ned to opt-in for that.


dpvc:
OK, I added the comments.

PR #1352, opened 2025-09-11
Update Operator Dictionary from MathML 4 spec (mathjax/MathJax#3363)

dpvc:
This PR updates the operator dictionary using the one from the MathML4 specification.  The MathML spacing values are different from the ones in the MathML3 specification that was used to create our operator table originally, but these only affect the output when MathML spacing is requested.  They are also used to define the TeX class in cases where the TeX class isn't explicitly set, and that only changed the values in some obscure instances (like geometric shapes where we used a fairly coarse approach to setting the class in the past).  The characters for which there are macros (like `\blacktriangle`) have their classes set by the TeX input jax, but the change in the operator dictionary may mean they get wrapped in a TeXAtom that overrides the operator dictionary value where it didn't have to before.  There are only a few of these.

The MathML4 dictionary no longer include `accent` or `fence` attributes, and since we use these, I needed to add those back in, so uses some heuristics to do so.

the main difference is that the `|` operator is not stretchy when used in infix form, while it used to be stretchy in MathML3.  This won't affect the TeX output, as the TeX jax sets `stretchy="true"` automatically when needed, but it can affect MathML input.  That is,  something like

``` xml
  <mi>x</mi>
  <mo>+</mo>
  <mo>|</mo>
  <mfrac>
    <mi>a</mi>
    <mi>b</mi>
  </mfrac>
  <mo>|</mo>
  <mo>+</mo>
  <mi>y</mi>
</math>
```

will no longer have stretchy `|` around the fraction.  This corresponds to the current MathML-Core output, and MathML4's expected output.  On the other hand,

``` xml
  <mi>x</mi>
  <mo>+</mo>
  <mrow>
    <mo>|</mo>
    <mfrac>
      <mi>a</mi>
      <mi>b</mi>
    </mfrac>
    <mo>|</mo>
  </mrow>
  <mo>+</mo>
  <mi>y</mi>
</math>
```

will have stretchy `|` as they are in prefix and postfix form.  Since SRE adds the `mrow` even to the MathML, it should not be much of an issue, even for those using MathML input, unless they disable the semantic enrichment.

Other changes include making many more character stretchy and some symmetric, and reclassifying some BIN and REL operators.  It is a little hard to tell these changes, as the MathML spacing value changes meant that the predefined operator definitions in the `MO` object had to change, and so it appears that there are lots of changes when most are not significant.  Ignoring spacing differences helps, but still isn't great.  I can provide some better diffs if you want to see them.

These changes required updates to 120 tests (mostly adding or removing `stretchy` attributes).  I couldn't remember how to get your emacs-based diff tools to work, so I ended up writing some scripts to do the updates for me.  They need a little more work to be generally applicable, but I'm hoping they will help with these changes in the future.

Resolves issue mathjax/MathJax#3363.

### DIFF
--- a/testsuite/tests/input/tex/Ams.test.ts
+++ b/testsuite/tests/input/tex/Ams.test.ts
@@ -4396,9 +4396,7 @@ describe('Ams symbols', () => {
       `<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\blacktriangleleft\\blacktriangleright\\therefore\\because\\eqsim\\vartriangle\\Join\\nless\\ngtr\\nleq\\ngeq" display="block">
          <mo data-latex="\\blacktriangleleft">&#x25C2;</mo>
          <mo data-latex="\\blacktriangleright">&#x25B8;</mo>
-         <mo data-latex="\\therefore">&#x2234;</mo>
-         <mo data-latex="\\because">&#x2235;</mo>
-         <mo data-latex="\\eqsim">&#x2242;</mo>
+         <mo data-mjx-texclass="REL" data-latex="\\therefore\\because\\eqsim">&#x2234;&#x2235;&#x2242;</mo>
          <mo data-mjx-alternate="1" data-latex="\\vartriangle">&#x25B3;</mo>
          <mo data-mjx-texclass="REL" data-latex="\\Join\\nless\\ngtr\\nleq\\ngeq">&#x22C8;&#x226E;&#x226F;&#x2270;&#x2271;</mo>
        </math>`


dpvc:
I added a commit that updates some of the TeX classes (in particular, the class for a period, which is incorrectly set as PUNC but should be ORD, as in actual TeX).  The changes to the tests indicated to me that there was an unneeded TeXAtom node being used around some character that don't need them.  This was coming from the `fixStretchy` filter, and is left over from when we didn't always have a TeX class assigned, if I recall correctly.  That is no longer the case, so there is no need for that TeXAtom wrapper.  I have updated the tests to accommodate the change.

dpvc:
I realized that I used the wrong issue number with this.  It should be mathjax/MathJax#3363.  I changed the title, but not the branch name, to reflect the correct issue.

PR #1354, opened 2025-09-16
Implement a method of enclosing multiple elements when there are extra ones selected.

dpvc:
This PR implements a method for enclosing multiple elements when extra ones are included rather than having multiple outlines and backgrounds.

This works by collecting the split nodes into groups by which line they are in (when there are line breaks) and enclosing those collections separately.  This is done by inserting a new, temporary node that represents the bounding box of all the elements in each collection.  Most of the work is done in the highlighter through the new method `encloseNodes()`, which does the separation into collections by line, computes the bounding box in each line, and creates and inserts the new node for that bounding box.  

How that bounding box node is created and inserted differs in the CHTML and SVG highlighters, and that is handled in the new `createEnclosure()` method.  For HTML elements, and new `mjx-bbox` element is created with the proper size and offsets, and is prepended to the `mjx-container`.  For SVG, a rect is created and inserted before the first selected element in each line.  The creation of the SVG rectangles is abstracted into a `createRectangle()` method that is used for both the enclosure rectangle as well as the individual background rectangles for the highlighter.  If an enclosing rectangle is created, it is used for highlighting as well, so no need to create a highlight rectangle as well.

Selected nodes that are enclosed get marked with a `data-mjx-enclosed` attribute and their backgrounds are not colored and they don't get the `mjx-selected` attribute (since the enclosure will get the background and outline).

The unhighlight methods are modified to remove any added enclosures when they become unselected.

The CSS for the CHTML outline is moved from the explorer.ts to the chtml.ts file, to correspond to the CSS in the svg.ts file, though I suppose the CSS from svg.ts could be moved to explorer.ts instead, if that seems better.  I also added 'data-' to the `sre-highlight` attributes, since they are not valid on SVG elements.

### DIFF
--- a/ts/a11y/explorer/Highlighter.ts
+++ b/ts/a11y/explorer/Highlighter.ts
@@ -467,21 +467,28 @@ class SvgHighlighter extends AbstractHighlighter {
    */
   public unhighlightNode(info: Highlight) {
     const node = info.node;
-    const previous = node.previousSibling as HTMLElement;
     if (node.hasAttribute('data-sre-highlighter-bbox')) {
       node.remove();
       return;
     }
-    node.removeAttribute('data-mjx-enclosed');
-    if (previous && previous.hasAttribute('data-sre-highlighter-added')) {
-      info.foreground
-        ? node.setAttribute('fill', info.foreground)
-        : node.removeAttribute('fill');
-      previous.remove();
+    if (node.tagName === 'svg' || node.tagName === 'MJX-CONTAINER') {
+      if (!node.hasAttribute('data-mjx-enclosed')) {
+        node.style.backgroundColor = info.background;
+      }
+      node.removeAttribute('data-mjx-enclosed');
+      node.style.color = info.foreground;
       return;
     }
-    node.style.backgroundColor = info.background;
-    node.style.color = info.foreground;
+    const previous = node.previousSibling as HTMLElement;
+    if (previous?.hasAttribute('data-sre-highlighter-added')) {
+      previous.remove();
+    }
+    node.removeAttribute('data-mjx-enclosed');
+    if (info.foreground) {
+      node.setAttribute('fill', info.foreground);
+    } else {
+      node.removeAttribute('fill');
+    }
   }
 
   /**
--- a/ts/a11y/explorer/KeyExplorer.ts
+++ b/ts/a11y/explorer/KeyExplorer.ts
@@ -1023,7 +1023,7 @@ export class SpeechExplorer
     this.current = node;
     this.currentMark = -1;
     if (this.current) {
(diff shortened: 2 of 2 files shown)


dpvc:
Oh, I also added a filter to the `subtree()` method to remove `null` entries, since there currently are ids in the structure that have no corresponding DOM nodes in the tree.  If that gets fixed, the filter should be removed.

zorkow:
> Oh, I also added a filter to the `subtree()` method to remove `null` entries, since there currently are ids in the structure that have no corresponding DOM nodes in the tree. If that gets fixed, the filter should be removed.

I thought that does not happen anymore. Do you have an example expression for this?

dpvc:
> Do you have an example expression for this?

Well, `x_0^1` has `data-semantic-structure="(4 (3 0 1) 2)"`, but there is no node with `data-semantic-id="3"`, and that ends up leaving `3` in the `rest` array, leading to `null` after the `this.getChild(child)` call when the whole expression is selected.

dpvc:
That is on the current develop branch of SRE.

ts/a11y/explorer/Highlighter.ts
```diff
@@ -99,6 +99,15 @@ export interface Highlighter {
    */
   unhighlightAll(): void;
 
+  /**
+   * Encloses multiple nodes if they in the same line
```
zorkow:
```suggestion
   * Encloses multiple nodes if they are in the same line
```

ts/a11y/explorer/Highlighter.ts
```diff
@@ -378,16 +466,96 @@ class SvgHighlighter extends AbstractHighlighter {
...
+      node.remove();
+      return;
+    }
+    node.removeAttribute('data-mjx-enclosed');
+    if (previous && previous.hasAttribute('data-sre-highlighter-added')) {
       info.foreground
-        ? info.node.setAttribute('fill', info.foreground)
-        : info.node.removeAttribute('fill');
-      info.node.parentNode.removeChild(previous);
+        ? node.setAttribute('fill', info.foreground)
```
zorkow: Here we are resetting only on node to the original foreground color. However, that should be all the nodes in the enclosed element.

zorkow:
PS: Not sure if this would be a blocker for 4.0.1

dpvc:
> The solution is probably to retain the `parts` in the `info` structure.

That is not necessary.  Each part already has its own `info` structure, and `unhighlightNode()` is being called for all of them, but the resetting of the foreground color was being skipped for the nodes when there are multiple enclosed nodes.  I'm fixing that in a new commit.

zorkow (approved):
lgtm.

PR #1360, opened 2025-09-27
Allow digits in \operatorname (mathjax/MathJax#2991)

GWDx:
Extend operatornamePattern to include `0-9`, so operators like `\operatorname{atan2}` are parsed as a single identifier.

Fixes mathjax/MathJax#2991

<br>

The bug first appeared in commit 2cd7f7cc, and was fixed in 7545df4c, which was later reverted in c8834c61.


dpvc:
Sorry to take so long to respond to this PR.  There is actually a different issue at play as well, and I didn't want to merge this until I had straightened that out.  That is done in PR #1367.  That PR does take care of the display issue, but I think it is fine to include digits here anyway.  So I'm merged this now.

PR #1361, opened 2025-10-01
Fix path handling to work with Windows and add tests for it

dpvc:
This PR makes it possible to run MathJax via node in Windows (without needing to use a unix subsystem).  The main issue is, as usual, the use of `\` rather than `/` in Windows path names, so we introduce a new function in the `context` object that converts Windows (absolute) paths to unix-style paths when needed, and use this whenever `__dirname` or `import.meta.url` or similar values are used.  That way, the management of paths can use `/` as it currently does.

In addition, the `context` setup for `context.os` now handles the situation within node as well (it used to always return `undefined`).  This is done using the `process.platform` value to make to the generic `Unix`, `Windows`, and `MacOS` values.  I wasn't sure how to treat `cygwin`, but have decided to set it to `Windows`, since it should not hurt to do the path adjustment there, and my understanding is that cygwin can use Windows paths, so it might be useful to do the translation anyway.

The `context` tests are adjusted for the new values, and I've added several new files to handle testing Windows and to get full coverage in the tests.

This resolves issue mathjax/MathJax#3439.

### DIFF
--- a/components/bin/build
+++ b/components/bin/build
@@ -193,7 +193,7 @@ function processParts(parts) {
 function processLines(file, objects) {
   if (objects.length === 0) return [];
   const base = path.dirname(file).replace(/^\.$/, '');
-  const dir = (PREFIX ? noramlize(path.join(PREFIX, base)) : base);
+  const dir = (PREFIX ? normalize(path.join(PREFIX, base)) : base);
   const dots = dir.replace(/[^\/]+/g, '..') || '.';
   const relative = normalize(path.join(dots, '..', JS, dir, path.basename(file))).replace(/\.ts$/, '.js');
   const name = path.parse(file).name;


dpvc:
I have added a commit that updates the build tools to work in Windows as well.  As always, the main issue is the `\` vs. `/` problem.  This is most important in `build`, where the file names in the `import` command in the `lib` directory would otherwise have `\` in them, which node will interpret as string escapes, so that something like `..\util\Styles.js` would try to use `\u` as indicating a unicode character and choke in the `til` that comes after it.

You also need to tell Windows to use a shell for `execSync()` and `spawn()`.  In most cases, `shell: true` will work, but in `makeAll`, because the arguments are in `'...'`, that is a problem for Windows, because the standard shell uses `"..."` to quote arguments, and `'...'` are take to be part of the argument.  I could have switched to `"..."`, but that would mean more unix steel characters would have to be quoted, and I didn't want to have to do that, and didn't want to have to deal with potentially different shell escapes for Windows.  Instead, I specify the bash shell explicitly for Windows.  I think that should be OK, though it relies on `git` being installed.

Since our package scripts expect to be run in bash, a Windows shell user will need to specify

``` shell
npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"
```

before running our build scripts.

Finally, I changed the package scripts to use a literal ESC character, so that it works in Windows as well as unix.

dpvc:
I should mention that this was [discussed](https://groups.google.com/g/mathjax-users/c/OSw268S4eU0) in the MathJax User's forum, where the original poster has confirmed (after a bit of back and forth) that the changes worked for him.

dpvc:
> When running pnpm build
> ...
> 'log' is not recognized as an internal or external command,
operable program or batch file.

Did you set the `script-shell` configuration as mentioned [above](https://github.com/mathjax/MathJax-src/pull/1361#issuecomment-3365732048)?  This is due to the fact that a number of our scripts, like the logging script, rely on the shell being `bash`, not the default windows command shell.  You need to do

``` shell
npm config set script-shell "C:\\Program Files\\git\\bin\\bash.exe"
```

or something similar (depending on where your copy of `bash` is).  Before this will work. 

I have run it successfully in my Windows VM, and the original poster has run it after making the needed patches, so I'm pretty sure this works.

zorkow:
> Did you set the `script-shell` configuration as mentioned [above](https://github.com/mathjax/MathJax-src/pull/1361#issuecomment-3365732048)? 

Yes, I did. Still did not work. Maybe there is something wrong with the path. I will check again.

zorkow:
> > Did you set the `script-shell` configuration as mentioned [above](https://github.com/mathjax/MathJax-src/pull/1361#issuecomment-3365732048)?
> 
> Yes, I did. Still did not work. Maybe there is something wrong with the path. I will check again.

OK, I got it to work:
I ran in powershell as administrator (as I had done before).
I set the path with a capitalized `G` for `git`. I thought that should not make difference in Windows?
I also needed to change the execution policy:
```
Set-ExecutionPolicy Unrestricted
```
I think it was `RemoteSigned` before.

dpvc:
I've added more details to the build instructions for Windows users.

PR #1368, opened 2025-10-16
Allow cleanAttributes to work on nodes with no attributes (e.g., XML nodes)

dpvc:
This PR allows the TeX input jax's `cleanAttributes()` filter to work with nodes that don't have attributes (like XML nodes).  Currently, it will crash with an error about not being able to call `get()` on an undefined value.


zorkow (approved):
lgtm.

PR #1373, opened 2025-10-24
Add a fonts loader path and have fonts use that for accessing their data

dpvc:
This PR creates a `fonts` loader path and has the font components use that to access their data rather than using a direct reference to jesdelivr.net.  That way, if a site wants to host their own fonts, it is easy to redirect to the site location rather than having to give a path for every font (e.g., the mhchem, bbm, dsfont, etc. fonts).  This also puts the reference to jesdelivr in only one place, which is helpful.


zorkow (approved):
lgtm.

PR #1377, opened 2025-10-26
Add missing preLoaded() call for input/mml when a11y is used.

dpvc:
The `semantic-enrich` extension has `input/mml` as a dependency (so that it can re-parse the SRE output), but it wasn't being listed in the preloaded items when the A11y utilities were being included in a combined component.  That means that loading `tex-chtml.js`, for example, and adding `input/mml` to the loaded items could cause the extension to be loaded a second time without sharing the common code, leading to problems with objects and options not being properly shared.


zorkow (approved):
lgtm.

PR #1378, opened 2025-10-28
Update tests to improve coverage of some modules

dpvc:
Thus PR adds some tests to get more coverage of some of the recent changes.

It also adjusts `BaseItems.ts` to remove some unneeded `?`, as the TeXAtom gets an inferred mrow child automatically when created, so (unless someone deletes them explicitly), they will always be available.


zorkow (approved):
lgtm.

PR #1387, opened 2025-11-10
Remove use of `*` in CSS declarations.  (mathjax/MathJax#3463)

dpvc:
This PR removes the usage of `*` in CHTML's CSS selectors.  Apparently, that is a performance issue in some cases, so these were made specific to the types of nodes that can be contained in the parent node.  Some of the more complex ones are constructed from parts.

Resolves issue mathjax/MathJax#3463.


dpvc:
> I did not know that `*` was a performance issue.

I didn't either.  Luckily, it was straight-forward to fix.

PR #1388, opened 2025-11-16
Make assistive-mml unset speech/braille and vice-versa

dpvc:
This PR adjusts the menu items for speech/Braille and hidden MathML so that selecting hidden MathML will turn off speech and Braille, and vice versa.  That makes is a little easier to switch the type of assistive support to be used.  (It really doesn't make sense to have both speech/braille and assistive MathML on at once in any case.)


dpvc:
I've added the function you requested (I had meant to do something like that and forgot).  I also added some text to the help dialog indicating the new menu behavior.

PR #1389, opened 2025-11-16
Add localstorage flag needed for node v25

dpvc:
Node v25 makes the `--localstorage-file` a requirement for apps that use `localStorage`, so I've added it here.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,7 +5,7 @@ on:
     types: [published]
 
 jobs:
-  setup:
+  publish:
     runs-on: ubuntu-latest
     name: Build and publish MathJax
     permissions:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
       - develop
 
 jobs:
-  setup:
+  testing:
     runs-on: ubuntu-latest
     name: Compile and test MathJax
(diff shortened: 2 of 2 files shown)


dpvc:
Well, it looks like our workflow doesn't allow this flag.  I'm not sure what node version is being used, but I guess it is note v25.  So we can ignore this PR for now, but may need to make it in the future.

zorkow:
> Well, it looks like our workflow doesn't allow this flag. I'm not sure what node version is being used, but I guess it is note v25. So we can ignore this PR for now, but may need to make it in the future.

According to [blog post](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) node24 should replace node20 as default as of this autumn. Start of winter is Dec 21, so there is still a bit of time...
Anyway, in the meantime we can force the use of node24 explicitly.

I will push a fix to this branch.

dpvc:
Thanks for taking care of this.

zorkow:
@dpvc tests are running now. I think this PR can be merged.

* `test.yml` now uses explicitly `node24`. We could remove this again in the near future or keep to have finer control over our test environment.
* In addition, I renamed the jobs in `test.yml` and `publish.yml` to something more meaningful than `setup` and also to be able to distinguish them to be able to select them when running locally with `act`.

PR #1390, opened 2025-11-16
Add hidden option to retain regions and highlighting, as requested by Volker

dpvc:
This adds an option that prevents the highlighting and speech/Braille regions from being removed when the expression loses focus, as requested by Volker.  You can use the developer's console to enter

``` js
MathJax.startup.document.options.keepRegions = true;
```

to set it.  I will see about adding a checkbox to the lab to make that easier.


zorkow (approved):
Thanks, very helpful!

PR #1420, opened 2026-01-21
Start moving AsciiMath to native

drlippman:
This is by no means a complete PR.  I'm just making it to share the progress I made in case it might be useful, and to see whether it would be worth me spending any more time on this.  

This is an attempt to start rewriting the AsciiMath input jax using MathJax 4 native methods, rather than needing the legacy shim.  I've verified the MathML generated by this code (via `asciimath2mml`) match the asciimath repo unit tests for expected output.  This is obviously not super important or a priority, since the existing legacy method is working. 

Some things that still need work:
- Implementing/fixing the OptionList.  I used AI to help with the wrapper code, and it added a bunch of the internal AsciiMath options to the OptionList, but I don't think all are needed and I don't think the ones that do make sense to implement are actually used.
- Checking spacing. I did not reimplement any of the legacy spacing mapping (OPTABLE), as it seems like MathJax was handling that automatically.
- Handling mathvariant.  AsciiMath originally used code mappings for `\mathbb` and other font change commands.  It's not clear to me what the correct way to handle this is now, as it seems like MathJax has built-in mappings for these, but they didn't seem to work.
- Unit tests.  I didn't implement unit tests in testsuite yet, partly because I couldn't make sense of how to do it.


drlippman:
Made another commit to get the options working for decimalsign and displaystyle, and added a new one for additionalSymbols to add symbols in a less hacky way.

drlippman:
And another change implementing alternate handling of font changes.  Original asciimath used <mstyle> tag with mathvariant for font changes, with some limited character mapping. MathML has deprecated using mathvariant and prefers Unicode. The MathJax output processor was handling some of that automatically, but this change builds the mapping of characters to their Unicode values into the Asciimath parser, and doesn't use mstyle or mathvariant at all.

dpvc:
David, thanks for putting the effort into making this PR.  Moving Asciimath to a proper v3/v4 input jax has been at the back of my mind for a while, so I'm glad you have made this start.  I have only had the chance to look briefly at the code, but it looks pretty good.  I will have to look deeper, of course.

I'm wondering about the relationship of this to the main AsciiMath codebase, and how it is to be kept in sync.  one option you might consider is making a npm package that exports the needed parser and tables so that other javascript code (including MathJax) could call it.  Then all the maintenance would be on the AsciiMath side, and MathJax could just be a consumer of that.  Since you have already converted the code to Typescript for MathJax, you might consider using that for AsciiMath as well.  What would allow you to compile to ES5 or ES6, for example (or both).  

In the past, keeping MathJax's AsciiMath input jax up-to-date with the AsciiMath repository meant copying changes from your repository into the slightly modified copy in MathJax, and that was fragile and not always done because of that.  it wold be great if we could work out a mechanism were that wasn't necessary.  We have an attachment like that with the mhchem parser, for example, which is not maintained by us, but we just link to the npm package version.

drlippman:
I'm not sure how to do what you're suggesting, but I'm open to the idea :)

drlippman:
This took me an embarrassingly long time to figure out and debug, but I think I got an npm-based version of the asciimath parser put together than can be both built in the asciimath repo into a asciimath-to-mathml converter like the original script, or can be used here in MathJax.

One thing I did get stuck on that you can probably tell me how to do quickly:  Even though the parser returns the content wrapped in `<mstyle displaystyle="true">`, MathJax is displaying things in non-displaystyle form.  I tried setting the second argument to `setInheritedAttributes` to `true`, but that didn't seem to make a difference.

drlippman:
Another thing I'd need your input on if we were going to proceed with this:  I've got the typescript asciimath parser mapping bold alphanumeric characters to their Unicode codepoints.  But to implement something equivalent to the TeX `\boldsymbol{}`, non-alphanumeric characters and brackets should be bold as well.  While not part of MathML Core, should that be handled using `mathvariant="bold"` on the individual tags?  If so, should that be in combination with the Unicode mapping, or should the Unicode mapping be left to MathJax to execute?

dpvc:
I'm sorry to have left you go without an answer for so long.  I haven't had the chance to look carefully through your code, but it looks like you were able to work out the npm-based idea that I suggested.  Thanks for your work on this.  I do hope to get to look more closely soon!

> While not part of MathML Core, should that be handled using `mathvariant="bold"` on the individual tags?

That is what MathJax currently does, so I would say yes.  Note that AsciiMath's MathML is only used internally, so it doesn't have to conform to MathML-Core, as MathJax's internal form is MathML3.

> If so, should that be in combination with the Unicode mapping, or should the Unicode mapping be left to MathJax to execute?

Again, since MathJax's output is not MathML, but either HTML or SVG, the MathML-Core requirement to use the Math Alphanumeric characters is not necessary.  MathJax does use those characters in its normal operation, and will convert `<mi mathvariant="bold">A</mi>` to use the character at U+1D400 in one of its fonts.  So you don't have to convert the output yourself.  I have been following the discussion in your repository, and it seems that you have done that already, which is fine.  MathJax will work either way.

Note that MathJax will also treat `<mi mathvariant="bold">&#x1D400;</mi>` the same as `<mi>&#x1D400;</mi>`. Because the Math Alphanumeric characters are already styled (bold, italic, etc.), the `mathvariant` attribute is ignored for those characters.  My recollection is that the MathML spec indicated that was how it should be handled, though I'm not able to find it now, so I may be remembering it from somewhere else.

drlippman:
Thanks for that info. Sounds like what I have set up should work for that.

There was an earlier question I still haven't figured out:  Even though the parser returns the content wrapped in `<mstyle displaystyle="true">`, MathJax is displaying things in non-displaystyle form.

drlippman:
Nevermind on my last question. It appears the `<mstyle>` was actually the problem. Removing it and using `setInheritedAttributes` seems to have done the trick.

PR #1424, opened 2026-01-26
Add support for delimiters that are specific to a mathvariant and use it for mhchem arrows.  (mathjax/MathJax#3486)

dpvc:
This PR adds support for fonts and font extensions to define delimiters that are specific to a mathvariant.  This is needed for the `mhchem` extension to handle some of its stretchy arrows without needing to use the PUA for them, which causes problems for speech generation.

Resolves part of issue mathjax/MathJax#3486.

-----

## Details

This PR touches a large number of files, but many are because the `VariantData` type now is dependent on the `DelimiterData` type, so has an additional template variable.  That means all the Wrapper definitions need to have that template variable added, and that is a large number of files.

The `mhchem` extension now uses the standard arrow positions (e.g., U+2192) rather than PUA positions, so the test file for `mhchem` needed to be updated.

The `xArrow` method in `AmsMethods.ts` file is modified to accept a `mathvariant` in which the arrow is to be found.  

That is then used in the `mhchem` extension to specify that the arrows are to be taken from the `-mhchem` variant.  The arrows are now in the usual Unicode positions rather than the PUA.  The left-right harpoons use the newly proposed U+1F8D2 and U+1F8D3 positions.

The CHTML font data now stores the variant as part of the delimiter usage data (so that it can create the proper CSS), and uses the variant in the calls that get delimiters and stretchy variants.  The `mo` wrapper records the variant when a stretchy delimiter is used.  Similarly, the SVG mo wrapper passes the variant when it stretches a delimiter.

The main changes are on the `common/FontData.ts` file.  Here, the `VariantData` now can include delimiters specific to the variant, and the `DelimiterData` can include a variant specification.  A new `VDelimiterMap` type is used to specify delimiters for variants, and the `FontExtensionData` is extended to include that, which is now processed by the static and instance `addExtension()` functions.

The `FontData` object itself gets a new `defaultVariantDelimiters` object to hold any variant-specific delimiters, and these are now processed in the `constructor()` function.  The `createVariant()` method now creates a `variant.delims` object similar to the `variant.chars` one that has a prototype chain that links delimiters in the same way the characters are links to fallback variants (so, for example, `bold` has `normal` as a fallback).  The `normal` delimiters use the original `this.delimiters` so that object is still the final fallback for all variants.  The `variant.linked` value is now an array consisting of the linked chars and delims objects rather than just the chars.  The `defineChars()` method now takes this array into account.

The `defineDelimiters()` method now takes a variant name, and sets the `delim.v` value for variants other than `normal`.  It now assigns the delimiters to the variants delimiter list (which will be `this.delimiters` for `normal` variant), and processes linked variants in the same way that `variant.chars` are handled.

The `getDelimiter()` method now takes an optional variant name as an argument and looks up the delimiter from the variant.  Again, note that the `normal` variant will have `this.delimiters` as its delimiters, and all variants have the normal variants as the final fallback.  So in most cases, the variant-specific list will be empty and will fall back on the normal list (`this.delimiters`), and this will operate exactly as before.  The only change will be when there are variant-specific delimiters that will be found first.

The `getSizeVariant()`, `getStretchVariant()`, and `getStretchVariants()` methods all now take an optional variant name and look up the delimiter from that variant.


dpvc:
This may want to wait for v4.2?

Also, to test it, you will need to get new versions of the fonts and font-tools repositories and rebuild the `mathjax-mhchem` font using the `variant-delims` branches of each.  I am just putting this in place now.

dpvc:
I've rebased this onto the current `develop` branch, so it is ready to review.

PR #1438, opened 2026-02-27
Fix erroneous definition for --mjx-bg1-color.

dpvc:
This PR fixes an incorrect CSS setting for the `--mjx-bg1-color`.  It is usually replaced when MathJax initializes itself, but the initial value should really be made correct anyway.


zorkow (approved):
lgtm.

PR #1440, opened 2026-02-27
Fix problem with \vert becoming stretchy after enrichment. (mathjax/MathJax#3530)

dpvc:
This PR fixes a problem with `\vert` losing its `stretchy="false"` attribute when in infix form (where this is the default), but after semantic enrichment, being in pre- or postfix position (where it is not).  This is similar to #1426, which handled explicit `|` characters, but missed handling `\vert`.  The tests were updated to account for the difference.

An alternative to this (and #1426) would be to adjust the filter that removes redundant attributes to leave explicit `stretchy` attributes when the `mo` content is `|`.  If you prefer that, I will redo this PR.

Resolves issue mathjax/MathJax#3530


zorkow (approved):
lgtm.

PR #1441, opened 2026-02-27
Fix several issues with multi-character stretchy assemblies. (mathjax/MathJax#3528, mathjax/MathJax#3531)

dpvc:
This PR fixes several issues with multi-character stretchy assemblies.  There were two main problems:

1. In Windows with CHTML, the extenders could show gaps due to the overlap between pieces not being big enough.
2. If an assembly height is requested that is less than the sum of the top and bottom pieces, they need to overlap, but that wasn't happening.

The first is fixed in the `ts/output/chtml/FontData.ts` file by reducing the `line-height` by a small fuzz factor.  The SVG output was using this, and the value for that (and a horizontal fuzz factor) have been moved to the common FontData.ts file so they can be used by both output formats.  

The positioning of the extender is also fixed so that the overlap with the upper glyph is more consistent.

The horizontal extenders in CHTML now uses the `HFUZZ` value to add a little overlap as well, since glyphs with rounded ends were producing some gaps.

The second issue is addressed in the `ts/output/chtml/Wrappers/mo.ts` file.  It now uses the named `VFUZZ` and `HFUZZ` values, but the main change is in the `createPart()` method, which checks if the requested height is less than the sum of the top and bottom parts (`W < 0`) in a vertical assembly (`nl` is not null), and if so, it overrides the top and bottom margins, which are what reserve the space for the top and bottom glyphs, to make the total height be correct.  The `createAssembly()` function is adjusted to add the needed heights to the data that is passed to `createPart()`

The `ts/output/svg/Wrapers/mo.ts` file is modified to remove the `prettier-ignore` comment, and allowing the function to be prettified.  That is really the only change, other than moving the definitions of the fuzz values to the common `FontData.ts` file.

Resolves issues mathjax/MathJax#3528 and mathjax/MathJax#3531.

### DIFF
--- a/ts/output/svg/Wrappers/mo.ts
+++ b/ts/output/svg/Wrappers/mo.ts
@@ -266,7 +266,7 @@ export const SvgMo = (function <N, T, D>(): SvgMoClass<N, T, D> {
     /**
      * @param {number} n         The number of the character to look up
      * @param {string} variant   The variant for the character to look up
-     * @returns {SvgCharData}     The full CharData object, with CharOptions guaranteed to be defined
+     * @returns {SvgCharData}    The full CharData object, with CharOptions guaranteed to be defined
      */
     protected getChar(n: number, variant: string): SvgCharData {
       const char = this.font.getChar(variant, n) || [0, 0, 0, null];
@@ -284,7 +284,7 @@ export const SvgMo = (function <N, T, D>(): SvgMoClass<N, T, D> {
      * @param {number} x         The x position of the glyph
      * @param {number} y         The y position of the glyph
      * @param {N} parent         The container for the glyph
-     * @returns {number}          The width of the character placed
+     * @returns {number}         The width of the character placed
      */
     protected addGlyph(
       n: number,
@@ -312,7 +312,7 @@ export const SvgMo = (function <N, T, D>(): SvgMoClass<N, T, D> {
      * @param {string} v    The variant for the top glyph
      * @param {number} H    The height of the stretched delimiter
      * @param {number} W    The width of the stretched delimiter
-     * @returns {number}     The total height of the top glyph
+     * @returns {number}    The total height of the top glyph
      */
     protected addTop(n: number, v: string, H: number, W: number): number {
       if (!n) return 0;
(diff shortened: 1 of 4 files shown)


ts/output/svg/Wrappers/mo.ts
```diff
@@ -333,27 +330,40 @@ export const SvgMo = (function <N, T, D>(): SvgMoClass<N, T, D> {
...
+    protected addExtV(
+      n: number,
+      v: string,
+      H: number,
+      D: number,
+      T: number,
+      B: number,
+      W: number
+    ) {
       if (!n) return;
```
zorkow, line 342: Add `// prettier-ignore-start` here

ts/output/svg/Wrappers/mo.ts
```diff
@@ -333,27 +330,40 @@ export const SvgMo = (function <N, T, D>(): SvgMoClass<N, T, D> {
...
-      const Y = H + D - T - B;                 // The height of the extender
-      const s = 1.5 * Y / (h + d);             // Scale height by 1.5 to avoid bad ends
-                                               //   (glyphs with rounded or anti-aliased ends don't stretch well,
-                                               //    so this makes for sharper ends)
-      const y = (s * (h - d) - Y) / 2;         // The bottom point to clip the extender
+      const Y = H + D - T - B; //                // The height of the extender
+      const s = (1.5 * Y) / (h + d); //          // Scale height by 1.5 to avoid bad ends
+      //                                         //   (glyphs with rounded or anti-aliased ends don't stretch well,
+      //                                         //    so this makes for sharper ends)
+      const y = (s * (h - d) - Y) / 2; //        // The bottom point to clip the extender
```
zorkow: Add `// prettier-ignore-end` here.

ts/output/svg/Wrappers/mo.ts
```diff
@@ -333,27 +330,40 @@ export const SvgMo = (function <N, T, D>(): SvgMoClass<N, T, D> {
      * @param {number} B    The height of the bottom glyph in the delimiter
      * @param {number} W    The width of the stretched delimiter
      */
-    /* prettier-ignore */
```
zorkow, line 336: ` prettier-ignore` here was probably a bit overkill at the time. But I don't like the double use of `//` either.
We should just use `prettier-ignore-start|end` instead. See below.

zorkow (approved):
There is one formatting issue wrt. the comment alignment.
Otherwise fine.

PR #1442, opened 2026-03-12
Tests/rewrite to snapshots

zorkow:
Moves all tests that use our custom `toXmlMatch` method to use Jest's snapshot matching. The only exceptions are the tests in `ConfigMacros`. I need to think a bit about those.
Other tests that could probably be done with snapshot matching are those using toXmlArrayMatch. 
Both are for a follow-up PR.


dpvc (approved):
This looks great.  Too bad we didn't know about this earlier!

Now that the tests are so much simpler to look at, the extra comment separators between tests probably aren't needed any more.  Those were useful to help see where one test stopped and the next started when there were large MathML sections, but since those are no longer in the files, the separators are probably not needed.  What do you think?

PR #1443, opened 2026-03-13
moves thests in ConfigMacros to use snapshots

zorkow:
This PR removes the remaining occurrences of `toXmlMatches` from the `ConfigMacros.test.ts` file and handles those using snapshots.


dpvc (approved):
Looks good.

PR #1444, opened 2026-03-13
Tests/rewrite array matches

zorkow:
This PR removes all occurrences of `toXmlArrayMatch` from the `Tag` and `Tex` tests.

This now allows us to get rid of all helper functions related to XML matching, which was done via conversion to JSON. This has the advantage that
1. we can get rid of the dependency of the `xml-js` library,
2. tests run faster as we skip the `xml2json` conversion entirely.

This completes the refactoring of the tests to use snapshots instead of customised XML comparison.


testsuite/tests/input/tex/Tag.test.ts
```diff
@@ -604,61 +603,15 @@ describe('Page References', () => {
...
-        `<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\eqref{eq1}">
-           <mrow href="#mjx-eqn%3Aeq1" class="MathJax_ref" data-latex="\\eqref{eq1}">
-             <mtext>(1)</mtext>
-           </mrow>
-         </math>`
-      ]
-    );
+    expect(
+      await page2mml('<p>$$a=b\\label{eq1}\\tag{1}$$</p><p>Refer to \\eqref{eq1}</p>')).
+      toMatchSnapshot();
```
dpvc: I'm wondering if the `).` from the end of line 607 shouldn't be moved to the beginning of 608.

Similarly for the others.  Some I think can be put on a single line, like 177+178.

dpvc (approved):
Looks good.  Slight reformatting might be nice, but not crucial.

testsuite/tests/input/tex/Tag.test.ts
zorkow: I ran prettier on the `Tag.test.ts` file. 
As we are now using snapshots we should include the test files in our prettier/linter routine.

PR #1446, opened 2026-03-23
removes separators between single tests

zorkow:
PR removes all separators between tests. 
It leaves those that separate tests from setup or tear down functions etc.  Those were previously double separator lines.


dpvc (approved):
This looks good.

PR #1447, opened 2026-03-23
Introduce linting and formatting on testsuite

zorkow:
PR introduces formatting and linting on the testsuite. In particular, it 
* adapts the configuration to include the `testsuite` (details below)
* reformats the code according to `prettier`
* does all the automatic linting (mainly `let` to `const`)
* adds some of the missing jsdoc in the test sources
* manually fixes a number of linting issues (see [last commit](https://github.com/mathjax/MathJax-src/commit/eb6740e540edbcbc432592c87540ee7d15770f1c) for details)

There are still four lint errors left of the type `no-useless-escape`. These seem to be LaTeX errors, which have an effect on the actual test output. I'll take care of those in a separate PR.

## Configuration and Setup Changes

‎* Introduces `pnpm-workspace.yaml` file so we can refer to scripts in the workspace root from internal packages.
* Adds linting and formatting scripts to the `package.json` in `testsuite` that calls the respective workspace scripts.
* Relaxes file restrictions in `eslint.config.mjs`
* Adds a second `eslint` configuration to switch off jsdoc rules for `*.test.ts` files. This unfortunately leads to a reformatting. But that is all just whitspace.


testsuite/src/setupTex.ts
```diff
@@ -118,34 +133,42 @@ export const throwCompileErrors = {
...
+ * @param {Function} fn The function used to typeset the TeX (tex2mml, typeset2mml, etc)
+ * @returns {any} The test output
  */
 export function expectTexError(
   tex: string,
   display: boolean = true,
-  fn: (((tex: string, display?: boolean) => any) | ((tex: string) => any)) = tex2mml,
+  fn:
+    | ((tex: string, display?: boolean) => any)
+    | ((tex: string) => any) = tex2mml
```
dpvc, line 153: Kind of weird to have the first `|` with nothing on its left, but if it works, I guess it is to line up the two options.

testsuite/tests/util/LinkedList.test.ts
```diff
@@ -74,7 +84,8 @@ describe('LinkedList functionality', () => {
     const list = new LinkedList(0, 1, 2, 3, 4, 5, 6);
     let j = 0;
     for (const item of list) {
-      if (item === j) j++; else break;
+      if (item === j) j++;
+      else break;
```
dpvc: These probably want braces, no?

testsuite/tests/util/LinkedList.test.ts
```diff
@@ -83,7 +94,8 @@ describe('LinkedList functionality', () => {
     const list = new LinkedList(0, 1, 2, 3, 4, 5, 6);
     let j = 6;
     for (const item of list.reversed()) {
-      if (item === j) j--; else break;
+      if (item === j) j--;
+      else break;
```
dpvc: Same here.

dpvc (commented):
Looks good.  Two comments about a strange use of `|`.

As for the four lint errors, they are all actual errors.  

The first one in `Base.test.ts` should have one of the `\` removed (should only be `\\end` not `\\\end`), and the second one needs an extra `\` (so that it is `\\cdot`).

The `\$` in `Bussproofs.test.ts` should just be `$`.

The `\\\end` should be `\\end` again in `Newcommand.test.ts`.

dpvc (approved):

testsuite/tests/util/LinkedList.test.ts
zorkow: Indeed. I thought that would be exactly the stuff the linter would flag. Odd.

testsuite/src/setupTex.ts
zorkow: Apparently that is what passes for pretty.

testsuite/tests/util/LinkedList.test.ts
zorkow: Resolved both by negating the condition: `if ... break`  
This way we keep the same number of lines.

zorkow:
Also needed to push a new version of `pnpm-lock.yaml` after running `pnpm install`. The old version made the tests fail.

PR #1449, opened 2026-03-23
Update font-size macros to correct LaTeX values and make fontsizev3 package for backward compatibility

dpvc:
This PR fixes several incorrect sizes for the font-sizing commands (`\tiny` and `\Tiny` were switched, and the larger sizes were all off by one).  It also adds `\SMALL` and `\Small`, which were missing, and `\footnotesize`, since several people have asked for that.  Finally, since the largest size got shifted down one size, I've added `\HUGE`, which seems that it should have been there, but isn't, at least in the `amsart` class that is the basis for MathJax's definitions.  It does seem to be defined in several other packages, however.

Resolves issue mathjax/MathJax#3280.


zorkow (approved):
lgtm.

PR #1451, opened 2026-04-02
Update/sre xmldom

zorkow:
Updates 
* xmldom to v0.9.9, due to security vulnerabilities
* sre to v5.0.0-beta.7 which itself has the xmldom update


dpvc (approved):
Yes, I saw the dependably alert the other day.  Go for it.

PR #1455, opened 2026-04-05
Properly handle trailing spaces in \def. (mathjax/MathJax#3550)

dpvc:
This PR fixes a problem with the `\def` command where trailing spaces are treated incorrectly as part of a template macro definition.  E.g.

``` latex
\def\x#1  {[#1]}
```

would include the space as part of a template, so `$$\x{y}$$` would fail (it would have to be `$$\x{y} $$`.  The trailing spaces should only be included in a template if there are other template entries like

``` latex
\def\x#1,#2 {[#1,#2]}
```

or 

``` latex
\def\x#1 #2 {[#1,#2]}
```

Resolves issue mathjax/MathJax#3550.


zorkow (commented):
lgtm.

zorkow (commented):
lgtm.

PR #1456, opened 2026-04-07
Properly handle margin-left and margin-right specified on the input elements.

dpvc:
This PR fixes the output jax to take `margin` css into account when positioning elements and computing their widths.  The code already handles `padding` and `border` CSS, so this adds similar support for `margin`.  This is needed in order to support the next PR to work around a bug in Safari that affects bussproof layouts.

### DIFF
--- a/ts/output/common/Wrapper.ts
+++ b/ts/output/common/Wrapper.ts
@@ -104,6 +104,7 @@ export const SPACE: StringMap = {
  * Padding and border data from the style attribute
  */
 export type StyleData = {
+  margin: [number, number, number, number];
   padding: [number, number, number, number];
   border: {
     width: [number, number, number, number];
@@ -596,9 +597,10 @@ export class CommonWrapper<
     if (!this.styleData) return bbox;
     const padding = this.styleData.padding;
     const border = this.styleData.border?.width || [0, 0, 0, 0];
+    const margin = this.styleData.margin || [0, 0, 0, 0];
     const obox = bbox.copy();
     for (const [, i, side] of BBox.boxSides) {
-      (obox as any)[side] += padding[i] + border[i];
+      (obox as any)[side] += padding[i] + border[i] + margin[i];
     }
     return obox;
   }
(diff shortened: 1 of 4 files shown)


ts/output/common/Wrapper.ts
```diff
@@ -748,7 +750,9 @@ export class CommonWrapper<
     if (!this.styleData) return;
     const border = this.styleData.border;
     const padding = this.styleData.padding;
-    bbox.w += (border?.width?.[3] || 0) + (padding?.[3] || 0);
+    const margin = this.styleData.margin;
+    bbox.w +=
+      (border?.width?.[3] || 0) + (padding?.[3] || 0) + (margin?.[3] || 0);
```
zorkow: This code is used repeatedly, just with a different index. Could you put this into a separate method, that is just called with the index?

ts/output/common/Wrapper.ts
```diff
@@ -758,8 +762,11 @@ export class CommonWrapper<
     if (!this.styleData) return;
     const border = this.styleData.border;
     const padding = this.styleData.padding;
-    bbox.h += (border?.width?.[0] || 0) + (padding?.[0] || 0);
-    bbox.d += (border?.width?.[2] || 0) + (padding?.[2] || 0);
+    const margin = this.styleData.margin;
+    bbox.h +=
+      (border?.width?.[0] || 0) + (padding?.[0] || 0) + (margin?.[0] || 0);
+    bbox.d +=
+      (border?.width?.[2] || 0) + (padding?.[2] || 0) + (margin?.[2] || 0);
```
zorkow: See above.

ts/output/common/Wrapper.ts
```diff
@@ -769,7 +776,9 @@ export class CommonWrapper<
     if (!this.styleData) return;
     const border = this.styleData.border;
     const padding = this.styleData.padding;
-    bbox.w += (border?.width?.[1] || 0) + (padding?.[1] || 0);
+    const margin = this.styleData.margin;
+    bbox.w +=
+      (border?.width?.[1] || 0) + (padding?.[1] || 0) + (margin?.[1] || 0);
```
zorkow: And here.

zorkow (changes requested):
I'd suggest adding a dedicated method fro combining `border, padding, margin` for the various `bbox` dimentions.

dpvc:
I've added the requested method.  Can you re-review?

ts/output/common/Wrapper.ts
dpvc: Good idea.  I've added it.

zorkow (approved):
lgtm.

PR #1457, opened 2026-04-07
Work around Safari bug with negative margins used in bussproofs.  (mathjax/MathJax#3547)

dpvc:
This PR works around a bug in Safari that causes the negative spacing at the right-hand side of bussproof layouts to be ignored when computing the widths of the parent element.  This does away with the `mspace` elements at the left and right of the layouts and uses `margin-left` and `margin-right` CSS attributes instead.  I have resisted using CSS in this way in the past, trying to keep everything in MathML directly, but could not find another solution that worked in all browsers.

Resolves issue mathjax/MathJax#3547.


zorkow (approved):
Nice. That makes the structures a lot cleaner.

dpvc:
I've made another commit to this branch to remove a check for `mspace` elements that are no longer used.  Do you want to re-review?

PR #1462, opened 2026-04-28
Update assistive-mml to mark all child elements as aria-hidden.  (mathjax/MathJax#3555)

dpvc:
This PR fixes a problem with the `assistive-mml` extension when in-line breaks occur in SVG output.  The extension was only marking the first child SVG as `aria-hidden`, but in-line breaks can cause multiple SVG and `mjx-break` elements, and the second an subsequent ones are not being hidden.

The fix is easy:  loop through all the children instead of just the first one.

Resolves issue mathjax/MathJax#3555.


zorkow (approved):
lgtm.

PR #1464, opened 2026-04-30
Fix prettier format missed in earlier PR.

dpvc:
This fixes a prettier warning that was missed in an earlier PR.


zorkow (approved):
lgtm.

PR #1470, opened 2026-05-15
TexError Localisation 1

zorkow:
PR 
* Merges the old Error message into the new framework
* Moves the JSON files into the correct place
* Adds components constants for loading json
* Fixes all erros to work with the new framework. 

There is some overlap/duplication between messages in `[tex]` and `[tex]/base` components.

NOTE: 02b3b4a09bcade0cac01564f69452a23e4348de7 (the actual merge) is in the wrong order due to a timing issue between my laptop and my desktop.


ts/input/tex/base/BaseMethods.ts
```diff
@@ -63,27 +65,19 @@ const MmlTokenAllow: { [key: string]: number } = {
  * @param {number} n       The number of expected alignment characters
  * @returns {string} String with space separated alignment characters
  */
-export function splitAlignArray(align: string, n: number = Infinity): string {
+export function splitAlignArray(align: string, n: number = Infinity) {
```
dpvc: Is there a reason to remove the `: string` return type?

ts/input/tex/bussproofs/BussproofsMethods.ts
```diff
@@ -136,21 +138,12 @@ function createRule(
...
   parser.i++;
-  const axiom = parser.GetUpTo(name, '$');
-  if (!axiom.includes('\\fCenter')) {
-    throw new TexError(
-      'MissingProofCommand',
-      'Missing %1 in %2.',
-      '\\fCenter',
-      name
-    );
+  let axiom = parser.GetUpTo(name, '$');
```
dpvc: Shouldn't this remain a `const`?  (I assume this is fixed in one of the later PRs?)

ts/input/tex/mhchem/MhchemConfiguration.ts
```diff
@@ -105,7 +105,7 @@ export const MhchemMethods: { [key: string]: ParseMethod } = {
         tex = tex.replace(pattern, name as string);
       }
     } catch (err) {
-      throw new TexError(err[0], err[1]);
+      throw new TexError(null, err[0], err[1]);
```
dpvc, line 108: We can probably make an error file for this since it looks like the errors are being passed with id's and strings.  I'll look into it.

ts/input/tex/unicode/UnicodeConfiguration.ts
```diff
@@ -138,7 +130,19 @@ const UnicodeMethods: { [key: string]: ParseMethod } = {
...
+          c = match[3];
+        } else {
+          parser.i += 2;
+          const cs = [...parser.GetCS()];
+          if (cs.length > 1) {
+            throw new TexError(COMPONENT, 'InvalidAlphanumeric', parser.currentCS);
+          }
+          c = cs[0];
+          match = [''];
+        }
```
dpvc: I'm not sure I understand this change.  Could something have gone wrong with the automated updates (perhaps due to the `'` in the regex above)?

If `match` is non-empty, it will always have `match[1]`, so the original line will aways be performed, and the rest will never be run.  Note that there is never a `match[3]` here, as there is only one set go grouping parentheses in the regex.  Plus the `else` clause makes no sense for the `\char'` case, since you can't use  a control sequence after `'` (as you can with `` ` ``).

So I think the original line should be retained and this added section removed.

ts/input/tex/verb/VerbConfiguration.ts
```diff
@@ -28,20 +28,23 @@ import TexParser from '../TexParser.js';
...
+import { Locale } from '../../../util/Locale.js';
+
+export const COMPONENT = '[tex]/verb';
+Locale.registerLocaleFiles(COMPONENT, '../ts/input/tex/verb');
 
 // Namespace
 const VerbMethods: { [key: string]: ParseMethod } = {
   /**
    * Implements the verbatim notation \verb|...|.
-   *
```
dpvc, line 36: Why remove this (when other instances of the same thing are retained)?

ts/input/tex/HandlerTypes.ts
```diff
@@ -34,6 +34,7 @@ export enum ConfigurationType {
   CONFIG = 'config',
   PRIORITY = 'priority',
   PARSER = 'parser',
+  ERRORS = 'errors',
```
dpvc: Is this still needed?  I don't recall seeing any configurations that include it.

ts/input/tex/StackItem.ts
```diff
@@ -568,9 +568,9 @@ export abstract class BaseItem extends MmlStack implements StackItem {
    * @param {string} kind The stack item type.
    * @returns {string[]} The list of arguments for the TeXError.
```
dpvc: The return type needs to be `string`.  Alternatively, they could remain an array and allow the error to include parameters, so that a message like `'Extra \\middle'` could be treated as `['Extra %1', '\\middle']` to make sure that no code is accidentally translated when the messages are translated.  The `CLASS.errors` list could be `{[id: string]: string | string[]}` and the `getErrors()` method could put the result into an array if it isn't already.  That would mean you don't have to edit the error lists other than to extract the code that shouldn't be translated.

ts/input/tex/StackItem.ts
```diff
@@ -568,9 +568,9 @@ export abstract class BaseItem extends MmlStack implements StackItem {
...
    */
-  public getErrors(kind: string): string[] {
+  public getErrors(kind: string): string {
```
dpvc: This could probably be renamed `getError()` since it only gets one error.  That would mean that the call to it in `base/BaseItems.ts` would need to be fixed.  There may be others as well in other Item definitions.

ts/input/tex/StackItem.ts
```diff
@@ -568,9 +568,9 @@ export abstract class BaseItem extends MmlStack implements StackItem {
...
     const CLASS = this.constructor as typeof BaseItem;
-    return CLASS.errors[kind] || BaseItem.errors[kind];
+    return (CLASS.errors || {})[kind] || BaseItem.errors[kind];
```
dpvc: This is probably better as `CLASS.errors?.[kind] ?? BaseItem.errors[kind]`.

dpvc (changes requested):
This looks good, modulo the issues we discussed face-to-face, and that you have fixed in later PRs.  I have made a few small suggestions, and point out one place where there seems to be an issue with added code that shouldn't be there.

zorkow:
Thanks for rebasing and going through the PR in details. 
Many of the issues you pointed out are probably down to merging of old versions of code that was later linted or incorrectly resolved merge conflicts. In particular, I have now recollection of making any major changes to `UnicodeConfiguration`!

ts/input/tex/base/BaseMethods.ts
zorkow: old code before we did the linting?

ts/input/tex/bussproofs/BussproofsMethods.ts
zorkow: ditto.

ts/input/tex/mhchem/MhchemConfiguration.ts
zorkow: Do in PR #1477

ts/input/tex/unicode/UnicodeConfiguration.ts
zorkow: Definitely a merge error. I had a script for "keeping mine" when `TexError` is involved.

ts/input/tex/StackItem.ts
zorkow: At the moment this returns the id of the error message. We might change this to include component later, as we are discussing.

ts/input/tex/StackItem.ts
zorkow: Renamed.

ts/input/tex/StackItem.ts
zorkow: done.

ts/input/tex/HandlerTypes.ts
zorkow: Removed.

zorkow:
Worked on the comments. PTAL.

dpvc (approved):
This looks good.

PR #1471, opened 2026-05-15
Feature/locale plus2

zorkow:
PR
* Refactors component calls into a single loading structure that sits inside the `locales` folder. This means we do not need to specify the component name in every file that uses `TexError`.
* Actually uses default locale when the current one cannot be found and adds some tests for this.
* Sorts all locale files by key.


dpvc (approved):
Looks good, but should wait to merge until `feature/locale_plus` is ready to merge.

PR #1473, opened 2026-05-15
Renames Locales Directory

zorkow:
PR renames `locales` directory to `__locales__`. The main reason is that if we have the `tex` directory with all the extensions, each in a subdirectory, having `locales` there makes it look like it is just another extension. `__locales__` makes it stand out better and possibly filter it more easily. This is similar to having `__mocks__` directories for `jest`.


dpvc (approved):
Feels a bit funny to see this name, but I understand the motivation.

PR #1475, opened 2026-05-16
Add an excludes option to the copy configuration block for components.

dpvc:
This PR implements an `excludes` option in the `copy` block of a component's `config.json` file.  It also uses that option to exclude the `__locales__/Component.ts` file in the sample `bbox` configuration file.  Similar blocks will need to be added to the other component configuration files.


components/bin/copy
```diff
@@ -70,19 +70,21 @@ const bundleDir = path.resolve(parent, bundle);
...
  * Copy a file or directory tree
  *
- * @param {string} from     The directory to copy from
- * @param {string} to       The directory to copy to
- * @param {string} name     The name of the file or directory to copy
- * @param {string} space    The indentation for output
+ * @param {string} from       The directory to copy from
+ * @param {string} to         The directory to copy to
+ * @param {string} name       The name of the file or directory to copy
+ * @param {string[]} eclude   The files to exclude
```
zorkow: Typo.

zorkow (approved):
One typo. Otherwise fine.

components/bin/copy
dpvc: Thanks. Fixing it.

PR #1476, opened 2026-05-17
Copy __locales__ for all components, and improve default json loading.

dpvc:
This PR adds `copy` blocks to the component configuration files for the TeX packages that have `__locales__` files.

In addition, it fixes the component name for the main TeX `__locales__` file : `[tex]` translates to `[mathjax]/input/tex/extensions`, which is where the webpacked versions go.  The correct component name is 'input/tex'.  In order to get this to work, the `components/mjs/core/core.js` file is modified to always use `Package.resolvePath()`, not just when the package name starts with `[`.  We also add a `mathjax.json` specification in `core.js` so that it works in node applications (that use the components framework) without their having to override the `MathJax.loader.json` configuration value.  We include it here so that the webpack comment is present, as this `.js` file doesn't get transpiled, which would remove comments.  (Alternatively, it could be in `ts/mathjax.ts`, with a separate `tsconfig` file like the one used for the web worker, but I think this is easier.)  The components `tsconfig` file is modified to not remove that comment when compiling the CJS versions.

The main `input/tex` config file copies both the `input/tex/__locales__` directory and the `input/tex/base/__locales__` directory, since the base package doesn't have a separate component to do the copying (though I suppose we could make a component directory for it with just a config file for copying the locales, if you thin that is cleaner).

The `components/mjs/startup/init.js` file is modified to load the core component, so that it will share the same `Locale` object as the core component.  Without this, people use use `<script src="https://cdn.jsdelivr.net/npm/mathjax@4/startup.js>` and the `loader.load` array to load individual components would end up having a separate `Locale` object where the strings get registered, and when startup does `Locale.setLocale()`, it would be with its own copy, and there would be no registered packages.

The `Bbox.test.ts` doesn't need to specify `async`, as the `setupTex()` command already returns a promise.

Finally, there are new `mjs/json.ts` and `cjs/json.ts` files to implement the `mathjax.json` method that supports both browser and node use.  Node users are supposed to load one of the `asyncLoad` implementations, but in case they don't this will allow the locales to be loaded anyway.  Without that, the locale loading would break existing node applications that don't currently import one of the `asyncLoad` files.  Because the CJS typescript settings don't currently allow `import()` to include `{with: {type: 'json'}}`, we use a separate mjs and cjs implementation, where the cjs version uses `require()` rather than `import()`, since that handles JSON files automatically.


zorkow (approved):
Lgtm. I've already added copy blocks to the `config.json` files yesterday (not yet pushed), so I might get some conflicts...

PR #1477, opened 2026-05-18
Add localization for mhchem error messages.

dpvc:
This PR adds localization for the `mhchem` package, for both English and German.  It uses the error id and message thrown by the mhchem parser and translates them into the format needed for the new localized `TeXError` object.  Some of the messages have data inserted into the message strings that has to be extracted to be used as a parameter to `TexError`, and one of the messages is re-using the `[tex]/base` message `Extra close brace or missing open brace`.

I've updated the test file for mhchem, but have not been able to get full coverage.  The bond error message actually can never be produced, but I've included it in case the mhchem parser ever changes to permit it.  I wasn't able to generate one of the two-parameter bug messages (they seem to be for actual bug conditions, so may not be producible).  So the coverage for mhchem is not as complete as it was.

Note that if you add

``` js
    json: (file) => import(
      file.replace('/bundle/input/tex/extensions/', '/ts/input/tex/'),
      {with: {type: 'json'}}
    ),
```

to the `lib/v4-lab.js` file in the MathJax configuration's `loader` block (just after `require: ...` for example), then this will cause the json files to be loaded directly from the `ts/input/tex/<package>/__locales__` directory, so you won't have to keep doing the copy script.  Also note that

``` shell
./components/bin/makeAll --copy --terse components/mjs/input/tex/extensions/<package>
```

will perform the copying to `bundle` for a given component, while

``` shell
./components/bin/makeAll --copy --terse components/mjs/input/tex/extensions
```

would copy all of them.

### DIFF
--- a/ts/input/tex/mhchem/__locales__/Component.ts
+++ b/ts/input/tex/mhchem/__locales__/Component.ts
@@ -1,0 +1,28 @@
+/*************************************************************
+ *
+ *  Copyright (c) 2026 The MathJax Consortium
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+/**
+ * @file  Locale component registration for [tex]/mhchem
+ *
+ * @author dpvc@mathjax.org (Davide P. Cervone)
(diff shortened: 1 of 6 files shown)


ts/input/tex/mhchem/__locales__/Component.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @file  Locale component registration for [tex]/mhchem
+ *
+ * @author dpvc@mathjax.org (Davide P. Cervone)
+ */
+
+import { Locale } from '../../../../util/Locale.js';
+
+export const COMPONENT = '[tex]/mhchem';
+
+Locale.registerLocaleFiles(COMPONENT, '../ts/input/tex/bbox');
```
zorkow: Should that be:

```suggestion
Locale.registerLocaleFiles(COMPONENT, '../ts/input/tex/mhchem);
```

zorkow (changes requested):
I believe you are registering the wrong locale folder.

dpvc:
> I believe you are registering the wrong locale folder.

Thanks for catching that.  Copy-and-paste error.

PR #1478, opened 2026-05-18
Localize menu entries, dialogs, and error messages

dpvc:
This PR adds localization to the MathJax contextual menu and its dialogs and errors.  It does this by adding the localization into the functions that create the menu items (and removing the menu text strings from those calls).  The Clearspeak dynamic menu and preferences dialog are localized, but not the names of the properties or their possible values, since these are the values needed by SRE to process the preferences, and translating them would break that.  (If you think they should be translates, I can do that.)

Some of the menu identifiers were changed, because there were some duplicates that didn't use the same message string, so they had to be made to use separate identifiers.  The `radioGroup()` entries use their values as their identifiers. That means there are things like `100%` as ids.  These could be filtered out and not translated, but it didn't seem to hurt to keep them in place.  If you want them removed, I can do that.

Finally, the package script for copying locale directories has been split into parts (the tex and menus locales are copied in separate scripts, as there are going to be more of these commands as we make more localizations).  I fixed a problem with the copying of the `ts/input/tex/__locales__` directory, which was not being placed in the correct location.

Finally, you should use

``` js
    json: (file) => import(file.replace('/bundle/', '/ts/').replace('/tex/extensions/', '/tex/'),
      {with: {type: 'json'}}
    ),
```

in the lab's `lib/v4-lab.js` file's `loader` section of the MathJax configuration object in order to handle both the tex locale directories (where the `extensions` subdirectory must be removed) and the menu (and other future) locale directories.


dpvc:
There are still a couple of things that need to be added, here, so hold off with the review for now.  I'll let you know when I have it completed.

ts/ui/menu/__locales__/de.json
zorkow, line 1: There are a couple of odd formulations and all the dashes in `MathJax-` menu items are certainly not correct.

dpvc:
I have made the additional changes that I mentioned above.

The bigger issue involved the timing of when the locale files are loaded, which needs to be done before any of the localizations can be used.  That means `Locale.setLocale()` needs to be done as early as possible.  It was being done at the beginning of the `MathJax.startup.defaultReady()` command, but that is too late for some usages, such as messages from the loader about failed loading of components (though these are not yet localized).  So I've inserted a `Locale.setLocale()` call into the startup and loader components so that it is performed before any files are loaded.  (The loader component is never actually used, but I've fixed it there anyway.)  Since this is done before nearly any action MathJax takes, that should mean that localized messages should be ready when they are needed.

In addition, the loader now calls `Locale.setLocale()` after it loads any file or collection of files, so that if any of them register localization files, they will get loaded as well for the current locale.  That means that the `require` extension doesn't have to do that itself, since the loader will do it automatically.

A related issue is that if the user changes the locale via the MathJax contextual menu's `Math Settings -> Language` submenu, and the page is reloaded (or a new one is opened) that menu setting isn't processed until the menu code runs, which occurs when the `MathDocument` is created.  The document is created as part of the `defaultReady()` chain, but that is too late (and in particular, it isn't until after the menu structure is created, so its localization would not be in the locale selected by the user).

To resolve this issue, the menu code now saves the locale in `localStorage` and both the startup and loader components look for that value when they do their initial `Locale.setLocale()` call.  That means everything will be in the user-selected (or page-configured) locale.

I've enabled the `Language` submenu, and included English and German entries, since these are our test languages.  At some point, we will want to combine the language submenu and the speech language submenu so that both MathJax and SRE use the same localization, but I haven't done anything about that for now.

I added the `de` locale for the menu messages and dialogs, and localized a couple of items that I had missed.

Finally, in `ts/util/Locale.ts`, I changed a couple of objects to ones with null prototypes, since we don't want to end up with keys that are inherited from `Object`.  These should probably be converted to `Map` objects, but I didn't take the time to do that, here.

So I think this should be ready to go now, and can be reviewed.

dpvc:
PS, some of the changes in `ts/components/loader.ts` are best viewed with whitespace change being hidden, since prettier indented a fairly large block that is otherwise unchanged.

zorkow (approved):
My main complaint is the German localization. I can take care of that when this PR has landed.

ts/ui/menu/__locales__/de.json
dpvc: OK, thanks. I'm afraid my German is not strong enough to recognize the problems.  I just used the DeepL translations.  I would appreciate repairs.

PR #1479, opened 2026-05-19
Finishes TeX Error localisation

zorkow:
PR (nearly) finished the localisation of the `tex` module. In particular, it 
* removes all duplicates from the extension packages
* puts all error messages into the right place
* `errors` in `BaseItem` and subclasses now get the `Component` string and `id`.
* introduces a `texError` function that throws the localised `TexError`. This way we do not have a breaking change. `texError` gets type `never`, which is needed to avoid typescript errors, but which makes it a bit awkward to test.

Finally, the only point still to be discusses is how to deal with the `getPackage` errors, in particular since there is also a warning. Should we make the Error a  TexError and localise the warning?

Note, that I have added `copy` blocks to all the necessary `config.json` files in the components. Once PR #1476 hits develop I will merge that in and resolve the conflicts.


ts/util/Locale.ts
```diff
@@ -136,14 +136,31 @@ export class Locale {
     let message = '';
     if (component) {
       message = this.lookupMessage(component, id);
```
dpvc: Since `message` is only used in the `if (component)` block, line 136 can be removed and line 138 can be made `const message = ...`.

dpvc (approved):
This all looks good, with one recommended change.

dpvc:
Looks like this needs to have `pnpm -s format:fix` run on it.

ts/input/tex/physics/PhysicsItems.ts
```diff
@@ -41,6 +42,11 @@ export class AutoOpen extends BaseItem {
    */
   public openCount: number = 0;
 
+  /**
+   * @override
+   */
+  protected component = COMPONENT;
+
```
dpvc: Is this still needed?  I don't see where it is used.

dpvc:
Does this line

https://github.com/mathjax/MathJax-src/blob/fcf47d91c07e2666f64e1c968a323c7d4f06f4b1/ts/input/tex/Tags.ts#L699

need to be localized?

And also this one?

https://github.com/mathjax/MathJax-src/blob/fcf47d91c07e2666f64e1c968a323c7d4f06f4b1/ts/input/tex/MapHandler.ts#L205

The first can be done through `Locale.error()`.  Should there also be a `Locale` function for the second?  Like `Locale.log()`?  (Or should `Locale.error()` use `console.error()` and a new function `Locale.throw()` use `throw`?)

dpvc:
Here are a couple more:

https://github.com/mathjax/MathJax-src/blob/fcf47d91c07e2666f64e1c968a323c7d4f06f4b1/ts/input/tex/bussproofs/BussproofsUtil.ts#L654-L656

https://github.com/mathjax/MathJax-src/blob/fcf47d91c07e2666f64e1c968a323c7d4f06f4b1/ts/input/tex/require/RequireConfiguration.ts#L215

https://github.com/mathjax/MathJax-src/blob/fcf47d91c07e2666f64e1c968a323c7d4f06f4b1/ts/input/tex/Configuration.ts#L441

ts/util/Locale.ts
```diff
@@ -136,14 +136,31 @@ export class Locale {
...
+
+  /**
+   * Process a message string by substituting the given arguments. The arguments
+   * can be positional, or a data mapping of names to values.
+   *
+   * @param {string} message The message string to process.
+   * @param {string | namedData = {}} data The first argument or the object of
+   *     names arguments
+   * @param {string[]} ...args Additional arguments (if data is a string)
+   * @param {...any} args
```
dpvc: The JSDoc comments are messed up, here.  The `= {}` in the `data` argument should be removed, as should the `...` in line 154.  Line 155 should be removed entirely.

``` suggestion
   * @param {string|namedData} data   The first argument or the object of named arguments
   * @param {string[]} args           Additional arguments (if data is a string)
```

dpvc (changes requested):
There are problems with the jsdoc comments, here.

zorkow:
> 
> The first can be done through `Locale.error()`. Should there also be a `Locale` function for the second? Like `Locale.log()`? (Or should `Locale.error()` use `console.error()` and a new function `Locale.throw()` use `throw`?)

There are a couple of issues here:
1. Both `MapHandler` and `Configuration` have a `warn` method. Those should be removed. 
2. I feel an error function should always throw an error. Otherwise, why call it an error.
3. We could have a general `Locale.warn` method that gives localised information on the console (via `console.info`).
4. I don't think logging should be localised.

zorkow:
> Does this line
> 
> https://github.com/mathjax/MathJax-src/blob/fcf47d91c07e2666f64e1c968a323c7d4f06f4b1/ts/input/tex/Tags.ts#L699
> 
> need to be localized?
> 

This is now localised.

ts/util/Locale.ts
zorkow: That's my emacs macro for generating jsdoc blocks that simply copies parameters and types.
Linting then introduces the extra `args` line.

dpvc (approved):
OK, this is a go for now.  I will handle the other throw commands in my PR that we discussed in our meeting.

dpvc:
Merging `feature/mhchem-locale` into `feature/locale` had introduced a merge conflict here.  I can fix that, if you like.

zorkow:
> > Does this line
> > https://github.com/mathjax/MathJax-src/blob/fcf47d91c07e2666f64e1c968a323c7d4f06f4b1/ts/input/tex/Tags.ts#L699
> > 
> > need to be localized?
> 
> This is now localised.

This now has a localisation as `UnknownTag` but is not a `TexError`.

zorkow:
Review suggestions incorporated. I left the regular `Error` messages alone as discussed yesterday to do them in a separate PR.
PTAL

dpvc (approved):
Looks good.

PR #1480, opened 2026-05-20
Update node-main to initialize the locale, and properly find the locale json files

dpvc:
This PR fixes `node-main` to initialize the locale before it starts loading anything (like the changes in `feature/menu-locale`).  it also adjusts the `asyncLoad()` function to properly find the JSON files, even when the package name doesn't start with `[` (it currently fails to find the localization files for `input/tex` because of that).  Also, mark the `asyncLoad()` method as being synchronous.


zorkow (approved):
lgtm.

PR #1481, opened 2026-05-20
Update Locale to handle synchronous JSON file loading, when possible.

dpvc:
This PR allows the `Locale` object to load the localization files synchronously, when the `mathjax.asyncLoad` method is marked as synchronous.  It also fixes the JSDoc issue that I mentioned in the review of #1479.

### DIFF
--- a/testsuite/tests/util/Locale.test.ts
+++ b/testsuite/tests/util/Locale.test.ts
@@ -1,4 +1,5 @@
 import { describe, test, expect } from '@jest/globals';
+import { mathjax } from '#js/mathjax.js';
 import { Locale } from '#js/util/Locale.js';
 import '#js/util/asyncLoad/esm.js';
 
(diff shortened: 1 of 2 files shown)


dpvc:
I've added the `Locale.throw()` method I mentioned in #1479, and made `Locale.error()` use `console.error()` rather than `throw`.  I also added `Locale.warn()` and `Locale.log()`, and added tests for these, as well as for the synchronous loading.

dpvc:
I will need to change this PR in light of our meeting today.  So wait with it for now.

dpvc:
OK, Ive reverted this branch to the original purpose of adding the synchronous loading of locale files, when possible.  I will do the other changes in a separate PR, so this one can be reviewed at this point.

dpvc:
It turns out that once more of MathJax is localized, this has to be handled differently to avoid cyclical dependencies.  I am closing this and will include the feature in a later PR.

PR #1482, opened 2026-05-24
Add Locale.warn() and Locale.throw(), add tests for them, and use them in several TeX packages.

dpvc:
This PR adds the `Locale.warn()` function and renames `Locale.error()` as `Locale.throw()`, and uses these in the TeX packages that I mentioned in the finalize-tex branch.  It updates the tests to cover these, and moves some output-trapping functions from the `testsuite/src/setupTex.ts` file to a new `traps.ts` file.  The reason for this is that `setupTeX.ts` loads `node-main`, which does two two things that affect the `Locale` package:  because it is component-based, it means `Locale.isComponent` gets set, which we don't want (that changes the location where `Locale` looks for the localization files), second, it alters the `mathjax.asyncLoad` which ends up causing different error messages.  I could have gotten around these issues, but it seemed reasonable to separate out the non-tex-based output trapping functions from `setupTex.ts` in any case, and that avoids the problem.

The change to `package.js` is to allow access to the new `trap.ts` file via `#helpers/trap.js`.  The `index.js` file is modified to export the traps from the new file rather than `setupTex.js`.  The needed traps are loaded into `setupTex.ts` rather than being defined there.  I also fixed some JSDoc `@returns` that changed after adding the `Locale.setLocale()` calls (in an earlier PR).  

Some tests in `Tex.test.ts` are adjusted to accommodate the use of `Locale.warn()` and the removal of the `warn()` methods that we discussed in our meeting.  The `Locale.test.ts` file is implified to use the `trapOutput` and `trapAsyncOutput()` functions from the new `traps.ts` file, rather that chaining the `console.error` and `console.warn` functions by hand, and a new section to test `Locale.warn()` is added.

The usage of `console` and `throw new Error` in several TeX packages are changes to `Locale.warn()` and `Locale.throw()` and their `warn()` methods removed, and the needed messages are added to the localization files.

### DIFF
--- a/testsuite/src/traps.ts
+++ b/testsuite/src/traps.ts
@@ -1,4 +1,4 @@
-import {jest} from '@jest/globals';
+import { jest } from '@jest/globals';
 
 type consoleMethod = 'log' | 'warn' | 'error';
 
@@ -10,15 +10,17 @@ type consoleMethod = 'log' | 'warn' | 'error';
  * @returns {string} The output sent to the given method.
  */
 export function trapOutput(method: string, code: () => void): string {
-  const messages: string[] = [];
-  const spy = jest.spyOn(console, method as consoleMethod)
-    .mockImplementation((...msg) => messages.push(msg.join(' ')));
+  const spy = jest
+    .spyOn(console, method as consoleMethod)
+    .mockImplementation(() => {});
+  let message = '(no message)';
   try {
     code();
+    message = spy.mock.calls.map((msg: string[]) => msg.join(' ')).join('\n');
   } finally {
     spy.mockRestore();
   }
-  return messages.join('\n');
+  return message;
 }
 
 /**
@@ -50,15 +52,17 @@ export async function trapAsyncOutput(
   method: string,
   code: () => Promise<void>
 ): Promise<string> {
-  const messages: string[] = [];
-  const spy = jest.spyOn(console, method as consoleMethod)
-    .mockImplementation((...msg) => messages.push(msg.join(' ')));
+  const spy = jest
+    .spyOn(console, method as consoleMethod)
+    .mockImplementation(() => {});
+  let message = '(no message)';
   try {
     await code();
+    message = spy.mock.calls.map((msg: string[]) => msg.join(' ')).join('\n');
   } finally {
     spy.mockRestore();
   }
-  return messages.join('\n');
+  return message;
 }
 
 /**
--- a/testsuite/src/traps.ts
+++ b/testsuite/src/traps.ts
@@ -1,3 +1,7 @@
+import {jest} from '@jest/globals';
+
+type consoleMethod = 'log' | 'warn' | 'error';
+
 /**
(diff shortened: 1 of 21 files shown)


dpvc:
I've added a commit that converts `console.log()` messages to `Locale.warn()` or `console.warn()` in the menu code.

testsuite/src/traps.ts
```diff
@@ -0,0 +1,103 @@
...
+  const saved = (console as any)[method];
+  let message = '';
+  (console as any)[method] = (...msg: any[]) => {
+    message += (message ? '\n' : '') + msg.join(' ');
+  };
+  code();
+  (console as any)[method] = saved;
+  return message;
+}
+
```
zorkow, line 25: If `code` throws an error  `console[method]` is never restored. I believe this could be more effectively and safer done with `jest.spyOn`.  E.g.,

```typescript
 export function trapOutput(method: string, code: () => void): string {
    const spy = jest.spyOn(console, method).mockImplementation(() => {});
    try {
      code();
    } finally {
      spy.mockRestore();
    }
    return spy.mock.calls.map((msg: any[]) => msg.join(' ')).join('\n');
  }
```

testsuite/src/traps.ts
```diff
@@ -0,0 +1,103 @@
...
+): Promise<string> {
+  const saved = (console as any)[method];
+  let message = '';
+  (console as any)[method] = (...msg: any[]) => {
+    message += (message ? '\n' : '') + msg.join(' ');
+  };
+  await code();
+  (console as any)[method] = saved;
+  return message;
+}
```
zorkow, line 66: same here.

ts/util/Locale.ts
```diff
@@ -215,15 +215,32 @@ export class Locale {
...
   ) {
     throw Error(this.message(component, id, data, ...args));
   }
 
+  /**
+   * Report a warning with a given string substituting the given parameters
+   *
+   * @param {string} component        The component whose message is requested
+   * @param {string} id               The id of the message
+   * @param {string|namedData} data   The first argument or the object of names arguments
```
zorkow:
```suggestion
   * @param {string|namedData} data   The first argument or the object of named arguments
```

zorkow (changes requested):
I suggest to use `jest.spyOn` on `console` methods rather than redefining the functions.

And one typo.

testsuite/src/traps.ts
dpvc: Looks good.  I was not aware of `spyOn()`.  This is much better.

dpvc:
I have made the requested changes (though I needed to alter the code slightly, since you have to use the `spy.mock.calls` data before `spy.mockRestore()` apparently).  I also changed a couple of tests that explicitly saved/restored the console methods.  

I'm going to merge this without additional review.  If you don't like the changes I made, we can do another PR after the fact.  I would like to merge the other reviewed PRs and don't want this to load them up.

PR #1484, opened 2026-05-25
Produce message if locale isn't set when using direct access to MathJax modules.

dpvc:
This PR introduces an error message about not having set up localization, if that is the case.  This applies only to node applications that use direct access to the MathJax modules (since the component-based apps in node or the browser will have `Locale.setLocale()` performed automatically).  In this PR, when a MathDocument is created, the Locale is checked and it `setLocale()` has never been called, the message is generated.   When we have a URL to point to, it should be added to the message.


zorkow (commented):
lgtm. 
I think this is the branch I need to merge for PR #1496

zorkow (approved):

PR #1485, opened 2026-05-25
Add localization for loader and startup via core/__locales__

dpvc:
this PR adds localization to the `loader` and `startup` modules.  It does so by adding a `core/__locales__` directory for the `core` component, and using that for their messages, since the `loader` and `startup` components include the `core` component.  These components now load `core/locale.js` before `lib/loader.js` or `lib/startup.js` so that the `Locale.isComponent` value is set before the files imported by the `lib` code (in particular, their `Component.js` files) are loaded.  Otherwise `Locale` might use the wrong directory when trying to find the localization files.

Copying the `core` localization has been added to the package scripts, and I've also taken this opportunity to adjust the `copy:assets` and `copy:locales` scripts to use a list of sub-scripts to run rather than having them explicitly called.  This allows new ones to be added easier, and also to make `mjs` the default for `$1` so you can just use `pnpm -s copy:assets` without needing to specify `mjs` directly.

The rest is just the changes to localize the messages in `loader.ts` and `startup.ts`.


zorkow (approved):
Some issues with the German, but that'll be in my PR #1498 .
O/w fine.

PR #1486, opened 2026-05-25
Update menu localization IDs (as per our developers' meeting) and sort them

dpvc:
This PR adjusts the localization IDs used by the menu component and sorts the result.  The various dialog messages get prefixes based on the dialog, so they sort together.  The percentage IDs are removed and are not localized.  The `120%` is removed from the `ScalePercent` message and included as a parameter in order to avoid it being accidentally translated.


zorkow (approved):
lgtm.

Some changes and German corrections are in #1496

PR #1487, opened 2026-05-31
Add localization for remaining core functions and add tests

dpvc:
This PR localizes the files that are part of the `core` component, which includes the `ts/util` files and the files in `ts/core`.  It also include the messages from `ts/output/common` even though those aren't in the core component, but since there are only a few of them, and the `common` code is in more than one output component, they are put in core to avoid duplication in `output/chtml` and `output/svg`.

## Details

A new `testsuite/src/locale.ts` is added that sets up the locales.  Because the `setupTex` is included in the `index.js` file in `#helpers`, and it loads `node-main`, which sets up `mathjax.asyncLoad` to be a synchronous `require`-based call, that altered some of the Locale test behavior.  This file can be used in non-tex tests (like the util tests) to initialize the locale without altering `asyncLoad`.

In `setupTex.ts`, we pre-load `AsyncLoad` and inform `Locale` that it is synchronous so that `Locale.setLocale()`will be synchronous (more about this later).  That way, we don't have to make the `setupTex()` and related functions be `async`, which would require changing a bunch of tests that use it.  (So we no longer return promises from those.)  We move the `Locale.setLocale()` to before the creation of the MathDocument in order to avoid the message about `Locale` not being initialized.

The `util` tests are modified to load the new `src/locale.js` file.  Tests are added to `Locale.test.ts` to test the synchronous loading and the new setup code (described when discussing `Locale` changes below).  Because the `core` component is now being registered whenever a core module is loaded, `Locale.setLocale()` will try to load a `core/__locales__/*.json` file, so some tests that use undefined locales end up throwing that error for that.  So we trap those errors when we don't care about them, and remove the `MathJax(component)` prefix from some tests, since the actual component that throws the error may be different (e.g., `MathJax(core)`).

A few other tests are changed to accommodate changes in the error strings (like `"` to `'`, so make the json files easier to deal with).

Most of the rest of the changes are the localizations of the messages in various files.

For the `ts/core/MmlNode` files, the `mError()` function is modified to take a message ID and an array of substitutions as its arguments rather than the message itself.  Technically, this is a breaking change, but unless someone has created their own internal MathML node, it should not affect anyone.

Finally, the changes to `ts/util/Locale.ts` are for two things:  to allow for synchronous loading of locale files, when possible, and to avoid a cyclical dependency.  For the latter, now that `ts/util/AsyncLoad.ts` has been localized by importing the `core/__locales__/Component.ts` file (which calls `Locale.registerLocaleFiles()`), importing `AsynchLoad` into `Locale.ts` would cause `Locale.registerLocalFiles()` to be called before `Locale` is set up, causing an error.

To avoid this, `Locale` now puts off loading `AsynchLoad` until the `Locale.setLocale()` call, and uses `import()` to do so asynchronously.  It checks whether `mathjax.asyncIsSynchronous` is true, and save that as `Locale.syncLoad` if it is.  The `getLocaleData()` method is modified to use `this.syncLoad()` when it is available in order to do the file loading synchronously, so that `Locale.setLocale()` is synchronous rather than asynchronous (though it still returns a promise).

Note, however, that the first call to `Locale.setLocale()` would still be asynchronous in order to wait for `mathjax.ts` and `AsyncLoad.ts` to load.  To really make `Locale.setLocale()` completely synchronous, one needs to set `Locale.syncLoad` by hand before `Locale.setLocale()` is called.  That is why `setupTex.ts` does that for the testsuite, thus allowing `setupTex()` and related functions to be synchronous, avoiding the need for `await` and `async` functions in the tests that call `setupTex()`, as described above.


ts/util/Options.ts
```diff
@@ -85,9 +89,9 @@ export const OPTIONS = {
    */
   optionError: (message: string, _key: string) => {
     if (OPTIONS.invalidOption === 'fatal') {
-      throw new Error(message);
+      throw Error(message);
     }
-    console.warn('MathJax: ' + message);
+    console.warn(message);
```
zorkow, line 94: We should probably mention this in the release notes. Just in case someone is looking for this in their code, a bit similar to what we do in the tests.

ts/core/__locales__/de.json
```diff
@@ -1,9 +1,37 @@
...
+  "MML/BadAttr": "Unbekannte Attribute für den Knoten '%1': %2",
+  "MML/BadNesting": "Falsche Verschachtelung von Math-Tags",
+  "MML/BadVariant": "Ungültige mathvariant: %1",
+  "MML/EqualPrescripts": "Es muss eine gleiche Anzahl von Vorzeichen jedes Typs vorhanden sein",
+  "MML/EqualScripts": "Es muss eine gleiche Anzahl von Skripten jedes Typs vorhanden sein",
+  "MML/NoVariant": "Es sollten keine untergeordneten Elemente für '%1'-Knoten vorhanden sein",
+  "MML/OnlyOnce": "%1 darf in %2 nur einmal vorkommen",
+  "MML/TokenChild": "%1 darf nur ein untergeordnetes Element eines Token-Elements sein",
+  "MML/mglyphAttr": "mglyph muss entweder die Attribute src oder fontfamily sowie das Attribut index aufweisen",
+  "MML/mmultscriptsChild": "%1 muss ein untergeordnetes Element von mmultiscripts sein",
```
zorkow:
```suggestion
  "MML/mmultiscriptsChild": "%1 muss ein untergeordnetes Element von mmultiscripts sein",
```

ts/core/__locales__/en.json
```diff
@@ -1,9 +1,37 @@
...
+  "MML/BadAttr": "Unknown attributes for '%1' node: %2",
+  "MML/BadNesting": "Improper nesting of math tags",
+  "MML/BadVariant": "Invalid mathvariant: %1",
+  "MML/EqualPrescripts": "There must be an equal number of prescripts of each type",
+  "MML/EqualScripts": "There must be an equal number of scripts of each type",
+  "MML/NoVariant": "There should not be children for '%1' nodes",
+  "MML/OnlyOnce": "%1 can only appear once in %2",
+  "MML/TokenChild": "%1 can only be a child of a token element",
+  "MML/mglyphAttr": "mglyph must have either src or fontfamily and index attributes",
+  "MML/mmultscriptsChild": "%1 must be a child of mmultiscripts",
```
zorkow:
```suggestion
  "MML/mmultiscriptsChild": "%1 must be a child of mmultiscripts",
```

ts/input/mathml/MathMLCompile.ts
```diff
@@ -299,7 +299,8 @@ export class MathMLCompile<N, T, D> {
             this.addChildren(mml, child);
           } else {
             childMml.mError(
-              'There should not be children for ' + childMml.kind + ' nodes',
+              'MML/NoChildren',
```
zorkow, line 304: That key does not exist.

ts/core/__locales__/en.json
```diff
@@ -1,9 +1,37 @@
...
   "HandlerNotDefined": "Handler '%1' is not defined (has it been loaded?)",
+  "InputError": "Math input error",
   "InputJaxNotDefined": "Input Jax '%1' is not defined (has it been loaded?)",
+  "MML/Arity": "Wrong number of children for '%1' node",
+  "MML/BadAttr": "Unknown attributes for '%1' node: %2",
+  "MML/BadNesting": "Improper nesting of math tags",
+  "MML/BadVariant": "Invalid mathvariant: %1",
+  "MML/EqualPrescripts": "There must be an equal number of prescripts of each type",
+  "MML/EqualScripts": "There must be an equal number of scripts of each type",
+  "MML/NoVariant": "There should not be children for '%1' nodes",
```
zorkow: Is that `MML/NoChildren`?

ts/core/__locales__/de.json
```diff
@@ -1,9 +1,37 @@
...
   "HandlerNotDefined": "Handler '%1' ist nicht definiert (wurde es geladen?)",
+  "InputError": "Fehler bei der mathematischen Eingabe",
   "InputJaxNotDefined": "Input Jax '%1' ist nicht definiert (wurde es geladen?)",
+  "MML/Arity": "Falsche Anzahl von untergeordneten Elementen für den Knoten '%1'",
+  "MML/BadAttr": "Unbekannte Attribute für den Knoten '%1': %2",
+  "MML/BadNesting": "Falsche Verschachtelung von Math-Tags",
+  "MML/BadVariant": "Ungültige mathvariant: %1",
+  "MML/EqualPrescripts": "Es muss eine gleiche Anzahl von Vorzeichen jedes Typs vorhanden sein",
+  "MML/EqualScripts": "Es muss eine gleiche Anzahl von Skripten jedes Typs vorhanden sein",
+  "MML/NoVariant": "Es sollten keine untergeordneten Elemente für '%1'-Knoten vorhanden sein",
```
zorkow: Is that `MML/NoChildren`?

ts/core/MmlTree/MmlNodes/mtr.ts
```diff
@@ -112,19 +112,15 @@ export class MmlMtr extends AbstractMmlNode {
...
+      this.mError('MML/mtableChild', [this.kind], options, true);
       return;
     }
     for (const child of this.childNodes) {
       if (!child.isKind('mtd')) {
         const mtd = this.replaceChild(this.factory.create('mtd'), child);
         mtd.appendChild(child);
         if (!options['fixMtables']) {
-          child.mError('Children of ' + this.kind + ' must be mtd', options);
+          child.mError('MML/mtrChildren', [], options);
```
zorkow: `this.kind` is missing.

```suggestion
          child.mError('MML/mtrChildren', [this.kind], options);
```

zorkow (changes requested):
I found a couple of issues with messages and their arguments.

ts/input/mathml/MathMLCompile.ts
dpvc: Yes.  Thanks for catching that.

ts/core/__locales__/de.json
dpvc: Yes.  I'll change it.

ts/core/__locales__/en.json
dpvc: Yes.  I'll change it.

PR #1489, opened 2026-05-31
Localize the MathML input jax

dpvc:
This PR localizes the MathML input jax and its extensions.  It also fixes a comment in the TeX `Component.ts` file.


zorkow (approved):
lgtm.

German changes in PR #1498

PR #1490, opened 2026-05-31
Localize a11y/semantic-enrich

dpvc:
This PR localizes the `a11y/semantic-enrich` component.


zorkow (approved):
lgtm.
German sounds a bit weird, but I can't think of anything better.

PR #1491, opened 2026-05-31
Localize a11y/speech

dpvc:
This PR localizes the `a11y/speech` component.  As discussed in the developers' meeting, we do not localize the messages from the web-worker itself.


zorkow (approved):
lgtm.

PR #1492, opened 2026-05-31
Localize a11y/explorer

dpvc:
This PR localizes the `a11y/explorer` component.  Moving the dialog box messages to the localization files means that the code for producing the message needed to localize the pieces that were used to construct the message first and the insert them into the full message.  This might need to be adjusted once we have other localizations so that the pieces work more naturally in other languages.


zorkow (approved):
lgtm.

PR #1494, opened 2026-05-31
Preload en locale in component definitions

dpvc:
This PR adds files to the `components/mjs` subdirectories in order to pre-load the `en` locale definitions when the components are webpacked (or loaded into a node application).  This required the addition of a `#ts` pseudo-path in order to be able to access the definitions from the `ts` directory (since they might not have been moved to the `mjs` or `bundle` directories yet).

This is done through the use of separate `en.js` files for each component so that it would be possible to use the webpack configuration to replace them easily with files for another locale, if a version with a different built-in default wanted to be created.

An alternative would be to move these to the `ts/**/__locales__` directories and have the `Component.ts` files load them directly.  Then even node applications with direct calls to the modules would have the default `en` localization automatically, and there would not be a need for the warning message if `Locale.setLocale()` wasn't called first.  But then node applications that change the default locale will end up loading both `en` and the new one, whereas with the files where they are now, that would not be the case for direct module loading.  I'm not sure which is the better approach.

### DIFF
--- a/ts/util/Locale.ts
+++ b/ts/util/Locale.ts
@@ -117,6 +117,7 @@ export class Locale {
       cdata[locale] = Object.create(null);
     }
     Object.assign(cdata[locale], data);
+    this.locations[component][1].add(locale);
   }
 
   /**
--- a/components/bin/package.json
+++ b/components/bin/package.json
@@ -1,6 +1,7 @@
 {
   "type": "commonjs",
   "imports": {
+    "#ts/*": "@mathjax/src/ts/*",
     "#js/*": "@mathjax/src/cjs/*",
     "#source/*": "@mathjax/src/components/cjs/*",
     "#root/*": "@mathjax/src/cjs/components/cjs/*",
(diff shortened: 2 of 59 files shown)


ts/util/Locale.ts
```diff
@@ -117,6 +117,7 @@ export class Locale {
       cdata[locale] = Object.create(null);
     }
     Object.assign(cdata[locale], data);
+    this.locations[component][1].add(locale);
```
zorkow, line 120:
```suggestion
    this.locations[component]?[1].add(locale);
```
This would avoid a cryptic error message if `registerMessages` is called when the component does not exist.

zorkow (approved):
One minor change. Otherwise lgtm.

ts/util/Locale.ts
dpvc: Is there a use case you have in mind for when this might happen?

Also, shouldn't we produce an error message for this rather than silently ignore it?

dpvc:
This is updated in #1509, and that needs to be changed in order to accommodate CJS versions, so I am closing this and #1509 in favor of a new branch that I will make later.

PR #1497, opened 2026-06-09
Fix problem with `\limits` and `\nolimits` when used after a script. (mathjax/MathJax#3569)

dpvc:
This PR fixes a problem with `\limits` and `\nolimits` when they are used after a super- or subscript is already specified and the operator has a coefficient.  E.g. `1 \sum_n\nolimits`.  The problem is that the wrong stack item is being replaced.

Resolves issue mathjax/MathJax#3569.


zorkow (approved):
lgtm.

PR #1498, opened 2026-06-10
Fix more for German locale files

zorkow:
Improves German in locales for

* `core`
* `tex`
* `mathml`
* `worker`
* `explorer`


dpvc (approved):
Looks good.

Issue #1458, opened 2026-04-09
Can't load .../svg/dynamic/math.js


dmca-glasgow:
Hello,

I have a 2 part problem with the following maths: `K^\perp` which seems to lead to unexpected behaviour in mathjax.

Given the following minimal-as-possible demo where I statically load packages and use 2 fonts:

```ts
import { mathjax } from '@mathjax/src/js/mathjax.js';

// TeX to MathML
import { liteAdaptor } from '@mathjax/src/js/adaptors/liteAdaptor.js';
import { STATE } from '@mathjax/src/js/core/MathItem.js';
import { SerializedMmlVisitor } from '@mathjax/src/js/core/MmlTree/SerializedMmlVisitor.js';
import { RegisterHTMLHandler } from '@mathjax/src/js/handlers/html.js';
import { TeX } from '@mathjax/src/js/input/tex.js';

import '@mathjax/src/js/input/tex/base/BaseConfiguration.js';

// MathML to SVG
import { MathML } from '@mathjax/src/js/input/mathml.js';
import { SVG } from '@mathjax/src/js/output/svg.js';

import { MathJaxNewcmFont } from '@mathjax/mathjax-newcm-font/js/svg.js';
import '@mathjax/mathjax-newcm-font/js/svg/dynamic/math.js';

import { MathJaxFiraFont } from '@mathjax/mathjax-fira-font/js/svg.js';
// import '@mathjax/mathjax-fira-font/js/svg/dynamic/math.js';

// TeX to MathML ---------------------------------------

const adaptor = liteAdaptor();
const visitor = new SerializedMmlVisitor();
RegisterHTMLHandler(adaptor);

const packages: string[] = [
  'base',
];

const tex = new TeX({ packages });

const mmlDoc = mathjax.document('', {
  InputJax: tex,
});

function texToMathML(tex: string) {
  const mmlNode = mmlDoc.convert(tex, { end: STATE.CONVERT });
  return visitor.visitTree(mmlNode)
}

// MathML to SVG ----------------------------------

const adaptor2 = liteAdaptor();
RegisterHTMLHandler(adaptor2);

const htmlDoc = mathjax.document('', {
  InputJax: new MathML(),
});

const NewcmFont = new MathJaxNewcmFont();
const FiraFont = new MathJaxFiraFont();

const packages2: string[] = [
  // 'math',
];

packages2.forEach((name) => {
  // @ts-expect-error Property 'dynamicFiles' is protected
[...]

PR #1500, opened 2026-06-11
Load font files synchronously, when possible.  (#1458)

dpvc:
This PR allows font ranges to be loaded synchronously if the `mathjax.asyncLoad()` method is marked as synchronous.  That makes it possible to use the synchronous typesetting and conversion calls in more situations.

Resolves issue #1458.

### DIFF
--- a/ts/output/common/FontData.ts
+++ b/ts/output/common/FontData.ts
@@ -1333,10 +1333,9 @@ export class FontData<
       if (mathjax.asyncIsSynchronous) {
         this.loadDynamicFileSync(delim);
         return this.getDelimiter(n);
-      } else {
-        retryAfter(this.loadDynamicFile(delim));
-        return null;
       }
+      retryAfter(this.loadDynamicFile(delim));
+      return null;
     }
     return delim as DelimiterData;
   }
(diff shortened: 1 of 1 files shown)


ts/output/common/FontData.ts
```diff
@@ -1330,8 +1330,13 @@ export class FontData<
     const delim = this.delimiters[n];
     if (delim && !('dir' in delim)) {
       this.delimiters[n] = null;
-      retryAfter(this.loadDynamicFile(delim));
-      return null;
+      if (mathjax.asyncIsSynchronous) {
+        this.loadDynamicFileSync(delim);
+        return this.getDelimiter(n);
+      } else {
```
zorkow: No need for `else` clause as `if` ends in `return`.

ts/output/common/FontData.ts
```diff
@@ -1378,8 +1383,13 @@ export class FontData<
       const variant = this.variant[name];
       delete variant.chars[n];
       variant.linked.forEach((link) => delete link[n]);
-      retryAfter(this.loadDynamicFile(char));
-      return null;
+      if (mathjax.asyncIsSynchronous) {
+        this.loadDynamicFileSync(char);
+        return this.getChar(name, n);
+      } else {
```
zorkow: ditto.

zorkow (changes requested):
See comments.

ts/output/common/FontData.ts
dpvc: OK, fixed.

dpvc:
I've fixed the issues.  Do you want to re-review, or should I merge?

zorkow (approved):
lgtm.

PR #1501, opened 2026-06-11
Don't rerender the output during menu initialization.

dpvc:
This PR adjusts the menu code to **not** re-render while the menu is being initialized and user preferences are being set (the actions attached to the menu variables could cause re-rendering).


zorkow (approved):
lgtm.

Issue #1474, opened 2026-05-16
textmacros: TextParser instances accumulate in nested ParseOptions.parsers across html.convert() calls


mackdelany:
## Summary

In long-lived `MathDocument` instances, every `html.convert(tex)` that contains `\text{...}` leaves one or more `TextParser` instances stranded on the nested `ParseOptions.parsers` array stashed in `tex.parseOptions.packageData.get('textmacros')`. Each leaked `TextParser` retains its `Stack`, `StackItem`s, and the parsed MathML subtree, so memory grows linearly with the number of conversions. No documented public API (`html.clear()`, `html.reset({ all: true })`, `tex.reset()`) clears this pool.

Consequence for downstream users: a browser app that calls `html.convert()` to render equations on demand (rather than batch-typesetting an HTML document) accumulates ~1 parser, ~1 Stack, and the equation's full MathML tree per `\text{...}` per render, indefinitely. We caught this via Chrome heap snapshots in a real product: `MmlMo`, `MmlMi`, `Attributes`, and `TextNode` counts growing ~5× over 60 seconds of UI interaction, with `Stack` count rising from 105 → 482 in the same window.

## Affected versions

- `mathjax-full@3.2.1` — confirmed via the repro below
- `mathjax-full@3.2.2` — source diff against 3.2.1 shows no relevant changes
- `mathjax-full@4.0.0-beta.7` — same buggy code paths present in `cjs/input/tex/TexParser.js` and `cjs/input/tex/textmacros/TextMacrosConfiguration.js` (verified by inspection; couldn't run the repro against 4.x because `AllPackages` was reorganized, but the underlying mechanism is identical)

## Reproduction

```js
// node mathjax-textmacros-leak-repro.mjs
// npm install mathjax-full@3.2.1

import { TeX } from "mathjax-full/js/input/tex.js"
import { SVG } from "mathjax-full/js/output/svg.js"
import { liteAdaptor } from "mathjax-full/js/adaptors/liteAdaptor.js"
import { HTMLDocument } from "mathjax-full/js/handlers/html/HTMLDocument.js"
import { AllPackages } from "mathjax-full/js/input/tex/AllPackages.js"

const adaptor = liteAdaptor()
const doc = adaptor.createDocument()
adaptor.document = doc

const tex = new TeX({ packages: AllPackages })
const svg = new SVG({ fontCache: "none" })
const html = new HTMLDocument(doc, adaptor, { InputJax: tex, OutputJax: svg })

[...]

PR #1502, opened 2026-06-11
Make sure the parsers array is cleared for tex and text parsers.  (#1474)

dpvc:
This PR moves the `popParser()` call in the `mml()` function of the `TeXParser` object so that it will be performed even when the function exits early.  It also adds `parseOptions.clear()` to the TeX input jax `reset()` function, as well as the TextMacros preprocessor, just like is done in the TeX input jax `compile()` function.

Resolves issue #1474.


zorkow (approved):
lgtm.

PR #1503, opened 2026-06-11
Fix `\bbox` problem with padding values that are decimals.  (mathjax/MathJax#3568)

dpvc:
This PR fixes a problem with `\bbox` where padding values that are decimals can lead to `NaN` values.  This was caused by using `parseInt()` instead of `parseFloat()`, so the fix is easy.

Resolves issue mathjax/MathJax#3568.


zorkow (approved):
lgtm.

PR #1504, opened 2026-06-13
Update to de menu fix

dpvc:
This PR implements a number of adjustments to the menu localization code, including:

1. Updating radio group submenus to remove the prefixes for their values (so that the code that uses their values doesn't have to be changed, and so that user configurations will not need to add the prefixes).
2. Make the a11y ARIA component a "fake" one (just for holding the localizations), rather than one that has an actual component file.
3. Changes some localization ids to match the original ones so that they are backward compatible, and add more prefixes so that nearly all the subitems have prefixes for the menus that they are in.  Then sorted the lists again.

As part of (3) I have added a conversion from the old to new IDs for the few that could not be returned to their originals.  These included ones that included spaces (which we removed), and a few others that needed updating.

-----

## Details

The a11y component files are removed from `components/mjs`, since there is no longer an actual component being created.

The explorer and menu configurations now copy the aria localization files (since both components use them and copying them twice doesn't really hurt).

The ts/a11y/aria.ts file is no longer needed, so is removed.

The `a11y/__locales__` directory is moved to `a11y/aria/__locales__` for the new fake `a11y/aria` component, and the `Component.ts` file is modified accordingly.  It now loads the `en` locale data (rathe than doing the in a separate `en.js` file in the component, since there is not component, and we are going to move the rest of the `en.js` component files into their respective `Component.js` files in a future PR anyway).  The `Component.ts` file now also exports a `roleDescPrefix` constant that gives the prefix for the aria role description values (this is used in the explorer and menu code as we will see below).

 In `explorer.ts`, since the values of the radio group variables are now the un-prefixed values (as they were originally), the initial value is changed back to `'math'` and the check for "none" is for the value `'none'`.  There is no longer a need for the `AriaRoleDescription` object, since the prefixes are not part of the data any longer.  Instead, we add the `roleDescPrefix` at the time the role description is being localized.

A small change in `KeyExplorer.ts` prevents a comma from being added to the speech string when there is no description.

In `MJContextMenu.ts`, the main change is to the `findID()` function in order to allow it to locate the menu items via either their un-prefixed or fully prefixed keys.  This is done by testing the name against the id and the id with prefixes removed.  We also have a remapping feature that translates the old keys to the new ones in the few cases where they now differ.  If you think this is overkill, the remapping can be removed.  I suspect that there are not may instances of people using these to adjust the menus, but I have provided some people with examples of how to do this, so there might be.

With the new `findID()`, the calls to this can have their prefixes removed (and the original shorter IDs used again).

In `Menu.ts`, the a11y/aria component data is loaded as `Aria`, and the original role description values are used.  Some localization ids are changed back to their original keys (now that they are prefixed, they no longer conflict, which was why I changed them in the original menu localization PR).

The `radioGroup()` and `radio()` functions now have a new argument that gives the prefix to use for its values.  This means you don't have to include the prefix in each menu entry bub hand, and shorten the strings used to define the menu items.  If this prefix is the empty string, no prefix is added.  If it is `'.'`, then the value is used un-localized as the content (e.g., for the precent strings, and for the renderer names).  If the prefix is the aria role description prefix, the aria localization is used.

Finally, I added prefixes to some additional menu items, since that seemed to be more consistent, and helped get some other keys to be able to use their original ids again.  There are still a few un-prefixed items that could get prefixes if we wanted to be more consistent yet.

In `MenuUtil.js`, I removed a dot from a prefix since those are no longer used.

The ids were changed in the localization files, and the files were sorted again.


ts/a11y/aria/__locales__/Component.ts
```diff
@@ -16,16 +16,23 @@
...
 /**
- * @file  Locale component registration for a11y/explorer
+ * @file  Locale component registration for a11y/aria
  *
  * @author dpvc@mathjax.org (Davide P. Cervone)
  */
 
-import { Locale, namedData } from '../../util/Locale.js';
+import { Locale, namedData } from '../../../util/Locale.js';
+import data from '#ts/a11y/aria/__locales__/en.json' with { type: 'json' };
```
zorkow, line 25: This does not work for me. Should this just be `./en.json`?

zorkow (approved):

PR #1505, opened 2026-06-13
Make sure multi-byte UTF8 characters are handled properly in data-latex attributes.  (mathjax/MathJax#3575)

dpvc:
This PR fixes a problem where unicode characters from planes other than Plane0 are not properly entered in some `data-latex` attributes.  The problem is that for a character that requires multiple string character at the beginning of the attribute, we need to back up the correct number of characters (not just 1).  The fix uses the string length of the unicode character passed to it.

Resolves issue mathjax/MathJax#3575.


zorkow (approved):
lgtm.

PR #1506, opened 2026-06-13
Set box-sizing in dialogs to avoid CSS bleed through that affects close and help buttons.

dpvc:
I've seen several sites that use CSS like

``` css
* {
  box-sizing: border-box;
}
```

and this CSS affects the layout of MathJax's dialog boxes (it distorts the close and help buttons in the top corners).

This PR adds CSS to reset the `box-sizing` to `content-box` so that those button will not be distorted.


zorkow (approved):
lgtm.

PR #1507, opened 2026-06-14
Fix label ids for expressions with forward refs.  (mathjax/MathJax#3562)

dpvc:
This PR fixes a problem where the ID attribute used for a tagged equation is incorrect if the equation includes a forward reference.  This turns out to be due to the `HandleLabel()` base method skipping the setting of the tag's label during the `refUpdate` pass.  The check for `refUpdate` occurs too early, and the fix is to move that check until after the label is set.

This may cause a merge conflict when the locale branch is merged, since the altered code includes a tex error call.  We will have to deal with that at when it occurs.

Resolves issue mathjax/MathJax#3562.


zorkow (approved):
lgtm.

PR #1508, opened 2026-06-15
Ignore braces in array preamble.  (mathjax/MathJax#3565)

dpvc:
This PR adds rules to the array preamble column parser to ignore braces within the preamble (when they aren't used to enclose an argument to one of the column specifications).  This isn't exactly in line with LaTeX, which does seem to process the braces in some way, but I can't figure out what it is, and don't see any documentation for it.  It looks like it might be to allow multi-character column specifiers, but then it also seem to allow multiple copies of a single column specifier, but not mixed ones.  So I'm not sure what it actually does.  In any case, ignoring them seems to handle the situation given in the original issue tracker.

I also added test for this situation.

Resolves issue mathjax/MathJax#3565.

### DIFF
--- a/testsuite/tests/input/tex/Base.test.ts
+++ b/testsuite/tests/input/tex/Base.test.ts
@@ -4014,6 +4014,14 @@ describe('Complete Array', () => {
     expect(tex2mml('\\begin{array}{{{c}}} X \\end{array}')).toMatchSnapshot();
   });
 
+  it('column newline', () => {
+    expect(tex2mml('\\begin{array}{c\nc} X & Y\\end{array}')).toMatchSnapshot();
+  });
+
+  it('column > > c < <', () => {
+    expect(tex2mml('\\begin{array}{>{a}>{b}c<{x}<{y}} X \\end{array}')).toMatchSnapshot();
+  });
+
   it('column brace errors', () => {
     expectTexError(tex2mml('\\begin{array}{c{xx}} X Y \\end{array}')).toBe('Illegal pream-token (xx)');
     expectTexError(tex2mml('\\begin{array}{c{{c}}} X Y \\end{array}')).toBe('Illegal pream-token ({c})');
--- a/testsuite/tests/input/tex/Base.test.ts
+++ b/testsuite/tests/input/tex/Base.test.ts
@@ -4002,6 +4002,10 @@ describe('Complete Array', () => {
     expect(tex2mml('\\begin{array}{> {x} c} X \\end{array}')).toMatchSnapshot();
   });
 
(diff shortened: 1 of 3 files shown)


zorkow:
The current fix does indeed solve the issue, but it seems to be very generous, as it allows any combination of well-braced expressions in the preamble, including empty ones, arbitrarily nested ones etc.

After some experimentation, here are the rules I believe LaTeX follows:
1. If there is a braced entry with multiple characters that is the only entry allowed. E.g., `{{clc}}` is allowed, but `{{cl}{c}}` or `{{cl}c}` are not.
2. The entry can consist of a mix of braced and unbraced single characters, e.g., `{{c}{l}{c}{r}}` or `{{c}l{c}r}`
3. Multiple nesting of braces, like `{{{c}}}` is not allowed, with the exception
4. The only time nested braces are allowed is in the case of `p` entries.

Having said that, I am not sure if it makes sense to expend time on implementing these rules, since we do not know 
1. if these are really the exact rules
2. what the intention of this syntax is or its use cases
3. if not throwing an error on multiple nested `{{{c}}}` or empty `{}` entries is, could be a problem

dpvc:
OK, I looked into the details of the code for the actual LaTeX array package, and it looks like the preamble is tokenized into either single characters or braced groups, and then those are processed as column specifiers (where some can take the following token as a parameter, as in `p{10em}`).  Using braces around single characters works, as you point out, but around other things can lead to some weird error messages (like "missing number", or even parsing on to the `\end{array}` for the *next* `\begin{array}`; very strange).

I have pushed a commit that better matches this behavior, though not the strange messages (it will always be "Illegal param-token").  I also remove one set of braces from the entire expression, if there is such a pair.

I think this matches LaTeX pretty well, and if there are other edge cases, we can deal with them later.

dpvc:
While looking through the documentation for `array.sty`, I found that the order for multiple `>` or `<` entries are done in the reverse order.  This was not being done in MathJax, so I've pushed another commit to reverse the order to match LaTeX.  I also added a test for it.

I also forgot to mention that in the previous commit I added support for newlines in the preamble, and I added a test for that.

zorkow:
Here's one example I found that works in LaTeX, but not in MathJax:

```LaTeX
\left(\begin{array}{{r}c{l}}a&bb& ccc\\ ddd& e & ff\end{array}\right)
```

dpvc:
Right, thanks.  The removal of outer braces was too simplistic.  I had convinced myself that that would work, but of course I should have known better.

I've made a new commit to fix the issue, and add a new test for your example.

zorkow (approved):
lgtm.

PR #1509, opened 2026-06-17
Move loading of `en` locale into `Component.js` files, use `__locales__` for Locale messages, and update tests.

dpvc:
This PR moves the loading of the `en` locale from the component definitions (using the `en.js` files in `components/mjs/...`) to their `__locales__/Component.ts` files instead.  That means that those building MathJax apps using direct imports of the MathJax modules will get the English error messages loaded automatically, and we don't have to have a message about not having called `Locale.setLocale()`.  This means the localization isn't a breaking change.

A second change is that the localization of the messages in the `Locale` object itself are now in a new `ts/util/Locale/__locales__` directory.  This means we don't have to have all the languages' messages in the `Locale` object initially, and since we now load the `en` messages automatically, that has the same effect as having them in line.  The `locale` fake component is the first one registered, so it will always be loaded first when the locale is changed, so locale messages should be available if they are needed once a locale is changed (if not, the fallback to English will at least give a proper, if unlocalized, message).  A copy command for the utility locales is added to `package.json` to accommodate this, and a new `components/mjs/locale` component is created that only copies the files (like the `components/mjs/sre` component does) is added.

The Locale test file is updated to get coverage for some changes in `Locale.ts` and the clean up registered components or messages after tests that set them.

The `ts/util/Locale.ts` file was modified to define a `COMPONENT` like the ones from `__locales__/Component.ts`, but we can't use ` Component.ts` file for `Locale` because the `Locale` object has to be defined before we can call `Locale.registerLocaleFiles()`, like we do on the other `Component.ts` files.  Instead, we import the `en.json` file directly and register it in this file after `Locale` is defined.  We also change `registerLocaleFiles()` to avoid overwriting a previous registration if there was already one made earlier (e.g., if the ui/menu and a11y/explorer components are loaded separately, both of which register the a11y/aria fake component).

The `registerMessages()` method now produces an error message if you try to register messages for a component that isn't registered (addressing your review comment in #1494.

Finally, I shortened the localization keys, and added the new message to the `en.json` and `de.json` files.

I know this looks like a lot of files, but most of them are just deleting the `en.js` files from `components/mjs/...`, removing the imports of those files from the component definitions, and adding the loading and registering of `en` to all the `__locales__/Component.ts` files.


dpvc:
This doesn't work with CJS compiling, so I will make a new PR separately to fix that.

PR #1510, opened 2026-06-17
Combine language and speech language menus, and add empty locales for those that aren't implemented yet.

dpvc:
This PR combines the two language menus in the Mathjax contextual menu into a single one that sets both the locale for SRE and the one used for MathJax's `Locale` framework.  

It looks like it has a lot of files, but most are just empty JSON files for the locales that SRE has but that MathJax doesn't yet support.  This allows the menu to include languages that are available in SRE, but not in the rest of MathJax, to be selected without causing error messages about missing localization files.  I looked at the old v2 localizations to see if we could leverage any of that to fill in some of the new files, but it looked like there was not all that much overlap, even in the TeX messages, since there are now a lot more of them.  I haven't tried to AI generate any of the missing data.

Currently the language menu is in the "Math Settings" submenu, but it's not really a setting about math *per se*.  It might be better to move it up to the top-level menu right under "Math Settings".  Thoughts?


dpvc:
I've moved the Language menu to the top level, as discussed in our developer's meeting.  This should be ready to go at this point.

zorkow (approved):
lgtm.

PR #1511, opened 2026-06-18
Work around issue with Safari 26 not processing math inside otherwise empty containers with overflow: auto.  (mathjax/MathJax#3579)

dpvc:
This PR adds some CSS that works around an issue with Safari 26 where math that is inside a container that is otherwise empty and has `overflow: auto` will not by typeset by the lazy typeset component.  It turns out that the `mjx-lazy` elements are being picked up by the intersection observer, but are marked as `isInsersecting: false`.  This seems to be due to the fact that the `mjx-lazy` node has zero height/width inside a container of zero height that has its own block formatting context (and is clipped to its height), so the intersection contains no pixels?

In any case, changing the `mjx-lazy` element to `display: inline-block` resolves the problem without otherwise affecting the display.

Resolves issue mathjax/MathJax#3579.


zorkow (approved):
lgtm.

PR #1514, opened 2026-06-26
Allow label ids to pass ui/safe filters.  (mathjax/MathJax#3580)

dpvc:
This PR fixes the id filter for `ui/safe` to allow the ids generated by `\label{}` to be used.  Without this PR, the ids needed for `\ref` to link to the labeled equation would be filtered out.

Resolves issue mathjax/MathJax#3580.


zorkow (approved):
lgtm.

PR #1515, opened 2026-06-26
Handle inline line breaking with nested embellished operators.  (mathjax/MathJax#3581)

dpvc:
This PR corrects a problem with in-line line breaking in SVG output involving embellished operators involving `mspace` or `mtext` elements containing spaces.  The problem has to do with the inferred `mrow` that contains the spaces; because the breaks occur *outside* the `mrow`, it should always put its output in the first (and only) parent element rather than advance through the parent elements.

Resolves issue mathjax/MathJax#3581.


zorkow (approved):
lgtm.

PR #1519, opened 2026-07-02
Don't add file:// to windows path in CommonJS files. (mathjax/MathJax#3585)

dpvc:
This PR fixes a problem with using MathJax CommonJS modules in node applications on windows.  Apparently the `file://` should not be added to absolute paths for CommonJS, but only should be for ESM modules.

Resolves issue mathjax/MathJax#3585.

### DIFF
--- a/ts/util/context.ts
+++ b/ts/util/context.ts
@@ -22,11 +22,17 @@
  */
 
 declare const process: { platform: string };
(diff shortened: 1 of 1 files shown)


zorkow (approved):
I could not test. But lgtm.

PR #1523, opened 2026-07-07
Update/publish workflow

zorkow:
* Updates the publish workflow to action versions running with node 24
* Also merges in `master` to make latest tags reachable from `develop` branch


dpvc (approved):
Looks good.

PR #1524, opened 2026-07-07
Handle in-line breaks in a math node that is an embellished operator in SVG output. (mathjax/MathJax#3587)

dpvc:
This PR addresses a problem in SVG output for an inline `math` element that is an embellished operator.  Such an operator should not have line breaks within it (unless they are explicit breaks), but the SVG output was processing them anyway.  The fix is to skip the marking of in-line breaks when the `math` node is an embellished operator.

Resolves issue mathjax/MathJax#3587.


zorkow (approved):
lgtm

PR #1525, opened 2026-07-08
Fix/bussproofs

zorkow:
## New Commands for (partially) complete Package

  `\DisplayProof` / `\DP` was unsupported: proof commands used outside a
  prooftree environment now open an implicit proof tree on the parser stack, and
  `\DisplayProof` terminates and displays it — so plain-TeX-style
  `\AxiomC{$A$}\UnaryInfC{$B$}\DisplayProof` works, including inline within
  other math, e.g. `X + \AxiomC...\DP + Y`. Inside the environment `\DisplayProof`
  is redundant and ignored.  Forgetting it produces a `"Missing \DisplayProof"`
  error; using it with a malformed tree gives the .sty`s `"Proof tree badly
  specified"`.

  All remaining abbreviations: `\AX`, `\UI`, `\BI`, `\TI` (the `\fCenter`
  sequent forms), `\QI`, `\QIC`, `\QuI`, `\QuIC`, `\DP`, plus
  `\EnableBpAbbreviations` as an empty operation as abbreviations are always
  enabled but we don`t want to throw an error.

  `\dottedLine` / `\alwaysDottedLine` works now in CHTML and SVG. `\doubleLine`
  works in CHTML, although with less spacing than in LaTeX, but does not work in
  SVG. What does not work are combinations as in the actual package (e.g.,
  `\dottedLine\doubelLine` leads to double dotted lines)

  `\insertBetweenHyps`: inserts spacing between hyps in the usually empty `mtd` element.

  `\kernHyps`: adds an mspace to the first premise, sliding the rule premises to the right.

  \bottomAlignProof / \centerAlignProof / \normalAlignProof: applied at proof
  finalization as the root mtable`s align attribute.

## Sequent Calculus Fix
 
Alignment along `\fCenter` over multiple sequents was only ever applied to a
pair of sequent lines, and only when the upper line of the pair was an axiom. So
it worked for a two-line axiom-plus-inference proof but not really anywhere
else.

Consequently, I have replaced `adjustSequentPairwise` with `alignSequents`,
which measures every sequent in the chain once, takes the maximum antecedent
width and maximum succedent width, and pads each line's narrower sides up to
those maxima.

### DIFF
--- a/ts/input/tex/bussproofs/BussproofsItems.ts
+++ b/ts/input/tex/bussproofs/BussproofsItems.ts
@@ -96,7 +96,7 @@ export class ProofTreeItem extends BaseItem {
    *
    * @param {MmlNode} tree The proof tree.
    */
-  private alignProof(tree: MmlNode) {
+  protected alignProof(tree: MmlNode) {
     const align = this.getProperty('proofAlign') as string;
     if (!align || align === 'normal') {
       return;
--- a/ts/input/tex/bussproofs/BussproofsItems.ts
+++ b/ts/input/tex/bussproofs/BussproofsItems.ts
@@ -22,6 +22,7 @@
  */
 
 import TexError from '../TexError.js';
+import NodeUtil from '../NodeUtil.js';
 import { BaseItem, CheckType, StackItem } from '../StackItem.js';
 import { MmlNode } from '../../../core/MmlTree/MmlNode.js';
 import Stack from '../Stack.js';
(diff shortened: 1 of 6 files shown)


ts/input/tex/bussproofs/BussproofsItems.ts
```diff
@@ -80,4 +89,33 @@ export class ProofTreeItem extends BaseItem {
...
     return this.create('node', 'mrow', [prefix, tree], {});
   }
+
+  /**
+   * Adjusts the vertical alignment of the finished proof tree with respect
+   * to the baseline, as requested by \bottomAlignProof or \centerAlignProof.
+   *
+   * @param {MmlNode} tree The proof tree.
+   */
+  private alignProof(tree: MmlNode) {
```
dpvc: I prefer `protected` to `private` for these, so that the item class could be subclassed and the subclass could have access to the method, if needed (e.g., if `toMml()` is being replaced and needs to call it).

dpvc (commented):
One minor request, otherwise looks good.  (I admit that I don't know these proof trees, so I can't really check anything myself.)

dpvc (approved):
Looks good.

PR #1528, opened 2026-07-12
Fixes data-latex content for Table Rows

zorkow:
Rows in Tables did not get a meaningful `data-latex` attribute. E.g., for `\begin{array}{c}a\end{array}` we got `data-latex="{c}"`.
This PR fixes this by explicitly collecting the LaTeX elements and then adding them to the final node, combined with `&`.


testsuite/tests/input/tex/__snapshots__/Cases.test.ts.snap
```diff
@@ -3,7 +3,7 @@
 exports[`Cases Numcases 1`] = `
 "<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\begin{numcases}{f(x)=} 1 &amp; if $x &gt; 0$ \\\\ 0 &amp; otherwise \\end{numcases}" display="block">
   <mtable columnalign="right left left left" columnspacing="0em 1em" rowspacing=".2em" data-break-align="top top top" data-latex="\\begin{numcases}{f(x)=} 1 &amp; if $x &gt; 0$ \\\\ 0 &amp; otherwise \\end{numcases}">
-    <mlabeledtr data-latex="{f(x)=}">
+    <mlabeledtr data-latex="1&amp;&amp;">
```
dpvc: The result here is a bit strange.  The second column is blank (probably because `numcases` is handling the text-mode material internally), but the second `&amp;` is wrong.  Is this due to the `mlabeledtr` having an extra child for the tag?

The same is true for all the other tests as well.

testsuite/tests/input/tex/__snapshots__/Cases.test.ts.snap
```diff
@@ -3,7 +3,7 @@
...
       <mtd id="mjx-eqn:1">
```
dpvc, line 7: Should this `mtd` have a `data-latex` attribute for the tag?

testsuite/tests/input/tex/__snapshots__/Colortbl.test.ts.snap
```diff
@@ -3,15 +3,15 @@
 exports[`Colortbl cellcolor 1`] = `
 "<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\begin{array}{cc} a &amp; \\cellcolor{red} b \\\\ \\cellcolor{yellow} c &amp; d \\end{array}" display="block">
   <mtable columnspacing="1em" rowspacing="4pt" columnalign="center center" data-frame-styles="" framespacing=".5em .125em" data-latex="\\begin{array}{cc} a &amp; \\cellcolor{red} b \\\\ \\cellcolor{yellow} c &amp; d \\end{array}">
-    <mtr data-latex="{cc}">
+    <mtr data-latex="a&amp;b">
```
dpvc: This is missing the `\cellcolor{red}` before the `b`.  Can that be captured as well?

testsuite/tests/input/tex/__snapshots__/Colortbl.test.ts.snap
```diff
@@ -3,15 +3,15 @@
...
       <mtd>
         <mi data-latex="a">a</mi>
       </mtd>
       <mtd mathbackground="red">
         <mi data-latex="b">b</mi>
       </mtd>
     </mtr>
-    <mtr data-latex="{cc}">
+    <mtr data-latex="c&amp;d">
```
dpvc: Similarly, `\cellcolor{yellow}` is missing before `c`.

testsuite/tests/input/tex/__snapshots__/Colortbl.test.ts.snap
```diff
@@ -65,7 +65,7 @@ exports[`Colortbl cellcolor in preamble 1`] = `
 exports[`Colortbl cellcolor late 1`] = `
 "<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\begin{array}{cc} a &amp; b \\cellcolor{red} \\end{array}" display="block">
   <mtable columnspacing="1em" rowspacing="4pt" columnalign="center center" data-frame-styles="" framespacing=".5em .125em" data-latex="\\begin{array}{cc} a &amp; b \\cellcolor{red} \\end{array}">
-    <mtr data-latex="{cc}">
+    <mtr data-latex="a&amp;\\cellcolor{red}">
```
dpvc: Here you capture `\cellcolor{red}` but missing the preceding `b`.  Of course, that is probably due to the `b` being missing from the `data-latex` attribute of the `mi` at line 73 below.

testsuite/tests/input/tex/__snapshots__/Newcommand.test.ts.snap
```diff
@@ -15,18 +15,18 @@ exports[`Nested Environments Environments Intermixed 1`] = `
 exports[`Nested Environments Environments Nested 1`] = `
 "<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\newenvironment{boxed}{\\begin{array}{|c|c|}\\hline}{\\\\\\hline\\end{array}}\\begin{boxed}\\begin{boxed}a&amp;b\\\\c&amp;d\\end{boxed} &amp; X \\end{boxed}" display="block">
   <mtable columnspacing="1em" rowspacing="4pt" columnalign="center center" columnlines="solid" framespacing=".5em .125em" frame="solid" data-latex="{array}">
-    <mtr data-latex="{|c|c|}">
+    <mtr data-latex="{array}&amp;X">
```
dpvc: The `{array}` doesn't seem right, here.

testsuite/tests/input/tex/__snapshots__/Tag.test.ts.snap
```diff
@@ -89,7 +89,7 @@ exports[`TagAll Align 1`] = `
 exports[`TagAll Align Two labels 1`] = `
 "<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\begin{align}a=b\\label{A}\\\\ c&amp;=d \\label{B}\\end{align}" display="block">
   <mtable displaystyle="true" columnalign="right left" columnspacing="0em" rowspacing="3pt" data-break-align="bottom top" data-latex="\\begin{align}a=b\\label{A}\\\\ c&amp;=d \\label{B}\\end{align}">
-    <mlabeledtr>
+    <mlabeledtr data-latex="a=\\label{A}">
```
dpvc: The `b` is missing before the `\label`, here.  Similarly in the tests below.

testsuite/tests/input/tex/__snapshots__/Amscd.test.ts.snap
```diff
@@ -3,7 +3,7 @@
 exports[`AmsCD AmsCD-1 1`] = `
 "<math xmlns="http://www.w3.org/1998/Math/MathML" data-latex="\\begin{CD}A @&gt;a&gt;&gt; B\\\\@VVbV @VVcV\\\\C @&gt;d&gt;&gt; D\\end{CD}" display="block">
   <mtable columnspacing="5pt" rowspacing="5pt" displaystyle="true" data-latex="\\begin{CD}A @&gt;a&gt;&gt; B\\\\@VVbV @VVcV\\\\C @&gt;d&gt;&gt; D\\end{CD}">
-    <mtr>
+    <mtr data-latex="A&amp;&amp;B">
```
dpvc, line 6: I'm wondering if the blank column can capture the arrows somehow.  These are constructed as raw MathML, so that may be hard to get.

dpvc (commented):
This is a big improvement, but there still seem to be some issues.  Those are probably actually problems with the attributes of the child nodes in many cases.  Perhaps they can be improved?  (I haven't looked into the problem, but it may be due to modifying the `parser.string` value for substitutions.)

zorkow:
Thanks for looking at these tests in more detail. I have to admit I mainly focussed on the expressions in `Base` and `Ams`. I believe for tag elements (coming from `\label` commands) we probably need to skip them when accumulating the row. 
I am also wondering if it would make sense to have `latex` labels on the `mtd` elements.

zorkow:
## Summary (cleaned up and greatly enhanced by co-pilot)

  Fixes several cases where the `data-latex` attribute on `mtable` (and related nodes)
  did not reflect the actual LaTeX source, plus a couple of related source-tracking bugs  uncovered along the way.

  ## Problems

  1. **Column templates garble `data-latex`.** Templates like `>{...}` rewrite the
  parser's string entry-by-entry as it goes, so the array's `data-latex` ends up built
  from a leftover fragment instead of the real source.
  2. **Nested arrays lose their source.** When a table is nested inside another
  environment (e.g. a tagged equation, or another array), `data-latex` was set from just  the `\end{name}` argument, producing a bogus value like `{array}`.
  3. **Non-rendering macros clobber the preceding node's source.** `\label`, `\tag`,
  `\notag`, and colortbl's `\cellcolor`/`\rowcolor`/`\columncolor` produce no Mml node
  of their own, but the generic source-tracking in `TexParser.updateResult` was still
  attaching their text to whatever node came before them.
  4. **`\ref`/`\eqref` recorded the wrong source.** The generic tracking captured the
  raw `\ref{...}` call instead of the resolved tag text actually used to build the
  reference node.
  5. **Tagged/labeled rows and `multline` rows had no correct row-level `data-latex`.**
  6. **`\newenvironment`-defined environments could invalidate cached source positions**  in an enclosing array when their body is spliced into the parser string mid-parse.
  
  ## Fixes

  - **Raw source capture for arrays** (`ArrayItem.captureLatex` / `setTableLatex`,
  `FilterUtil.fixArrayLatex`): at `\begin{name}`, grab the true
  `\begin{name}...\end{name}` text up front (or stash a start position if the end isn't
  in the string yet, e.g. for `\newenvironment`). At table-close, recover this text and
  store it as `rawLatex`/`rawLatexName` on the `mtable`. A new postprocessing filter
  then replaces `data-latex` on the table only when the existing value looks broken
  (matches the `{name}` or dangling-`\end` patterns), so already-correct values are left  untouched. Falls back to reconstructing from row-level `data-latex` if nothing else
  is available.
  - **`envSplice` generation counter**: since expanding a `\newenvironment` splices its
  body into `parser.string` (invalidating any previously-stashed absolute positions), a
  global counter is bumped on every such splice and snapshotted alongside a stashed
  start position, so stale positions are detected and skipped rather than misused. 
  - **Blanking `\end{name}` with whitespace**: When a user-defined environment
  closes, `BeginEnv` takes everything already consumed and appends `edef` to it,
  continuing to parse from the same position. Leaving that literal `\end{name}`
  text in place leads to incorrect `data-latex`. Simply deleting it messes up
  subsequent parsing. Replacing `\end{name}` with an equal-length run of
  whitespace and removing it later in `setTableLatex` fixes both problems.
  - **`TexParser.skipLatex()`**: a one-shot flag that suppresses the next
  `updateResult()` source attachment, used by `\label`, `\tag`(the tagging path),
  `\notag`, and colortbl's color macros so they no longer overwrite the preceding node's  `data-latex`.
  - **Row-level LaTeX tracking** (`rowLatex` on `ArrayItem`/`MultlineItem`, `colorLatex`  on `ColorArrayItem`): cell/row source is accumulated as rows are built, including
  folding in colortbl macro text and label/tag content, and used to set
  `data-latex`/`data-latex-item` on `mtr`/`mlabeledtr` rows (also used in `Tags.ts` for
  the label row). 
  - **`\ref`/`\eqref` fix**: builds the reference node's content via an explicit
  internal `TexParser` call and sets its `data-latex`/`data-latex-item` directly from
  the resolved tag text, then calls `skipLatex()` so generic tracking doesn't overwrite
  it with the raw macro call.
  - **`NodeUtil.getChildrenLatex`**: shared helper (replacing the old inline logic in
  `ArrayItem.cellLatex`) that prefers a child's `rawLatex` property over its
  `data-latex` attribute when reconstructing source from content nodes.
  
## New `setProperty` keys

  | Property | Set where | Purpose |
  |---|---|---|
  | **`rawLatex`** | `ArrayItem.captureLatex()`, and copied onto the `mtable`/wrapping `mstyle` node in `setTableLatex()` / `EndTable()` | Caches the raw, unmodified LaTeX source of the array/table environment (`\begin{name}...\end{name}`), captured before column-template substitutions or nested-environment parsing can corrupt the parser's string. | 
  | **`rawLatexName`** | Alongside `rawLatex` everywhere it's set | The environment name  (`array`, `matrix`, etc.), needed to detect "broken" `data-latex` values and to rebuild `\begin{name}...\end{name}` wrappers. |
  | **`rawLatexStart`** | `captureLatex()`, only when the matching `\end{name}` can't be  found up front (e.g. `\newenvironment`-defined environments) | Parser position right after `\begin{name}`, so the source can be sliced out later once the environment has actually finished parsing. |
  | **`rawLatexGen`** | `captureLatex()`, alongside `rawLatexStart` | Snapshot of the `envSplice` generation counter, used to detect whether the stashed `rawLatexStart` position has been invalidated before trusting it. |

  ### How they fit together

  1. On `\begin{name}`, `captureLatex()` grabs the raw source immediately, or stashes a
  start position + generation if the end isn't available yet. 
  2. On table close, `setTableLatex()` uses `rawLatex` if captured, else recovers it
  from `rawLatexStart`/`rawLatexGen` (if the generation still matches), else 
  reconstructs it from the rows' own `data-latex`.
  3. `FilterUtil.fixArrayLatex` (a new postprocessing filter in the `TeX.ts` pipeline)
  walks all `mtable` nodes and replaces `data-latex` with the cached `rawLatex` only
  when the existing value looks broken/missing (`FilterUtil.isBrokenArrayLatex`).

zorkow:
PS:

As discussed earlier, what is still missing:
* Dealing with the prefix rows in the `cases` package
* Also there is no solution for `table` expressions generated with `amscd` 

The `findEnvEnd` parsing method, to find `\end{name}` if it exists, is very similar to a number of other parsing functions we have (in fact I started by simplifying `Entry`) that we already have. I wonder if there is some way of refactoring them into a single utility.

testsuite/tests/input/tex/__snapshots__/Amscd.test.ts.snap
zorkow: Future PR.

testsuite/tests/input/tex/__snapshots__/Cases.test.ts.snap
zorkow: This could potentially be resolved if we had `data-latex` entries in the `mtd` elements. But I leave that for a separate PR.

testsuite/tests/input/tex/__snapshots__/Cases.test.ts.snap
zorkow: Dito. For the use case of SRE translations to Euro-Braille cells are handled with a speech rule.

testsuite/tests/input/tex/__snapshots__/Colortbl.test.ts.snap
zorkow: fixed

testsuite/tests/input/tex/__snapshots__/Colortbl.test.ts.snap
zorkow: fixed.

testsuite/tests/input/tex/__snapshots__/Colortbl.test.ts.snap
zorkow: fixed

testsuite/tests/input/tex/__snapshots__/Newcommand.test.ts.snap
zorkow: fixed

testsuite/tests/input/tex/__snapshots__/Tag.test.ts.snap
zorkow: fixed

ts/input/tex/newcommand/NewcommandMethods.ts
```diff
@@ -267,11 +267,20 @@ const NewcommandMethods: { [key: string]: ParseMethod } = {
...
           const rest = parser.string.slice(parser.i);
-          parser.string = ParseUtil.addArgs(
-            parser,
-            parser.string.substring(0, parser.i),
-            edef
+          let prefix = parser.string.substring(0, parser.i);
+          // Here `\end{name}` has already been parsed, but left in place it
+          // ends up in the LaTeX source. Remove it using whitespace to keep the
+          // counter correct.
+          const escaped = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
dpvc: There is a `quotePattern()` function in `ts/util/string.ts` for doing this.

ts/input/tex/newcommand/NewcommandMethods.ts
```diff
@@ -89,17 +89,17 @@ const NewcommandMethods: { [key: string]: ParseMethod } = {
     const def = parser.GetArgument(name);
     !(params instanceof Array)
       ? // @test Def DoubleLet, DefReDef
-        NewcommandUtil.addMacro(parser, cs, NewcommandMethods.Macro, [
-          def,
-          params,
-        ])
+      NewcommandUtil.addMacro(parser, cs, NewcommandMethods.Macro, [
```
dpvc: Prettier will put these spaces back.

ts/input/tex/FilterUtil.ts
```diff
@@ -341,6 +341,62 @@ const FilterUtil = {
...
+   * parsing:
+   *
+   * - Column templates rewrite the parser's string entry by entry, so we end up
+   *   using a leftover fragment instead of the actual source.
+   *
+   * - When the table is nested inside another environment (e.g., inside a
+   *   tagged equation, or another array), `addLatexItem` sets `data-latex` from
+   *   the `\end{name}` argument alone (e.g., `{array}`), which is wrong.
+   *
+   * This filter only replace a `data-latex` value that looks missing or
```
dpvc: replace => replaces

ts/input/tex/base/BaseItems.ts
```diff
@@ -1362,12 +1378,160 @@ export class ArrayItem extends BaseItem {
...
+   *     -1 if no matching end was found.
+   */
+  private static findEnvEnd(str: string, name: string): number {
+    const escaped = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+    const begin = new RegExp(`^\\\\begin\\s*\\{${escaped}\\}`);
+    const end = new RegExp(`^\\\\end\\s*\\{${escaped}\\}`);
+    let depth = 0;
+    let nested = 0;
+    let i = 0;
+    while (i < str.length) {
```
dpvc, line 1488: PR #1539 has replaced loops that look through the parser string by hand so that they use `GetNext()` instead.  That way, they will handle comments properly, which this loop doesn't.  It would be good to adjust this loop now, so that we don't forget it later when #1539 lands.

The alternative would be to handle comments here, taking into account the `legacyComments` option (which isn't in place yet, but will be when that PR is merged).  But I tried to reduce the number of places where you had to take comments into account by hand, so it would be best to use `GetNext()` instead.  It works like `str.charAt(i)`, but skips spaces and comments, and takes multi-character unicode characters into account (so if you use `c = parser.getNext()`, you would use `parser.i += c.length` to move past `c`).

ts/input/tex/base/BaseItems.ts
```diff
@@ -1362,12 +1378,160 @@ export class ArrayItem extends BaseItem {
...
+   * right after the corresponding `\begin{name}` (skipping nested environments
+   * of that name).
+   *
+   * @param {string} str The string to search.
+   * @param {string} name The environment name to look for.
+   * @returns {number} The index right after the matching `\end{name}`, or
+   *     -1 if no matching end was found.
+   */
+  private static findEnvEnd(str: string, name: string): number {
+    const escaped = name.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
dpvc: This could also use the `quotePattern()` function from `ts/util/strings.ts`.

dpvc (changes requested):
This looks a lot better, and is a big help for handling tables.

I do recommend an important change in the `findEnvEnd()` method to make it compatible with #1539 and the handling of comments.  The other comments are minor.

ts/input/tex/newcommand/NewcommandMethods.ts
zorkow: Actually prettier did a whole number of changes, but oddly enough it did not change this one.

dpvc: Weird, it did for me when I checked before making the comment.

dpvc (approved):
This looks good now.  If you merge now, I'll make the changes needed for the loop in the PR for the comments support.

PR #1530, opened 2026-07-17
Fix magnification of HTML embedded in MathML or TeX.

dpvc:
This PR fixes a problem with the keyboard or mouse magnification of expressions that include HTML embedded in the MathML or TeX expressions.  The HTML was not being properly enlarged in the magnification window.  This is due to the use of `px` font sizing in the CHTML output, and a pad interaction with the scaling in SVG output.

The solution for CHTML is to override the `getFontStyles()` method to use a scaling in `em` units (in this case, the parent elements already do the relative scaling, so `1em` keeps it at that scale).  For SVG, setting `font-size` to initial and leaving the `px` values seems to do the trick.

### DIFF
--- a/ts/output/chtml/Wrappers/semantics.ts
+++ b/ts/output/chtml/Wrappers/semantics.ts
@@ -243,7 +243,9 @@ export interface ChtmlXmlNodeNTD<N, T, D>
       ChtmlDelimiterData,
       ChtmlFontData,
       ChtmlFontDataClass
-    > {}
+    > {
+  getFontStyles(): {'font-family': string, 'font-size': string};
+}
 
 /**
  * The ChtmlXmlNodeClass interface for the CHTML XmlNode wrapper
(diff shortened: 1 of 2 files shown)


ts/output/chtml/Wrappers/semantics.ts
```diff
@@ -243,7 +243,9 @@ export interface ChtmlXmlNodeNTD<N, T, D>
       ChtmlDelimiterData,
       ChtmlFontData,
       ChtmlFontDataClass
-    > {}
+    > {
+  getFontStyles(): {'font-family': string, 'font-size': string};
```
zorkow: Missing jsdoc.

ts/output/chtml/Wrappers/semantics.ts
```diff
@@ -313,6 +315,12 @@ export const ChtmlXmlNode = (function <N, T, D>(): ChtmlWrapperClass<N, T, D> {
       this.dom = [this.adaptor.append(parents[0], this.getHTML()) as N];
     }
 
+    public getFontStyles(): { 'font-family': string; 'font-size': string } {
```
zorkow: Missing `@override` jsdoc.

zorkow (approved):
Missing jsdoc. Otherwise fine.

PR #1531, opened 2026-07-22
Allow entity definitions to load synchronously, and add tests for it.

dpvc:
This PR adds the ability to synchronously load entity extensions from node applications that have `asyncIsSyncronous` set.  This is similar to the synchronous loading of dynamic font data font.  A number of new tests are added, and the current single function is broken into several pieces.

It also removes a deprecated use of `RegExp.$1`.


dpvc:
Sorry, accidentally deleted this branch while making a fix, so had to open a new PR.

PR #1532, opened 2026-07-22
Allow entity definitions to load synchronously, and add tests for it.

dpvc:
This PR adds the ability to synchronously load entity extensions from node applications that have `asyncIsSyncronous` set.  This is similar to the synchronous loading of dynamic font data font.  A number of new tests are added, and the current single function is broken into several pieces.

It also removes a deprecated use of `RegExp.$1`.

### DIFF
--- a/testsuite/tests/util/Entities.test.ts
+++ b/testsuite/tests/util/Entities.test.ts
@@ -1,25 +1,64 @@
 import { describe, test, expect } from '@jest/globals';
 import * as Entities from '#js/util/Entities.js';
-import { handleRetriesFor } from '#js/util/Retries.js';
+import { mathjax } from '#js/mathjax.js';
 import '#js/util/asyncLoad/esm.js';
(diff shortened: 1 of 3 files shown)


zorkow (approved):
lgtm.

But I am wondering if the jsdoc in `mathjax.ts`
```Typescript
  /**
   * When asyncLoad uses require(), it actually operates synchronously and this is true
   */
  asyncIsSynchronous: false,
```
should be updated. `asyncLoad` could be a function using `readFileSync` in Node.

dpvc:
> `asyncLoad` could be a function using `readFileSync` in Node.

Only if the user replaces `asyncLoad`; of our `ts/utility/asyncLoad/*.ts` files, the only one that sets `asyncIsSynchronous` to `true` is `node.ts`, which uses `require()`.  None of our versions use `readFileSync()`.

But I will modify the comment to be more generic.

PR #1533, opened 2026-07-22
Clear the SVG stylesheet during reset.

dpvc:
This PR adds a line to the SVG output jax's `reset()` function, so that the stylesheet will be regenerated after a reset.  This helps support reusing the SVG output jax in sequentially created math documents (so that in the second document, the SVG output jax doesn't think it already has a stylesheet in place)


zorkow (approved):
lgtm.

PR #1537, opened 2026-08-05
Make sure SRE options are set if only a convert call is used.

dpvc:
This PR fixes some comment typos, and makes sure that the SRE options are in place if a conversion action is taken before the page is typeset.  E.g., if `MathJax.tex2chtml()` or `doc.convert()` is called from a command-line application that just converts math strings without processing a page as a whole.


dpvc:
I removed the commit for the `asyncLoad()` change as I already have it in another branch whose PR will be made shortly.

zorkow (approved):
lgtm.

PR #1538, opened 2026-08-05
Improve handling of webworker failures.

dpvc:
This PR sets things up to better handle failures in landing the speech worker or its maps, rather than just crash.

In `KeyExplorer.ts`, we remove initial `, ` (for when the speech is empty, as it will be if speech has failed).  The `Word/Math` message is used when there should be speech, but speech hasn't been attached (i.e., speech has failed to be produced).  This way, if the page is being read, you will at least get "math" as the output rather than nothing at all.

In `speech.ts`, the errors are always strings, not Error objects, so fixed that.  In `speechError()`, we only report non-empty messages (this is to help avoid getting a speech error for every math expression on the page when SRE fails; see `WebWorker.ts` for more).

In `GeneratorPool.ts`, if the worker failed to start, we produce a message indicating that rather than no speech or braille.  An alternative (in a comment) is to use the text content of the selected element, but that might be pretty ugly.  On the other hand, it would at least be specific to the node being walked.

In `WebWorker.ts`, we only report a termination warning once if the warning is the same.  That way, if the worker doesn't load, you don't get a redundant warning for every expression on the page, but just one warning.  We also don't throw an error when terminating if the worker hasn't started (since we need to terminate the tasks).  Finally, we add some commands to handle failure to start the worker, and failure to load a map file.  We check to see if the error already includes the path that failed, and if not add it ourselves.

The `speech-worker.ts` file adds error handling to the `getMap()` methods so that errors are reported back to MathJax rather than just logged to the console.

Similarly, the HTMLAdaptor and NodeMixin adaptor are modified to include error trapping when the worker is started up.

Finally, the `asyncLoad()` function is fixed to handle `node:` references, which was lost somewhere along the line in the locale branch.


ts/a11y/speech/GeneratorPool.ts
```diff
@@ -212,6 +213,10 @@ export class GeneratorPool<N, T, D> {
    */
   public getLabel(node: N, _center: string = '', sep: string = ' '): string {
     const adaptor = this.adaptor;
+    if (!this.webworker.ready) {
+      // return document.querySelector('.mjx-selected').textContent;
```
zorkow, line 217: Remove comment.

ts/a11y/speech/GeneratorPool.ts
```diff
@@ -231,6 +236,10 @@ export class GeneratorPool<N, T, D> {
    */
   public getBraille(node: N): string {
     const adaptor = this.adaptor;
+    if (!this.webworker.ready) {
+      // return document.querySelector('.mjx-selected').textContent;
```
zorkow, line 240: Remove comment.

zorkow (approved):
lgtm.

ts/a11y/speech/GeneratorPool.ts
dpvc: These comments are in connection to my comment:

> An alternative (in a comment) is to use the text content of the selected element, but that might be pretty ugly. On the other hand, it would at least be specific to the node being walked.

I meant that as a question about whether to do it the way I have it, or to use the text content instead.  I left those comment in there in case you wanted to try the other alternative.  Let me know which you think is best.

dpvc: I'm leaving these in for now, as discussed in our developer's meeting.

dpvc:
I have added a live-region message when the web worker fails to start up.  This again brings up the question about how to add new localized messages.  This branch only has the English and German localizations, but develop now has AI-generated versions for the other locales.  How do we handle that situation?  Should I merge develop into this branch, and add the missing key into those other files?

dpvc:
In any case, please re-review the changes for the live region.

ts/a11y/speech/__locales__/de.json
```diff
@@ -6,6 +6,7 @@
   "Worker/Cancelled": "Task %1 abgebrochen",
   "Worker/Failed": "MathJax Webworker konnte nicht geladen werden: %1",
   "Worker/FailedFrom": "MathJax Webworker konnte nicht von %2 geladen werden: %1",
+  "Worker/Warning": "Warnung: MathJax kann für die mathematischen Formeln auf dieser Seite keine Sprachausgabe erzeugen.",
```
zorkow:
```suggestion
  "Worker/Warning": "Warnung: MathJax kann für mathematische Formeln auf dieser Seite keine Sprachausgabe erzeugen.",
```

zorkow:
> I have added a live-region message when the web worker fails to start up. This again brings up the question about how to add new localized messages. This branch only has the English and German localizations, but develop now has AI-generated versions for the other locales. How do we handle that situation? Should I merge develop into this branch, and add the missing key into those other files?

We let CrowdIn handle this as soon as this PR has landed on develop.

zorkow (approved):
Changes look good. Just one suggestion for the German message.

dpvc:
> We let CrowdIn handle this as soon as this PR has landed on develop.

So I assume that means that CrowdIn takes care of propagating the new key to the other files, and marks them as needing translation?

PR #1539, opened 2026-08-05
Properly handle comments in arguments and other places, and spaces in macro templates. (mathjax/MathJax#3577)

dpvc:
This PR brings MathJax's handing of percent-sign comments for macros more in line with actual LaTeX, both in terms of calling a macro and defining one.  In the past, percent signs could be part of a template macro's delimiter strings, and multiple spaces counted as significant in the template, but in actual LaTeX, percent signs act as comments within the template (the content is not part of the template) and multiple spaces are collapsed.  Also, when calling a macro, percent signs were not properly acting as comments within the arguments.

These issues are resolved here, but since this could be a breaking change to existing content, two new options are added, `legacyComments` and `legacyTemplateMacros` to control whether the new processing or the legacy processing is to be performed.

Resolves issue mathjax/MathJax#3577.

-----

## Details

A number of tests needed to be added to cover the new functionality, and some old tests were updated that operate differently under the new comment processing.  In `ConfigMacros.test.ts` the `runMacroTests()` method doesn't need to run two tests, since the results are compared to snapshots rather than to each other, as they were in the past.  That means the control test doesn't actually have to be performed, so it (and its parameter) is removed.

The new options are added in the `input/tex.ts` file.

In `input/tex/NodeUtil.ts`, `setAttribute()` and `setProperty()` now use the more-appropriate `Property` type rather than `Args`, since that's what the node's property object uses, and because `Args` needs to be more general in order to handle the new approach to template macros.

Much of the work for percent signs is done in `input/tex/TexParser.ts`.  The `GetNext()` macro now handles removing comments, so that everything that uses `GetNext()` will properly handle comments.  (We then adjust other functions that used to walk the TeX string by hand to now use `GetNext()`.)  A new `clipComment()` function is used to remove a comment from the input string.  A new `unescapePercents()` is defined for use in handling arguments that are not further processed as TeX, like the contents of `\mmlToken` commands, so you can do `\mmlToken{mo}{\%}` to get `<mo>%</mo>`, or `\mmlToken{text}{\\\%}` to get `<mtext>\%</mtext>`.

The `GetNext()`, `GetArgument()`, and `GetBrackets()` functions take a new boolean argument that determines if the new percent-processing rules are to be used.  (This allows the `legacyComments` option to get the old behavior.)

The extra `return` values are to avoid a warning that the function doesn't always have a return value.  I don't like it, but I don't think it hurts.

In `input/tex/Types.ts`, the `Args` type is extended to allow `string[]` values.  These are needed for template macros, which now use an array of tokens to look for rather than a fixed string.  That allows multiple spaces to be collapsed, and comments to be removed from both the template itself, and the usage within a macro call.

In `input/tex/base/BaseItems.ts`, the `getEntry()` macro, which scans ahead for the end of the entry, now skips comments properly.  (This isn't currently controlled by `legacyComments`, but I suppose it should be.)

In `input/tex/base/BaseMethods.ts`, the `\mmlToken` argument handles escaped percents.  The loop in the `Entry()` method is modified to use `GetNext()` in order to handle comments.  That means that we have to switch the roles of `i` and `parser.i`, since `GetNext()` uses `parser.i`.

The `input/tex/bbox/BBoxConfiguration.ts` file handles escaped percents in its argument (because it allows CSS specifier that might need to be percentages).

The loop in `input/tex/cases/CasesConfiguration.ts` is modified to use `GetNext()` as above, again requiring that `i` and `parser.i` be swapped.

In `input/tex/configmacros/ConfigMacrosConfiguration.ts`, if a macro is a templated one (def[2] is an array), we tokenize the elements in the array, since that is how the macro templates are now stored.

In `input/tex/html/HtmlMethods.ts`, the arguments are now processed for escaped percents, other than the CSS id, which is take verbatim.  (I'm not sure why I did that, but it probably should be unescaped as well.  Perhaps I just missed it when I added the service routine.)

In `input/tex/newcommand/NewcommandMethods.ts`, I replace a `x ? y() : z()` construct with an if-then.  The other change is to use template token arrays rather than template strings.  This uses the updates `getParameter()` call rather than the (removed) `MatchParam()` method (see `NewcommandUtil.ts` below).

The main changes are in `input/tex/newcommand/NewcommandMethods.ts`.  Here we add methods for tokenizing a string that acts as one of the template separators.  The tokens newlines converted to space, multiple spaces collapsed to one, and only a single (actual) space at the end.

Uses of `this` were changes to `NewcommandUtil` for consistency.

The `GetTemplate()` method is modified to return the parameter separators as token arrays instead of a single string.  The loop is reorganized a bit to accommodate that.

The `GetParameter()` method is likewise modified in order to accept arrays of tokens instead of strings, and to look through the input string by tokens rather than looking for the exact string.  That lets it handle comments and multiple spaces.

The `SetOptions` method in `input/tex/setoptions/SetOptionsConfiguration.ts` handles escaped percent signs, since the option may need a percentage (e.g., for multline widths).

Finally, the loop in `input/tex/verb/VerbConfiguration.ts` is modified to use `GetNext(false)`.  This allows you to do `\verb|%|` to get a monospaced percent sign, but note that you can't do this inside a macro definition, or within an argument that is passed to a macro, as the percent would eb traded as a comment in those cases.  This is consistent with actual LaTeX.


ts/input/tex.ts
```diff
@@ -74,6 +74,8 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
     maxTemplateSubtitutions: 10000,
     // math-style to use for Latin and Greek letters
     mathStyle: 'TeX', // one of TeX, ISO, French, or upright
+    legacyComments: false, // true to handle comments as ealier v4 versions (not properly processed in arguments)
```
zorkow:
```suggestion
    legacyComments: false, // true to handle comments as earlier v4 versions (not properly processed in arguments)
```

ts/input/tex.ts
```diff
@@ -74,6 +74,8 @@ export class TeX<N, T, D> extends AbstractInputJax<N, T, D> {
...
+    legacyMacroTemplates: false, // true to handle spaces in macro as ealier v4 versions (not collapsed)
```
zorkow:
```suggestion
    legacyMacroTemplates: false, // true to handle spaces in macro as earlier v4 versions (not collapsed)
```

dpvc:
I just retested this PR with the newline versus space issue that we discussed face-to-face this morning, and it worked properly, so I must have not changed the branch or recompiled (or perhaps not reloaded the lab) properly.  It works as expected now, and there are tests in the `Newcommand.test.ts` file for that, so it does work as expected.

Also, I tried commenting out the `GetNext()` command that we were discussing, and only one test fails, and it is one where the initial spaces are in the `\def` but not the call to the defined macro, so that illustrates the need for that `GetNext()` so that the initial spaces of the template are skipped, as they should be.

I will make the changes in this branch for the loop I flagged in #1528 once you have merged it in, since those changes relate to the comment handling here.

zorkow (approved):
Thanks for looking into the `def` issue.
Otherwise, only a couple of typos and a merge conflict.

dpvc:
I've resolved the merge conflicts, but am waiting to merge until I add the changes to the loop in #1528.

dpvc:
I've adjusted the loop discussed in #1528 to use `GetNext()` so that comments are properly handled.  I've also changed it to match all environment `\begin`/`\end` pairs, since a mismatched `\end` will end a `\begin` (with an error) so there is no need to look further.  Similarly, a mismatched close brace will do the same.  of course, these are edge cases.  Another one would be having the `\end` hidden in a macro, but there is really nothing to be done about that.  These kinds of look-aheads are notoriously fragile.

In any case, do you want to re-review?

dpvc:
Note that the tests are not passing because the changes in b57a05138 are not in this branch (I suppose those should have been made to the `develop` branch, not `feature/locale_others`, so they could be merged into pending branches like this).

ts/input/tex/base/BaseItems.ts
```diff
@@ -1475,50 +1478,45 @@ export class ArrayItem extends BaseItem {
...
-    let i = 0;
-    while (i < str.length) {
-      const c = str.charAt(i);
+   *     -1 if no matching end was found or if there is incorrect nesting of
+   *     environments or braces.
+   */
+  protected findEnvEnd(parser: TexParser, name: string): number {
+    const nesting = [name];
+    let c;
+    while ((c = parser.GetNext())) {
```
zorkow, line 1492: I think the issue is here: `GetNext` calls `clipComment` which modifies `parser.string`. But that's used in line [1421](https://github.com/mathjax/MathJax-src/pull/1539/changes/8e983f6a5511e7a964520a32650c88c55a935ce9#diff-3b76864a0c40d4c4dec7445ac69c7bc2386c2679f200ae53cdeb8ac1c9144621R1421).

zorkow (commented):
The issue I can see with this approach is that we are now clipping the comment, which affects the `data-latex` output. Here's an example:
```LaTeX
\begin{array}{cc}
a& b %comment
\\
c&d
\end{array}
```
The `%comment` is now missing on the `data-latex` attribute of the `mtable` element.

On the other hand, in the old version, the `b` is dropped from the expression in the `data-latex` attribute, which is arguably worse. After all the comment is still visible in the top level `data-latex`.

dpvc:
> The issue I can see with this approach is that we are now clipping the comment, which affects the `data-latex` output. 

Note that this is happening in lots of other cases already.  For example

``` latex
\text% a command
{the text}
```

produces 

``` xml
<mtext data-latex="\text{the text}">the text</mtext>
```

If the `data-latex` attributes are to be used by SRE or other code, it seems that this is actually an advantage, as this makes the LaTeX more useful, since the consuming code would not have to handle comments itself.  I see no reason why the comments need to be retained in the `data-latex` attributes.  They aren't meant to reproduce the original LaTeX expression (which, as you point out, is in the top-level `data-latex`), but to represent LaTeX that produced the output that the node contains.  Comments are not part of that.

One of the purposes of the changes to `GetNext()` was precisely to remove the comments so that other parts of the code don't have to worry about them.  In order to retain this in `data-latex`, this loop (and every other one that was modified in this PR) would need to take comments into account, and would have to take the `legacyComments` option into account.  That would lead to lots of redundant code, as well as potential errors and future problems as similar loops are created later without thinking about comments.

So I think this is the correct result.  Let me know if you disagree.

PR #1541, opened 2026-08-07
Fix existing linting errors

zorkow:
Fixes existing warning regarding `jsdoc/require-returns-check`. They occur in functions that both return a value or throw an exception, where the latter is hidden in the `texError` method. There are two ways to fix this:
1. Use `return texError(...)`
2. Add `undefined` to the return type in jsdoc, such as 
```TypeScript
@returns {string|undefined} ...
```
Although `texError` has return type `never`,  
```Typescript 
@returns {string|never} ...
``` 
does not fix the error.

3. Final *unrealistic* alternative is to never have `texError` as the final command in a function. Example is `getBraces` in `ColumnParser.ts`, where the first `texError` is ignored by the linter while the last is not (hence there is a `return texError`).

Have a look which if you like the jsdoc solution and we can adapt the `getBraces` similarly.

### DIFF
--- a/ts/input/tex/TexParser.ts
+++ b/ts/input/tex/TexParser.ts
@@ -377,7 +377,7 @@ export default class TexParser {
    * @param {string} _name Name of the current control sequence.
    * @param {string?} def The default value for the optional argument.
    * @param {boolean=} matchBrackets True if indernal brackets must match.
-   * @returns {string} The optional argument.
+   * @returns {string|undefined} The optional argument.
    */
   public GetBrackets(
     _name: string,
@@ -426,7 +426,7 @@ export default class TexParser {
    *
    * @param {string} name Name of the current control sequence.
    * @param {boolean=} braceOK Are braces around the delimiter OK.
-   * @returns {string} The delimiter name.
+   * @returns {string|undefined} The delimiter name.
    */
   public GetDelimiter(name: string, braceOK: boolean = false): string {
     let c = this.GetNext();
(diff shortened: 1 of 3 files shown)


dpvc:
I'm wondering if `TexError.ts` could

```
export type texerror = undefined;
```

and then use

```
  @returns {string|texerror}
```

in order to be more explicit about what is happening.  I haven't tested if that actually prevents the error or not, but it should.

zorkow:
> I'm wondering if `TexError.ts` could
> 
> ```
> export type texerror = undefined;
> ```
> 
> and then use
> 
> ```
>   @returns {string|texerror}
> ```
> 
> in order to be more explicit about what is happening. I haven't tested if that actually prevents the error or not, but it should.

I am afraid, I cannot get that to work. 
* If we only us it in the jsdoc we get the `value is never read` TS error. And we still get the linting error `return expression not available`.
* If we use it also in the return value (i.e., `string | texerror`) we still get the same linting error.
* Finally if we use it as return value for `texError` we still get a number of ` Argument of type 'void | MmlNode' is not assignable to parameter of type 'MmlNode'` errors.

So I don't see an elegant way around it.

dpvc:
OK, thanks for checking.  Sorry that didn't work.  So I guess `@returns {...|undefined}` is the way to go.  I will approve.

dpvc (approved):
Does the trick.  One of my PRs added `return` to avoid this problem, I think in `TexParser.ts`, so that one will need to remove those returns before merging.

dpvc:
I've removed the `return` commands from my `fix/comments` branch, so I think everything will merge cleanly.

PR #1542, opened 2026-08-07
Update packages and linting

zorkow:
* Updates all packages to latest version except for `typescript`
* Fixes all linting errors of the form `no-useless-assignment`. These generally make things clearer. The only exceptions are in `Wrappers/mo.ts` and `Wrappers/mmultiscripts.ts` where it makes things slightly uglier.


ts/input/tex/newcommand/NewcommandUtil.ts
```diff
@@ -119,12 +119,12 @@ export const NewcommandUtil = {
    */
   GetTemplate(parser: TexParser, cmd: string, cs: string): number | string[] {
     // @test Def Double Let, Def ReDef, Def Let
-    let c = parser.GetNext();
+    parser.GetNext(); // Remove initial '{' from the string
```
dpvc: I think this is removing spaces, not `{`.  After #1539, it will also remove comments.

dpvc (approved):
One issue with an incorrect comment, but otherwise ok.

ts/input/tex/newcommand/NewcommandUtil.ts
zorkow: `GetNext` removes spaces and gets the next non-whitespace characters. But then we ignore that and then `c` to the next character. So does that not mean that we are effectively ignore the initial characters, which in the tests was always `{` ?

dpvc: No, `GetNext()` does return the next character, but doesn't move past it.  And the next character isn't a `{` in the case where you have a template, like `\def\x#1{...}` or `\def\x#1,#2\end{...}`.

https://github.com/mathjax/MathJax-src/blob/fb987178c2d279a99b0db5ea02e751691703955e/ts/input/tex/TexParser.ts#L298-L303

zorkow: OK, the `{` is probably wrong here. But the purpose of this `GetNext` is to move past the first non-whitespace character so `c` in the loop is bound to the second non-whitespace character.
I will clarify the comment.

zorkow: Changed comment to 

```Typescript
// Move past the first non-whitespace character.
```

dpvc: > the purpose of this GetNext is to move past the first non-whitespace character so c in the loop is bound to the second non-whitespace character.

That's still not correct.  The `GetNext()` call skips whitespace and returns the next character *without moving past it*, so two calls to `GetNext()` in a row both return the same character.  Note that `GetNext()` returns `getCodePoint()`, and neither function advances `i` past the character it returns.

https://github.com/mathjax/MathJax-src/blob/fb987178c2d279a99b0db5ea02e751691703955e/ts/input/tex/TexParser.ts#L298-L303

https://github.com/mathjax/MathJax-src/blob/fb987178c2d279a99b0db5ea02e751691703955e/ts/input/tex/TexParser.ts#L283-L286

The first `GetNext()` is just to move past the whitespace, so that `let i = parser.i` will be the position of the first non-whitespace character.  Inside the loop, the `c = this.GetNext()` will get that first non-whitespace character again.

The comment probably should be `Remove initial white-space`.

PR #1545, opened 2026-08-09
Fix synchronous loading to avoid potential loop.  (mathjax/MathJax#3597)

dpvc:
This PR prevents a potential infinite loop with synchronous file loading of dynamic font ranges, or when a dynamic range fails to load.  It also adjusts how dynamic range failures are reported.

The `FontData/NoLoad` message now shows the error that occurred, and we use `Locale.warn()` to produce it when the failure is trapped rather than later.  If a file is requested again when it has failed, we just skip it silently (that could occur if a second character from that range was requested, and we don't need a message for each character in the range).  This goes for both the synchronous and asynchronous loading methods.

The synchronous method now does the `setup()` call even if the file doesn't have to be loaded, so that it will work if you create new output jax instances, where the font data for the dynamic range hasn't been set up yet, but the file has been loaded.

There was a second issue involved in the problem from mathjax/MathJax#3597, which is that the `getChar()` method was supposed to remove the data that would have caused the range to be loaded, but it was failing to do so when the range was part of an inherited variant definition (like when `double-struck` inherits from `normal`).  The `delete` command doesn't delete from the `__proto__` chain, only from the top level, so we have to loop through the chain in order to find the correct one to delete from.  This is handled via a new `deleteDef()` method.

Resolves issue mathjax/MathJax#3597.


zorkow (approved):
lgtm.

zorkow:
> The `FontData/NoLoad` message now shows the error that occurred, and we use `Locale.warn()` to produce it when the failure is trapped rather than later. If a file is requested again when it has failed, we just skip it silently (that could occur is a second character from that range was requested, and we don't need a message for each character in the range). This goes for both the synchronous and asynchronous loading methods.
> 

What this PR highlights is that we need a strategy for changing existing messages (or adding new ones) in the light of existing localisations. So once we merge and branch `feature/locale_others` we have to be able to flag the `FontData/CantLoad` messages in the locale that have not been updated.

dpvc:
Yes, I was thinking that we needed to be able to identify new and changes messages as well.  Would it make sense to change the message to just `''` for new ones and ones that need updating?  Then it is easy to see them.  Alternatively, we could add something like `[update]` at the beginning if you don't want to lose the old translation.

PR #1546, opened 2026-08-09
Remove locale warning that is no longer needed, and fix a prettier issue.

dpvc:
This PR removes the locale warning, now that the `en` definitions are loaded automatically, even for direct loading of the MathJax modules.  It also fixes a prettier issue from a previous PR.


zorkow (approved):
lgtm.

PR #1547, opened 2026-08-09
Use retryAfter() and handleRetriesFor() from `mathjax` so they are easier to override

dpvc:
This PR moves the references to `retryAfter()` and `handleRetriesFor()` to the version in the `mathjax` variable rather than from the `Retries.ts` module directly.  This makes it easier to override them.  I have an example of making MathJax load tex extensions synchronously that would be helped by this, as that means there would be fewer methods that need to be overridden.  Together with the synchronous loading of dynamic font data and of entity definitions, this would allow MathJax to be run synchronously in node applications, where synchronous file loading can be achieved.

Note that `asyncLoad` uses `mathjax.asyncLoad` internally, so we don't have to use `mathjax` for that.

### DIFF
--- a/ts/output/common/FontData.ts
+++ b/ts/output/common/FontData.ts
@@ -1313,7 +1313,7 @@ export class FontData<
     if (!dynamic.promise) {
       dynamic.promise = Promise.resolve();
       try {
-        asyncLoad(this.dynamicFileName(dynamic));
+        mathjax.asyncLoad(this.dynamicFileName(dynamic));
       } catch (err) {
         dynamic.failed = true;
         console.warn(err);
--- a/ts/output/common/FontData.ts
+++ b/ts/output/common/FontData.ts
@@ -26,7 +26,6 @@ import { mathjax } from '../../mathjax.js';
 import { OptionList, defaultOptions, userOptions } from '../../util/Options.js';
 import { StyleJson } from '../../util/StyleJson.js';
 import { asyncLoad } from '../../util/AsyncLoad.js';
-import { retryAfter } from '../../util/Retries.js';
 import { Locale } from '../../util/Locale.js';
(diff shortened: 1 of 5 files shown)


ts/output/common/FontData.ts
```diff
@@ -1314,7 +1313,7 @@ export class FontData<
     if (!dynamic.promise) {
       dynamic.promise = Promise.resolve();
       try {
-        mathjax.asyncLoad(this.dynamicFileName(dynamic));
+        asyncLoad(this.dynamicFileName(dynamic));
```
zorkow: This now returns a promise regardless of if it fails or not. So the `try...catch` does not really have an effect on this. But that means that `dynamic.failed` is never set. That would mean the file is set to loaded
even if the promise failed.

ts/util/Entities.ts
```diff
@@ -507,7 +508,7 @@ function replace(match: string, entity: string): string {
       entity.charAt(0).toLowerCase();
     if (!loaded[file]) {
       loaded[file] = true;
-      const promise = mathjax.asyncLoad(`./util/entities/${file}.js`);
+      const promise = asyncLoad(`./util/entities/${file}.js`);
```
zorkow: Similarly, should the promise failure here not be handled explicitly?

zorkow (changes requested):
I have two concerns about the new async loading. Please have a look.

dpvc:
I have put back the `mathjax.asyncLoad()` calls to what they used to be.  Do you need to re-review?

zorkow:
> I have put back the `mathjax.asyncLoad()` calls to what they used to be. Do you need to re-review?

No, that looks good now.

zorkow (approved):

PR #1548, opened 2026-08-09
Update safe extension to remove event listeners in MathML input

dpvc:
This PR adds support for filtering out event listeners like `onclick` from MathML input.  By default these are now filtered out.   This behavior is controlled by `on*` entry in the `filterAttributes` map.  One could use

``` js
MathJax = {
  loader: {load: ['ui/safe']},
  startup: {
    ready() {
      MathJax.startup.document.safe.filterAttributes.delete('on*');
    }
  }
}
```

to remove this filter.  There is no configuration parameter for this, as there are completely unsafe, so removing this filter should be hard to do.

The PR also fixes the alignment of `@returns` jsDoc declarations.

### DIFF
--- a/ts/ui/safe/SafeMethods.ts
+++ b/ts/ui/safe/SafeMethods.ts
@@ -334,7 +334,7 @@ export const SafeMethods: { [name: string]: FilterFunction<any, any, any> } = {
   },
 
   /**
-   * Filter a on* attribute (don't allow them)
+   * Filter an on* attribute (don't allow them)
    *
    * @param {Safe<N,T,D>} _safe  The Safe object being used
    * @param {string} _value      The attribute's value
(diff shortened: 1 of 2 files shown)


ts/ui/safe/SafeMethods.ts
```diff
@@ -332,4 +332,24 @@ export const SafeMethods: { [name: string]: FilterFunction<any, any, any> } = {
   ): string | null {
     return id.match(safe.options.dataPattern) ? value : null;
   },
+
+  /**
+   * Filter a on* attribute (don't allow them)
```
zorkow:
```suggestion
   * Filter an on* attribute (don't allow them)
```

zorkow (approved):
lgtm.
One typo, but I might be wrong.

PR #1549, opened 2026-08-09
Refactor maction output wrappers to move common code to the common wrapper.

dpvc:
This PR refactors the `maction` wrapper functions to move common functions into the `common/Wrappers/maction.ts` file from the `svg` and `chtml` wrappers when they are the same.  In addition, a couple of service routines are moved to the `common/Wrapper.ts` function so that the functions moved to common `maction` can use them.

The motivation for this is the fact that the `toggle` and `statusline` actions in the `actions` map are essentially the same between the SVG and CHTML output versions, so could be moved to the common one.  Only the `tooltip` implementation differs, so the two jax-specific wrappers add that in individually.  The `setEventHandler()`, `Em()`, and `Px()` functions are moved to the common `maction` wrapper, and the `text()` function is moved to the common `Wrapper`.  A new `prefix` getter is added to the common `maction` wrapper so that the SVG output can use attributes like `data-toggle` while the CHTML output can just use `toggle`, as is presently the case.  

Finally, some comments were re-aligned, so you might want to view without white-space changes.

### DIFF
--- a/ts/output/chtml/Wrapper.ts
+++ b/ts/output/chtml/Wrapper.ts
@@ -518,14 +518,6 @@ export class ChtmlWrapper<N, T, D> extends CommonWrapper<
   }
 
   /**
-   * @param {string} text  The text from which to create an HTML text node
-   * @returns {T}           The generated text node with the given text
-   */
-  public text(text: string): T {
-    return this.jax.text(text);
-  }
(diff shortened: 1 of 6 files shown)


zorkow (approved):
lgtm.

PR #1550, opened 2026-08-12
Update/pnpm

zorkow:
This takes care of the change in `pnpm` from v10 to v11, where the `-s` option is no longer equivalent to silent.
See the notes here (https://pnpm.io/cli/run#--sequential--s).


dpvc (approved):
Well, I don't like that they changed the meaning of `-s`, as I type that automatically now, and don't want to have to do `--silent`, so I guess I'll have to make and alias for `pnpm --silent`.  Sigh.

In any case, thanks for fixing this.

PR #1553, opened 2026-08-17
Feature/locale others

zorkow:
PR fills in files for all existing locales with automatic translations by copilot and claude.
Also adds the initial crowdin configuration file.
The idea is that once this is all merged into `develop`, we spawn the CrowdIn support from `develop` leading to a `develop_l10n` branch, thereby replacing the current branch `l10n_l10n`, which only has some not so useful corrections to Spanish translations.
We would then manage all translations in CrowdIn, using English as a basis and distinguishing the following categories for managing changes:

```text
NEW
  message ID doesn't exist in locale

CHANGED
  English source changed since translation

MISSING_ARGUMENT
  translation doesn't contain all required arguments

OBSOLETE
  ID no longer exists in English

OK
  translation corresponds to current English source
```


dpvc (approved):
Looks good.

PR #1554, opened 2026-08-27
Implement component-based typing for the MathJax variable

dpvc:
This PR implements a mechanism for providing Typescript types for the `MathJax` global object.  This is complicated by several factors.  First, the `MathJax` object is used for initially for configuration and, then is altered by the startup process to include the appropriate commands based on the input and output jax that are loaded, and the types are different for those two uses.  Second, the final form of the `MathJax` object depends on the components that are loaded, and Typescript doesn't know this at compile time (the `loader.load` array gives the components, by their code isn't loaded until run time), and so the typing needed for the initial `MathJax` as a configuration object can't be determined (the `loader.load` is part of that object, but the types need to be known before that object's value is given).

To handle these difficulties, you need to tell MathJax what components you are going to load as parameters to a type that can be used to specify the `MathJax` object, as well as list them in the `ladder.load` array.  It is possible to list them only once, however, as shown in the examples that I will post in separate comments below.  It is probably best to look at these first, before diving into the code changes, in order to see how the typing are to be used in practice.

The typing mechanism can be used in all four of the main ways of using MathJax illustrated in the MathJax-demos-node repository:  the simple node-only method of using `node-main`, the component-based approach, the direct linking to MathJax modules, and the mixed direct-component approach, though each uses its own technique.  They are all based on the same typing code, however, just used in different ways.

The typing works by adding explicit typing for the `options` objects for all the components that allow option settings in the `MathJax` configuration object.  That involves nearly all the components, and so this PR touches a *lot* of files, and adds some new `.d.ts` files for the component definition files.  Fortunately, however, most of the changes follow a common pattern, so that should help with the review.

The changes in the `a11y` and `ui` directories are the most complicated (as they are the ones that have to extend the MathDocument class and its options), so it is probably best to start with the `core` directory, with the input and output jax, and the MathDocument first.  Then the input jax subclasses, their support files, and the TeX and MathML extensions.  Then the output jax, and finally the `a11y` and `ui` components.

The main idea is to export an `OPTIONS` type (e.g., `TEX_OPTIONS`, `CONFIGMACROS_OPTIONS`, `CHTML_OPTIONS`, etc.) that specify the types for the options of the class that is being defined, then a `const options: CHTML_OPTIONS = {...}` (for example) that lists the default options.  That is done as a separate object so that full type checking is performed (if it was done as `public OPTIONS = { ... } as CHTML_OPTIONS`, that would flag any incorrect or missing options, but would not flag extra ones, and we can't do `public OPTIONS: CHTML_OPTIONS = {...}` as the options from the parent class are given through `...Base.OPTIONS`, and so aren't listed explicitly, and get flagged.  As a separate object, it must match the option types exactly, and we can either use it directly as the `OPTIONS` value, or use `...options` to include it as part of a larger `OPTIONS` object.  (You'll see how it works in the code below).

For the various Handler objects, like the `a11y` and `ui` files that extend the `MathItem` and `MathDocument` classes, the situation is a little more complicated, as they have to add their options into the existing ones for the document class that they extend, and it is not always clear what that base class is going to be.  So, for example, the menu code needs to be aware of the explorer options as well as the complexity options, so it has to build on several other document classes. For these, ther his an `OPTIONS` type that lists the new option types for the updated MathDocument subclass, with an associated `options` that gives the default values for those options (fully checked), as usual, plus an additional interface that extends the base document class's options with the new `OPTIONS`, and that is used to set the subclass' `options` property's type.  A bit complicated, but it does the trick.

The SRE and A11Y options pose a similar challenge, as they get extended in several cases as well.  A similar process is used (newline the new options, then extends an interface with the new options being added to the old).

The building of the `MathJax` object typings is done using the files in the new `ts/stypes` directory.  The `Types.ts` file has the main machinery for combing the various types for the individual components into a larger single type that is what `MathJax` needs based on the components being loaded.  Components get a `COMPONENT_DEF` that describes the configuration options and any properties that get added to the `MathJax` object, plus whether the component defines an input or output jax (since those produce conversion methods in `MathJax`).  These objects get combined in the `type2mjx` types near the bottom of the file.  This has some fancy type definitions to handle things like creating the `input2outout` conversion functions (which rely on both the input and output components), and turing the config options from being required pro0erties into optional ones, and putting everything into the correct place in the final `MathJax` object type definition.  I'm actually astonished that this can be made to work.

The `Components.ts` file creates a type that lists all the components and created the `COMPONENT_DEF` object for each.  This file could potentially be generated automatically as part of the MathJax build process, but currently that isn't being done.  These component definitions are combined later to create the type for the `MathJax` object.  That is done using the definitions in the `mjx.ts` file.  That allows you to use a list of component names to produce the combination of the component definitions for those types, and then use the utilities in `Types.ts` to merge those into the final `MathJax` type.

Because the types depend on the DOM node types (the `N`, `T`, and `D` template values in the code), there is a `ts/types/dom` directory that has dom-specific versions of the typing macros that include the proper DOM types.  The `html.ts` file has the ones for in browser use (or with one of the adaptors that uses HTMLElements, like the jsdom and linkedom adaptors), and `lite.ts` for the LiteDOM adaptor.  That means that the results of the output conversions, for example, will be properly typed, and things like `MathJax.chtmlStylesheet()` will return an `HTMLElement` or `LiteElement` appropriately (rather than `any`).

There is a lot of pretty complicated Typescript in these files that we may want to look at together.  I will write up more details later, but want to get this PR posted first.  Look to the example files below to see how this can be used in practice.

Using MathJax in a Typescript project can involve loading MathJax component file (either from the `components` directory for the source versions, or from `bundle` for the packed versions), typescript will complain about the fact that there are no `.d.ts` files for these `.js` files.  So this PR includes new `.d.ts` files for all the `.js` files in `components/mjs`, and modifies the build tools to generate `.d.ts` files in the `bundle` directory.  Very few of the `components/mjs` files actually export anything, so most of those `.d.ts` files are just `export {}` in order to appease Typescript, but a few of them do include the actual exports for those files.  The only really interesting one is for `components/mjs/node-main`, which exports the `init()` function.  This is defined to take the list of components that you are loading and return the `MathJax` object properly typed for those, so that makes that node-only "simple" approach very straight-forward.


ts/types/Components.ts
```diff
@@ -0,0 +1,264 @@
...
+    'chtml',
+    CHTML_OPTIONS<DOM> & CHTML_FONT_OPTIONS,
+    DOM
+  >;
+  'output/svg': OUTPUTJAX<'svg', SVG_OPTIONS<DOM>, DOM>;
+
+  'a11y/semantic-enrich':
+    | DOC_OPTIONS<ENRICH_OPTIONS<DOM>>
+    | DOC_TYPE<EnrichedMathDocument<N<DOM>, T<DOM>, D<DOM>>>;
+  'a11y/assisitive-mml':
```
zorkow:
```suggestion
  'a11y/assistive-mml':
```

zorkow (approved):
Another long one...
I found one typo. Not too bad, as the module is no longer used anyway. 

Should we consider creating tests for the types/options? I am not sure how one would go about that, though.

PR #1557, opened 2026-09-01
Filter \t from URL protocols, same as \n and \r

dpvc:
This PR adds `\t` to the characters that are filtered by the Safe extension in URL protocols.  This prevents using `\t` to bypass the safe filtering.


zorkow (approved):
lgtm.

PR #1560, opened 2026-09-02
Update use of SPEECH and BRAILLE strings where they were missed earlier

dpvc:
This PR changes to explicit strings for their `SPEECH` and BRAILLE` visions, and corrects a typo in one `SEM` string.  Should be a quick one.


zorkow (approved):
lgtm.

PR #1561, opened 2026-09-02
Use the ui/safe data URL filter for the values used in `\data` TeX macros

dpvc:
This PR causes the `\data` macro in TeX to use the `ui/safe` data filter when the safe extension is loaded.  This prevents the `\data` macro from entering data attributes that could be harmful within frameworks that use data attributes for code control.

It also removes a reference that looks up a `data-MJX` attribute that is never actually set anywhere in the code.

In addition, it cleans up the mechanism used to set the MathML attribute and class filters by adding a service method to the MathML and MathMLCompile objects to set those values rather that casting the jax to `any` and setting them by hand.

Finally, it corrects an incorrect type that appears in the jsDoc for one of the safe methods.


zorkow (approved):
lgtm.

PR #1564, opened 2026-09-10
Revert change from feature/component-types that prevents multiple input jax from registering properly.

dpvc:
A change to `startup.ts` in the `feature/component-types` branch causes problems when multiple input jax are loaded, as in `tex-mml-chtml`, etc.  Only the first gets registered.  This reverts the change so that it works as before.


zorkow (approved):
lgtm.

PR #1565, opened 2026-09-10
Prevent webpacking from producing invalid names for region stylesheets

dpvc:
This PR fixes a problem with the id attribute for the style-sheets used for the various Region classes.  It currently is based on the class's `name` property, which should be the class name, but when webpacked the name can be changes to some random characters, and it is possible for that to produce an invalid element id.  E.g., I had one of the region classes be named `$E`, and that lead to an error in the browser for `MJX-$E-styles` being used as the `id` attribute.

The solution used here is to use the `className` property rather than `name`, since that is something like `MJX_hover` and won't be changed by webpack.


zorkow (approved):
lgtm.

PR #1567, opened 2026-09-12
Correct the type for enableMenu

dpvc:
This aPR fixes the type of `enableMenu` that wasn't changed from its specific `true` value to the generic `boolean` during the `feature/component-types` update.


zorkow (approved):
lgtm.

PR #1568, opened 2026-09-15
Feat/typst support

cvanelteren:
As it says on the tin


dpvc:
Thank you for the work you have put into this PR.  I have only had a quick look, but will do a more formal review when time permits.  I did bring this up in our developers' meeting this week, though we need to investigate it more closely to better understand your contribution.

One thing that we wondered is whether this might be something that would be better suited as part of the Typst community rather than MathJax itself.  For example, while there are many MathJax plugins for various learning management systems and blogging software, we do not maintain (or even catalog) those; they are maintained by their respective communities.  It might be that this input jax should be a third-party one that is made available through Typst rather than MathJax.  The fact that you need to separately load Typst code to pass to MathJax makes it feel more natural that way to me.

In any case, that is something to consider.  I will get back to you with more when I can.

cvanelteren:
@dpvc I see what you mean. The src that is imported is the webassembly stuff to interface with typst. I can see if I can add this to their repo in parallel. The implementation here complies with the interface that mathjax provides and is similar to Asciidoc's implementation.

cvanelteren:
Added a standalone here in the meantime for posterity: https://github.com/cvanelteren/mathjax-typst

PR #1569, opened 2026-09-16
Add tab key support to contextual menu, and add an aria-label

dpvc:
This PR adds a check for the Tab key in the key event handler and unposts the message when that occurs.  It also adds an ARIA label to the menu, as requested in mathjax/MathJax#3609.  As we know, the menu code needs an overhaul, which should be done after it is moved into the MathJax-src repository.

Resolves issue mathjax/MathJax#3609.


ts/ui/menu/MJContextMenu.ts
```diff
@@ -147,8 +147,21 @@ export class MJContextMenu extends ContextMenu {
     this.nofocus = false;
   }
 
+  public keydown(event: KeyboardEvent) {
+    console.log(event);
```
zorkow: This should be removed.

zorkow (approved):
Just one left over logging. Otherwise fine.

PR #1573, opened 2026-09-20
Prevent `mtextInheritFont`/`merrorInheritFont` from overriding explicit font attributes

dpvc:
This PR fixes a problem where `mtextInheritFont` being `true` would override any explicit font settings on an `mtext` element, such as `style="font-family: Courier"` or `fontfamily="Courer"` attributes.  This also font weight and font style settings.  Similarly, if `mtextFont` is set, or for an `mtext` that is a direct descendent of `merror`, the `merrorInheritFont` and `merrorFont` values would work in a similar way.

This PR fixes these problems by using the `removedStyles` object to store the font, font weight, and font style to use when `mtextInheritFont` or any of the others is set.  This object is consulted in the common wrapper's `getVariant()` function to get the proper font/weight/style, so this leverages the existing algorithm rather than replacing it.

You can test this using

``` xml
<mtext fontfamily="arial" fontweight="bold" fontstyle="italic">Testing</mtext>
</math>
```

and

``` xml
<mtext style="font-family: Arial; font-weight: bold; font-style: italic">Testing</mtext>
</math>
```

with `mtextInheritFont` set to `true`, which should produce the same result as when it is `false`.  Similarly for setting `mtextFont`.  In addition, removing the explicit font in the ones above should add the given weight/style to the default font.  Put an `merror` around the `mtext` to test the same for the `merror` options.

Note that the lab allows you to control these options, so they are easy to change.


ts/output/common/Wrappers/mtext.ts
```diff
@@ -231,35 +231,34 @@ export function CommonMtextMixin<
...
+        this.node.attributes.getExplicit('fontfamily')
+      ) {
         return;
       }
-      super.getVariant();
+      this.removedStyles ??= {};
+      this.removedStyles.fontFamily = font;
+      const variant = this.node.attributes.get('mathvariant') as string;
+      const [, italic, bold] =
+        (this.constructor as typeof CommonMtextMixin).INHERITFONTS[variant] ||
```
zorkow: I am surprised that works as `CommonMtextMixin` is declared as a function earlier in the file.
I've always viewed mixins more as class factories.

zorkow (approved):
Lgtm.
Just one comment.

ts/output/common/Wrappers/mtext.ts
dpvc: I'd because of the lines

https://github.com/mathjax/MathJax-src/blob/7281b94ce431452f57e8832233c739cdef96ef83/ts/output/common/Wrappers/mtext.ts#L183-L185

within that function, so there is a local `CommonMtextMixin` class in that function, and the `typeof CommonTextMixin` in line 254 refers to that local class, not the outer function.  All the mixins return classes with the same name as the constructor function (so when debugging, we will see objects of class `CommonMtextMixin` rather than `anonymous class` or something like that).

dpvc: And you are right, the mixin is a class factory.  It's just that the factory and the returned class have the same name.

PR #1574, opened 2026-09-22
Allow font family, weight, and style CSS to be inherited by token elements

dpvc:
For an expression like

``` xml
<mrow style="font-family: arial">
  <mn>1</mn>
  <mo>+</mo>
  <mn>2</mn>
</mrow>
```
 the font family should affect all three elements that it contains.  But currently in MathJax, that is not the case.

This PR fixes that issue and propagates the font family, weight, and style CSS styles of a container to its child nodes.  For an `mstyle` element, the `fontfamily`, `fontweight` and `fontstyle` are also propagated.  For an `mi` element with a single character content, the style will be set to italic for the specified font automatically.

This inheritance is accomplished by using the `setInheritedAttributes()` method.  The `font-family`, `font-weight`, and `font-style`, if specified, are moved to the inherited `fontfamily`, `fontweight`, and `fontstyle` attributes (these are inherited values, not explicit attributes, so don't show up in the MathML output directly).  The output jax wrapper's `getVariant()` method then picks these up and uses them when determine the math variant to use (which can be an explicit font).  This handling is simplified a bit as well.


zorkow (approved):
lgtm.

